PR #141, opened 2013-07-26
C syntax fixes

rygorous:
Fixes for C and C-like syntax highlighting.

* C-like and C: Exponents may contain '+' too
  (e.g. "1.234e+3" instead of just "1.234e3").
* C: Numbers can have "float" (f), "unsigned" (u), "long" (l) 
  or "long long" (ll) suffixes (case insensitive).

Real C compilers have a few additional restrictions; there
are no unsigned floats, for example. This regexp accepts
a reasonable superset of the valid suffixes.


Golmote:
I merged it manually. I messed up a bit though, so there are two commits: 4d64d077265b6e229701f486ce704fb921c69384 and 071c3ddf767d3d30bc17e4ee0b27f3543eca7a15.

PR #167, opened 2013-09-18
Make line-highlight plugin work with line-numbers plugin

ryanzec:
Right nwo the line-highlight and line-numbers plugins don't work together.  The line-highlight plugin first adds the line number for the highlighted lines.  The line-highlight plugin add itself to the code element which is problematic when the code element is positioned relatively (with the line-numbers plugin does).

This change checks for the line-numbers class on the pre element and if it find it, is does not add the data-start/data-end attributes (since the line-numbers plugin is already handling that) and it does appends the line-highlight element to the pre element no matter what (which fixes the position issue.


benjaminapetersen:
Ping on this one!  I just tried Prism for the first time, made a plain index.html file, installed Prism with bower and added a JS & CSS block.  Here is a screenshot if it helps:

prism-

Golmote:
@benjaminapetersen Can you provide an online example that reproduces the issue please?

citkane:
Another ping on this one given the caveat that I have prism well embedded within an angular app, so I have not tested the default behaviour. This is what I am getting when line numbering and line highlighting are enabled together:

prism2

And how it looks with a dirty CSS hack:

prism1

By overriding the css thus:

```css
.line-highlight{
	min-height:28px;
}
.line-highlight::before, .line-highlight[data-end]::after {
	border-radius:0;
	top:0;
	left:0;
	height:28px;
	width:38px;
	box-shadow:none;
}
```
It would be nice if the two can play together reliably - thanks for a great product.

PR #191, opened 2013-12-19
Added highlighting of $ shell prompt

kbjr:
When using bash highlighting to show command line use like this:

``` bash
$ npm install foo
```

The opening prompt `$` is now highlighted.


Golmote:
I think this can be closed, in favor of the [Command Line plugin](http://prismjs.com/plugins/command-line/) which provides a nice display for the prompt and is configurable.

PR #208, opened 2014-02-11
Detect GitHub file links.

ahluntang:
Detect GitHub link and fetch the raw file through the GitHub API. This is because in some cases GitHub blocks fetching files through the raw.github.com feature.
## Usage

```
<pre data-src="https://github.com/LeaVerou/prism/blob/gh-pages/prism.js"></pre>
```


mAAdhaTTah:
I think this is now handled by the JSONP plugin.

PR #308, opened 2014-07-29
Add APL Language (Thanks to github.com/ngn for APL Regex)

JasonRivers:
Adding APL Language to Prism with examples and extra style colours.


Golmote:
I merged this manually. Added some aliases too.
Thanks for contributing! ;)

PR #313, opened 2014-08-05
Add SBP langage

kylekatarnls:
I used Prism on [http://sbp.selfbuild.fr/](http://sbp.selfbuild.fr/) and extended PHP Prism component to handle SBP, the langage this website presents (not yet translated), SBP is an open-source langage that adds syntaxic sugar to PHP. So I propose in this pull request the code I used to display SBP in Prism.
Here is the SBP repository:
[GitHub repository](https://github.com/kylekatarnls/sbp)


kylekatarnls:
Oh, thanks to consider this. But I have to say SBP is no longer maintained. Nowadays, it would not be useful I think even if I can't say how many other people could have used it.

Golmote:
Then we'll close the PR. If anyone needs the language definition, it's still here anyway.

PR #320, opened 2014-08-10
Extend Latex support

zeitgeist87:
Hi,

I am not a Latex expert, but I use it from time to time. I tried to add a few features, that I would like to have. It is far from complete, but it should be easily extendible by people who really know Latex ;). I am open for suggestions to improve this patch.

This patch extends the support for Latex in the following way:

* It introduces a helper function to highlight Latex environments that
  are enclosed with `\begin{something} ... \end{something}`. There are a
  lot of these environments in the Latex language that could get special
  highlighting in the future.
* Certain functions like `\ref` or `\begin` always contain a reference
  or a keyword and they are highlighted as such.
* Chapter and section headlines are written in bold to make them stand
  out more, because they structure the document.
* Since `selector` and `string` have the same color in the default
  theme, the token type `function` is used to highlight Latex commands.


Golmote:
@zeitgeist87 Sorry for the delay. I merged this manually in 942a6ec7694ecd343b44e3f7b6fc9079d3b8bbfe. Would you mind updating the examples in `examples/prism-latex.html` to reflect the additions you made?

PR #355, opened 2014-09-02
[Git] Add of some jasmine tests

lgiraudel:
Add of some jasmine tests to check the HTML output.

The tests can be executed with `gulp test` or automatically with `gulp watch`.


TimothyGu:
Is there a reason why you used Jasmine instead of Mocha? I don't intend to start a flame war but am just wondering since Mocha is a lot more popular (1620 vs 302240 npm downloads in last week)

PR #362, opened 2014-09-07
Fixed 'self' variable in Safari (desktop & iOS)

planerist:
The old method to set 'self' variable worked incorrectly in Safari web browsers. So I propose the fix to make it work good on all browsers & nodejs as well.


Golmote:
I assume this is no longer relevant. Since we do no longer overwrite the `self` global directly, and instead use a `_self` variable.

PR #370, opened 2014-09-22
JSON Language

CupOfTea696:
Added JSON support to prism.


ibc:
Will it be finally pulled?

LeaVerou:
Holy crap, I’m so sorry for taking this long. 
It looks fine and can always be fixed later. However:
a) We need a minified version and an entry in components.js otherwise it won’t show up anywhere.
b) A few examples would be nice, but not required. @Golmote do we have any docs for doing that?

foxx:
Just a heads up, this version doesn't seem to work as nicely as the one on
Stripe. I was able to rip out their Prism.languages.json code but it needs
work. I'll submit an alternative later today so you guys can compare the
differences.

Golmote:
No there is no actual guidelines for writing the examples. Just try to follow the format used in the current examples and to provide a pretty complete overview of the features of the component.

chris-martin:
JSON is the only thing I'm using Prism for, so I'm not sure whether this is a bug in this plugin or in the library, but FWIW: This doesn't work well with strings that contain numbers.

json

Golmote:
@chris-martin: This component has never made it into Prism. I think you'd better use the JS component to highlight your JSON:

LeaVerou:
@CupOfTea696 Reversing the order of the number and string tokens shall fix this. Please refer to the JS language definition for the optimal order.

CupOfTea696:
That should've fixed it.

chris-martin:
I'm confused about how this can make a difference. Aren't JS objects unordered?

Golmote:
In theory they are. In practice all engines follow the order in which the properties were defined. And Prism relies on this behaviour.

uranusjr:
(See also discussion in #441)

markcarver:
This works rather well.

The only thing that I see missing is that if you use the `show-language` plugin, it only capitalizes the first letter. This will need to be added to the `Languages` object in https://github.com/PrismJS/prism/blob/gh-pages/plugins/show-language/prism-show-language.js#L8

Here is a temporary work around:
```js
  Prism.hooks.add('before-highlight', function(env) {
    var pre = env.element.parentNode;
    if (!pre || !/pre/i.test(pre.nodeName)) return;
    if (/^json/i.test(env.language)) {
      pre.setAttribute('data-language', 'JSON');
    }
  });
```

nauzilus:
This is done automatically via the gulp build task: https://github.com/PrismJS/prism/blob/gh-pages/gulpfile.js#L52-L89

darobin:
Is there any hope of this landing soonish? Highlighting with JS isn't very helpful as pretty much everything ends up being pretty much the same colour :) Also, I've been thinking of making it possible to style keys starting with `@` differently so it works for JSON-LD too, if that's acceptable.

Golmote:
@darobin I'll try to make a proper review of this PR by the end of the week.

vsimonian:
Hate to be that guy (I understand this is an OSS project), but any updates on a code review? This PR has been around for over a year now.

drodil:
Anything new to this?

components/prism-json.js
```diff
@@ -0,0 +1,17 @@
+Prism.languages.json = {
+    'property': /"(\b|\B)[\w-]+"(?=\s*:)/ig,
+    'string': /"(?!:)(\\?[^'"])*?"(?!:)/g,
+    'number': /\b-?(0x[\dA-Fa-f]+|\d*\.?\d+([Ee]-?\d+)?)\b/g,
+    'function': {
+		pattern: /[a-z0-9_]+\(/ig,
+		inside: {
+			punctuation: /\(/
+		}
+	},
```
LeaVerou: JSON doesn't have functions…

components.js
```diff
@@ -181,6 +181,11 @@ var components = {
 			"require": "clike",
 			"owner": "sherblot"
 		},
+		"json": {
+			"title": "JSON",
+			"option": "default",
+            "owner": "CupOfTea696"
```
LeaVerou: This seems misaligned. Different indent, perhaps?

components/prism-json.js
```diff
@@ -0,0 +1,17 @@
+Prism.languages.json = {
+    'property': /"(\b|\B)[\w-]+"(?=\s*:)/ig,
+    'string': /"(?!:)(\\?[^'"])*?"(?!:)/g,
```
LeaVerou: JSON doesn't support single quoted strings.

components/prism-json.js
```diff
@@ -0,0 +1,17 @@
+Prism.languages.json = {
+    'property': /"(\b|\B)[\w-]+"(?=\s*:)/ig,
```
LeaVerou: Any valid string is a key, not just `[\w-]`.

LeaVerou:
Sorry this has taken so long. I just took a look and added a few comments.

CupOfTea696:
@LeaVerou no worries! Applied the necessary changes.

LeaVerou:
Merged, thanks!
@darobin I think a change to accommodate JSON-LD would be awesome!! Please submit a PR for it, I would love this.

drodil:
Hi, good work! Is this going to be released anytime soon?

LeaVerou:
@drodil Any merged PR is immediately available to the live site.

Btw, @CupOfTea696, I noticed you made this enabled by default. While I did merge it to include it in Prism, I do not agree with it being enabled by default. Could you or someone else please send another PR to remove that?

zeitgeist87:
> Could you or someone else please send another PR to remove option: default?

I've removed the default option.

LeaVerou:
Thank you Andrea!!

CupOfTea696:
@LeaVerou That's probably because I copied it over from javascript. Sorry about that. Probably wasn't sure on what it did and just left it that way :p

Issue #368, opened 2014-09-21
Alternative language names


CupOfTea696:
Would it be possible to add alternative language names, so you'd be able to use language-html for markup as well as language-markup.

PR #371, opened 2014-09-22
language aliases

CupOfTea696:
Added common language aliases as suggested in #368


Golmote:
I added some very common language aliases in 2f9fe1efb80e4744092cfd1e0ad245df18e59c0d. Please submit another PR if you wish to add any.

PR #375, opened 2014-09-28
CSS Extra Enhancement

CupOfTea696:
Added string & punctuation to css.selector.inside


mAAdhaTTah:
I don't think is needed anymore.

PR #385, opened 2014-10-14
correct calculation for number of lines

### LINKED ISSUE #382: Scrolling over code block with scoll wheel or touch pad scrolls block of code down by one line when last line of code terminates with newline in pre source.
If a page author has a block of code such as:

    <pre><code>some.source()
    </code></pre>

The line break at the end of the code block results in a blank line at the end of the processed block (that gets a line number), but which the code block doesn't expand to show. This results in a scroll bar appearing, and when the block is scrolled over (as when scrolling down the page) causes the code area to scroll down by one line, obscuring the first line of code in a way that can be confusing to the viewer.

See http://www.airpair.com/angularjs for examples.

Note: to replicate this by scrolling over a code block, the mouse pointer must land inside the code block while scrolling (so if you scroll over a code block with a line height of 1 and the mouse 'jumps' over that line of code while scrolling, it won't cause the text inside the code block to scroll)

haarg:
A final \n in a `<pre>` is not rendered by browsers, so we shouldn't
include it in the calculation for the number of lines.

Fixes #382.

### DIFF
--- a/plugins/line-numbers/prism-line-numbers.js
+++ b/plugins/line-numbers/prism-line-numbers.js
@@ -5,10 +5,10 @@ Prism.hooks.add('after-highlight', function (env) {
 		return;
 	}
 
-	var linesNum = (1 + env.code.split('\n').length);
+	var linesNum = env.code.match(/\n(?!$)/g).length + 1
 	var lineNumbersWrapper;
 
-	lines = new Array(linesNum);
+	lines = new Array(linesNum + 1);
 	lines = lines.join('<span></span>');
 
 	lineNumbersWrapper = document.createElement('span');


aldanor:
Wondering about this one as well; any reason why it hasn't been merged in?

aldanor:
@LeaVerou @haarg 

Here's the simplest minimal example in JSFiddle: http://jsfiddle.net/g8zwu9fr/

Note that there's a use case for it since e.g. it's exactly what Jekyll (Kramdown) outputs when you have a fenced code block like this:

    ~~~ python
    print 'foo'
    print 'bar'
    ~~~

Golmote:
Couldn't it be handled the same way an initial line feed is?

See https://github.com/PrismJS/prism/blob/gh-pages/components/prism-core.js#L185

haarg:
Why not just use this patch as is?  Rather than trying to hack around the issue, it just does the appropriate thing and fixes the line number calculation to match browsers.

Golmote:
I'm just trying to find the best way to solve it.

plugins/line-numbers/prism-line-numbers.js
```diff
@@ -5,10 +5,10 @@ Prism.hooks.add('after-highlight', function (env) {
 		return;
 	}
 
-	var linesNum = (1 + env.code.split('\n').length);
+	var linesNum = env.code.match(/\n(?!$)/g).length + 1
```
apfelbox: Nitpicking: could you please re-add the final `;`?

LeaVerou:
@Golmote The way we're handling the initial line feed is terrible (sorry, I think it's my fault) and is causing problems in dabblet currently (try pressing enter in the beginning of your code). I think we should remove that.

As for the last line break, the issue is that browsers don't display that last line, so we shouldn't calculate it. This is not the same as the initial line break, which is displayed (which is why I was "conveniently" removing it).

Golmote:
@LeaVerou: Thank you for the feedback on this. I might create a plugin to remove the initial line feed then, so that those who want that behaviour can have it.

PR #397, opened 2014-11-16
highlightAll accepting an element to query

AlexeiDrake:
Currently, the `highlightAll` function queries `document` for all highlightable elements, then calls `highlightElement` on all of them. Sometimes we may wish to do so for a specific element deeper down the DOM tree.
There is already a `highlightElement` function, but it exits for elements that are not immediately highlightable. The `highlightAll` function seems to be a good candidate for a simple generalisation.
If no element is passed, the function can default to querying `document`, hopefully requiring minimal changes to existing code.

### DIFF
--- a/components/prism-core.js
+++ b/components/prism-core.js
@@ -140,8 +140,11 @@ var _ = self.Prism = {
 		}
 	},
 
-	highlightAll: function(async, callback) {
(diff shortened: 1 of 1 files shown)


tanakahisateru:
I created another pull-request #1230

Golmote:
#1230 fixed this issue in a more compatible way.

PR #405, opened 2014-12-01
+ Build and Publish

tenbits:
Hello Team,

there is still no npm package, components and plugins are not ready to use in node env., and the bower package is from the 3rd part source. 
I'm proposing here automatically build and release the prism. 

**Build**: For the `node` environment, factory wrapper is used for each component and plugin, so it is then possible to `require` it.
**Release**: Just run `npm run release` from the cli and the script bumps both `bower/npm` versions, check-outs new `release` branch, builds the library, creates proper `.gitignore`,  publishes the new release branch and also the new git tag. If the maintainer uncomments `npm publish` [line](https://github.com/tenbits/prism/blob/builder/tools/release/process.js#L37) this will also publish the package. Bower's package updates after the new tag is created (_but we should ask the current bower maintainer to update the repo url_).

There is how release branch looks then: [prism/release](https://github.com/tenbits/prism/tree/release) 

Should you have any questions please give me a note, Alex.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 hide-*.js
 node_modules
 .idea/
+lib/
--- a/bower.json
+++ b/bower.json
@@ -4,6 +4,8 @@
   "main": [
       "prism.js",
       "themes/prism.css"
+      "lib/prism.js",
+      "themes/prism.css"
   ],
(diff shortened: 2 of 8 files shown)


Mrman:
+1

PR #416, opened 2014-12-10
Added Markdown language

### LINKED ISSUE #30: Add a markdown syntax highlighter
I'm working on a project for offline markdown editing, and I'm totally willing to change the highlighting system for prism.

I'm suggesting a code highlighting for markdown. Probably I will end up implementing this one :)

Thanks

Golmote:
Fixes #30 

Not really sure about the use of the aliases that don't make real sense, but I wasn't confident enough to modify the themes...


vkbansal:
and what about this PR? When will this be merged?

Golmote:
@vkbansal When @LeaVerou will have time to review it.

LeaVerou:
Looks good to me. Conflicts need to be resolved.

However, for MD, it might be a good idea to apply some formatting instead of just highlighting it. E.g. make bold tokens bold, make italic tokens italic, Make headings bigger, de-emphasize the MD symbols to make the content stand out (like Byword does) etc.

Golmote:
@LeaVerou Does it look better now?

prism-markdown

LeaVerou:
Nice!
Btw, we can fix some of the issues by:
- Giving priority to MD over HTML highlighting
- Adding nested rules after we define the grammar, for some common cases

Golmote:
Giving priority to MD over HTML will raise other issues, like the inability to comment markdown:
```markdown
<!-- *italic* -->
```
It will also break HTML highlighting in some cases like this one:
```markdown
<div class="_test_"></div>
```

Are we sure we want to do this?

LeaVerou:
If commenting is the main conflict, we can add it as a separate token. But if cases like the second are common, then probably not a good idea. I’m really not sure :(

Golmote:
I don't think people really make extensive use of HTML attributes in markup. So I guess it's ok to give priority to MD.

LeaVerou:
Then we should probably add a comment token early on, referencing HTML’s comment token.

Golmote:
I inserted markdown before the `prolog` key of markup, so HTML comments have priority.

I added some nested patterns with `inside` but since the bold, italic and link patterns are in competition, the result is not perfect.

prism-markdown

mAAdhaTTah:
Just FYI, looks like some of the edge cases are no longer (italic looks like it's working inside bold).

Golmote:
@mAAdhaTTah Yes, that is why the text says "italic inside bold does work". ^^

mAAdhaTTah:
Reading comprehension FTW. :-p

Golmote:
Is it more clear like this?

prism-markdown

mAAdhaTTah:
Works for me.

Golmote:
@LeaVerou Looks good for you?

LeaVerou:
Sure! It’s not perfect, but then again, what is :)

PR #423, opened 2014-12-12
File highlight start/end line number option

nfriedly:
I incorporated the `data-start` attribute from the line-numbs plugin and added a corresponding `data-end` attribute. If either is present, the download file is split into lines and slice'd to the correct subset before highlighting.

Plays nice with line-numbers plugin, and treats negative `data-start`'s as `0`.

I included an example on the page.

### DIFF
--- a/plugins/file-highlight/index.html
+++ b/plugins/file-highlight/index.html
@@ -8,6 +8,7 @@
 <base href="../.." />
 <link rel="stylesheet" href="style.css" />
 <link rel="stylesheet" href="themes/prism.css" data-noprefix />
+<link rel="stylesheet" href="plugins/line-numbers/prism-line-numbers.css" data-noprefix />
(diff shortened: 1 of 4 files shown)


Golmote:
I like the original idea. But IMO the regexp option would only solve some uncommon use cases and so is not really nedded.

nfriedly:
@Golmote hah, I got pretty much the exact opposite response from a couple of my coworkers - they thought that line numbers were relatively annoying and would much prefer being able to add a comment to the code to mark the start and end of the snippet to show. 

However, I'll admit that it might be a little bit uncommon, and that's part of why I hadn't included it yet. (Also, I was hoping to hear from @LeaVerou before putting much more work into this.)

nfriedly:
Hey, we should probably just close this. I don't work at the same employer anymore and I don't think they're still using any of this :/

Issue #412, opened 2014-12-09
Thoughts about strings, comments, regexps and other "enclosing" patterns


Golmote:
Hi there,

As you are aware of, Prism has problems with all patterns that are "enclosing" ones (couldn't think of a better word to describe them...) like strings, comments and regexps.
If we take the JS component, we soon encounter the following problems, regarding comments, strings and regexps :

As far as the tokens appear in the order defined in the components, all is fine :
```javascript
/*te"/st*/ "*/test", /te"st*/;
```
renders as <img src="http://puu.sh/dneVs/3b17d1fdc7.png" />

But if the string comes first :
```javascript
"/*test", /te"st*/;  /*te"/st*/
```
it renders as <img src="http://puu.sh/dneZg/a6ba40c526.png" />

The same kind of problems occurs when the regexp comes before the string, etc.

Currently, Prism will start by matching the full string against the comment pattern, resulting in the creation of a token that will prevent the string to be properly matched on next iteration.

I was thinking lately about making a distinction between "enclosing" patterns and other ones. If enclosing patterns were to be known, it could be possible to overcome the problem by changing the way Prism handles the matching : it could be possible to start by matching the full string against all "enclosing" patterns, and apply the one that matched the closer to the start of the string. I think this would be a safe behaviour for most (if not all) of the languages.

What do you think of this theory ? (And am I making myself clear ?)

foo123:
Check [prism-grammar](https://github.com/foo123/prism-grammar) external parsing add-on for Prism, which uses a grammar to highlight solving all these issues (and more)

Golmote:
Yes I know about your addon. I don't think it's a solution we want to include into Prism though.

foo123:
@Golmote  eventualy if this is going to be solved adequately sth like that add-on will have to be created since simple regexes cannot account for that complexity of laguage, it needs a grammar (or sth equivalent to that, it doesnt have to be the prism-grammar add-on, see [here](https://github.com/foo123/editor-grammar/blob/master/grammar-reference.md#modularity-and-future-directions) as well for further options).

Golmote:
You are right that regexes can't solve the issue completely but it can definitely do better than now. And Prism will keep working this way. We won't add parsing abilities to it.

foo123:
[...]

Issue #307, opened 2014-07-29
Higlighting issue with strings & comments


apfelbox:
This is the combined issue for all related issues (#187, #183, #155, #153).

There is a fight between the highlighting of strings vs. comments.
Currently only one of them can work

```
// this is "a string"
"string with /* comment";
```

I will take a look whether this issue can be "fixed" (it can't be fixed for all cases due to the limited power of regular expressions though).

foo123:
Check [prism-grammar](https://github.com/foo123/prism-grammar) external parsing add-on for Prism, which uses a grammar to highlight solving all these issues (and more)

WebReflection:
chiming in with a proof of concept that I've no idea if is worth exploring.

Basically, it removes strings, highlight them, then put them back.

```js
function safeHighlight(content, highgligh) {
  var str = [];
  function pre(content) {
    return content.replace(
      /(["'])(?:(?=(\\?))\2.)*?\1/g,
      ($0, $1) => $1 + (str.push($0.slice(1, -1)) - 1) + $1
    );
  }
  function post(content) {
    return content.replace(
      /(["'])(\d+)\1/g,
      ($0, $1, $2) => $1 + str[$2] + $1
    );
  }
  return post(highgligh(pre(content)));
}

var content = 'var str = "hello/**/world//!";';

// use prism or the actual highlighter instead of String
safeHighlight(content, String);
```
I haven't done many benchmarks but it seems like a simple/safe solution for strings in comments and/or in code: it will put them back in their context, if that was highlighted, strings will result highlighted too.

Best Regards

zeitgeist87:
@WebReflection your solution would certainly work for strings, although it seems more like a hack. Do you have any proposal for turning it into a general purpose solution?

We have recently merged a partial solution for that problem #779. It can correctly handle strings with one comment inside but not more (e.g.: `"hello/**/world!"`)

WebReflection:
well, the general purpose solution is what you call the hack:

  * find all strings and put a string placeholder
  * parse everything without problems and ignoring all strings content
  * put strings back using the placeholder

A char-by-char parser wouldn't have this problem because it would know it's inside a string ... my work-around does exactly the same making the "_inside a string_" a no-brainer for the parser.

I could write a more robust example with a better placeholder but I honestly don't know how to improve every language parsed by `prism`

zeitgeist87:
> well, the general purpose solution is what you call the hack:
> 
> find all strings and put a string placeholder
[...]

Issue #187, opened 2013-12-12
String incorrectly highlight when /*


Freyskeyd:
Hey,

String are not detected as string when we use `/*` in it.

``` javascript
var cfg = [
  './working/string.js',
  './not/**/working.string.js'
];
```

see ya!

Issue #155, opened 2013-08-14
JavaScript: /* inside a string variable considered as comment start


Dither:
For example try to highlight this script: http://github.global.ssl.fastly.net/assets/frameworks-eae23340ab2a6ba722166712e699c87aaf60ad8f.js

Somewhere in the middle of that script there is the code:

``` javascript
}.call(this), function () {
    $(document).on("ajaxBeforeSend", function (a, b, c) {
        if (!c.dataType) return b.setRequestHeader("Accept", "*/*;q=0.5, " + c.accepts.script)
    })
}.call(this), function () {
```

And from `"*/*;q=0.5, " +` it's considered as comment by Prism until the next mention of `"*/*"`.

PR #424, opened 2014-12-12
Competition between patterns to solve the comment/string/regexp problem

Golmote:
This is a proposition to solve the main problem of Prism. (#307, #412, #187, #183, #155, #153...)

The PR adds the notion of competition between patterns.

Before highlighting the code, we loop through the patterns to group together the ones that are in competition with each other (each group contains what I call competitors).
Then we loop through the groups and for each part of the string, we try to match each competitor. The winner is the one which matches the closest from the start of the substring.

To indicate that some patterns are in competition, you need to add a "competition" key on the pattern, specifying an identifier for the competition. All patterns in the grammar that have the same competition identifier are in competition with each other.

If you don't change the language definitions, the behaviour is the same as before.

A behaviour that needs to be mentionned is that the position of the first competitor in the grammar determines the position all the whole competition with that identifier (i.e. all the competitors), wherever the other competitors are defined in the grammar. This should normally not be a problem, as candidates for competition are most of the time consecutives in the grammar.

I configured C-like and JS as an example, by setting the same competition key for the tokens "comment", "string" and "regexp" (that are all three "enclosing" patterns). Here is a before/after comparison :
prism-before prism-after

What do you think of it ?


Golmote:
Hi there!

I thought a lot about this stuff during the holidays. Your suggestion using `start` and `end` patterns would raise some tricky problems...
To match the whole symbol, we'd need to merge the `start`, `contents` and `end` regexps together, leading to the same problems encountered in PR #444! What happens if the user sets the `i` modifier only on the `start` token, for example?

What do you think? Can we sort it out or should we go back to the boolean solution, and just find a better name than `competition`?

zeitgeist87:
You don't have to combine them. You search for `start` and get an index, then you search for `end` and get an index, then you check if `contents` matches inbetween. If it's possible, you combine them of course.

zeitgeist87:
You could even optimize a bit, by matching only `start` for tokens that are in competition. Then sort the resulting list of start indices in decending order. Then go through the list and match `end` and `contents`. The first match is the new token.

Golmote:
This looked like an awesome idea until I tested strings like `"foo \" bar"`. When searching for `end`, we find the escaped quote.
Even if we add some lookbehind check to ensure the quote is not escaped, it will fail on `"foo \\"` by thinking the quote is escaped when it's not...

So we might need to consume `contents` before searching for `end`, but this would require `contents` to be more precise. Like `/(\\.|[^"])*/` instead of just `/(\\?.)*/` for strings, for example. This is probably not easy enough for "a random dev, new to Prism, trying to hack together a language definition".

What do you think?

zeitgeist87:
@Golmote yes you are absolutely right. I hadn't thought of escaped end patterns. You could also not use back references to refer to the string that was matched with `start`.

A `boolean` flag `competition` would probably be best, although I like your idea of allowing different scopes of competition by nameing them with a string.

LeaVerou:
@Golmote: If the dev gives us the escape character too (which would default to `\`, but other languages may have different escape characters), can’t we build the necessary regex ourselves in the background?

I really think the flag idea is confusing for devs trying to hack together language definitions. Prism’s priorities are: 1. API & code simplicity 2. Performance 3. Small code size, in that order.

Also, have we done any performance comparison between this compared to previous versions of Prism that failed on such nested structures? A small performance hit is acceptable, but if Prism has to become e.g. 4 times slower to become a bit more correct, it wouldn't be worth it. It's a tradeoff.

Golmote:
I just ran a quick benchmark on the three concerned versions of Prism: the original (current state on Prism website), the one from this PR (in the current state) and the one from PR #444 (at commit zeitgeist87/prism@0187761a56d8abdc83288921e4c4782ecbab955d).

I did not benchmark your proposed API as I could not yet make it work properly.

The benchmark highlights the following code:
 ```javascript
/*fo"/o*/ "*/bar", /ba"z*/;
"Foo // bar"
"/foo/"; /* "bar" /baz/ */
/"foo"/, "/*bar", "baz*/";
/foo"/, "bar", "baz";
// foo "bar" baz
"foo /* bar */ baz";
```

I repeated this code fragment multiple times to test for longer codes.

The code was highlighted a first time before the benchmark runs so that I could check the result and ignore the grammar optimization time in PR #444.

Here are the results in Chrome 39.0.2171.95:

Prism version | Code length | Number of runs | Average time
---|---|---|---
Original | 155 chars | 1000 | 0.29 ms
PR #424 | 155 chars | 1000 | 0.37 ms (27% slower)
PR #444 | 155 chars | 1000 | 0.35 ms (20% slower)
Original | 1550 chars | 100 | 1.86 ms
PR #424 | 1550 chars | 100 | 2.64 ms (41% slower)
PR #444 | 1550 chars | 100 | 2.38 ms (28% slower)
Original | 15500 chars | 50 | 9.99 ms
PR #424 | 15500 chars | 50 | 14.28 ms (43% slower)
PR #444 | 15500 chars | 50 | 13.67 ms (37% slower)
Original | 155000 chars | 10 | 171.87 ms
PR #424 | 155000 chars | 10 | 208.23 ms (21% slower)
PR #444 | 155000 chars | 10 | 200.53 ms (17% slower)

Here are the results in Firefox 34.0.5:

Prism version | Code length | Number of runs | Average time
---|---|---|---
Original | 155 chars | 1000 | 1.11 ms
PR #424 | 155 chars | 1000 | 1.46 ms (31% slower)
PR #444 | 155 chars | 1000 | 1.39 ms (25% slower)
Original | 1550 chars | 100 | 5.53 ms
PR #424 | 1550 chars | 100 | 7.08 ms (28% slower)
PR #444 | 1550 chars | 100 | 7.12 ms (28% slower)
Original | 15500 chars | 50 | 51.35 ms
PR #424 | 15500 chars | 50 | 74.60 ms (45% slower)
PR #444 | 15500 chars | 50 | 69.39 ms (35% slower)
Original | 155000 chars | 10 | 1022.95 ms
PR #424 | 155000 chars | 10 | 1705.92 ms (66% slower)
PR #444 | 155000 chars | 10 | 2153.13 ms (110% slower)

Finally, here is the benchmark code I used:

```html
<pre><code id="benchmark-core" class="language-javascript">/*fo"/o*/ "*/bar", /ba"z*/;
"Foo // bar"
"/foo/"; /* "bar" /baz/ */
/"foo"/, "/*bar", "baz*/";
/foo"/, "bar", "baz";
// foo "bar" baz
"foo /* bar */ baz";
</code></pre>

<script src="components/prism-core.js"></script>
<script src="components/prism-clike.js"></script>
<script src="components/prism-javascript.js"></script>
<script>
	(function() {
		var elt = document.getElementById('benchmark-core'),
			code = elt.textContent;

		Prism.highlightElement(elt);

		function avg(arr) {
			return arr.reduce(function(a,b) { return a+b; })/arr.length;
		}
		function str_repeat(s, n) {
			var res = '';
			while(n--) {
				res += s;
			}
			return res;
		}

		function benchmark(code, cpt, cb) {
			var logs = [],
				f = function() {
					if(cpt--) {
						var start = performance.now();
						Prism.highlight(code, Prism.languages.javascript, 'javascript');
						logs.push(performance.now() - start);
						requestAnimationFrame(f);
					} else {
						console.log('Average time: ' + avg(logs) + ' ms');
						cb && cb();
					}
				};
			console.log('Highlighting '+code.length+'-char code '+cpt+' times.');
			f();
		}

		window.onload = function() {
			benchmark(code, 1000, function () {
				benchmark(str_repeat(code, 10), 100, function () {
					benchmark(str_repeat(code, 100), 50, function () {
						benchmark(str_repeat(code, 1000), 10, function () {
							console.log('Done');
						});
					});
				});
			});
		};
	}());
</script>
```

zeitgeist87:
@Golmote Thanks for the performance tests. Really nice work!

Golmote:
> @Golmote: If the dev gives us the escape character too (which would default to \, but other languages may have different escape characters), can’t we build the necessary regex ourselves in the background?

@LeaVerou If we need to build the regexp by combining the `start`, `content` and `end` patterns, with or without any escape character, we will encouter the same problems with the regexp flags (see previously https://github.com/LeaVerou/prism/pull/424#issuecomment-68592959, or yours here https://github.com/LeaVerou/prism/pull/444#issuecomment-67894079)

If we try to add the escape character to zeitgeist87's suggestion (https://github.com/LeaVerou/prism/pull/424#issuecomment-68594600), we could indeed ensure the `start` and `end` pattern are not preceded by the escaped character, but it would fail on cases like `"foo \\"`.

Another idea, to ease the merging of the different parts, we could ask the dev to input "regexp-strings" instead of real regexps so that don't have any flag problems. For example, the C-like string pattern would become something like `{start: '("|\')', contents: '(\\\\?.)*?', end: '\\1'}`. But IMO, all these backslashes are quite ugly, and the whole thing is really easy to get.

Finally, as the C-like string pattern shows it, there is the problem of the back references, which are particularly needed for patterns that allow any identifier to be used as a delimiter (e.g Heredoc strings in PHP:
```php
$str = <<<PRISM_IS_GREAT
Example of string
spanning multiple lines
using heredoc syntax.
PRISM_IS_GREAT;
```
This needs back references, to be matched properly.)

Golmote:
I'm still trying to think about a better word for an hypothetical boolean flag but I'm quite stuck with my lack of english vocabulary. Maybe something like `matchFirst` would be more appropriate?

Also remember that not setting the boolean flag would just keep the current behaviour of Prism and devs will write new language definitions by looking at the existing ones so I guess they will quickly understand what the boolean is meant for...

LeaVerou:
Wow, @Golmote you never cease to impress!! Nice work!!

zeitgeist87:
I kind of like the name `greedy` for the flag.

Golmote:
Wouldn't `greedy` be confusing regarding its meaning when talking about regexps?

zeitgeist87:
@Golmote I think `greedy` is a good choice, because it has essentially the same meaning as in the context of a regex, but on the scope of the whole language. So people who know what geediness means for regexes can intuitively understand what it means on a larger scale. But maybe I am wrong. I am awful at nameing things...

Golmote:
>  I am awful at nameing things...
Haha! So I am.

Well, if we stay with the flag solution, maybe @LeaVerou would have a better idea for the name.

PR #428, opened 2014-12-14
Add Polymer language

### LINKED ISSUE #426: CSS Highlighting Error within a custom element
```
.some-class {
   width:  {{shareSize}}px;
   height: {{shareSize}}px;
}
```

{{shareSize}} is an attribute for my custom element (polymer) which is used as follow:
`<my-custom-element shareSize=32></my-custom-element>`

so... yes... the hightlighting breaks at the first `{{shareSize}}px;`

Golmote:
To be more precise, this is to highlight the data-binding syntax used in Polymer templates.

Fixes #426 

Heavily based on Handlebars component.


paladox:
As polymer is based in html, this may not work as the lang returned would be html or markup. (im thinking of when users use the Prism.highlight function.

paladox:
Also polymer 2 is out now which uses es6 syntax. Polymer 1.x still uses the old syntax :).

Should we do polymer 2 seperate?

Golmote:
Ideally we should be able to make only one component for all versions of Polymer. This PR certainly needs to be reworked anyway, considering its age. I'll have to dive into this at some point.

paladox:
Also considering this is in html + javascript this should extend the javascript language so that there's no need for Prism.language.prism

mAAdhaTTah:
Does this need to use `markup-templating` now?

Golmote:
@mAAdhaTTah Yes, this PR needs to be rewritten from scratch and it will certainly use `markup-templating`.

I never took the time to work on this PR again, because I know @LeaVerou was reticent to include it into Prism, although I don't remember exactly why. I don't know if she has changed her mind since then.

Golmote:
Oh, here it is: https://github.com/PrismJS/prism/pull/439#issuecomment-67732158. I guess this comment is no longer relevant in that case, since we've been using this exact pattern in templating languages for some time now. I'll try to update this PR anytime soon.

LeaVerou:
I'm fine either way, up to you @Golmote!

mAAdhaTTah:
@Golmote I think this would be great to include, but if it needs to be rewritten from scratch, I'm going to close this PR for now.

PR #444, opened 2014-12-21
Combine regexes in arrays if possible

zeitgeist87:
It is often useful to split up a big regex into separate logical
units. Unfortunately this is bad for performance, because the input
string has to be separately traversed for every unit.

This patch adds a simple optimization function, which walks through a
grammar and combines regexes that are split up in an array into one
single regex.

This can only be done, if the array contains only regexes, which all
have the same modifiers and no back references.


zeitgeist87:
I could remove support for strings. That would significantly simplify it....

LeaVerou:
Strings definitely need to be supported, as my plan for this is to convert keywords in languages to arrays of strings and then push/remove as necessary in inherited languages instead of overriding the entire list to add/remove a few.

Btw happy new year!!

zeitgeist87:
We could combine this with #424. If every element of an array is in "competition" with the other elements of the same array, the order wouldn't matter any more. The code for `optimizeGrammar()` would be much simpler, we could optimize more aggressively and it would be more consistent, because the user expects the order to be not significant.

The only problem is of course performance.

> Btw happy new year!!

@LeaVerou Happy new year to you too!!

Golmote:
I also feel like both PR could be merged somehow, but we first need to find some rock-solid solutions for the problems...

zeitgeist87:
I combined this PR with #424 and the result is very nice. I basically just plagiarised @Golmote's idea and work, so I don't want to take credit for it. But the combination just works so beatifully, that I had to try it. I gave credit to @Golmote in the commit message. We could also merge the two PRs in some other way...

components/prism-core.js
```diff
@@ -140,6 +140,67 @@ var _ = self.Prism = {
...
+					competition.push(patterns);
+				}
+				continue;
+			}
+
+			for (var i = 0; i < patterns.length; ++i) {
+				var pattern = patterns[i],
+					type = _.util.type(pattern);
+
+				if (patterns.competition) {
```
Golmote: Shouldn't it be `pattern` (singular) ? We won't set a `competition` property on the array.

zeitgeist87: Yes you are correct. It's a copy&paste error.

components/prism-core.js
```diff
@@ -248,62 +312,64 @@ var _ = self.Prism = {
...
 			var patterns = grammar[token];
-			patterns = (_.util.type(patterns) === "Array") ? patterns : [patterns];
 
-			for (var j = 0; j < patterns.length; ++j) {
-				var pattern = patterns[j],
-					inside = pattern.inside,
-					lookbehind = !!pattern.lookbehind,
-					lookbehindLength = 0,
-					alias = pattern.alias;
+			patterns = (_.util.type(patterns) === "Array") ? patterns : patterns.competition ? grammar._competition : [patterns];
```
Golmote: What about arrays containing competition patterns? Like the c-like comments :
```javascript
[
	{
		pattern: /(^|[^\\])\/\*[\w\W]*?\*\//g,
		lookbehind: true,
		competition: true
	},
	{
		pattern: /(^|[^\\:])\/\/.*?(\r?\n|$)/g,
		lookbehind: true,
		competition: true
	}
]
```

components/prism-core.js
```diff
@@ -235,6 +297,8 @@ var _ = self.Prism = {
 		var rest = grammar.rest;
 
 		if (rest) {
+			_.optimizeGrammar(rest);
```
Golmote: The non-enumerable property `_competition` won't be added to the grammar in the for/in loop that follows.
This seems to raise errors when the code tries to handle competition in `inside` scenarios (like embedded JS in HTML).

components/prism-core.js
zeitgeist87: Yes that is also an error. Sorry for that. I first implemented the competition inside of arrays and then patched on the global competition without enough testing as it seems. I have to think about how to solve this. Thanks for your review!

Golmote:
It also looks like the competition behaviour is applied everytime a competition pattern is encountered. If I understand your code properly, the first time already matches all the patterns, so the others are not required.

components/prism-core.js
zeitgeist87: That is a good point. I could just concatinate it to the `_competition` property of grammar.
```js
if (rest) {
	_.optimizeGrammar(grammar);
	_.optimizeGrammar(rest);

	for (var token in rest) {
		grammar[token] = rest[token];
	}
	delete grammar.rest;
	grammar._competition.concat(rest._competition);
}
```

zeitgeist87:
> It also looks like the competition behaviour is applied everytime a competition pattern is encountered. If I understand your code properly, the first time already matches all the patterns, so the others are not required.

Could you explain a bit further. I don't know exactly what you mean.

For every partial string, all the patterns that are in competition are applied once. The best result is chosen. For the next partial string all the patterns have to be tried again. Maybe I miss something here...

Btw, thanks for your thorough review.

components/prism-core.js
Golmote: Yes, this should work. (But `concat` returns a new array so you must assign the result. And thus the property must be writable.)

Golmote:
> For every partial string, all the patterns that are in competition are applied once.
That's the point I don't agree with.

The `_competition` property contains all the competition patterns found in the grammar.

On line 316, when looping through the patterns, if the current pattern is a competition pattern, you "replace" it with all the competition patterns stored in the `_competition` property. So, after this iteration, all competition patterns are processed.

So when you'll encounter another competition pattern after the first one, you'll replace it again with the `_competition` property, although this pattern has already been processed with the other competition patterns.

Do you get my point?

LeaVerou:
Happy new year and thanks for working on this!

Please do not merge this or the competition PR before I have time to take a proper look, as these are very substantial changes.

Golmote:
Of course I won't. ^^ And happy new year to you too :)

zeitgeist87:
> Do you get my point?

Yes thanks. Now I get it.

components/prism-core.js
zeitgeist87: But I can use 
```js
Array.prototype.push.apply(grammar._competition, rest._competition);
```

zeitgeist87:
I think I fixed all of the issues. I no longer use `_competition` to store an array of competing patterns. Instead I directly insert them into the grammar in the `optimizeGrammar()` function. This way I can avoid the issue pointed out by @Golmote of the competition patters getting applied repeatedly. It is also a much nicer design.

zeitgeist87:
There are synergetic effects between the combination of regexes and the idea of competition, because to combine regexes more aggressively, the arrays of regexes have to be greedy. To make them greedy I use competition. To introduce globale competition with a flag I only have to collect them in an array and insert it at the right position into the grammar. Because arrays are already greedy, we get the global competition basically for free.

Golmote:
I'm mostly worried about one edge case, now.
Because the optimization process modifies the structure of the grammar, what would happen if an asynchronously loaded component tries to extend an already optimized grammar?

zeitgeist87:
> Are we sure we want to push it, and not unshift it?

What difference would that make? I don't see how adding them to the front of the array would be beneficial.

> Also, this seems to mean that the other elements in the array will be in competition with the "competition patterns", even if they were not supposed to be. If I'm right, could this lead to unexpected behaviours?

You are right. I can't think of an example, where it would lead to unexpected behaviour though, but I can't rule out the possibility either.

> Because the optimization process modifies the structure of the grammar, what would happen if an asynchronously loaded component tries to extend an already optimized grammar?

That's a good point. Some tokens in the grammer will be set to `null`. This shouldn't lead to any problems except in `_.util.clone()` and in `_.languages.DFS()`, but I think I can fix that.

Golmote:
Did you see my edit about the elements in the same array? I think the order should not be a problem here.

About the extension mechanism, maybe we could store the optimized grammar in a non-enumerable property of the original grammar (kept unchanged)?
In the `tokenize` function, we would work on the optimized grammar, but the extension mechanism would use the original grammar.

zeitgeist87:
> About the extension mechanism, maybe we could store the optimized grammar in a non-enumerable property of the original grammar (kept unchanged)?
> In the tokenize function, we would work on the optimized grammar, but the extension mechanism would use the original grammar.

I just had the same idea. Then I could also easily add a special token like `_competition` before the element, instead of pushing it to the end of the existing array.

zeitgeist87:
I have updated the PR. Now the original grammar is not modified by `optimizeGrammar()` any more. Instead the optimized grammar is stored in the non enumerable property `_optimized`. Now it is also possible to "deoptimize" by simply deleting the property `_optimized`. Since `_.languages.DFS()` can do in place modifications of the grammar I deoptimize at this point. The changed grammar will be optimized again next time it is passed to `_.tokenize()`.

Golmote:
While doing a benchmark, I noticed that this PR does not highlight properly the following two cases:
```javascript
"/*bar baz*/";
"//bar";
```

If there is a space between the opening quote and the slash, the highlighting is fine.

Screenshot:

zeitgeist87:
@Golmote Thanks. It should be fixed now. The previous version did not account for the lookbehind in the sort function. Therefore the index was wrong.

Golmote:
It is fixed, indeed! Nice work

zeitgeist87:
I have created a performance test on http://jsperf.com to compare #334, #444, #424, which all solve the same problem:

http://jsperf.com/prism-parser-performance/3

You are welcome to make revisions of the test or change the parameters.

For some reason I can't get Firefox to give consistent results. This has probably something to do with memory management issues. So firefox should perform more consistently with less demo code.

vkbansal:
@zeitgeist87 I'm getting error on Prism PR #444

zeitgeist87:
@vkbansal I used https://rawgit.com to load the files directly from github, but this service sometimes failes. You could try to refresh the page and try again. It certainly works for me.

But maybe there is something else wrong. Could you post the error log from the js console?

vkbansal:
@zeitgeist87 it seems that the problem was asynchronous loading of the scripts.
The log was:
```
Cannot set property 'keyword' of undefined
```

After couple of dozens of reloads, I've managed to get it work!
Great work :+1:

zeitgeist87:
@vkbansal could you run the tests again. I updated the test and now the results are gone. Thanks

vkbansal:
@zeitgeist87 Ran on Chrome, Firefox, Opera and IE11

Golmote:
@zeitgeist87 I think @LeaVerou will not agree to add the "enclosing" boolean (at least for now). Would you mind refactor this PR to focus only on regexps combination? I think she might be ok to merge it, then.

I'll keep thinking about enclosing patterns, so we can handle them properly in the future.

LeaVerou:
What happened with this? It was a nice improvement. :/

Golmote:
@LeaVerou I think @zeitgeist87 never had the time to refactor the PR to keep only the regexps combination. We might have to refactor it ourselves, but that means closing it and opening a new one.

mAAdhaTTah:
Going close this, but happy to revisit if we want to include these changes.

PR #450, opened 2014-12-31
Add diff syntax definition

uranusjr:
Largely based on `git`’s definition for `inserted` and `deleted`, but expanded to support

1. All `coord` styles including `diff -c` and `diff -u`.
2. The `>`/`<` syntax.
3. The `!` syntax in context diff.

Example page included.

For the `!` syntax I aliased `important` to `diff`.


TimothyGu:
In general, +1, although I would appreciate it if you can make Git depend on diff instead of still being stand-alone.

uranusjr:
I don’t own the Git syntax though. Would be happy to help if @lgiraudel thinks this is a good idea.

components/prism-diff.js
```diff
@@ -0,0 +1,20 @@
...
+		/^(\*{3}|-{3}|\+{3}).*$/m,
+		// Match "@@ ... @@" coord lines in unified diff.
+		/^@@.*@@$/m,
+		// Match coord lines in normal diff (starts with a number).
+		/^\d+.*$/m,
+	],
+
+	// Match inserted and deleted lines. Support both +/- and >/< styles.
+	'deleted': /^[-\<](?![-\<]).+$/m,
+	'inserted': /^[\+\>](?![\+\>]).+$/m,
```
LeaVerou: Why escape the angle brackets? Also `+` doesn’t need escaping inside a character class. 
Also, what’s the negative lookahead trying to prevent? What happens if the line starts with e.g. ++ or >>?

LeaVerou:
Overall looks nice, see a few minor comments.
Thanks for adding examples too!
Please fix the merge conflicts?

Re:git, once I merge this, feel free to submit a PR to make it dependent on diff. It makes sense. Just because you don’t own it doesn’t mean you can’t suggest changes to it!

uranusjr:
Thanks @LeaVerou, amended. Sorry for the changes taking so long, I totally forgot about this. :(

Golmote:
(You should not have included the autohotkey changes in this PR)
Thanks, I'm merging.

Golmote:
@uranusjr Is it intended that those two lines in unified diff are matched as "coord"?

```
--- qcli.yml	2014-12-16 11:43:41.000000000 +0800
+++ /Users/uranusjr/Desktop/qcli.yml	2014-12-31 11:28:08.000000000 +0800
```

uranusjr:
@Golmote Yes that is what I intended, since they sort of act as separators between diff chunks, like the `***` lines in context diff. That’s just my opinion though; I wouldn’t mind if you want it be done differently.

Golmote:
I'm telling because of #401 and #769. I just modified the git component to match those lines as "inserted" and "deleted".
I'll try to think of a way to distinguish between those lines containing file names and the lines containing coords.

PR #451, opened 2014-12-31
Added syntax highlighting for Qore language

temnoregg:


components/prism-qore.js
```diff
@@ -0,0 +1,20 @@
+Prism.languages.qore = Prism.languages.extend('clike', {
+  'comment': {
+    pattern: /(^|[^\\])(\/\*[\w\W]*?\*\/|((\/\/)|#).*?(\r?\n|$))/g, // /(^|[^\\])([\w\W]*?\*\/|(^|[^:])\/\/.*?(\r?\n|$))/g,
```
Golmote: Is it important to consume the line feed ? Otherwise, I think the regexp could end simply with `.+`

temnoregg: If you think it would be better then change it - I am not good at regexp and I have probably copy pasted from somewhere - actually it is year old or so - so I don't exactly remember where or how I got it - I have just started using bower so I wanted to add it to primary repository to use it with the project

Golmote:
Merged manually in 125e91ff1db2d385150d5e63c4779933e851f70c.

PR #452, opened 2015-01-04
make comment non-greedy in handlebars

vkbansal:
When a comment is used inside a document, the document was marked as comment from the start of the comment until last occurrence of `}}`. This fixes that behavior.


Golmote:
Thanks!

PR #467, opened 2015-01-11
Added basic support for the julia language

cdagnino:
I added a component for the Julia language. I based it on the Python component, but changed the booleans and keywords. Let me know if I need to change anything


Golmote:
Could you please use tabs for indentation instead of spaces?

LeaVerou:
Also, if it’s based off Python, it might be a good idea to make it inherit from Python. I don’t know much about Julia though and how related it is to Python…

cdagnino:
@Golmote : changed the indentation to tabs (sorry about that, should have been more careful)

@LeaVerou : I based it off the Python component just to get me started and because they share a few things, but Julia takes things from many other languages. A second iteration should depart more from the Python component, so I don't think it's useful to make it inherit from Python.

LeaVerou:
Indentation of "title" is still off compared to the others.

Apart from that, LGTM. @Golmote, what do you think?

Golmote:
Apart from the "ignore" pattern, which I don't think will ever match (because of the "operator" pattern), it looks good to me.

cdagnino:
Let me know if you need anything else from my end, thanks.

LeaVerou:
The ignore pattern was a hack that was needed in the early days that Prism escaped stuff. It hasn’t been needed for a while now. Similarly, the `&lt;` etc tokens won't match, because Prism operates on `textContent` and doesn't escape stuff any more (it’s not needed). You need to remove the ignore token and convert entities to their equivalent symbol.

We also need to fix the Python language for the above issues, how did we miss that when the big conversion happened, I have no idea…

cdagnino:
I remove the ignore token and changed the entities.
Let me know if I need to pull/rebase any changes (and I'd appreciate if you are explicit on what I need to do, my git knowledge isn't that solid)

LeaVerou:
LGTM. Thanks!

PR #468, opened 2015-01-12
Stop ie8 from breaking with prism-core

Pinpickle:
Can't say anything about plugins, but this stops IE8 from breaking (read some discussion at #9).
All I did was add a `var` statement to the declaration of `self` at the beginning of the file and IE stopped complaining. Nothing is broken, because that still makes it a global declaration.

Please ignore my typo about issue 8 in my commit message, `git amend` didn't work for some reason.


Golmote:
Does it still work when used in a Worker?

Pinpickle:
@Golmote I'm afraid I don't know and I'm not even sure how to test that. I'm fairly sure the new code means [exactly the same thing](http://stackoverflow.com/a/1470494/1136593), so this shouldn't change any functionality anywhere.

Happy to test with some instructions!

Golmote:
You should be able to test it by adding the `data-manual` attribute on the script and then calling `highlightAll` yourself, with `true` as the first parameter (`async`).

Pinpickle:
Tested it on Chrome, Firefox, Safari (Mac), and Chrome, IE11 (Windows). Everything looked to be working. An error was thrown about an undefined `document` (because workers don't have access to the DOM). I was getting these with or without the change in my tests so that's an issue for another day.

So yes, it works when used in a Worker.

Joeynoh:
Came here to submit the same pull request. This issue keeps failing our IE tests on deploy, and now we cannot use a package manager to install Prism.

PR #471, opened 2015-01-13
Add class-name highlighting for Eiffel

Conaclos:
Hi!

The creation of external theme, in particular a themes, in particular the [EiffelStudio IDE](https://www.eiffel.com/) theme, the Eiffel language definition must enable highlighting of the class-names.

I based the regex on the unanimously used Eiffel convention for class naming (i.e. class-names are always all upper-case characters).
The only known issue is that the formal generics are considered as class names...

Thanks.

### DIFF
--- a/components/prism-eiffel.js
+++ b/components/prism-eiffel.js
@@ -13,6 +13,8 @@ Prism.languages.eiffel = {
 	'char': /'(?:%'|.)+?'/,
 	'keyword': /\b(?:across|agent|alias|all|and|attached|as|assign|attribute|check|class|convert|create|Current|debug|deferred|detachable|do|else|elseif|end|ensure|expanded|export|external|feature|from|frozen|if|implies|inherit|inspect|invariant|like|local|loop|not|note|obsolete|old|once|or|Precursor|redefine|rename|require|rescue|Result|retry|select|separate|some|then|undefine|until|variant|Void|when|xor)\b/i,
 	'boolean': /\b(?:True|False)\b/i,
+	// Convention: class-names are always all upper-case characters
+	'class-name': /\b[A-Z][\dA-Z_]*\b/g,
 	'number': [
 		// hexa | octal | bin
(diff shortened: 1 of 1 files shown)


LeaVerou:
How big of an issue is that? What are formal generics? Is there anything to distinguish them on the syntax level? 

(also, thanks!)

Conaclos:
### Formal generic parameter and class-name

[Eiffel Standard (ECMA 367)](http://www.ecma-international.org/publications/standards/Ecma-367.htm) 8.12:
> The genericity mechanism, still based on classes, gives us a new level of flexibility through type parameterization. You may for example define a class as LIST [G], yielding not just one type but many: LIST [INTEGER], LIST [AIRPLANE] and so on, parameterized by G.

The only way to know if the identifier is a class-name or a formal generic parameter is to know the class declaration.
e.g.
```eiffel
deferred class
    A [G]

feature
    items: G
        deferred  end

end
```
The declaration ``` class A [G] ``` enables to say that ``` G ``` is a formal generic parameter. All occurrences of ``` G ``` in the class denotes this generic.

### Semantic issue

I think that use the same color for class-anmes and formal generic parameter names is not really a problem. The notion are related...
However using `class-name' as CSS class for both is semantically wrong.

What do you think if I use `entity' CSS class instead of `class-name'?

Conaclos:
@LeaVerou 
### Is it really an issue?

Using the same CSS class for class-names and formal generic parameter names is not really a problem. The notion are related...

What do you think about it?

Conaclos:
@Golmote Hi!

A new commit from @carlfriess is now integrated (See #827).

Eiffel users which use Prism for Eiffel highlighting gave me a positive feedback about this PR.
For example http://eiffel-guide.com/ uses a modified version of Prism in order to benefit of these changes.

Golmote:
Sorry for the long time waiting. Thank you for contributing!

PR #474, opened 2015-01-14
Deep clone arrays

nauzilus:
Using slice won't deep clone arrays of objects. For instance when extending clike, the comments definition won't be deep cloned.


LeaVerou:
Ouchie, that must have been a tough one to debug. Thanks, merged!!

PR #475, opened 2015-01-14
Use explicit language class if given.

rbuch:
Using the file highlight plugin with a custom file type didn't work for me, generating `code` with a class name of `language-<file extension>`, even with explicit language specification in the `pre`.  This did not highlight properly.  Copying the class from the `pre` to `code` fixed the problem.


Golmote:
This has been fixed in 8030db914a1e7331d245db0b9b787b4364994440. Thanks for contributing.

PR #478, opened 2015-01-16
fix class-names starting with space

vkbansal:
This is a quick fix to remove leading space from class names

**Before:**
prism01

**After:**
prism02


LeaVerou:
Is there any actual problem with these class names besides aesthetics?

vkbansal:
It's just aesthetics. 
I have compulsion for such thing :p

LeaVerou:
Heh, fair enough. However, I have a bigger compulsion with the aesthetics of the code we work with, rather than the generated markup one only sees in the dev toops. Hope u understand!

vkbansal:
No problem :)

PR #481, opened 2015-01-19
Added Highlight Keywords plugin

vkbansal:
This plugin allows fine-grained control over highlighting keywords.

**Before:**
capture2

**After:**
capture3

### DIFF
--- a/plugins/highlight-keywords/index.html
+++ b/plugins/highlight-keywords/index.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+
+<meta charset="utf-8" />
+<link rel="shortcut icon" href="favicon.png" />
+<title>Show Language ▲ Prism plugins</title>
+<base href="../.." />
+<link rel="stylesheet" href="style.css" />
+<link rel="stylesheet" href="themes/prism.css" data-noprefix />
+<link rel="stylesheet" href="plugins/highlight-keywords/prism-highlight-keywords.css" data-noprefix />
+<script src="prefixfree.min.js"></script>
+
+<script>var _gaq = [['_setAccount', 'UA-33746269-1'], ['_trackPageview']];</script>
+<script src="http://www.google-analytics.com/ga.js" async></script>
+</head>
+<body>
+
+<header>
+	<div class="intro" data-src="templates/header-plugins.html" data-type="text/html"></div>
+
+	<h2>Show Language</h2>
+	<p>Allows fine-grained control over highlighting keywords.</p>
+</header>
+
+<section>
+	<h1>Examples</h1>
+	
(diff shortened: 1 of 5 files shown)


LeaVerou:
Lovely idea! Two things:

- I’d suggest prepending the classes with `keyword-`, otherwise there will be a lot of clashes and unexpected results.
- Please explain a bit more how this is used. Just a couple of sentences should be enough. I had to read the plugin code to understand what it does exactly.

vkbansal:
> I’d suggest prepending the classes with keyword-, otherwise there will be a lot of clashes and unexpected results.

fixed it.

> Please explain a bit more how this is used. Just a couple of sentences should be enough. I had to read the plugin code to understand what it does exactly.

I'm not good at writing :p 
May be @Golmote could can help

Golmote:
Just something like:
"This plugins adds special CSS classes for each keyword matched in the code. For example, the keyword `if` has the class `keyword-if`. You can customize the appearance of each keyword by providing your own stylesheet."
?

I'm not particularly good at writing english. This is not my native language. ^^'

Also, I'm not sure I like the idea of including a stylesheet with the plugin. IMO, you should provide no stylesheet and simply include an inline `<style>` in the HTML file to highlight your example.

vkbansal:
@LeaVerou updated the description

plugins/highlight-keywords/index.html
```diff
@@ -0,0 +1,51 @@
...
+<script>var _gaq = [['_setAccount', 'UA-33746269-1'], ['_trackPageview']];</script>
+<script src="http://www.google-analytics.com/ga.js" async></script>
+</head>
+<body>
+
+<header>
+	<div class="intro" data-src="templates/header-plugins.html" data-type="text/html"></div>
+
+	<h2>Highlight Keywords</h2>
+	<p>This plugins adds special CSS classes for each keyword matched in the code. For example, the keyword <code>if</code> will have the class <code>keyword-if</code>. You can have fine grained control over the appearance of each keyword by providing your own stylesheet.</p>
```
LeaVerou: "plugin", singular please :)

Also, "will have the class `keyword-if` **as well**". It will still have the class `keyword` too, right?

stylesheet -> CSS rules (stylesheet implies separate file, which is not necessary)

Apart from that, looks ok!

vkbansal:
@LeaVerou done. Anything else?

mAAdhaTTah:
Just to throw it out there, but is this worth merging into prism-core, rather than as a plugin?

LeaVerou:
Merged, thanks!

@mAAdhaTTah I don’t think that’s needed commonly enough to be in the core. Also, I'd rather the core be agnostic regarding specific tokens, like `keyword`.

PR #483, opened 2015-01-20
Added TypeScript definition

vkbansal:
typescript


LeaVerou:
thanks, merged!

At some point we should convert all keywords to arrays, so we can push and remove keywords when overriding languages instead of having to repeat the whole list for just a few changes… (but that should happen after we optimize arrays)

Golmote:
> At some point we should convert all keywords to arrays, so we can push and remove keywords when overriding languages instead of having to repeat the whole list for just a few changes… (but that should happen after we optimize arrays)

#444 is still pending.

Golmote:
@vkbansal About this grammar definition... why are the `Array` and `Function` constructors considered as keywords but not the other native constructors (`Object`, `Date`, `RegExp`...)?

vkbansal:
@golmote because they are not constructors. They are static types just like string or boolean. It's just how typescript is.

Golmote:
Alright then ^^

LeaVerou:
> #444 is still pending.

Yes, I know, and it’s my fault :(

PR #489, opened 2015-02-01
Initial definition for stylus

vkbansal:
capture


components/prism-stylus.js
```diff
@@ -0,0 +1,61 @@
...
+	'url': /url\((["']?).*?\1\)/gi,
+	'variable': /^\s*([\w-]+)(?=\s*[+-\\]?=)/gm,
+	'string': /("|')(\\\n|\\?.)*?\1/g,
+	'important': /\B!important\b/gi,
+	'hexcode': /#[\da-f]{3,6}/gi,
+	'entity': /\\[\da-f]{1,8}/gi,
+	'number': /\d+\.?\d*%?/g,
+	'selector': [
+		{
+			pattern: /:{1,2}after|before|first-letter|first-line|selection/g,
```
Golmote: I think you forgot parentheses around the keywords alternatives.
Also, `:{1,2}` could be written `::?`.

components/prism-stylus.js
```diff
@@ -0,0 +1,61 @@
...
+			alias: 'pseudo-element'
+		},{
+			pattern: /:(?:active|checked|disabled|empty|enabled|first-child|first-of-type|focus|hover|in-range|invalid|lang|last-child|last-of-type|link|not|nth-child|nth-last-child|nth-last-of-type|nth-of-type|only-of-type|only-child|optional|out-of-range|read-only|read-write|required|root|target|valid|visited)(?:\(.*\))?/g,
+			alias:'pseudo-class'
```
Golmote: Is it important to distinguish between pseudo-elements and pseudo-classes?
It would probably be more future-proof to match them both with a more generic pattern like `/::?[\w-]+(?:\(.*\))?/` or something.

I'd like to hear @LeaVerou's opinion on this.

vkbansal:
References:
- https://developer.mozilla.org/en-US/docs/Web/CSS/pseudo-elements
- https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes

components/prism-stylus.js
```diff
@@ -0,0 +1,61 @@
...
+		},{
+			pattern: /(\[[\w-]+?\s*[*~|$^]?(?:=\s*\S+)?\])/g,
```
Golmote: Why did you make the first pattern `[\w-]+` ungreedy? It can only be follow by an operator or a closing square bracket so it should be safe to keep it greedy.

What about selectors like `input[type]`? All the value-related part should be made optional, I think.

In the part for the value, `\S+` is greedy. Couldn't it be a problem with selectors like `div[foo=bar][baz=bar]`?

Finally, the parentheses around the whole regexp aren't needed.

components/prism-stylus.js
vkbansal: Keywords seems fine to me

components/prism-stylus.js
```diff
@@ -0,0 +1,61 @@
...
+			inside: {
+				"attr-name":
+				{
+					pattern: /(\[)([\w-]+)(?=\s*[~$^|=]{0,2})/g,
```
Golmote: Parentheses around `[\w-]+` are not needed.

Also, I think you forgot the `*=` operator. You should probably keep them always in the same order, for clarity.

components/prism-stylus.js
```diff
@@ -0,0 +1,61 @@
...
+					lookbehind: true
+				},
+				"punctuation": /\[|\]/g,
+				"operator": /[=$~|^]/g,
```
Golmote: You forgot the `*` again.

I think it would be better to match the whole selector instead of each characters separately: `[*~|$^]?=`

components/prism-stylus.js
```diff
@@ -0,0 +1,61 @@
...
+		{
+			pattern: /\.[a-z-]+/i,
+			alias: 'class'
+		},
+		{
+			pattern: /#[a-z-]+/i,
+			alias: 'id'
+		},
+		{
+			pattern: /\b(html|head|title|base|link|meta|style|script|noscript|template|body|section|nav|article|aside|h[1-6]|header|footer|address|main|p|hr|pre|blockquote|ol|ul|li|dl|dt|dd|figure|figcaption|div|a|em|strong|small|s|cite|q|dfn|abbr|data|time|code|var|samp|kbd|sub|sup|i|b|u|mark|ruby|rt|rp|bdi|dbo|span|br|wbr|ins|del|image|iframe|embed|object|param|video|audio|source|track|canvas|map|area|sv|math|table|caption|colgroup|col|tbody|thead|tfoot|tr|td|th|form|fieldset|legeng|label|input|button|select|datalist|optgroup|option|textarea|keygen|output|progress|meter|details|summary|menuitem|menu)\b/g,
```
Golmote: Again, listing all the elements is not very future-proof. Isn't there a way to match them more generically?

components/prism-stylus.js
```diff
@@ -0,0 +1,61 @@
...
+	'url': /url\((["']?).*?\1\)/gi,
+	'variable': /^\s*([\w-]+)(?=\s*[+-\\]?=)/gm,
+	'string': /("|')(\\\n|\\?.)*?\1/g,
+	'important': /\B!important\b/gi,
+	'hexcode': /#[\da-f]{3,6}/gi,
+	'entity': /\\[\da-f]{1,8}/gi,
+	'number': /\d+\.?\d*%?/g,
+	'selector': [
+		{
+			pattern: /::?after|before|first-letter|first-line|selection/g,
```
Golmote: What I meant about the parentheses is this:

` /::?(?:after|before|first-letter|first-line|selection)/g`

components/prism-stylus.js
vkbansal: This not possible in case of stylus as far as I know. You can think of a better way. 

Also how is it different from listing keywords?

Golmote: > Also how is it different from listing keywords?

It's not. If keywords could be matched generically, I would also ask them to be ^^
If it is not possible to match elements generically in Stylus, then it's fine.

components/prism-stylus.js
```diff
@@ -0,0 +1,61 @@
...
+	'number': /\d+\.?\d*%?/g,
+	'selector': [
+		{
+			pattern: /::?(after|before|first-letter|first-line|selection)/g,
+			alias: 'pseudo-element'
+		},{
+			pattern: /:(?:active|checked|disabled|empty|enabled|first-child|first-of-type|focus|hover|in-range|invalid|lang|last-child|last-of-type|link|not|nth-child|nth-last-child|nth-last-of-type|nth-of-type|only-of-type|only-child|optional|out-of-range|read-only|read-write|required|root|target|valid|visited)(?:\(.*\))?/g,
+			alias:'pseudo-class'
+		},{
+			pattern: /\[[\w-]+?\s*[*~|$^]?(?:=\s*\S+)?\]/g,
```
Golmote: Would you mind moving the `|` after the `^` so that it's coherent with the other usages below?

components/prism-stylus.js
```diff
@@ -0,0 +1,61 @@
+Prism.languages.stylus = {
+	'comment': {
+		pattern: /(^|[^\\])(\/\*[\w\W]*?\*\/|\/\/.*?(\r?\n|$))/g,
+		lookbehind: true
+	},
+	'keyword': /(px|r?em)\b|\b(is|defined|not|isnt|and|or|unless|for|in)\b/g,
```
Golmote: What about all the other CSS units? (http://www.w3.org/TR/css-values/#lengths chapters 5 and 6)

vkbansal:
@LeaVerou Fixed merge conflict

components.js
```diff
@@ -302,7 +302,11 @@ var components = {
 			"title": "F#",
 			"require": "clike",
 			"owner": "simonreynolds7"
-		}
+		},
+        "stylus" : {
+            "title": "Stylus",
+            "owner": "vkbansal"
+        }
```
Golmote: Please be careful about the indentation here. You must use tabs.

vkbansal:
@LeaVerou Please review this too

LeaVerou:
Isn’t there anything of value this can borrow from the CSS definition? I feel that a lot of stuff is repeated, but it might just be my impression.

vkbansal:
@LeaVerou That was what I tried at first. But, it didn't workout.

LeaVerou:
Ok, merged, though I have my doubts about the CSS thing.

Anyway, thanks for contributing!!

vkbansal:
Thanks! I'm open to suggestions.

Issue #447, opened 2014-12-25
f sharp support


radutoev:
Hey - it would be really cool (for me especially :) ) if you added f# support. 
Thanks.

Golmote:
Fixed by #491

PR #491, opened 2015-02-06
Added F# support and example

simonreynolds7:
Added support for F# and included example page


components/prism-fsharp.js
```diff
@@ -0,0 +1,16 @@
+Prism.languages.fsharp = Prism.languages.extend('clike', {
+	'comment': [
+		{
+			pattern: /(^|[^\\])\(\*[\w\W]*?\*\)/g,
+			lookbehind: true
+		},
+		{
+			pattern: /(^|[^\\:])\/\/.*?(\r?\n|$)/g,
```
Golmote: Is the last group needed? Do we want to consume the line feed?

components/prism-fsharp.js
```diff
@@ -0,0 +1,16 @@
...
+			lookbehind: true
+		}
+	],
+	'keyword': /\b(abstract|and|as|assert|base|begin|class|default|delegate|do|done|downcast|downto|elif|else|end|exception|extern|false|finally|for|fun|function|global|if|in|inherit|inline|interface|internal|lazy|let|let!|match|member|module|mutable|namespace|new|not|null|of|open|or|override|private|public|rec|return|return!|select|static|struct|then|to|true|try|type|upcast|use|use!|val|void|when|while|with|yield|yield!|asr|land|lor|lsl|lsr|lxor|mod|sig|atomic|break|checked|component|const|constraint|constructor|continue|eager|event|external|fixed|functor|include|method|mixin|object|parallel|process|protected|pure|sealed|tailcall|trait|virtual|volatile)\b/g,
+	'string': /@?("""|"|')((\\|\n)?.)*?\1(B)?/g,
+	'preprocessor': /^\s*#.*/gm,
+	'number': /\b-?(0x)?\d*\.?\d+(y|uy|s|us|l|ul|un|L|UL|F|f|lf|LF|I|M|m)?\b/g
```
Golmote: Shouldn't you allow hexadecimal characters too? `[\da-fA-F]`?

Golmote:
Thanks for submitting this.

(Note for later: this fixes #447 )

components/prism-fsharp.js
simonreynolds7: Good point. I've added support for hexadecimal characters

components/prism-fsharp.js
LeaVerou: We most certainly don’t. Think of CSS styling that adds a background too. This should be converted to a lookahead. Thanks for catching that Golmote!

simonreynolds7: The last group matches the pattern found in prism-clike.js for single line comments
F# uses the same syntax for single line comments, but the comment property had to copied across as F# uses (* multi-line comment*) instead of the usual C /* multi-line comment */

LeaVerou: I just checked and we are indeed using this in clike. This should be fixed there too…

Golmote: Even a lookahead should normally not be needed. `.*` should do the job, right?

components/prism-fsharp.js
```diff
@@ -0,0 +1,16 @@
...
+		},
+		{
+			pattern: /(^|[^\\:])\/\/.*?(\r?\n|$)/g,
+			lookbehind: true
+		}
+	],
+	'keyword': /\b(abstract|and|as|assert|base|begin|class|default|delegate|do|done|downcast|downto|elif|else|end|exception|extern|false|finally|for|fun|function|global|if|in|inherit|inline|interface|internal|lazy|let|let!|match|member|module|mutable|namespace|new|not|null|of|open|or|override|private|public|rec|return|return!|select|static|struct|then|to|true|try|type|upcast|use|use!|val|void|when|while|with|yield|yield!|asr|land|lor|lsl|lsr|lxor|mod|sig|atomic|break|checked|component|const|constraint|constructor|continue|eager|event|external|fixed|functor|include|method|mixin|object|parallel|process|protected|pure|sealed|tailcall|trait|virtual|volatile)\b/g,
+	'string': /@?("""|"|')((\\|\n)?.)*?\1(B)?/g,
+	'preprocessor': /^\s*#.*/gm,
+	'number': /\b-?(0x)?(\d[a-fA-F]?)*\.?((\d\.?[a-fA-F]?)(y|uy|s|us|l|ul|un|L|UL|F|f|lf|LF|I|M|m))?\b/g
```
Golmote: I'm not sure I get this new pattern properly. Why are there two dots? Can hexadecimal numbers use a floating point syntax in F#? If not, I guess `\d*\.?[\da-fA-F]+` should be enough.
Maybe you should add a "Numbers" part in the example file so that you can check they are highlighted properly. (Maybe check [those listed there](https://msdn.microsoft.com/en-us/library/dd233193.aspx), for example?)

components/prism-fsharp.js
LeaVerou: Ah yeah, even better!

components/prism-fsharp.js
simonreynolds7: It was getting a bit complex as F# floating point numbers can be declared by ending a number with a dot
I've instead split the numbers regex to handle different scenarios as per commit message. Good idea with the Numbers section in the example file, I've included that as well

components/prism-fsharp.js
```diff
@@ -0,0 +1,29 @@
...
+		{
+			pattern: /\b-?(0b((0|1))*(y|uy)?)\b/g
+		},
+		{
+			pattern: /\b-?(\d*\.(f|\d)?\d*(f|F|m|M|((E|e)(\+|-)?\d*))?)\b/g
+		},
+		{
+			pattern: /\b-?(\d{1,}(y|uy|s|us|l|u|ul|L|UL|I)?)\b/g
+		}
+	]
```
Golmote: I think you should simplify the 4 patterns here:
1st: `((\d*)|([a-fA-F]*))*` => `[\da-fA-F]*`. And there seems to be a pair of parentheses that can be removed.
2nd: `((0|1))*` => `[01]*`. There is also a pair of parentheses that's not needed.
3rd: I guess the `f` in the middle is for handling the numers ending with a dot? What about... `2.F` or `2.m`? Are they valid numbers in F#? Furthermore, `(E|e)` should be simplified as `[Ee]` and `f|F|m|M` as `[fFmM]`. And still a pair of unneeded parentheses.
4th: `\d{1,}` => `\d+`. An the pair of unneeded parentheses can be removed.

Try to keep the patterns as short as possible. This will actually make them more readable. Counting the parentheses is probably the hardest part in understanding a regexp ^^

examples.js
```diff
@@ -6,7 +6,7 @@
 
 var examples = {};
 
-var treeURL = 'https://api.github.com/repos/LeaVerou/prism/git/trees/gh-pages?recursive=1';
+var treeURL = 'https://api.github.com/repos/simonreynolds7/prism/git/trees/fsharp-support?recursive=1';
```
Golmote: Woops, this should probably not be committed.

examples/prism-fsharp.html
```diff
@@ -0,0 +1,95 @@
...
+4.14
+2.3E+32
+2.3e+32
+2.3e32
+0x0000000000000000LF
+//BigInt
+9999999999999999999999999999I
+//Decimal
+0.7833M
+0.7833m
```
Golmote: I don't see the numbers you were telling me about earlier. The ones that end with a dot?

components/prism-fsharp.js
```diff
@@ -0,0 +1,29 @@
...
+			pattern: /(^|[^\\])\(\*[\w\W]*?\*\)/g,
+			lookbehind: true
+		},
+		{
+			pattern: /(^|[^\\:])\/\/.*/g,
+			lookbehind: true
+		}
+	],
+	'keyword': /\b(abstract|and|as|assert|base|begin|class|default|delegate|do|done|downcast|downto|elif|else|end|exception|extern|false|finally|for|fun|function|global|if|in|inherit|inline|interface|internal|lazy|let|let!|match|member|module|mutable|namespace|new|not|null|of|open|or|override|private|public|rec|return|return!|select|static|struct|then|to|true|try|type|upcast|use|use!|val|void|when|while|with|yield|yield!|asr|land|lor|lsl|lsr|lxor|mod|sig|atomic|break|checked|component|const|constraint|constructor|continue|eager|event|external|fixed|functor|include|method|mixin|object|parallel|process|protected|pure|sealed|tailcall|trait|virtual|volatile)\b/g,
+	'string': /@?("""|"|')((\\|\n)?.)*?\1(B)?/g,
```
Golmote: You should remove the unneeded parentheses around the `B`.

examples.js
simonreynolds7: Woops indeed, reversing this

examples/prism-fsharp.html
simonreynolds7: Added now

components/prism-fsharp.js
simonreynolds7: Done

components/prism-fsharp.js
simonreynolds7: Thanks for the suggestions

components/prism-fsharp.js
```diff
@@ -0,0 +1,29 @@
...
+			pattern: /(^|[^\\:])\/\/.*/g,
+			lookbehind: true
+		}
+	],
+	'keyword': /\b(abstract|and|as|assert|base|begin|class|default|delegate|do|done|downcast|downto|elif|else|end|exception|extern|false|finally|for|fun|function|global|if|in|inherit|inline|interface|internal|lazy|let|let!|match|member|module|mutable|namespace|new|not|null|of|open|or|override|private|public|rec|return|return!|select|static|struct|then|to|true|try|type|upcast|use|use!|val|void|when|while|with|yield|yield!|asr|land|lor|lsl|lsr|lxor|mod|sig|atomic|break|checked|component|const|constraint|constructor|continue|eager|event|external|fixed|functor|include|method|mixin|object|parallel|process|protected|pure|sealed|tailcall|trait|virtual|volatile)\b/g,
+	'string': /@?("""|"|')((\\|\n)?.)*?\1B?/g,
+	'preprocessor': /^\s*#.*/gm,
+	'number': [
+		{
+			pattern: /\b-?(0x[\da-fA-F]*.(un|LF)?)\b/g
```
Golmote: Why the dot? Is it supposed to be an actual dot or any character? I don't get this.
Also, I think you should use `+` instead of `*` since we want at least one char.
There are still unneeded parentheses around the whole number.
Finally, [this page on the MSDN](https://msdn.microsoft.com/en-us/library/dd233193.aspx) show an example using `lf` in lowercase, that is not handled here.

components/prism-fsharp.js
```diff
@@ -0,0 +1,29 @@
...
+		},
+		{
+			pattern: /\b-?(0b[01]*(y|uy)?)\b/g
```
Golmote: Here again, I think we need a `+` instead of the `*`.
And you can remove the parentheses around the whole number.

components/prism-fsharp.js
```diff
@@ -0,0 +1,29 @@
...
+		},
+		{
+			pattern: /\b-?(\d*\.(\d*)?([fFmM]|[Ee](\+|-)?\d*)?)\b/g
```
Golmote: If I read this properly, I think this pattern would match a single dot.

Furthermore:
* Regarding the part `(\d*)?`, why the question mark? The `*` already makes it optional. Also note that if you remove the question mark, you should remove the parentheses too.
* `(\+|-)?` can be replaced with `[+-]?`
* And you can remove the unneeded parentheses around the whole number.

components/prism-fsharp.js
```diff
@@ -0,0 +1,29 @@
...
+		},
+		{
+			pattern: /\b-?(\d+.([uy]|[us]|[ul]|[UL]|I)?)\b/g
```
Golmote: Why the dot? Is it supposed to be an actual dot or any character?

Golmote:
Concretely I think the number patterns should be as followed:

```javascript
	[
		{
			pattern: /\b-?0x[\da-fA-F]+(un|lf|LF)\b/g
		},
```
(at least one hexadecimal char and mandatory suffix)
```javascript
		{
			pattern: /\b-?0b[01]+(y|uy)\b/g
		},
```
(at least one binary char and mandatory suffix)
```javascript
		{
			pattern: /\b-?(\d+\.|\d*\.?\d+)([fFmM]|[eE][+-]?\d+)?\b/g
		},
```
(a number followed by a dot alone or a decimal number, followed by optional suffix or optional exponential notation)
```javascript
		{
			pattern: /\b-?\d+(y|uy|s|us|l|u|ul|L|UL|I)?\b/g
		}
	]
```
(a number followed by optional suffix)

Do you think this handles all the cases?

simonreynolds7:
The patterns you suggest look good, better than what I could write
The only changes I made where to make the suffixes for binary and hexadecimal numbers optional as they are valid ints

Golmote:
Great!

This looks good to me, @LeaVerou.

LeaVerou:
Why are there objects with just a pattern attribute instead of just regexes? The object syntax is for when you need to set more attributes than just the pattern...

simonreynolds7:
Whoops, my bad. I've changed them to a simple array as per the documentation

LeaVerou:
LGTM, merging.
As a minor comment, the array would be more readable if every regex was on separate lines.

PR #498, opened 2015-02-18
Make use of method `_.util.type`.

ewaldgrusk:
The title says it all :smiley:


LeaVerou:
Thanks! Wow, how on Earth had nobody noticed this before?

PR #501, opened 2015-02-25
Fixed repository url in examples

hason:


Golmote:
Thanks!

LeaVerou:
Doesn't it have a redirect? I don't think we should be chasing old URLs if so.

Golmote:
It didn't look like it had. The website was broken since the change. I had to change the url on the Download page too.

Issue #390, opened 2014-11-02
File Highlight and dynamically loaded content


morkro:
Hey, 

first of all: great work here!

I got a documentation page where I load some views via AJAX. I know that I must call `Prism.highlightAll();` when I want to get Prism working on dynamically loaded content. But somehow this doesn't work with all the additional plugins like _File Highlight_ or _Line Highlight_. They just don't apply.

Is this on purpose, a bug or what can I do to get them also running on dynamically loaded content? :)

hiulit:
Hey @morkro can you try it? It should be fixed now.

Golmote:
I consider this issue was fixed by #514.

PR #514, opened 2015-02-26
Added fileHighlight function

hiulit:
This is what we were discussing on this issue https://github.com/PrismJS/prism/issues/390

I'm no JavaScript ninja so I don't even know if what I made is correct, but it works for me and it let's me call this function whenever I need it, so I can use it when loading dinamic content.

### DIFF
--- a/plugins/file-highlight/prism-file-highlight.js
+++ b/plugins/file-highlight/prism-file-highlight.js
@@ -1,4 +1,4 @@
-(function(){
+var fileHighlight = (function fileHighlight(){
 
 if (!self.Prism || !self.document || !document.querySelector) {
 	return;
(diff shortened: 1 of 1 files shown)


Golmote:
I'd prefer if the variable was not declared globally. Using it as a property of the Prism object would be better IMO (`Prism.fileHighlight(...)`).

hiulit:
I don't know if I feel capable to do the task... As I mentioned earlier, I'm no JavaScript ninja :P

If you think that would be useful, be my guest, or have someone else do it, for the sake of Prism! Hehe!

Golmote:
What I'm asking for is in fact just a little change:

Instead of doing this:
```javascript
var fileHighlight = (function fileHighlight(){
    if (!self.Prism || !self.document || !document.querySelector) {
        return;
    }

    var Extensions = /* ...
        Plugin's code...
    ... */

    return fileHighlight;
}());
```

I'm suggesting you do this:
```javascript
(function () {
    if (!self.Prism || !self.document || !document.querySelector) {
        return;
    }

    // Declare it as a property of the Prism object
    self.Prism.fileHighlight = function() {

        var Extensions = /* ...
            Plugin's code...
        ... */

    };

    // Call it immediately
    self.Prism.fileHighlight();
}());
```

LeaVerou:
Um, sorry for being thick, but what’s the purpose of this?

Golmote:
It's all explained in issue #390. The purpose is to be able to call fileHighlight dynamically...

hiulit:
Have I done it right? It's works for me.

Golmote:
This looks good to me now ;) Thanks for refactoring it.

Now it's up to @LeaVerou to decide whether to merge it or not.

LeaVerou:
Merged, thanks!

PR #517, opened 2015-03-01
Fix bug related to line-highlighting plugin with Coy theme

### LINKED ISSUE #318: Highlighted line not fixed when scrolling in COY theme on demo page
When scrolling in one of the code boxes with highlighted lines on http://prismjs.com/plugins/line-highlight/ (with Coy theme selected) the highlight stays fixed while it should scroll together with the targetted line. 

schermopname 20

schermopname 19

I was able to reproduce this on recent versions of FF,Chrome and IE. 

When trying the highlight plugin myself, I can't even get the highlighting to work (also Coy theme), but it is possible this is because of conflicting js/css with other plugins, so I'll investigate this further.

Golmote:
This is an attempt to fix #318.
This only solves the problem with the Line-highlight plugin used with Coy theme.
This does not solve the problem that occurs currently when using both the Light-highlight plugin and the Line-numbers plugin.

(The drop-shadow on the Coy theme, which implies the `<code>` is scrollable instead of the `<pre>` is a real pain in the neck.)

### DIFF
--- a/themes/prism-coy.css
+++ b/themes/prism-coy.css
@@ -215,3 +215,17 @@ pre[class*="language-"].line-numbers code {
 pre[class*="language-"].line-numbers .line-numbers-rows {
 	left: 0;
 }
+
+/* Plugin styles: Line Highlight */
+pre[class*="language-"][data-line] {
+	padding-top: 0;
+	padding-bottom: 0;
+	padding-left: 0;
+}
+pre[data-line] code {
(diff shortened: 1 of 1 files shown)


LeaVerou:
Merged, thanks!

Sorry for the delay, I've been getting tons of Github spam recently in another repo, and the legit notifications get lost in the trillions of spam ones. :(

PR #543, opened 2015-03-08
Tutorial link suggestion

9itish:


LeaVerou:
Um, I think you made a mistake with this PR and typed your message as the PR code. In any case, I wouldn’t be opposed to adding that tutorial as well (though it’s a bit small), but you need to send a PR with the change :)
Github lets you do this graphically, by pressing "Edit" on a file in the repo and then clicking "Send Pull Request". Perhaps this is what you did and it went wrong here?

Issue #469, opened 2015-01-12
Stack overflow encountered with custom build of prism.js when selecting Markdown + Markup language options


cookch10:
Given the following build URL:  http://prismjs.com/download.html?themes=prism&languages=markup+markdown&plugins=line-highlight+line-numbers+show-invisibles+autolinker+wpd+show-language

The generated prism.js file yields a stack overflow error when loaded on a web page:

![image](https://cloud.githubusercontent.com/assets/6051695/5708507/bb6bbce6-9a5f-11e4-9434-c74fe6fe5079.png)

Golmote:
@cookch10 It should be fixed now

vkbansal:
@Golmote Thanks for the quick fix!

cookch10:
@Golmote thanks for the fast follow-up.  Works without fail, now! 👍

LeaVerou:
Thanks for fixing this @Golmote!

PR #544, opened 2015-03-09
Add Textile language

Golmote:
Preview of the examples:

prism-textile


components/prism-textile.js
```diff
@@ -0,0 +1,283 @@
...
+	// Modifiers highlighting
+	Prism.languages.textile['phrase'].inside['block'].inside['modifier'].inside = Prism.util.clone(modifierTokens);
+	Prism.languages.textile['phrase'].inside['list'].inside['modifier'].inside = Prism.util.clone(modifierTokens);
+	Prism.languages.textile['phrase'].inside['table'].inside['modifier'].inside = Prism.util.clone(modifierTokens);
+	Prism.languages.textile['phrase'].inside['inline'].inside['modifier'].inside = Prism.util.clone(modifierTokens);
+	Prism.languages.textile['phrase'].inside['link'].inside['modifier'].inside = Prism.util.clone(modifierTokens);
+	Prism.languages.textile['phrase'].inside['image'].inside['modifier'].inside = Prism.util.clone(modifierTokens);
+
+	// Allow some nesting
+	Prism.languages.textile['phrase'].inside['inline'].inside['bold'].inside = {
```
LeaVerou: Wow, a lot of repetition here. Why not define one object and clone it?

Golmote:
@LeaVerou It should be more DRY now. Does it look good to you?

LeaVerou:
Is cloning really needed? So what if they refer to the same object? It’s not like it’s mutated at any point.

Golmote:
Well, I do not mutate it in any way but a plugin could. The last time I forgot to clone duplicated parts of the grammar, bad things happened (remember #469), so I thought prevention was better than cure.

PR #408, opened 2014-12-06
Add support for Yaml language

marcaube:
Hi Lea! Here's some code to add basic yml highlighting.

PR #546, opened 2015-03-11
Add YAML highlighter

hason:
Replaces #408


examples/prism-yaml.html
```diff
@@ -0,0 +1,109 @@
...
+<h2>Null and Boolean</h2>
+<pre><code>---
+A null: null
+A null: ~
+Also a null: # Empty
+Not a null: ""
+Booleans: [ true, True, false, FALSE ]
+</code></pre>
+
+<h2>Numbers a timestamps</h2>
```
Golmote: You meant "Numbers and timestamps", right?

components/prism-yaml.js
```diff
@@ -0,0 +1,44 @@
+Prism.languages.yaml = {
+	'scalar': {
+		pattern: /([\-:]\s*(![^\s]+)?[ \t]*[\|>])[ \t]*(?:(\n[ \t]+)[^\r\n]+)?(?:\3[^\r\n]+)*/,
```
Golmote: If I read this correctly, the `\3` refers to the group `(\n[ \t]+)` which is part of an optional group. Shoudn't the last bit `(?:\3[^\r\n]+)*` be included in the same optional group (as the `\3` is mandatory)?

components/prism-yaml.js
```diff
@@ -0,0 +1,44 @@
...
+		lookbehind: true,
+		alias: 'string'
+	},
+	'comment': /#[^\n]+/,
+	'key': {
+		pattern: /(\s*[:\-,\[\{\n?][ \t]*(![^\s]+)?[ \t]*)[^\n\{\[\]\},#]+?(?=\s*:\s)/i,
```
Golmote: There is no need to escape the braces `{` and `}` inside character classes.

Golmote: Also, is the `i` flag actually useful?

components/prism-yaml.js
```diff
@@ -0,0 +1,44 @@
...
+		lookbehind: true,
+		alias: 'atrule'
+	},
+	'directive': {
+		pattern: /(^|\n)[ \t]*%[^\n]+/,
+		lookbehind: true,
+		alias: 'important'
+	},
+	'datetime': {
+		pattern: /([:\-,\[\{]\s*(![^\s]+)?[ \t]*)(\d{4}-\d\d?-\d\d?(t|[ \t]+)\d\d?:\d{2}:\d{2}(\.\d*)?[ \t]*(Z|[-+]\d\d?(:\d{2})?)?|\d{4}-\d{2}-\d{2}|\d\d?:\d{2}(:\d{2}(\.\d*)?)?)(?=[ \t]*(\n|$|,|\]|\}))/i,
```
Golmote: Here again, no need to escape the curly brace in the first character class.

components/prism-yaml.js
```diff
@@ -0,0 +1,44 @@
...
+	},
+	'string': {
+		pattern: /([:\-,\[\{]\s*(![^\s]+)?[ \t]*)("(\\\\|\\"|[^"\\])*"|'(\\\\|\\'|[^'\\])*')(?=[ \t]*(\n|$|,|\]|\}))/,
+		lookbehind: true
+	},
+	'number': {
+		pattern: /([:\-,\[\{]\s*(![^\s]+)?[ \t]*)[\+\-]?(0x[\dA-Fa-f]+|0o[0-7]+|(\d+\.?\d*|\.?\d+)(e[\+\-]?\d+)?|\.inf|\.nan)[ \t]*(?=\n|$|,|\]|\})/i,
+		lookbehind: true
+	},
+	'tag': /![^\s]+/i,
```
Golmote: Is the `i` flag useful here?

components/prism-yaml.js
```diff
@@ -0,0 +1,44 @@
...
+	'important': /[&\*][\w]+/,
```
Golmote: The `*` does not need to be escaped inside a character class.

components/prism-yaml.js
```diff
@@ -0,0 +1,44 @@
...
+	'punctuation': /([:\[\]\{\}\-,\|>?]|---|\.\.\.)/
```
Golmote: The curly braces `{}` and the pipe `|` do not need to be escaped inside a character class.

hason:
@Golmote @LeaVerou What do you think?

components/prism-yaml.js
```diff
@@ -0,0 +1,44 @@
+Prism.languages.yaml = {
+	'scalar': {
+		pattern: /([\-:]\s*(![^\s]+)?[ \t]*[\|>])[ \t]*(?:(\n[ \t]+)[^\r\n]+(?:\3[^\r\n]+)*)/,
```
Golmote: I missed one here: the `|` does not need to be escaped inside character classes.

Golmote:
It looks good to me. Now it's up to @LeaVerou to decide.

LeaVerou:
LGTM, merged, thanks! :)

PR #556, opened 2015-03-20
Fixed Objective-C extend to be clike

joeworkman:


uranusjr:
Objective-C [is a strict superset of C](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/Introduction/Introduction.html), which is why I chose `c` as its dependency, instead of `clike`.

joeworkman:
Ha! You know what, I completely overlooked the fact that there was a C component. Not sure how I overlooked that about 5 times...

PR #557, opened 2015-03-21
Reorder languages in components.js (Fix #441)

### LINKED ISSUE #441: Reordering languages in components.js
Ordering them by time added doesn't work: It doesn't help in anything and it causes craploads of merge conflicts (essentially if I merge 1 PR with a new language, all others need to pull & merge conflicts to be merged).
I suggest ordering them either alphabetically or by language popularity. Thoughts?

Golmote:
(GitHub makes it look like a lot of change, but this is really just a reorder ^^)

Visually, I simply added a border to separate the "popular" languages from the others:

EDIT: Second commit fixes the ordering of Perl and PHP... ><

### DIFF
--- a/components.js
+++ b/components.js
@@ -41,57 +41,43 @@ var components = {
 			"title": "Markup",
 			"option": "default"
 		},
-		"twig": {
-			"title": "Twig",
-			"require": "markup",
-			"owner": "brandonkelly"
-		},
 		"css": {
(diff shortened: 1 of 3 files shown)


LeaVerou:
Nice one, thanks!
At some point we should do something better with all these languages, they’re so many now!! But that’s for the distant future…

PR #563, opened 2015-04-02
Don't force scrollbars to be shown unnecessarily

### LINKED ISSUE #559: Scrollbars are always enabled, even in single-line pieces of code
I downloaded prism and did not touch any of the code. When I enable it in my Ghost blog, all of the code blocks have scroll bars. I've noticed that this does not happen to other Ghost blogs. The features I've enabled are "line numbers" and and "show language". You can see an example of what I mean in this blog post: http://antoniojsantos.me/assigning-values-to-variables/. All the other posts in my blog also have this issue. I am using the Casper theme with my own modifications (see the changes in my [fork](https://github.com/FinestRocker/Casper) of the theme). If you need any more information I'll gladly provide it.

nauzilus:
The inner `code` element needs to handle scrolling, otherwise the shown language will scroll out of view (which is fixed to the outer `pre` container) if the user has to scroll to see more highlighted code.

Fixes #559 and potentially #382.

Demo at http://nauzilus-prism.azurewebsites.net/plugins/show-language/


Golmote:
LGTM. Merged.

PR #570, opened 2015-04-16
Remove `overflow: auto` from code element

### LINKED ISSUE #569: show-language conflicts with line-numbers
The show-language plugin sets `overflow: auto` on the `code` element. This causes the line numbers (which are bumped outside of the code element with a negative margin) to become invisible. The offending line is [here](https://github.com/PrismJS/prism/blob/gh-pages/plugins/show-language/prism-show-language.css#L4-5). I don't know why show-language needs `overflow: auto`: removing it fixes the problem<s>, but I'm sure there are edge cases I'm not considering</s>. Scrolling is supposed to happen on the `pre` element anyways, so this can be removed safely. I'll open a pull request.

dylanmackenzie:
Fixes #569. The pre element already has `overflow: auto` so the scrollbars will appear on it instead.

### DIFF
--- a/plugins/show-language/prism-show-language.css
+++ b/plugins/show-language/prism-show-language.css
@@ -2,7 +2,6 @@ pre[class*='language-'] {
 	position: relative;
 }
 pre[class*='language-'] > code[data-language] {
-	overflow: scroll;
 	max-height: 28em;
 	display: block;
 }


Golmote:
Poke @nauzilus, this is related to the change you made in #563. What do you think?

nauzilus:
Actually I think the problem was always there =(

As  mentioned in #563, this is to keep the language (in the generated content) stick to the top right of the `pre` element regardless of scrolling. But that doesn't seem to be practical without breaking things.

When I get home, I'll make a change to just put  the generated content on the `pre`,  and it'll just have to scroll, demonstrated here: http://nauzilus.net/show-lang-fix.html

nauzilus:
Submitted #571
Example: http://nauzilus-prism.azurewebsites.net/plugins/show-language

PR #571, opened 2015-04-17
Work nicely with line-numbers plugin

nauzilus:
Generated content on the `pre` means if the highlighted area needs to scroll, the language will scroll out of view with the code.
But that's better than completely breaking another plugin.


Golmote:
> But that's better than completely breaking another plugin.

Indeed.

PR #576, opened 2015-05-08
add dockerfile highlighting

JustinBeckwith:


components/prism-docker.js
```diff
@@ -0,0 +1,8 @@
+Prism.languages.docker = {
+	'keyword': {
+    pattern: /^\s*\b(ONBUILD|FROM|MAINTAINER|RUN|EXPOSE|ENV|ADD|COPY|VOLUME|USER|WORKDIR|CMD|ENTRYPOINT)\b\s/mi
+	},
+  'string': /("|')(\\\n|\\?.)*?\1/,
```
Golmote: Please be careful with the indentation. Prism uses tabs to indent, not spaces.

components/prism-docker.js
```diff
@@ -0,0 +1,8 @@
+Prism.languages.docker = {
+	'keyword': {
+    pattern: /^\s*\b(ONBUILD|FROM|MAINTAINER|RUN|EXPOSE|ENV|ADD|COPY|VOLUME|USER|WORKDIR|CMD|ENTRYPOINT)\b\s/mi
```
Golmote: The `\b` parts seems redundant with the `\s` parts (the latter implies the former). Also, do we really need to include those whitespaces in the match? What about something like:

```javascript
{
	pattern: /(^\s*)(?:ONBUILD|...|ENTRYPOINT)(?=\s)/mi,
	lookbehind: true
}
```

Finally, what about the [`LABEL` keyword](https://docs.docker.com/reference/builder/#label)?

examples/prism-dockerfile.html
```diff
@@ -0,0 +1,39 @@
...
+# Grab the source code and add it to the workspace.
+ADD . /go/src/github.com/JustinBeckwith/revel-appengine
+
+# Install revel and the revel CLI.
+RUN go get github.com/revel/revel
+RUN go get github.com/revel/cmd/revel
+</code></pre>
+
+
+<h2>Full example</h2>
```
Golmote: The "Instructions" and the "Full example" look more or less the same for me. I guess you could remove the "Instructions" one.
Also, I find the official example more complete than yours. https://docs.docker.com/reference/builder/#dockerfile-examples

Golmote:
Thanks for contributing! Would you mind taking a look at my comments?
Also, you need to rename the JS file from `prism-docker.js` to `prism-dockerfile.js`, since the component is `dockerfile`.

components/prism-docker.js
```diff
@@ -0,0 +1,8 @@
...
+	},
+  'string': /("|')(\\\n|\\?.)*?\1/,
+	'comment': /#[^\n]+/,
+  'punctuation': /([:[\]{}\-,|>?]|---|\.\.\.)/
```
Golmote: No need for the parentheses here.

components/prism-docker.js
```diff
@@ -0,0 +1,8 @@
+Prism.languages.docker = {
+	'keyword': {
+    pattern: /^\s*\b(ONBUILD|FROM|MAINTAINER|RUN|EXPOSE|ENV|ADD|COPY|VOLUME|USER|WORKDIR|CMD|ENTRYPOINT)\b\s/mi
+	},
+  'string': /("|')(\\\n|\\?.)*?\1/,
+	'comment': /#[^\n]+/,
```
Golmote: I think the `[^\n]+` could be replaced with `.+`. This would even add support for `\r`.

JustinBeckwith:
Thanks for the awesome feedback.  I updated the PR with all of the requested updates.  Let me know if there's anything else!

Golmote:
@JustinBeckwith Sorry for the delay. Your updates look good to me. You did not add the `LABEL` keyword though, is it intended?
@LeaVerou, if it looks good to you too, feel free to merge this. I'll run the gulp task to generate the minified file if you don't have the time to do so.

JustinBeckwith:
That was an oversight :) Thanks again!

egut:
Ohh exactly what I was looking for, any idea when it will be merged?

(As I'm currently using bower so its nifty if I don't have to modify files within bower components)

/BR

PR #583, opened 2015-05-23
Added Ruby string interpolation

westonganger:
The only case it doesnt handle is if the string interpolation is not inside a string

### DIFF
--- a/components/prism-ruby.js
+++ b/components/prism-ruby.js
@@ -21,7 +21,7 @@ Prism.languages.insertBefore('ruby', 'keyword', {
 });
 
 Prism.languages.ruby.string = {
-	pattern: /("|')(\\\n|\\?.)*?\1/,
+	pattern: /("|')(#\{[^}]+\}|\\\n|\\?.)*?\1/,
 	inside: {
 		'interpolation': {
 			pattern: /#\{[^}]+\}/,
--- a/components/prism-ruby.js
+++ b/components/prism-ruby.js
@@ -5,17 +5,33 @@
  * 		constant, builtin, variable, symbol, regex
  */
 Prism.languages.ruby = Prism.languages.extend('clike', {
-	'comment': /#[^\r\n]*(\r?\n|$)/,
-	'keyword': /\b(alias|and|BEGIN|begin|break|case|class|def|define_method|defined|do|each|else|elsif|END|end|ensure|false|for|if|in|module|new|next|nil|not|or|raise|redo|require|rescue|retry|return|self|super|then|throw|true|undef|unless|until|when|while|yield)\b/,
+        'comment': /#(?!\{[^\r\n]*?\})[^\r\n]*(\r?\n|$)/g,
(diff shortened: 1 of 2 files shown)


components/prism-ruby.js
```diff
@@ -5,17 +5,33 @@
  * 		constant, builtin, variable, symbol, regex
  */
 Prism.languages.ruby = Prism.languages.extend('clike', {
-	'comment': /#[^\r\n]*(\r?\n|$)/,
-	'keyword': /\b(alias|and|BEGIN|begin|break|case|class|def|define_method|defined|do|each|else|elsif|END|end|ensure|false|for|if|in|module|new|next|nil|not|or|raise|redo|require|rescue|retry|return|self|super|then|throw|true|undef|unless|until|when|while|yield)\b/,
+        'comment': /#(?!\{[^\r\n]*?\})[^\r\n]*(\r?\n|$)/g,
```
Golmote: Please use tabs to indent, not spaces.

Also, why the `g` flag everywhere? It is not used by Prism.

components/prism-ruby.js
```diff
@@ -5,17 +5,33 @@
...
+        pattern: /("|')(\\\n|\\?.)*?\1/,
+        inside: {
+          'interpolation': {
+            pattern: /#\{[^}]+\}/,
+            inside: {
+              'delimiter': {
+                pattern: /^#\{|\}$/,
+                alias: 'tag'
+              },
+              rest: Prism.languages.extend('ruby')
```
Golmote: You should use `Prism.util.clone` instead of `extend`.

components/prism-ruby.js
```diff
@@ -5,17 +5,33 @@
...
 		lookbehind: true
 	},
-	'variable': /[@$]+\b[a-zA-Z_][a-zA-Z_0-9]*[?!]?\b/,
-	'symbol': /:\b[a-zA-Z_][a-zA-Z_0-9]*[?!]?\b/
+	'variable': /[@$]+\b[a-zA-Z_][a-zA-Z_0-9]*[?!]?\b/g,
+	'symbol': /:\b[a-zA-Z_][a-zA-Z_0-9]*[?!]?\b/g
 });
+
+Prism.languages.ruby.string = {
+        pattern: /("|')(\\\n|\\?.)*?\1/,
```
Golmote: Again, be careful with the indentation. Prism uses tabs, not spaces.

Golmote:
I don't have time to test it right now, but I quickly reviewed the code. Please take a look at my comments ;)

Golmote:
I finally got some time to play around with your PR.
I think you could fix the second failure ("Double quoted strings inside of string interpolation") by changing the `string` pattern to `/("|')(#\{[^}]+\}|\\\n|\\?.)*?\1/`. What do you think?

LeaVerou:
@Golmote do you think this is good to merge?

Golmote:
Oh, I did not notice the last commit. So yes, it looks good to me now.

PR #587, opened 2015-06-01
Handlebars: support block names with dashes (or other special characters)

### LINKED ISSUE #586: Handlebars block name with dash
Looks like the color highlighting is not being correctly applied when the block name has a dash in it:

prismjs

apfelbox:
Hi,

as mentioned in #586 blocks with dashes should be supported.
As I was searching for a list of allowed characters in block names I couldn't find anything in the short time I was looking (`{{#this-is_ällowed✓-yep}} .. {{/this-is_ällowed✓-yep}}`).

[This hints](https://github.com/wycats/handlebars.js/blob/4bed826d0e210c336fb9e500835b1c1926562da5/lib/handlebars/base.js#L39) in the direction that any valid object key is allowed, but whitespace is obviously not allowed (as Handlebars couldn't determine whether this is still the block name or already an argument). So I went for `\S`, as basically only whitespace would error out (which was confirmed by my quick tests).

Fixes #586


miguelcobain:
Awesome!

PR #592, opened 2015-06-11
Updated test.html with support for testing the plugins

VitaliyR:
Separated PR from https://github.com/PrismJS/prism/pull/584 which is upgrading test.html so it can now test the plugins too.

Not the best solution is used, but it is working. Prism core object is being updated while any of plugin is loaded and for clearing it out from plugin's mixins we need to reload it. It's done via reattaching prism.js script file.


VitaliyR:
@Golmote I've merged your changes into this branch

VitaliyR:
@Golmote any updates?

Golmote:
I don't have much time right now to properly test this, for the moment.

VitaliyR:
@Golmote, sure, take your time. Maybe you have propositions how to make it better?

test.html
```diff
@@ -189,8 +365,17 @@ <h2>Test drive</h2>
...
+		textarea.textContent = savedCode;
+	}
 
-(textarea.oninput = function() {
+(textarea.onblur = function() {
+	if (window.localStorage){
+		localStorage.setItem('test_code', this.value);
+	}
+},
+textarea.oninput = function() {
```
mAAdhaTTah: Can we remove the `localStorage` related stuff and split that into a separate PR?

test.html
```diff
@@ -164,6 +281,65 @@ <h2>Test drive</h2>
...
+	$('#prism').remove();
+	$$('script[data-type="plugin"],link[data-type="plugin"]').forEach(function (script) {
+		script.remove();
+	});
+
+	// firstly loads prism core, after it - all selected plugins
+	return new Promise(function (resolve) {
+		$u.script('prism.js', function () {
+			this.setAttribute('id', 'prism');
+
```
mAAdhaTTah: We should `resolve` the `Promise` here and `return Promise.all([...])` within the `then` handler. It'll be a little cleaner.

test.html
```diff
@@ -164,6 +281,65 @@ <h2>Test drive</h2>
...
+		pluginPath = pluginPath.replace(/{id}/g, plugin);
+
+		$u.script(pluginPath + '.js', function() {
+			this.setAttribute('data-type', 'plugin');
+
+			var style = document.createElement('link');
+			style.setAttribute('rel', 'stylesheet');
+			style.setAttribute('href', pluginPath + '.css')
+			style.setAttribute('data-type', 'plugin');
+			document.body.appendChild(style);
```
mAAdhaTTah: Not all plugins have associated CSS, so we should check if it does first.

mAAdhaTTah:
@VitaliyR Our apologies on the delay! I know its been a while, but would you be interested in making the changes suggested here?

test.html
```diff
@@ -164,6 +281,65 @@ <h2>Test drive</h2>
...
+ * You can hit me for such solution, but there are no other way :)
+ */
+function reloadPlugins ()
+{
+	var selectedPlugins = $$('input[type="checkbox"]:checked', document.forms[0]);
+	selectedPlugins = selectedPlugins.map(function(plugin) { return plugin.value });
+
+	$('#prism').remove();
+	$$('script[data-type="plugin"],link[data-type="plugin"]').forEach(function (script) {
+		script.remove();
```
mAAdhaTTah: @Golmote @LeaVerou Removing this script tag doesn't undo what the script has done. I think we'd need to implement #1381 before we can do this. Any other ideas?

LeaVerou: We could use an iframe whose `srcset` is set dynamically for the demo :)

PS: Awwww, I wish we could remove `<script>` elements and have their effects undone, that would be so nicely reactive! :)

VitaliyR: Haha, yes, indeed. It was done (probably? 2 years has passed..) for more clean html structure, I guess.. But later I'm realoading Prism core js file - so it will be run again and we will receive clean object. 

Doing everything in iframes is better and nice idea also

test.html
VitaliyR: We can, but why? :)

test.html
VitaliyR: Indeed, but should I just add here `onerror` and silent the message in case css file is not found? I mean if it's not found - nothing critical will happen :)

test.html
VitaliyR: Something like:
```javascript
return new Promise(function (resolve) {
  $u.script('prism.js', function () { 
    this.setAttribute('id', 'prism');
    resolve();
  });
}).then(function() {
  return Promise.all(selectedPlugins.map());
});
```
?

mAAdhaTTah: Yeah, that's the idea.

test.html
mAAdhaTTah: It's unrelated to the main feature (testing plugins on the test page).

VitaliyR:
Oki, here're few changes.

* Promises fixes
* Check if plugin has css file and only then load it
* Removed localStorage stuff from this PR

RunDevelopment:
@mAAdhaTTah Is this going to get merged anytime soon?

mAAdhaTTah:
@RunDevelopment I'm not sure [this has been addressed](https://github.com/PrismJS/prism/pull/592#discussion_r194592133) (or even how to address it). Although I think the point of clobbering & replacing the global `Prism` object might be enough. We should test this before we merge, but I dig it.

test.html
```diff
@@ -164,12 +281,72 @@ <h2>Test drive</h2>
...
+ * Loads a plugin including it css file
+ *
+ * @param {string} pluginName the plugin to load
+ * @type {Promise} the promise which resolves as soon as plugin is loaded
+ */
+function loadPlugin (pluginName)
+{
+	var plugin = components.plugins[pluginName];
+	if (!plugin){
+		return;
```
mAAdhaTTah: I _think_ this should return `Promise.resolve()` to make the return value consistent?

test.html
```diff
@@ -164,12 +281,72 @@ <h2>Test drive</h2>
...
+
+	var pluginPath = components.plugins.meta.path;
+
+	return new Promise(function (resolve) {
+		pluginPath = pluginPath.replace(/{id}/g, pluginName);
+
+		$u.script(pluginPath + '.js', function() {
+			this.setAttribute('data-type', 'plugin');
+
+			var hasCss = !!!plugin.noCSS;
```
mAAdhaTTah: You can just use `!` here.

test.html
VitaliyR: Agree, updating

test.html
VitaliyR: What about if an object doesn't have `noCSS`? !undefined will be true - which is wrong

VitaliyR:
FYI: Updated PR and replied to your comment

test.html
RunDevelopment: But `!!!undefined` is also `true`

VitaliyR: Omg, indeed, sorry about that. So it will be `!!plugin.noCSS`. So for undefined (when noCSS attribute isn't provided) it will be false, while for true - it will be true

mAAdhaTTah:
@RunDevelopment We have a PR to add line-numbers to the test page. Do we / can we combine it with this?

mAAdhaTTah:
(Didn't mean to close)

RunDevelopment:
@mAAdhaTTah I don't see why we would need #1554 after this.

PR #595, opened 2015-06-12
adding vhdl language

a-rey:
#### Hello! 

I use some hardware description languages in a website of mine and was sad to see they are not supported by Prism. Here is my proposed Prism support for the [VHDL language](https://en.wikipedia.org/wiki/VHDL). Let me know if there are any issues!

sample code below borrowed from [here](http://www.csee.umbc.edu/portal/help/VHDL/samples/hello_proc.vhdl)

screen shot 2015-06-12 at 4 00 58 pm


components/prism-vhdl.js
```diff
@@ -0,0 +1,24 @@
+Prism.languages.vhdl = {
+	'comment': /(--[^\r\n]+)/,
```
Golmote: Why not `.+` instead of `[^\r\n]+`?

components/prism-vhdl.js
```diff
@@ -0,0 +1,24 @@
...
+	// support for all logic vectors
+	'vhdl-vectors': {
+		'pattern': /(\b[oOxXbB]"[\dA-Fa-f_]+"|"[UuXx01ZzWwLlHh-]+")/,
```
Golmote: It looks like all letters are uppercase or lowercase: you could use the `i` modifier to simplify the regexp.

components/prism-vhdl.js
```diff
@@ -0,0 +1,24 @@
...
+		'alias': 'number'
+	},
+	'string': /(")(\\\n|\\?.)*?"/,
```
Golmote: No need for the first couple of parentheses.

components/prism-vhdl.js
```diff
@@ -0,0 +1,24 @@
...
+	'constant': /\b(use|library)\b/i,
+	// support for predefined attributes included
+	'keyword': /\b(access|after|alias|all|architecture|array|assert|attribute|begin|block|body|buffer|bus|case|component|configuration|constant|disconnect|downto|else|elsif|end|entity|exit|file|for|function|generate|generic|group|guarded|if|impure|in|inertial|inout|is|label|library|linkage|literal|loop|map|new|next|null|of|on|open|others|out|package|port|postponed|procedure|process|pure|range|record|register|reject|report|return|select|severity|signal|shared|subtype|then|to|transport|type|unaffected|units|until|use|variable|wait|when|while|with|'base|'left|'right|'high|'low|'ascending|'image|'value|'pos|'val|'succ|'pred|'leftof|'rightof|'left|'left|'right|'right|'high|'high|'low|'low|'range|'range|'reverse_range|'reverse_range|'length|'length|'ascending|'ascending|'delayed|'stable|'stable|'quiet|'quiet|'transaction|'event|'active|'last_event|'last_active|'last_value|'driving|'driving_value|'simple_name|'instance_name|'path_name)\b/i,
```
Golmote: I think there are a lot of duplicates here (unless I'm missing something...?): `'left`, `'right`, etc.

components/prism-vhdl.js
```diff
@@ -0,0 +1,24 @@
...
+	'boolean': /\b(true|false)\b/i,
+	'function': {
+		// support for operator overloading included
+		pattern: /([a-z0-9_]+|"\S+")\(/i,
+		inside: {
+			punctuation: /\(/
+		}
+	},
+	// decimal, based, physical, and exponential numbers supported
+	'number': /('[UuXx01ZzWwLlHh-]'|\b\d+([_.]+)?(#[\dA-Fa-f_.]+#)?([eE][-+]?\d+)?)/,
```
Golmote: No need for the enclosing parentheses.

Why is the `'[UuXx01ZzWwLlHh-]'` part listed here? Couldn't it be handled in the `vhdl-vectors` pattern above?

Also, I think `([_.]+)?` can be written `[_.]*`.

Finally, if all letters are uppercase or lowercase (it looks like it's the case), you can use the `i` flag to simplify the regexp.

components/prism-vhdl.js
```diff
@@ -0,0 +1,24 @@
...
+	'operator': /[-+]{1}|<=?|>=?|=|:=|&|\*|\/|\b(abs|not|mod|rem|sll|srl|sla|sra|rol|ror|and|or|nand|xnor|xor|nor)\b/i,
```
Golmote: The `{1}` is not needed.

I think `:=` should be listed before `=`, otherwise it won't get highlighted properly... right?

`&`, `\*` and `\/` could be included in the first character class, like so: `[-+&*\/]`. You could even add `=` if you reorder the parts so that the character class is after the other operators that use `=`.

Golmote:
Thanks for contributing!

Please take a look at my comments ;)

a-rey:
Thank you for your comments! 

I made the changes you brought up. Let me know if you see anything else.

LeaVerou:
> I left the `[01uxzwlh-]` part in the `number` pattern and not in the `vhdl-vectors` pattern to avoid the issue of a `'0'` inside of a string (example: `"blah blah '1' blah"`) being highlighted before strings had their chance.

Not sure what you mean here exactly, but strings are processed before numbers, since the `string` token comes earlier in your definition.

Golmote:
Alright, this makes sense. Since vhdl-vectors are processed before strings, he wanted to prevent those single-quoted numbers from breaking the highlighting of strings.

components/prism-vhdl.js
```diff
@@ -0,0 +1,24 @@
+Prism.languages.vhdl = {
+	'comment': /--.+/,
+	// support for all logic vectors
+	'vhdl-vectors': {
+		'pattern': /(\b[oxb]"[\da-f_]+"|"[01uxzwlh-]+")/i,
```
Golmote: No need for the enclosing parentheses here.

components/prism-vhdl.js
```diff
@@ -0,0 +1,24 @@
...
+	'boolean': /\b(true|false)\b/i,
+	'function': {
+		// support for operator overloading included
+		pattern: /([a-z0-9_]+|"\S+")\(/i,
+		inside: {
+			punctuation: /\(/
+		}
+	},
+	// decimal, based, physical, and exponential numbers supported
+	'number': /'[01uxzwlh-]'|\b\d+[_.]*(#[\da-f_.]+#)?([e][-+]?\d+)?/i,
```
Golmote: No need for the brackets in `[e]`

Golmote:
I added a couple of comments. Apart from them, this looks good to me.

a-rey:
Made the requested changes. Thanks again!

LeaVerou:
LGTM. Please add a minified version? Also, examples in the [examples page](http://prismjs.com/examples.html) would be nice, but not required.

Golmote:
If needed, I can merge, build the minified version and add the example given in the first message of this PR, when I get back from work.

a-rey:
minified and example ready!

Golmote:
This looks great, thanks. As it looks good to @LeaVerou too, I'll merge now.

Issue #111, opened 2013-06-01
Support ES6 syntax


DavidBruant:
Among other things, ES6 added [arrow function syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/arrow_functions).
````js
setTimeout(() => {console.lol()}, 200)
````

Currently, Prism highlights the arrow as if it was the inequality operator.

loganfranken:
I put together a CodePen that shows the current level of support for ES6 features in Prism:
http://codepen.io/loganfranken/pen/aOwMXw

It actually looks like the arrow function syntax is now supported (wahoo!), but there are just a few missing pieces:

- [ ] Octal Literals (`0o543`)
- [ ] Binary Literals (`0b101011`)
- [ ] Template Strings
- [ ] Regex `y` Flag
- [ ] Regex `u` Flag
- [ ] `of` Keyword (used in `for..of` loops)
- [ ] `as` Keyword (used w/ `import` statement)
- [ ] `from` Keyword (used w/ `import` statement)

apfelbox:
For point 1 & 2 see #597.

Golmote:
Fat arrow and spread operators are highlighted, indeed, but not as whole tokens. `=>` gets highlighted because of `=` and `>` being operators, and `...` because of `.` being punctuation.
I wonder if we should add them on their own.

Golmote:
Just added support for template strings (04f72b1495278dccea1d6d994fbc885d5f841692). I updated the examples with a section illustrating those ES6 features:

![http://puu.sh/ioDbf/1642341e7c.png](http://puu.sh/ioDbf/1642341e7c.png)

I also updated the known failures:

![http://puu.sh/ioDdB/191d618d6c.png](http://puu.sh/ioDdB/191d618d6c.png)

loganfranken:
You're all so awesome!

I think we're done here, right? We can close this issue?

(Unless you wanted to keep it open to address the edge cases you described, @Golmote?)

Golmote:
No, I think we can close it for now. We'll think about those spread and fat arrow operators later, since they are currently highlighted, though not as semantically as expected.

darenju:
Any work on that? I understand these operators are already highlighted, but a drawback is for people who use fonts with ligatures.

Since Prism renders those operators in two separate `span`s, the ligature can't be processed.

PR #596, opened 2015-06-14
Add Missing ES6 Keywords and Regex Flags

loganfranken:
Adds the following ES6 keywords and regex flags:

- Regex `y` Flag
- Regex `u` Flag
- `of` Keyword (used in `for..of` loops)
- `as` Keyword (used w/ `import` statement)
- `from` Keyword (used w/ `import` statement)

(Fixes a good chunk of #111)


Golmote:
Thanks for contributing!

PR #597, opened 2015-06-14
Support for JavaScript octal and binary numbers (in ES6)

apfelbox:
This will add support octal and binary numbers:

```js 
var a = 0o01234567;
var b = 0o101;
```

I also removed a duplicate negative sign for `Infinity`.

Fixes part of #111


apfelbox:
You can merge #596 before this one, so I can rebase my PR based on these javascript changes.

Golmote:
I just merged #596. Let me know when this PR is updated.

apfelbox:
So, i just rebased and squashed the changes in one commit.

Ready to be merged @Golmote

Golmote:
Sorry, can you do it again?

apfelbox:
@Golmote no problem, done.

loganfranken:
Hey @vkbansal, maybe I'm wrong, but it doesn't look like operators are highlighted as part of numbers with Prism currently:
http://codepen.io/loganfranken/pen/gpxaLe

So it would seem to fall in line that `-Infinity` shouldn't be highlighted as a number (if the minus operator in `-123` isn't highlighted as number, for example).

Totally let me know if I'm missing something, though!

apfelbox:
@vkbansal @loganfranken even easier than that. If you take a closer look at the braces in the regex, that only `Infinity` has a negative sign. That is because that whole match already has a negative sign at the beginning – and therefore the one in `Infinity` is duplicate.

2015-06-14 at 21 31

@loganfranken you are right, the regex matches it, but it isn't included in the token. Because of this it will be highlighted as operator later on.

loganfranken:
@apfelbox Ahh, thanks for clearing that up!

vkbansal:
@apfelbox thanks for the clarification :)

PR #601, opened 2015-06-19
Enhancements in Python component

aviaryan:
1. Highlighting was added for function declaration. 
2. Fixed a bug where variables like a,b,c....f where detected as numbers and thus highlighted.


Golmote:
Thanks for contributing!

PR #602, opened 2015-06-20
Swallow XHR exception in File Highlight Plugin

### LINKED ISSUE #152: Why only first error shown in file highlight?
Hi,

the file highlight plug-in only shows the first error on page.

It would be better to fail for each error, no?

loganfranken:
Fixes #152


Golmote:
I wasn't aware that the `send` method could throw exceptions. In which case does it happen? Could you provide on online example of this issue before and after the fix?

loganfranken:
Augh, sorry about this: I can no longer get this issue to occur.

Last night, I swore I was able to recreate @opatry's example from #152: basically, the first `xhr` request would fail, causing Prism to bomb out and not attempt the second `xhr` request, leaving the second code block blank.

But I can no longer reproduce it; in fact, it works perfectly in this Fiddle:
https://jsfiddle.net/eyu3xo7r/

(Maybe I passed out and had a very specific Prism-related dream last night)

I'll close this pull request given that I can't recreate the behavior.

PR #609, opened 2015-06-27
Add Keyman language highlighting

mcdurdin:
Adds Keyman / KMFL language highlighting, as documented at http://help.keyman.com/developer/9.0/docs/

### DIFF
--- a/components/prism-keyman.js
+++ b/components/prism-keyman.js
@@ -8,7 +8,7 @@ Prism.languages.keyman = {
 		/&(baselayout|bitmap|capsononly|capsalwaysoff|shiftfreescaps|copyright|ethnologuecode|hotkey|includecodes|keyboardversion|kmw_embedcss|kmw_embdjs|kmw_helpfile|kmw_helptext|kmw_rtl|language|layer|layoutfile|message|mnemoniclayout|name|oldcharposmatching|platform|targets|version|visualkeyboard|windowslanguages)\b/i,
 		/\b(bitmap|bitmaps|caps on only|caps always off|shift frees caps|copyright|hotkey|language|layout|message|name|version)\b/i
 	],
-	'number': /\b(U\+[\dA-F]+|d\d+|x[\da-f]+|[0-7]+)\b/i, // U+####, x###, d### characters and numbers
+	'number': /\b(U\+[\dA-F]+|d\d+|x[\da-f]+|\d+)\b/i, // U+####, x###, d### characters and numbers
 	'operator': /[+>\\,()]/,
 	'tag': /\$(keyman|kmfl|silkey|keymanweb|keymanonly):/i   // prefixes
 };
--- a/components/prism-keyman.js
+++ b/components/prism-keyman.js
@@ -1,14 +1,14 @@
 Prism.languages.keyman = {
-	'comment': /\b[cC]\s.*?(\r?\n|$)/,
-	'function': /\[\s*((CTRL|SHIFT|ALT|LCTRL|RCTRL|LALT|RALT)\s+)*(([T|K|U]_[a-z0-9_]+)|(".+")|('.+'))\s*\]/i,  // virtual key
+	'comment': /\bc\s.*/i,
+	'function': /\[\s*((CTRL|SHIFT|ALT|LCTRL|RCTRL|LALT|RALT|CAPS|NCAPS)\s+)*([TKU]_[a-z0-9_?]+|".+?"|'.+?')\s*\]/i,  // virtual key
 	'string': /("|')((?!\1).)*\1/,
 	'keyword': /\b(any|beep|call|context|deadkey|dk|if|index|notany|nul|outs|return|reset|save|set|store|use)\b/i,  // rule keywords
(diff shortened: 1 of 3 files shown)


components/prism-keyman.js
```diff
@@ -0,0 +1,14 @@
+Prism.languages.keyman = {
+	'comment': /\b[cC]\s.*?(\r?\n|$)/,
```
Golmote: You can simplify the regexp by using the `i` flag.
Also, I think you can remove the last part `?(\r?\n|$)` since the dot `.` does not match line feeds. (Unless you intentionally wanted to consume the line feed?)

components/prism-keyman.js
```diff
@@ -0,0 +1,14 @@
...
+	'function': /\[\s*((CTRL|SHIFT|ALT|LCTRL|RCTRL|LALT|RALT)\s+)*(([T|K|U]_[a-z0-9_]+)|(".+")|('.+'))\s*\]/i,  // virtual key
```
Golmote: I think you meant `[TKU]` instead of `[T|K|U]`, even though I can't find any usage of `T` or `U` in the documentation...

Furthermore, if I'm reading this properly, you don't need the three pairs of inner parentheses in the last part: `([T|K|U]_[a-z0-9_]+|".+"|'.+')`.

Also, it looks like you forgot the `CAPS` and `NCAPS` shift codes.

Finally, the documentation mentions key codes like `K_KANJI?15` which do not appear to be handled here (due to the `?` char)

components/prism-keyman.js
```diff
@@ -0,0 +1,14 @@
...
+	'string': /("|')((?!\1).)*\1/,
+	'keyword': /\b(any|beep|call|context|deadkey|dk|if|index|notany|nul|outs|return|reset|save|set|store|use)\b/i,  // rule keywords
+	'atrule': /\b(ansi|begin|unicode|group|(using keys)|match|nomatch)\b/i,   // structural keywords
```
Golmote: The parentheses are not needed around `using keys`.

components/prism-keyman.js
```diff
@@ -0,0 +1,14 @@
...
+	'bold': [   // header statements
+		/\&(baselayout|bitmap|capsononly|capsalwaysoff|shiftfreescaps|copyright|ethnologuecode|hotkey|includecodes|keyboardversion|kmw_embedcss|kmw_embdjs|kmw_helpfile|kmw_helptext|kmw_rtl|language|layer|layoutfile|message|mnemoniclayout|name|oldcharposmatching|platform|targets|version|visualkeyboard|windowslanguages)\b/i,
+		/\b(bitmap|bitmaps|(caps on only)|(caps always off)|(shift frees caps)|copyright|hotkey|language|layout|message|name|version)\b/i
```
Golmote: The parentheses are not needed around `caps on only`, `caps always off` and `shift frees caps`

components/prism-keyman.js
```diff
@@ -0,0 +1,14 @@
...
+	],
+	'number': /\b(([uU]\+[\dA-Fa-f]+)|([dD]\d+)|([xX][\dA-Fa-f]+)|([0-7]+))\b/, // U+####, x###, d### characters and numbers
```
Golmote: You should use the `i` flag to simplify the regexp.
Also, the inner pairs of parentheses around each alternative are not needed.

components/prism-keyman.js
```diff
@@ -0,0 +1,14 @@
...
+	'operator': /[+>\\,\(\)]/,
```
Golmote: You don't need to escape `(` and `)` inside a character class.

examples/prism-keyman.html
```diff
@@ -0,0 +1,101 @@
...
+
+<h2>Comments</h2>
+<pre><code>c This is a comment</code></pre>
+
+<h2>Strings, numbers and characters</h2>
+<pre><code>"'this' is a string"
+'and so is "this"'
+U+0041 d65 x41   c these are all the letter A
+</code></pre>
+
```
Golmote: You should add small examples for virtual keys and prefixes, since there don't seem to be any in the large example below.

Golmote:
Thanks for contributing. Please take a look at my comments.

components/prism-keyman.js
mcdurdin: T, U are referenced in http://help.keyman.com/developer/9.0/guides/touch-layouts/creating-a-touch-keyboard-layout-for-amharic-the-nitty-gritty.php. 

Also will tweak quoted strings (non-greedy match).

Aware that ordering virtual keys ahead of strings will format "[K_A]" wrongly but in practice this would never be used so not worth supporting.

mcdurdin:
Thanks for the deep review. I think all comments addressed, a couple of extra minor tweaks as well.

Golmote:
Thanks, this looks good to me now.
@LeaVerou what do you think?

LeaVerou:
Merged, thank you both!

mcdurdin:
Thank you both as well -- that's how a pull request should always work :)  Smooth as.

PR #616, opened 2015-07-01
Add support for Inform 7

Golmote:
Basic support for Inform 7, based on documentation at http://inform7.com/learn/

Examples:

prism-inform7


loganfranken:
Dang! You are *killing* it, Golmote! Thanks for getting all of these languages incorporated!

Golmote:
@loganfranken Sorry for the spam if you get notified each time ^^
There is still a bunch of languages that I'd like to see included in Prism, and I have a bit of spare time these days. :)

loganfranken:
I do get notified each time, but I definitely don't see it as spam!

PR #617, opened 2015-07-01
Fix typo in keyword kmw_embedjs for prism-keyman.js

mcdurdin:


mcdurdin:
Added extra keywords baselayout, layer, platform, as well

PR #627, opened 2015-07-06
Add Ubuntu Mono font to font stack

polybuildr:
None of the other fonts are currently available on Ubuntu and the
addition of Ubuntu Mono to the font stack makes the code look much nicer
on Ubuntu systems.

Also added a missing newline to the end of prism-coy.css

### DIFF
--- a/themes/prism-coy.css
+++ b/themes/prism-coy.css
@@ -7,7 +7,7 @@
 code[class*="language-"],
 pre[class*="language-"] {
 	color: black;
-	font-family: Consolas, Monaco, 'Andale Mono', monospace;
+	font-family: Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace;
 	direction: ltr;
 	text-align: left;
 	white-space: pre;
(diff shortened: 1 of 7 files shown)


apfelbox:
I think this is a nice addition.

/cc @Golmote @LeaVerou ?

Golmote:
I don't have an Ubuntu to test it so I don't really have an opinion on this. I'll leave it to you.

LeaVerou:
Merged, thanks!
Don’t have Ubuntu either, but it won’t break anything :)

polybuildr:
Thank you! Now I'll have a nicer experience on all those websites using Prism. :D

uranusjr:
Just noticed this was merged into the master branch instead of gh-pages. Is this intended?

Golmote:
Dammit you're right. I'll try to fix this.

polybuildr:
Opened #646 in case it helps.

Golmote:
So I reverted the commit on the master branch. @LeaVerou, have we decided anything about this master branch? Should we sync it with gh-pages on a regular basis?

Golmote:
@LeaVerou I have synced it right now. But please tell me how you want to handle this.

PR #628, opened 2015-07-06
Add support for OCaml

Golmote:
Examples:

prism-ocaml


LeaVerou:
Nice work!
I’m not quite sure about the aliases; is there any semantic meaning in them, or are they purely to get highlighting for the current themes?

Golmote:
I'd say that the directives aliased as functions have some meaning.
But I must admit the `type` token aliased as variable is purely for visual purpose... I kinda thought that's what aliases were for.

Golmote:
The `builtin` alias could be the easy alternative (as it does not mean much by itself...), but this is just bypassing the problem, isn't it?

Issue #83, opened 2013-04-10
SCSS parser fails lots of common declarations


Snugug:
I'm finding that the SCSS parser is failing on lots of fairly standard declarations. For instance, media queries with a media type (`@media screen…` fails to highlight the `@media`), lots of sibling selectors fail to highlight the first half of the selector (`.foo > .bar` fails to highlight `.foo`), and absolutely no numbers (or their units) get properly highlighted.

Golmote:
This should be fixed now.

grayghostvisuals:
\o/

PR #631, opened 2015-07-08
Fix SCSS Media Query Regex

loganfranken:
A slight adjustment to the SCSS regex to capture the media types and logical operators in a media query.

Fixes part of #83.

Because the SCSS parser is different than the CSS parser in that it allows syntax highlighting *within* the media query, this PR *doesn't* fix cases where operators occur after parentheses, like so:

```css
@media tv and (min-width: 700px) and (orientation: landscape) { ...  }
```

(The `and` won't get properly highlighted here and I'm not sure it *can* under the current methodology since you need the regex to do a hop over the parentheses block)

But this at least gets the `@media tv and` part highlighted again and we can keep working through the other issues.


Golmote:
I guess we would need to match to whole query and then add `inside` rules to be able to highlight properly the conditions. I may check this out later. For now, I'll merge this, thanks!

PR #636, opened 2015-07-10
Add ABAP language

dellagustin:
Based on the work of @rmtiwari (https://github.com/rmtiwari/ABAPSyntaxHighlighter).
Supports:
* line comments
* end of line comments
* numbers
* operators
* strings
* string templates
* punctuation
* token operators
* huge (probably complete) list of keywords


components/prism-abap.js
```diff
@@ -0,0 +1,50 @@
+Prism.languages.abap= {
```
Golmote: May you add a space before the `=`?

components/prism-abap.js
```diff
@@ -0,0 +1,50 @@
...
+	'comment': {
+		pattern: /^[\*].*$/gm,
```
Golmote: There is no need to escape the `*` inside a character class.
Also, I think the `$` is useless here, since `.*` will consume everything until the end of line anyway.

Finally, the `g` flag can be removed: it is not used by Prism. (This applies to all your regexps)

components/prism-abap.js
```diff
@@ -0,0 +1,50 @@
...
+		lookbehind: false
+	},
+	'string' : /(`|')(\\?.)*?\1/gm,
+	'string_template': {
+	  pattern: /(\||\})(\\?.)*?(?=\||\{)/g,
+	  lookbehind: true,
+	  alias: 'string'
```
Golmote: Please be careful with the indentation here. Prism uses tabs, not spaces.

components/prism-abap.js
```diff
@@ -0,0 +1,50 @@
...
+	},
+	/* End Of Line comments should not interfere with strings when the  
+	quote character occurs within them. We assume a string being highlighted
+	inside an EOL comment is more acceptable than the oposite.
+	*/
+	'eol_comment': {
+		pattern: /(\s\").*$/gm,
+		lookbehind: false,
```
Golmote: There is no need for `lookbehind` to be `false`. It is `false` by default.

Golmote: Also, I think the `$` is useless, like I said before, because the `.*` pattern will consume characters until the end of line.

components/prism-abap.js
```diff
@@ -0,0 +1,50 @@
...
+		lookbehind: true,
```
Golmote: Please remove the extra comma here.

components/prism-abap.js
```diff
@@ -0,0 +1,50 @@
...
+	},
+	/* Numbers can be only integers. Decimal or Hex appear only as strings */
+	'number' : /\b\d+\b/g,
+	/* Operators must always be surounded by whitespace, they cannot be put
```
Golmote: "surounded" => "surrounded"

components/prism-abap.js
```diff
@@ -0,0 +1,50 @@
...
+		lookbehind: true,
+	},
+	/* Numbers can be only integers. Decimal or Hex appear only as strings */
+	'number' : /\b\d+\b/g,
+	/* Operators must always be surounded by whitespace, they cannot be put 
+	adjacent to operands. 
+	*/
+	'operator' : {
+		pattern: /(\s)(\+|-|\*{1,2}|\/|<|>|<=|>=|=|\?=|<>)(?=\s)/g,
+		lookbehind: true,
```
Golmote: Please remove the trailing comma here.

components/prism-abap.js
```diff
@@ -0,0 +1,50 @@
...
+	},
+	'string_operator' : {
+		pattern: /(\s)(&{1,2})(?=\s)/g,
+		lookbehind: true,
+		alias: "keyword"
```
Golmote: Why isn't this highlighted as "operator" instead of "keyword"?

components/prism-abap.js
```diff
@@ -0,0 +1,50 @@
...
+	},
+	'token_operator' : [{
+		/* Special operators used to access structure components, class methods/atributes, etc...*/
+		pattern: /(\w)(-|~|->|=>|\||\{|\})(?=\w)/g,
+		lookbehind: true,
+		alias: "punctuation"
+	}, {
+	    /*Special tokens used do delimit string templates*/
+	    pattern: /[\|\{\}]/g,
```
Golmote: There is no need to escape those characters inside a character class. The regexp can be rewritten as `/[|{}]/`.

components/prism-abap.js
```diff
@@ -0,0 +1,50 @@
...
+		alias: "punctuation"
+	}],
+	'punctuation' : /[,.:\(\)]/g
```
Golmote: Same here, no need to escape the parentheses.

components/prism-abap.js
```diff
@@ -0,0 +1,50 @@
...
+	  pattern: /(\||\})(\\?.)*?(?=\||\{)/g,
+	  lookbehind: true,
+	  alias: 'string'
+	},
+	/* End Of Line comments should not interfere with strings when the  
+	quote character occurs within them. We assume a string being highlighted
+	inside an EOL comment is more acceptable than the oposite.
+	*/
+	'eol_comment': {
+		pattern: /(\s\").*$/gm,
```
Golmote: There is no need to escape the double quote here.

Golmote:
Thanks for contributing! Please take a look at my comments.

components/prism-abap.js
dellagustin: That is because in the official ABAP Editor this specific kind of operator is highlighted as keyword.

dellagustin:
Thanks for the comments, I did the changes.

components/prism-abap.js
Golmote: Alright, then.

components/prism-abap.js
```diff
@@ -0,0 +1,47 @@
...
+	'number' : /\b\d+\b/,
+	/* Operators must always be surrounded by whitespace, they cannot be put 
+	adjacent to operands. 
+	*/
+	'operator' : {
+		pattern: /(\s)(\+|-|\*{1,2}|\/|<|>|<=|>=|=|\?=|<>)(?=\s)/,
+		lookbehind: true
+	},
+	'string_operator' : {
+		pattern: /(\s)(&{1,2})(?=\s)/,
```
Golmote: The parentheses aren't needed around `&{1,2}`. Also, it would be shorter written as `&&?`. (BTW this also apply to `\*{1,2}` in the operator regexp above.)

Golmote:
This looks good to me. @LeaVerou, can you take a look, too?

Golmote:
Sorry for the delay, I'm merging this.

PR #637, opened 2015-07-10
Added resource to template for quickstart

javatlacati:
Hope this can boost the number of languages supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,5 +18,6 @@ Prism depends on community contributions to expand and cover a wider array of us
 - My time is very limited these days, so it might take a long time to review longer PRs (short ones are usually merged very quickly), especially those modifying the Prism Core. This doesn't mean your PR is rejected.
 - If you contribute a new language definition, you will be responsible for handling bug reports about that language definition. Soon I plan to add usernames of project owners for themes, plugins and language definitions so this becomes more clear to users.
 - If you add a new language definition, theme or plugin, you need to add it to `components.js` as well, so that it becomes available to the download build page.
+- to help u get started you can test your new language definition with this [pen](http://codepen.io/javatlacati/pen/zGjWZd) fork it and keep going until you have something like [this] (http://jsfiddle.net/javatlacati/nu4ywjd8/).
 
 Thank you so much for contributing!!


Golmote:
I'm not sure about this. The pens you provide (why are they hosted on two different websites, BTW? oO) aren't really helpful...

I guess we could host a grammar template in the repo, but I don't really see the point, since there are lots of components already available to get inspired!

javatlacati:
Just not to favore one site, even if my favorite if plkr.
Well... I've checked some running examples and most of them fail because user's don't know how to integrate into code the functionality.
For example [this] (http://jsfiddle.net/PeterChaplin/ta9vy1e0/), or [this](http://stackoverflow.com/questions/18391425/markup-not-showing-up-using-prism-js), or this [one](http://codepen.io/chriscoyier/pen/cGhpf).

Golmote:
Your first example mostly fails because the included files ([[1]](https://dl.dropboxusercontent.com/u/313630820/js/prism.js/prism.js), [[2]](https://dl.dropboxusercontent.com/u/313630820/js/prism.js/prism.css)) do not exist...
Your third example works fine too, once you fix the wrong stylesheet URL.

For your second example... well, we can't do much about the JSFiddle issue. And regarding the fact that `<pre>` tags interpret HTML, as I already explained in #639, this is its normal behaviour, and Prism does not change anything to it. To display an HTML tag inside a `<pre>`, you *must* escape the "less-than" character as `&lt;`.

javatlacati:
I know, but I didn't created the examples as you may nottice. If people are making mistakes is because of a lack of a template they can easily reach trough google search. JSFiddle and codepen excels in ease to find in google search because of their tag functionality meanwhile github doesn't.
I know I don't have the best example ever made, if you can improve it we'll be pleased.
I consider that the repo needs more than a template a running example that we can modify and get the output instantly without the need to download and configure it locally as you can achieve with those kind of sites.
Those sites have also the advantage of being social aware, because they integrate social network & sharing functionalities, and many people use them as portfolio when getting some job, so.. they're serious and high audience with a huge community wiling to learn and share each one.

Golmote:
I get your point. I'll let @LeaVerou decide whether she wants to merge this.
Thanks for contributing, anyway.

mAAdhaTTah:
I don't think this is necessary, tbh. Appreciate the contribution, but going to close.

LeaVerou:
Also, why on earth does that codepen have the whole Prism code pasted in? Why not just add it as a resource from the CDN and get updates as they are released?
Yeah, I second @mAAdhaTTah’s decision.

PR #638, opened 2015-07-10
Number recognition improved and documented regex

javatlacati:
Explained the regex with natural language so people unfamiliar with it can understand it and easily improve the regex expressions

Tested with:

    0b001
    0x56ab16ec675.7648p34
    public static final double MAX_VALUE = 0x1.fffffffFfffffP+1023;
    public static final double MIN_NORMAL = 0x1.0p-1022;
    public static final double MIN_VALUE = 0x0.0000000000001P-1022; // 4.9e-324
    4.9e-324
    3.30e23
    3.30e+23
    3.30e+23f
    3.30e+23l
    6.67e−11
    9.870699812169277
    NumberFormat formatter = new DecimalFormat("0.00000000000");
    String string = formatter.format(9.870699812169277E-4);
    System.out.println(string);

### DIFF
--- a/components/prism-java.js
+++ b/components/prism-java.js
@@ -5,7 +5,7 @@ Prism.languages.java = Prism.languages.extend('clike', {
 	// a word consisting of 0x an optional letter from a to f or digit perhaps one dot at least one letter from a to f or digit or
 	// a word consisting of an optional digit perhaps one dot at least one digit perhaps the letter e a + or - symbol, optional digits and perhaps one of the letters d f l or
 	// a word consisting of an optional digit perhaps one dot at least one digit
-	'number': /\b0b[01]+\b|\b0x[\da-f]*\.?[\da-fp\-]+\b|\b\d+\.?\d*(e|E)?([\+]|[\-])?[\d]*[dfl]*\b|\b\d*\.?\d+\b/i,
+	'number': /\b0b[01]+\b|\b0x[\da-f]*\.?[\da-fp\-]+\b|\b\d+\.?\d*e?[+-]?[\d]*[dfl]*\b|\b\d*\.?\d+\b/i,
 	'operator': {
 		pattern: /(^|[^\.])(?:\+=|\+\+?|-=|--?|!=?|<{1,2}=?|>{1,3}=?|==?|&=|&&?|\|=|\|\|?|\?|\*=?|\/=?|%=?|\^=?|:|~)/m,
 		lookbehind: true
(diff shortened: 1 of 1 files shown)


components/prism-java.js
```diff
@@ -1,8 +1,13 @@
 Prism.languages.java = Prism.languages.extend('clike', {
 	'keyword': /\b(abstract|continue|for|new|switch|assert|default|goto|package|synchronized|boolean|do|if|private|this|break|double|implements|protected|throw|byte|else|import|public|throws|case|enum|instanceof|return|transient|catch|extends|int|short|try|char|final|interface|static|void|class|finally|long|strictfp|volatile|const|float|native|super|while)\b/,
-	'number': /\b0b[01]+\b|\b0x[\da-f]*\.?[\da-fp\-]+\b|\b\d*\.?\d+[e]?[\d]*[df]\b|\b\d*\.?\d+\b/i,
+	// number is either
+	// a word consisting of 0b the 0 or 1 number one or more times or
+	// a word consisting of 0x an optional letter from a to f or digit perhaps one dot at least one letter from a to f or digit or
+	// a word consisting of an optional digit perhaps one dot at least one digit perhaps the letter e a + or - symbol, optional digits and perhaps one of the letters d f l or
+	// a word consisting of an optional digit perhaps one dot at least one digit
+	'number': /\b0b[01]+\b|\b0x[\da-f]*\.?[\da-fp\-]+\b|\b\d+\.?\d*(e|E)?([\+]|[\-])?[\d]*[dfl]*\b|\b\d*\.?\d+\b/i,
```
Golmote: Why did you change `[e]` to `(e|E)`? The regexp is case-insensitive so there is no need to specify the `E`. Furthermore, it could be simplified to `e` without the brackets.

Also, `([\+]|[\-])?` should be written `[+-]?`.

Golmote:
I'm not certain explaining the regexp this precisely is useful.

Comments may be useful to explain a behaviour that would not be obvious otherwise but just describing the whole meaning of the regexp is not needed.

components/prism-java.js
javatlacati: You can fix this minor mistakes once integrated ( Actually is faster than place a comment about it )

javatlacati:
I'm sorry to telling you but... regex is not obvious to most IT people (at least 90% of the people I know for example). Even for me wasn't that clear.. because I wasn't used it in about 3 or 4 years and at college teached me to use it only 3 or 4 days.

Golmote:
Don't misunderstand me. I'm not telling regexps are obvious to everyone. I'm telling that people wishing to add a grammar in Prism need to know how to write regexps.

> [&laquo; Comments should provide additional information that is not readily obtainable from the code itself. They should never parrot the code. &raquo;](http://jazoon.com/history/Portals/0/Content/slides/tu_a5_1740-1830_henney_keynote.pdf#page=10&zoom=auto,31,-155)

IMO, this:

```javascript
// a word consisting of 0b the 0 or 1 number one or more times
/\b0b[01]+\b/
```

is really the same as this:

```javascript
// assign the number 5 to the variable foo
var foo = 5;
```

Golmote:
I took some time to think again about this. Your comments are definitely overkill.
I'm closing this PR. Feel free to update it if you wish to merge only the changes you made to the regexp.

PR #640, opened 2015-07-11
adding system verilog language

a-rey:
#### Hello! 

I use some hardware description languages in a website of mine and was sad to see they are not all supported by Prism. Here is my proposed Prism support for the [System Verilog](https://en.wikipedia.org/wiki/SystemVerilog) and [Verilog](https://en.wikipedia.org/wiki/Verilog) languages. Since System Verilog is just an extension of Verilog, I combined both into one. The file can be used to highlight either language. I also added an example for you guys to take a look at if you are not familiar with the syntax. [This handout](http://web.stanford.edu/class/ee183/handouts_win2003/VerilogQuickRef.pdf) is a great reference for syntax rules. Let me know if there are any issues!

Here is some sample code:
screen shot 2015-07-11 at 5 05 09 pm


components.js
```diff
@@ -328,6 +328,10 @@ var components = {
 			"require": "clike",
 			"owner": "chrischares"
 		},
+		"system_verilog": {
```
Golmote: Would you mind naming the language with a dash instead of an underscore?
Or maybe just name it `verilog`, since this is the base language?

components/prism-system_verilog.js
```diff
@@ -0,0 +1,23 @@
+Prism.languages.system_verilog = {
+  'comment': /\/\/.*|\/\*[\w\W]*?\*\//,
+  'string': /"(\\\n|\\?.)*?"/,
```
Golmote: Please add support for \r and \r\n too: `(\\(?:\r\n?|\n)|`

components/prism-system_verilog.js
```diff
@@ -0,0 +1,23 @@
...
+  // support for any kernel function (ex: $display())
+  'property': /(?!\b)\$\w+\b/,
```
Golmote: Why do you use a look-ahead here? Replacing `(?!\b)` with `\B` should do exactly the same result, I think...

components/prism-system_verilog.js
```diff
@@ -0,0 +1,23 @@
...
+  // support for user defined constants (ex: `define)
+  'constant': /(?!\b)`\w+\b/,
```
Golmote: Same as comment above.

components/prism-system_verilog.js
```diff
@@ -0,0 +1,23 @@
...
+  'function': {
+    'pattern': /[a-z0-9_]+\(/i,
+    'inside': {
+      'punctuation': /\(/
+    }
```
Golmote: Here, it would probably be better to use a look-ahead instead of the `inside` (since the parenthesis will be highlighted by the `punctuation` pattern anyway):

```
'function': {
    'pattern': /[a-z0-9_]+(?=\()/i
},
```

components/prism-system_verilog.js
```diff
@@ -0,0 +1,23 @@
...
+  // bold highlighting for all verilog and system verilog logic blocks
+  'important': /\b(always_latch|always_comb|always_ff|always) ?@?/,
+  // support for time ticks, vectors, and real numbers
+  'number': /(?!\b)#{1,2}\d+|(\b\d+)?'[odbh] ?[\da-f10zx_\?]+|\b\d+[._]?(e-?\d+)?/i,
```
Golmote: Same as before regarding the negative look-ahead.
`#{1,2}` could be simplified as `##?`.
In the second alternative, shouldn't there be a `\b` before the part `[\da-f10zx_\?]+` (since the `\b` just before is inside an optional capture)?
You don't need to escape the interrogation mark inside a character class.
In the last alternative, it seems you allow no digit between the decimal dot and the `e`... is that intended? Also, make sure you did not forget the `Xe+XX` syntax (I don't know whether Verilog allows it or not).

components/prism-system_verilog.js
```diff
@@ -0,0 +1,23 @@
...
+  'operator': /[-+\{\}]|!=?|<{1,2}=?|>{1,2}=?|={1,3}|\^|~|%|&{1,2}|\|?\||\?|\*|\//,
```
Golmote: You don't need to escape the curly braces inside the character class.
`<{1,2}` and the two other similar patterns could be simplified as `<<?` (Although I might agree that `={1,3}` may be more readable than `==?=?`).
Is there a reason you did not include those operators `^ ~ % ? * \/` in the first character class?

examples/prism-system_verilog.html
```diff
@@ -0,0 +1,104 @@
+<h1>System Verilog</h1>
+<p>To use this language, use the class "language-system_verilog".</p>
```
Golmote: Here, you might want to add a paragraph explaining that the component is also compatible with Verilog.

Golmote:
Thanks for contributing! Please take a look at my comments.

a-rey:
Renamed the component to `verilog`. Added more info to `components.js` and `prism-verilog.html` to note the crossover between the 2 languages. Made the requested changes and updated a few mistakes on my part with the languages. Let me know if you see anything else.

Golmote:
Thanks. This looks quite good to me now! @LeaVerou what do you think?

Golmote:
Sorry for the delay, I'm merging this now.

PR #646, opened 2015-07-26
Add Ubuntu Mono font for Ubuntu users

polybuildr:
None of the other fonts are currently available on Ubuntu and the
addition of Ubuntu Mono to the font stack makes the code look much nicer
on Ubuntu systems.


Golmote:
That's fine, I already patched it: ed9d7e354c20a95755af7641e3a2a0c77a713791

Thank you, anyway :)

PR #648, opened 2015-07-28
Add Composer support

RobLoach:
[Composer](https://getcomposer.org/) is a package manager for PHP. This PR allows Prism to be installed with Composer, exposing it to easily be consumed by PHP frameworks like Symfony, Wordpress, Drupal, and many others.


Golmote:
Sorry for the time it took for this PR to be merged. Thank you for contributing!

Issue #645, opened 2015-07-25
line-number plugin not working when using language-none


markg85:
Hi,

The line-number plugin only seems to be working if you set the highlighting to something other then none. While none is the advised one to use for plain text (if i read prismjs.com correctly).

What i "think" is going on is the hooks not being executed if highlighting wasn't used. Perhaps a simple fix would be to always execute the hooks?

Cheers,
Mark

loganfranken:
Hey @markg85!

(Thanks for reporting this!)

I *think* this is the same problem that was discussed in #181: basically, (as @Golmote explains), the line numbers plugin runs in the `after-highlight` hook, which is never called because no highlighting occurs.

@Golmote lays out two approaches to fix this in #181: add a `no-grammar` hook or a `complete` hook. I think the latter approach makes the most sense.

I'm wondering if we should close #181 and have this be the main issue for this problem since #181 is actually about two different issues, the first of which we already solved.

(Let me know if you feel like I got this wrong, @markg85!)

markg85:
hi @loganfranken,

You seem to be spot on. A "no-grammer" hook sounds a bit weird to me. I'd probably go for a "complete", "completed" or "done" hook which should probably be run when the core prismjs is done doing it's thing.

Golmote:
Hi both of you.

I followed your advice, @loganfranken, and closed the other issue.

@markg85 I agree that the `complete` hook makes more sense than my other proposal. PRs are welcome to solve this issue, since I won't have much time during the next weeks to look into this. But I'll give it a try after my holidays if nobody did it in the meantime.

markg85:
@Golmote I'm not quite familiar with how this script works and where it should be adapted. So I happily leave this to your (or anyone else who knows what needs to be done) capable hands :)

Golmote:
This is fixed, as you can see in the last example on the [plugin page](http://prismjs.com/plugins/line-numbers/).

brandoncc:
It still isn't working for me with the latest code downloaded from prismjs.com. Do I need to do something special to get it to work? I can't find any information on getting it to work.

PR #657, opened 2015-08-08
Complete hook

manfer:
If I understood correctly this is how you want to solve https://github.com/PrismJS/prism/issues/645. By adding a new complete hook and changing line numbers plugin to run in that hook instead of running on after highlight hook.

### DIFF
--- a/plugins/line-highlight/prism-line-highlight.js
+++ b/plugins/line-highlight/prism-line-highlight.js
@@ -85,7 +85,7 @@ function applyHash() {
 
 var fakeTimer = 0; // Hack to limit the number of times applyHash() runs
 
-Prism.hooks.add('after-highlight', function(env) {
+Prism.hooks.add('complete', function(env) {
 	var pre = env.element.parentNode;
 	var lines = pre && pre.getAttribute('data-line');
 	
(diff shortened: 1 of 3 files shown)


apfelbox:
:+1:

manfer:
> Just curious, why is this branch moved here? It can be argued that an empty code block should still trigger the complete hook.

Not sure if you mean that an empty code block now triggers the complete hook and it shouldn't or if you want to trigger it and it doesn't. Anyway this code has nothing to do with that issue.

If,

```
<pre><code></code></pre>
```

is a problem before or after this code,

```
<pre><code class="lang-xxx"></code></pre>
```

for a valid xxx language is/was either. And this code does not deal in any way with that issue if there is any.

The hook addition has been suggested on https://github.com/PrismJS/prism/issues/645 and the problem that tries to solve is to include line numbers on source code for which there is no grammar.

So the complete hook runs at the end of proccess. For source code for which there is a grammar that's at the end, line 233. But for source code for which there is no grammar (`!grammar`) that's on line 196 before the `return`.

I'm not too familiarized with the library so the PR can be wrong of course. So needs review.

Golmote:
I made a few changes to your PR (see fd549959cd57d1c5de8be2f6d7600fe9bd050857) to take @uranusjr's comment into account.

markg85:
Thank you folks for fixing this (issue #645). It looks perfectly fine to me now.

PR #660, opened 2015-08-13
Add tests for Markup component

Golmote:
A set of tests for the Markup component.
@apfelbox may you take a look at this, I'm not particularly familiar with writing tests. I might be testing too much, or too few. What do you think?


apfelbox:
Looks good to me :+1: 
I think thats exactly the right size for tests, it hits the balance between small enough to grasp in one or two looks but not too small so that we end up with a plethora of different tests files.

Golmote:
Great, thanks !

Issue #79, opened 2013-03-17
SCSS Class w/Pseudo Class Misparsed


Snugug:
I have the following using SCSS syntax highlighting:

``` scss
/*comment*/
.icon:before {
  font-family: 'Icon Font', sans-serif;
  content: 'icon';
}
```

The `.icon` is parsed into `.<span class="token property">icon</span>` producing incorrect highlighting for it.

Issue #75, opened 2013-02-18
Doesn't work if prism is asynchronously loaded


WickyNilliams:
If prism is loaded asynchronously then it doesn't highlight any code. It appears because it is being loaded *after* `DOMContentLoaded` event has already fired. 

* Example loading async: http://jsfiddle.net/5QPNN/3/
* Same example loading sync : http://jsfiddle.net/5QPNN/1/

Damn shame this does not work out of the box as I would definitely like this kind of thing loaded asynchronously as it's (technically) non-essential and a discrete piece of functionality.

Any help much appreciated

aykevl:
Can I make a pull request for that? I would really like prism to support async. I don't know how to test it though - is there any documentation how to build the prism.js file?

Golmote:
Sure. To build the prism.js file, you just need to run `gulp`.

Golmote:
#959 fixed this issue. First JSFiddle is working as expected.

Issue #40, opened 2012-10-04
(line-highlight) Bug in Safari if no line-height set on pre tag


kpzani:
If the pre tag doesnt have a line-height then getComputedStyle returns "normal" which cannot be parsed as a float and causes the lines and highlight to be placed underneath the code.

not sure how best to solve that except by including a default line-height in the stylesheets that come with prism and adding a little comment in the files to help warn users.  

tested with Safari 5.0.6 on Leopard.

PR #663, opened 2015-08-13
Keep the highlighted handlebars code if there are more plugging, e.g.…

janvanhelvoort:
… line-numbers.


Golmote:
I'm not sure I understand the issue this PR is supposed to fix... Can you provide an example?

janvanhelvoort:
if you have a code block with the following layout, E.g.
```
<pre class="line-numbers"><code class="language-handlebars">{{template}}</code></pre>
```
In the 'Line-Number' plugin, on the following lines [#40 - #42](https://github.com/PrismJS/prism/blob/95dc10296922e7666b75428ea89359ee93d95e62/plugins/line-numbers/prism-line-numbers.js#L40-L42), the output is set on `env.element`.

If you use it with handlebars syntax highlight, the added code from the 'Line-Numbers' plugging is removed, because on line [#75 - #79](https://github.com/PrismJS/prism/blob/95dc10296922e7666b75428ea89359ee93d95e62/components/prism-handlebars.js#L75-L79) the code is overwritten. It uses `env.highlightedCode` instead of `env.element`.

The fix of mine, takes `env.element.innerHTML` if it is there, otherwise it will take the `env.highlightedCode`

Golmote:
Can you take a look at this Fiddle? It uses the code you just provided with the line numbers plugin. And I can't see any problem.

janvanhelvoort:
Strange, in your example, it works, in our project, it isn't working,
Can I mail jou the project?

janvanhelvoort:
I see it, on you site: Prism-line-numbers.js rule #1: `Prism.hooks.add('complete', function (env) {`
If i get it from Bower, rule #1 `Prism.hooks.add('after-highlight', function (env) {`

but on git, the correct line is displayed ( `Prism.hooks.add('complete', function (env) {` )
Bower is serving old code?

and it uses the wrong git: `git://github.com/LeaVerou/prism.git#~1.0.1`

janvanhelvoort:
`$ bower install 'git://github.com/PrismJS/prism#gh-pages' --save` works for now :+1: 

```
  "dependencies": {
    "prism": "git://github.com/PrismJS/prism#gh-pages"
  },
```

Golmote:
The "complete" hook is a really recent change. It is not part of release 1.0.1.
gh-pages is always the most up-to-date branch, but we're trying to sync master regularly.

PR #674, opened 2015-08-17
components.json now return JSON data

malash:
Rename components.js to components.json. Now components.json return JSON data and the new components.js make a ajax request to get data from components.json so the Github Page still works.
The reason why I do this is I want to program a Angular Module for Prism that real-time analyze component. But the old components.js pollute the global variable and it's unfriendly to Angular.


apfelbox:
Nice idea, but this completely breaks the website, as it is using the the values in the `components` variable. So :-1:  for me.

If you want to use it in an external file, just strip the leading `var components = ` and parse it as JSON. The file is formatted, that this is possible.

malash:
Using JSON.parse may work, but is not perfect. Even code comments can cause parse failed.( See https://github.com/PrismJS/prism/blob/b8d098e1d420715366302274c511ab4aeaeeb2fe/components.js:58 // ---, Other situation like ' instead of " can also cause failed ). Because JSON is the subset of JavaScript. Contributors won't program in strict JSON format. So if I program the Angular Module depend on this repo, it will not work someday, or today :(
The new components.js get components.json data via Ajax, and it can work well on Github Page http://prismjs.com/. Maybe the problem is some 3rd-part websites include the components.js and read the global var components. In the new components.js the Ajax will cause the Cross-Domain error. So the problem is that how many 3rd-part wensites imclude the components.js?

apfelbox:
I know that this is not the best solution, but it is up to the maintainers to ensure that everything after the equals sign is valid JSON.

If you want to adapt the website, we could hover switch to JSON, although this would make the test and download pages more fragile.

malash:
Thank you. I'm waiting for your approval.

Golmote:
Thanks for contributing. This PR, in its current state, won't be merged though, since it breaks the website.
When Prism will go fully asynchronous - I'm confident in that it will, someday -, it would probably be a good idea to make the components fetching asynchronous too, but for now, I think you'll have to rely on the global object.
Feel free to submit a PR that makes the right part of the equals sign valid JSON: we'll ensure it stays valid in the future.

PR #676, opened 2015-08-17
Atom Dark Theme

gibsjose:
Added an atom-dark theme based on https://github.com/atom/atom-dark-syntax.


gibsjose:
Switching to dedicated branch

PR #677, opened 2015-08-17
Edited Python language: Added 'self' keyword and 'class-name'

gibsjose:
Added 'self' as keyword and added 'class-name' matching to Python language.


components/prism-python.js
```diff
@@ -8,9 +8,16 @@ Prism.languages.python= {
 		pattern: /((^|\s)def[ \t]+)([a-zA-Z_][a-zA-Z0-9_]*(?=\())/g,
 		lookbehind: true
 	},
-	'keyword' : /\b(as|assert|async|await|break|class|continue|def|del|elif|else|except|exec|finally|for|from|global|if|import|in|is|lambda|pass|print|raise|return|try|while|with|yield)\b/,
+	'keyword' : /\b(as|assert|async|await|break|class|continue|def|del|elif|else|except|exec|finally|for|from|global|if|import|in|is|lambda|pass|print|raise|return|self|try|while|with|yield)\b/,
```
gibsjose: Upon further review, it appears that `self` is not a keyword (which is likely why it wasn't there in the first place...).  I will correct this and commit an updated version only including the `class-name`.

components/prism-python.js
```diff
@@ -14,3 +14,10 @@ Prism.languages.python= {
 	'operator' : /[-+]|<=?|>=?|!|={1,2}|&{1,2}|\|?\||\?|\*|\/|@|~|\^|%|\b(or|and|not)\b/,
 	'punctuation' : /[{}[\];(),.:]/
 };
+
+Prism.languages.insertBefore('python', 'keyword', {
```
Golmote: I don't see the need for the `insertBefore()` here. The `class-name` pattern could go directly into before the `keyword` key in the object above.

Golmote:
Thanks for contributing! Please take a look at my comment.

gibsjose:
Fixed!

PR #725, opened 2015-08-28
Issue 675 - fixing inline style for Coy theme

AnujRNair:
This change fixes inline code blocks for the Coy theme


Golmote:
Thanks!

PR #731, opened 2015-08-28
Add tests for PowerShell

nauzilus:
There were some errors revealed which I've fixed up, and some missing operators/keywords.

I can split to different PRs if that's preferred.


components/prism-powershell.js
```diff
@@ -9,26 +9,38 @@ Prism.languages.powershell = {
...
+		{
+			pattern: /"(`?[\w\W])*?"/m,
+			inside: {
+				'function': {
+					pattern: /[^`]\$\(.*?\)/,
+					inside: {}
+				}
+			}
+		},
+		/'([\w\W])*?'/m
```
Golmote: You don't need the parentheses here. The `m` flag is not needed either.

Actually, I think this could be simplified as `/'[^']*'/`, even if we might want to handle the escaping better. (See my comment in the string test).

tests/languages/powershell/string_feature.test
```diff
@@ -0,0 +1,37 @@
...
+				["operator", "-and"],
+				["string", "'interpolation'"],
+				["punctuation", ")"]
+			]
+		],
+		"\""]
+	],
+	["string", ["\"string `\"with`\" escaping\""]],
+	["string", "'non-interpolated $string'"],
+	["string", "'also '"],["string", "'with'"],["string", "' escaping'"]
```
Golmote: This is not really a string, but multiple strings. Couldn't you handle the escaping so that this is only one token?

components/prism-powershell.js
```diff
@@ -9,26 +9,38 @@ Prism.languages.powershell = {
 			lookbehind: true
 		}
 	],
-	'string': {
-		pattern: /("|')(`?[\w\W])*?\1/m,
-		inside: {}
-	},
+	'string': [
+		{
+			pattern: /"(`?[\w\W])*?"/m,
```
Golmote: The `m` flag is not needed here.

components/prism-powershell.js
```diff
@@ -9,26 +9,38 @@ Prism.languages.powershell = {
...
+		/\b(ac|cat|chdir|clc|cli|clp|clv|compare|copy|cp|cpi|cpp|cvpa|dbp|del|diff|dir|ebp|echo|epal|epcsv|epsn|erase|fc|fl|ft|fw|gal|gbp|gc|gci|gcs|gdr|gi|gl|gm|gp|gps|group|gsv|gu|gv|gwmi|iex|ii|ipal|ipcsv|ipsn|irm|iwmi|iwr|kill|lp|ls|measure|mi|mount|move|mp|mv|nal|ndr|ni|nv|ogv|popd|ps|pushd|pwd|rbp|rd|rdr|ren|ri|rm|rmdir|rni|rnp|rp|rv|rvpa|rwmi|sal|saps|sasv|sbp|sc|select|set|shcm|si|sl|sleep|sls|sort|sp|spps|spsv|start|sv|swmi|tee|trcm|type|write)\b/i
+	],
+	// per http://technet.microsoft.com/en-us/library/hh847744.aspx
+	'keyword': /\b(Begin|Break|Catch|Class|Continue|Data|Define|Do|DynamicParam|Else|ElseIf|End|Exit|Filter|Finally|For|ForEach|From|Function|If|InlineScript|Parallel|Param|Process|Return|Sequence|Switch|Throw|Trap|Try|Until|Using|Var|While|Workflow)\b/i,
 	'operator': {
-		pattern: /(\W)-(and|x?or|not|eq|ne|gt|ge|lt|le|Like|(Not)?(Like|Match|Contains|In)|Replace)\b/i,
+		pattern: /(\W?)(!|-(eq|ne|gt|ge|lt|le|sh[lr]|not|b?(and|x?or)|(Not)?(Like|Match|Contains|In)|Replace|Join|is(Not)?|as)\b|[-+]{2}|[-+*\/%]=?)/i,
```
Golmote: `[-+]{2}` matches `--`, `-+`, `+-` and `++`.

components/prism-powershell.js
```diff
@@ -9,26 +9,38 @@ Prism.languages.powershell = {
 			lookbehind: true
```
Golmote: Here, we might want to get rid of the last group and just end the regexp with `#.+`, unless you really need to consume the line feed.

components/prism-powershell.js
```diff
@@ -9,26 +9,38 @@ Prism.languages.powershell = {
...
-		pattern: /("|')(`?[\w\W])*?\1/m,
-		inside: {}
-	},
+	'string': [
+		{
+			pattern: /"(`?[\w\W])*?"/m,
+			inside: {
+				'function': {
+					pattern: /[^`]\$\(.*?\)/,
+					inside: {}
```
Golmote: Could you add a comment like "See at the end" or something, so that this looks more intentional?

Golmote:
Yay! Thanks for submitting this!

> There were some errors revealed

I know, right? That's the great consequence of adding tests ^^

Would you mind taking a look at my comments?

tests/languages/powershell/string_feature.test
nauzilus: Ah yes, you're right :(

Using `/'([^']|'')*'/` seems to handle escaping (and empty `''` string) ok. If that's fine, I'll commit.

Golmote: Yup, that's fine.

nauzilus:
Have updated PR after incorporating feedback. Thanks so much for a thorough review! :+1:

Golmote:
Thanks!

Issue #324, opened 2014-08-11
Line numbers alignment problems


rmzelle:
I'm using Prism with both the Line Highlight and Line Numbers plugins, downloaded a few days ago, and am having several issues with line number alignments.

With Firefox, I get the following alignment:

![image](https://cloud.githubusercontent.com/assets/77951/3882781/7aaa8176-219a-11e4-84ab-79f19d2dee4c.png)

I tried fixing the height of the highlighted number with 

```
    .line-highlight {
    margin-top: 0em;
    }
```

which seemed to do the trick in Firefox:

![image](https://cloud.githubusercontent.com/assets/77951/3882798/beb39ea2-219a-11e4-986c-f999910eecaa.png)

But in Google Chrome, the alignment is wrong no matter what. Without setting "margin-top" to 0em, I get:

![image](https://cloud.githubusercontent.com/assets/77951/3882824/ffe63f06-219a-11e4-9c61-318b98994aea.png)

Setting "margin-top" to 0em lifts the number a little bit, but not enough.

A second issue is that resizing to a smaller window results in line wrapping of the code, but the line numbers don't change to accommodate:

![image](https://cloud.githubusercontent.com/assets/77951/3882892/81ef5aaa-219b-11e4-82a9-d82a893ff1a6.png)

Swissa:
You can add the following css:

pre {
  overflow: auto;
  word-wrap: normal;
  white-space: pre;
}

Golmote:
This should be fixed by #732.

PR #732, opened 2015-08-29
Issue 324 - Fixing word wrap on long code lines, which is breaking the line-numbers plugin

AnujRNair:
This commit fixes **part** of Issue https://github.com/PrismJS/prism/issues/324, which is breaking the line-numbers plugin when working with bootstrap.

Currently, if a line overflows onto the next line, the line numbers will not space out accordingly to compensate for this overflow.

This change rectifies this by not breaking long lines onto multiple lines
(It will be up to the developer to break their code onto multiple lines if they do not want the overflow)

### DIFF
--- a/themes/prism-coy.css
+++ b/themes/prism-coy.css
@@ -13,6 +13,7 @@ pre[class*="language-"] {
 	white-space: pre;
 	word-spacing: normal;
 	word-break: normal;
+	word-wrap: normal;
 	line-height: 1.5;
 
 	-moz-tab-size: 4;
--- a/themes/prism-dark.css
+++ b/themes/prism-dark.css
@@ -14,6 +14,7 @@ pre[class*="language-"] {
 	white-space: pre;
 	word-spacing: normal;
 	word-break: normal;
+	word-wrap: normal;
 	line-height: 1.5;
 
 	-moz-tab-size: 4;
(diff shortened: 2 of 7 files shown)


Golmote:
The `white-space: pre` property already prevents long lines from wrapping, doesn't it?

AnujRNair:
It will, however bootstrap sets ```word-wrap: break-word```, which means that if wrapping is needed, a word (or unbroken string) will be broken to allow text to wrap so it fits in it's container. 

From my understanding, the ```white-space: pre``` property only stops word wrapping on white spaces!

Setting both means that lines will overflow correctly and not break on whitespace, or words

Golmote:
Ok, I get your point. But I cannot reproduce the behaviour you're describing in Prism. Adding `word-wrap: break-word` has no effect when `white-space` is set to `pre`: see [this Fiddle](http://jsfiddle.net/8ru3e79w/).

Unless it is only happening in a specific browser?

AnujRNair:
I am using Chrome Version 44.0.2403.157 (64-bit) on a Mac.

I see these for ```break-word```:

1)
screen shot 2015-08-29 at 1 29 10 am
2)
screen shot 2015-08-29 at 1 29 27 am

Isn't the goal to always keep a line of text on one line, and not wrap it, so that the line numbers render correctly next to their respective lines?

Apologies if I am being confusing, or I am missing something!

Golmote:
No no, it's fine, I just couldn't reproduce the bug in Firefox. But you are right that it happens in Chrome.

Thanks for fixing this!

AnujRNair:
Awesome, thanks!

PR #750, opened 2015-09-01
Add the chinese readme translation

bimohxh:
awesomes.cn is an open source for github web repos, we offer translation for repo's readme and improve this to let our chinese users use the repo easily.In addition, we want more more developer join us to enrich the chinese readmes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # [Prism](http://prismjs.com/)
+[![中文说明](http://awesomes.oss-cn-beijing.aliyuncs.com/readme.png)](http://www.awesomes.cn/repo/PrismJS/prism)
 
 Prism is a lightweight, robust, elegant syntax highlighting library. It's a spin-off project from [Dabblet](http://dabblet.com/).
 


LeaVerou:
Hey, thanks for your work!
I might be missing something, but the website you linked to just has the same info as the readme here, in English (not Chinese). What am I missing?

awesomes-cn:
Just visit http://www.awesomes.cn/repo/PrismJS/prism you see the same info   as the readme here,because your browser language is English.Our website support both Chinese and English language,so we will detect the browser language ,if it is Chinese we will default show Chinese translation,otherwise show the English readme.And we offer a language switch on the top right as `EN / 中`,you can click `中` to read Chinese version.

brutalcrozt:
ping to update

LeaVerou:
Hi there @awesomes-cn, even when I select Chinese, there is a large blob of untranslated text and it's only very short bits of text that have been translated. Am I missing something?

LeaVerou:
Hi there,

Sorry for the delay.
I would be inclined to merge this, but not as the first thing immediately after the heading. How about a "Translations" section at the end of the readme, with a list that has this as the only item?

mAAdhaTTah:
@LeaVerou If we moved it to a list, should we keep the badge?

LeaVerou:
Sure, why not.

mAAdhaTTah:
@bimohxh Landed in https://github.com/PrismJS/prism/commit/f2b19644931a9b13428aab08839fada6ca3d9b46 – thank you for the contribution!

PR #754, opened 2015-09-02
Fix: node.js hacks for components and prism.js

valorkin:
@Golmote this 2 small hack allow to execute prismjs from nodejs without exceptions
I am using prismjs as loader for webpack
https://github.com/valor-software/prismjs-loader

### DIFF
--- a/components/prism-core.js
+++ b/components/prism-core.js
@@ -426,3 +426,8 @@ return _self.Prism;
 if (typeof module !== 'undefined' && module.exports) {
 	module.exports = Prism;
 }
+
+// hack for components to work correctly in node.js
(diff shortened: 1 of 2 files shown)


Golmote:
Please do not modify `prism.js` directly, it is used mainly for the website. Can you update the PR to modify the files `components/prism-core.js`, `plugins/file-highlight/prism-file-highlight.js`, etc. separately?

You can then run the default gulp task to build the minified versions and the file `prism.js`.

valorkin:
@Golmote updated, please review

Golmote:
Thanks for contributing ;)

PR #762, opened 2015-09-09
Fixing coy theme + line numbers plugin overflowing on long blocks of text

AnujRNair:
Currently, with the coy theme and line numbers plugin, due to the before and after pseudo elements, long blocks of text overflow outside of the code blocks like so:

screen shot 2015-09-08 at 10 42 48 pm

This commit fixes this, making the code element's max height slightly smaller than the pre element's max height. All code is shown properly, and code no longer overflows outside of the pre element.

Seen on Mac, Chrome Version 45.0.2454.85 (64-bit)

### DIFF
--- a/themes/prism-coy.css
+++ b/themes/prism-coy.css
@@ -44,10 +44,11 @@ pre[class*="language-"] {
 	background-origin: content-box;
 	overflow: visible;
 	max-height: 30em;
+	padding: 0;
 }
 
 code[class*="language"] {
-	max-height: 28.5em;
+	max-height: inherit;
 	height: 100%;
 	padding: 0 1em;
 	display: block;
--- a/themes/prism-coy.css
+++ b/themes/prism-coy.css
@@ -47,7 +47,7 @@ pre[class*="language-"] {
 }
 
 code[class*="language"] {
-	max-height: inherit;
+	max-height: 28.5em;
 	height: 100%;
 	padding: 0 1em;
 	display: block;


Golmote:
Hi @AnujRNair. Thanks for contributing again.
I am unable to reproduce the issue in the first place, though. Can you please provide an online example of the problem, before I merge the fix?

AnujRNair:
Hi @Golmote, no problem! Im loving PrismJS :)

I've realised the problem actually exists because Bootstrap3 is applying some styling to the ```pre``` blocks, as seen here:

http://codepen.io/anon/pen/jbPjwm

2 options can help solve this:
 * The proposed commit
 * Setting ```pre[class*="language-"] { padding: 0; }``` to reset the bootstrap styling

Solution 1 leaves some nice space at the top and bottom of the code block.
Solution 2 resets the problematic styling, however, doesn't leave any spacing at the top and bottom of the code block

My proposed solution would be solution 2, as this fixes the actual issue, and then allow the developer to add more styling on their own in they would like to fix spacing.

If you're happy with solution 2, I can kill this merge request and create a new one?

Golmote:
I'll go with solution 2 as well. The creator of this theme did not include spacing, so let's respect that choice and only fix the issue.

You don't have to create a new PR, though. Just update your fork and this PR will be updated automagically.

AnujRNair:
Thanks, done!

PR #767, opened 2015-09-12
Plugin: color previewer

Golmote:
This was obviously inspired by PR #372, though I tried to make a more generic implementation, like Dabblet does, so that we can easily add more previewers in the future.

13ca902c57

It is compatible with CSS, Markup attributes (like the old bgcolor), Less, Sass, Scss and Stylus.


LeaVerou:
Wow, amazing!!

LeaVerou:
Merged, thank you so much!!

PR #775, opened 2015-09-22
Highlight Elixir string markdown header correctly

danielberkompas:
Elixir allows markdown inside its triple-quote heredoc syntax. However,
the "#" used in markdown headers is currently misinterpreted as a
comment. So, if you write this:

```elixir
@doc """
Get the first name of a user.

## Parameters

- `user` - A User struct.

## Examples

    user = %User{name: "Alice Winston"}
    User.first_name(user)
    "Alice"

"""
```

Prism interprets the markdown headers as comments, not as part of the
string, and this messes up the coloring.

04__functions__modules__and_structs___learnelixir_tv

It's easy to fix, just put the string regular expressions before the
comment regular expression in the elixir component.


Golmote:
Unfortunately, this will lead to other issues (e.g. strings will be highlighted inside comments (breaking them))...
Comments might be used to comment code, and that's why we usually handle it first.

Would it be ok to only change the negative look-ahead in the comment pattern, so that we ignore comments starting with more than one hash?  `/#(?![{#]).*/`

danielberkompas:
I'll give it a try and let you know.

danielberkompas:
Changing the negative lookahead wasn't enough. If it was given this:

```elixir
@doc """
## Header
"""
```

It would match on `# Header`. Since Javascript doesn't support negative lookbehinds, I added a negated capture group to assert that the `#` is not preceded by any other `#`s. What do you think of this solution?

Golmote:
Damn you're perfectly right! Please use the `lookbehind` feature provided by Prism, though, so that the char before the hash is not consumed:

```javascript
'comment': {
	pattern: /([^#])#(?![{#]).*/,
	lookbehind: true
}
```

Can you also fix the indentation before the comments you added?

I can merge this once it's done ;)

danielberkompas:
Done! I amended the commit.

Golmote:
Thanks!

Golmote:
Oh, I have to add an alternative to handle the empty comment `#` and pass the tests.
The final pattern is: `/(^|[^#])#(?![{#]).*/m`

danielberkompas:
How do you run the tests?

Golmote:
By running `npm test`

danielberkompas:
I'm fixing this now.

Golmote:
It's fixed already! See ccb6566ccb14f365ab221525f09247d775a9bd06

danielberkompas:
:+1:

PR #776, opened 2015-09-22
Add Nginx

westonganger:
Adding Nginx highlighting.
Extracted from this gist https://gist.github.com/misiek08/6566480


Golmote:
Thanks for contributing! Would you mind adding an example file, that reflects most of the features and flaws, following the format of [the existing ones](https://github.com/PrismJS/prism/tree/gh-pages/examples) ?

Golmote:
Also, please keep your changes on separated branches, so that this PR contains only Nginx-related stuff. You just pushed vim support ^^

components/prism-nginx.js
```diff
@@ -0,0 +1,18 @@
+
+Prism.languages.nginx = Prism.languages.extend('clike', {
+        comment: {
+                pattern: /(^|[^"{\\])(#.*?(\r?\n|$))/,
```
Golmote: I think you'll need a `lookbehind: true` here, so that no char is consumed before the comment.
Also, you probably don't need the last group ; `/(^|[^"{\\])#.*/` should be enough (since it does not match line feeds).

components/prism-nginx.js
```diff
@@ -0,0 +1,18 @@
...
...hreshold|rtsig_signo|satisfy|satisfy_any|secure_link_secret|send_lowat|send_timeout|sendfile|sendfile_max_chunk|server|server_name|server_name_in_redirect|server_names_hash_bucket_size|server_names_hash_max_size|server_tokens|set|set_real_ip_from|smtp_auth|smtp_capabilities|so_keepalive|source_charset|split_clients|ssi|ssi_silent_errors|ssi_types|ssi_value_length|ssl|ssl_certificate|ssl_certificate_key|ssl_ciphers|ssl_client_certificate|ssl_crl|ssl_dhparam|ssl_engine|ssl_prefer_server_ciphers|ssl_protocols|ssl_session_cache|ssl_session_timeout|ssl_verify_client|ssl_verify_depth|starttls|stub_status|sub_filter|sub_filter_once|sub_filter_types|tcp_nodelay|tcp_nopush|timeout|timer_resolution|try_files|types|types_hash_bucket_size|types_hash_max_size|underscores_in_headers|uninitialized_variable_warn|upstream|use|user|userid|userid_domain|userid_expires|userid_name|userid_p3p|userid_path|userid_service|valid_referers|variables_hash_bucket_size|variables_hash_max_size|worker_connections|worker_cpu_affinity|worker_priority|worker_processes|worker_rlimit_core|worker_rlimit_nofile|worker_rlimit_sigpending|working_directory|xclient|xml_entities|xslt_entities|xslt_stylesheet|xslt_types)\b/i,
```
Golmote: `$` and `@` will only match when they are surrounded by word characters, since they are themselves non-word chars. Is that what you expect ? If not, they should not be included in the same group as the other keywords.

components/prism-nginx.js
```diff
@@ -0,0 +1,18 @@
...
+});
+
+Prism.languages.insertBefore('nginx', 'keyword', {
+        variable:/(\$[a-zA-Z_]+)/ig,
```
Golmote: You don't need the parentheses around the whole pattern. Also, you're using the `i` flag so you can simplify the character class a bit. Finally, there is no need for the `g` flag in Prism (it is not used).

components/prism-nginx.js
```diff
@@ -0,0 +1,18 @@
...
+        'package': {
+                pattern: /(\\|namespace\s+|use\s+)[\w\\]+/,
+                lookbehind: !0,
```
Golmote: Please use `true` here. We can leave this kind of compression for the generated minified version.

Golmote:
I did a quick review. Please take a look at my inline comments.

PR #778, opened 2015-09-23
Improve performance by doing more work in the worker

zeitgeist87:
Currently the worker tokenizes the code and then turns it into a
JSON string. This JSON string is sent back to the main thread as a
message and parsed. Then it is turned into a HTML string and
stored in the attribute env.highlightedCode.

These conversions into JSON are unnecessary, because we can create
the HTML string directly in the worker and send it as a message to
the main thread.

Signed-off-by: Andreas Rohner <andreas.rohner@gmx.net>


Golmote:
Hi! Nice catch, indeed. Thanks for contributing!

PR #779, opened 2015-09-24
Partial solution for the "Comment-like substrings"-problem

zeitgeist87:
This patch introduces a new attribute called `greedy`. The
attribute is a simple boolean flag. If there is no match
for a greedy pattern it can concatenate the next two tokens
into a single string and try to match on this string again.
If a match is found on the second attempt, then the old tokens are
deleted and replaced by the new match.

This solves the "Comment-like substrings"-problem for exactly one
comment at very little cost. With this patch the following code is
highlighted correctly:

```js
"foo /* bar */ baz"; "foo // bar";
/lala"test"sdf/;
```

This approach fails if there are more than one comments inside the
string:

```js
"foo /* bar */ baz /* bar */ baz";
```


zeitgeist87:
My previous attempts at this (#444, #334) were too big in terms of code size and the performance impact was too large. This time I reduced it to the bare minimum (41 lines) and the performance should be almost identical. It is only a partial solution, but it should fix most cases, where a comment-like substring appears unintentionally in a piece of code.

Golmote:
Don't think I didn't care about this PR. I just know @LeaVerou has a strong opinion about this so I consider only her can take a decision. I hope you're not closing because of a lack of interest.

zeitgeist87:
Sorry I closed it by accident. I locally updated my gh-pages branch to the newest version of prism and deleted this commit in the process. I should have put the commit in a separate branch. It is now reopened.

zeitgeist87:
Hi @LeaVerou, there seems to be renewed interest in solving this issue #307. I think this PR is a compromise between a full blown LR parser, which is what @foo123 seems to be suggesting, and the regex only approach of Prism. The performance impact should be negligible.

foo123:
@zeitgeist87, the prism-grammar suggestion is just an add-on (which handles these cases among others) but for a solution to this issue with prism itself, i suggested `delimited-block tokens` (as in here https://github.com/PrismJS/prism/issues/412), which simply require a small state machine (since they have to keep state untill their pattern closes), instead of a parser machine (as in the add-on)

zeitgeist87:
@foo123 sorry I didn't mean to misrepresent your prism-grammar project.

Concerning your suggestion of `delimited-block tokens`. I think something like this was already discussed a few months ago and the consensus was, that the added size and complexity isn't really worth it. Prism aims to be a pure small and fast regex highlighter. It does not have to be totally correct. If you want to parse a language without any errors, you need something at least as powerful as a LR(1) parser and Prism is not designed to do that.

foo123:
@zeitgeist87 no problem

Agree "*Prism aims to be a pure small and fast regex highlighter*", but disagree
"*It does not have to be totally correct*". One is not asking to do context-sensitive parsing but this is about lexical analysis, highlight comments CDATA, strings, literal regexes correctly (which are all over the place). If this is not done it is not good (and of course Prism is not alone in that, and afaik many people are not happy with these highlighters not highlighting). 

However since these tokens are `blocks` (in the sense mentioned earlier) it needs a kind of state to be kept so "parser" knows it is inise a block and match.

And it is not much additional code (for what i suggested dont know about other approach). One only has a generic block tokenizer and just passes to it the different start and end paterns (for the different block tokens represented in the language model), so it is just one additional function (and a small one that is)

A highlighter **needs to highlight** not parse in general, however if highlighting requires some parsing to be done correctly it should be done. Let me give a personal example, i was creating a template engine and wanted to avoid having a full parser for speed and size reasons, however i indeed implemented all the parser parts necessary for the engine to do what it was supposed to do (instead of working half the time, lets say) And it still does not have a full parser (for example function arguments are not tokenised separately but passed as one argument, but there is no problem there for what the engine is supposed to achieve, see what i mean).

zeitgeist87:
@foo123 Yes a full blown parser is probably not necessary. I would love to see something like delimited-block tokens for prism. You could implement it and make a PR...

foo123:
@zeitgeist87 
If the devs here want me to do a PR for this i could try, else it is just a suggestion, among others. Cheers!

LeaVerou:
Sorry I'm so late to this!
Great idea!!
Does this solve the `"foo /**/ bar"` problem? I think this type of thing is our most common highlighting fail that I see on random websites using Prism (especially recently, as I was reading gulp tutorials :p ).
In terms of size or quality, the code looks ok to me. :) I don’t currently have the time to read it carefully though. @Golmote, what do you think?

LeaVerou:
Btw @Golmote, if you, applying your strictest code review criteria, think this is ok (and it does indeed fix the `"foo /**/ bar"` problem), you have the ok from me to merge it :)
(Yoohoo, 3 years later we’re finally fixing this issue!! (at least for most cases). So happy!)

foo123:
@LeaVerou if you are refering to my last comment/suggestion, yes it solves that problem exactly with what is described

(if not ignore this comment)

LeaVerou:
@foo123 Another reason that Prism does not work with a parser, is that a parser generally fails in invalid code. A syntax highlighter needs to work well with invalid code (i.e. highlight most of it as correctly as possible), as any code being typed in an editor goes through invalid states. Not to mention that tons of code examples on the Web are actually invalid. Prism is not a linter, neither does it need to be. I see Prism all the time in websites I visit, and I notice very few failures (and almost always of the kind this PR solves).
That said, I would like being able to define tokens by start, end and escape and a state machine with 2 states shouldn’t take too much code. If you want to give it a shot, a PR would be nice. However, if it’s a huge PR that adds a couple hundred lines, chances are it won’t be merged. Not only because it increases complexity and download size, but also because every line added to Prism Core increases the effort required for future contributors and maintainers. We are very liberal with accepting PRs for language definitions, styles and plugins, but very strict when it comes to Core changes, especially extensive ones.

foo123:
@LeaVerou , ok i'm working on sth ele right now but i will make a PR as soon as i can for the delimited tokens suggestion

About the other things, every highlighter (invluding Prism) does some klind of parsing. For example using just regexes is parsing. The way invalid code is highlighted is an option (for example it can be bypassed, or highlighted as is). Personaly i would prefer the highlighter to highlight correct code and highlight incvalid code in a different way (i think this is the *reason d' etre* of highlighting, **highlight what is relevant to a given language semanticaly**, but opinions can vary on this)

zeitgeist87:
> Btw @Golmote, if you, applying your strictest code review criteria, think this is ok, you have the ok from me to merge it :)
> (Yoohoo, 3 years later we’re finally fixing this issue!! (at least for most cases). So happy!)
> 
> Edit: Just looked at the samples more clearly, yup, it definitely does work on "foo /**/ bar". Awesome!

@Golmote If you can find the time, could you please review this PR. I think it is a nice solution and it has potential, but it is also quite complicated and definitely needs a review. There is no rush, though. I just wanted to "ping" the PR again.

LeaVerou:
I hope we can get this merged at some point, I really like this functionality.
@zeitgeist87, it would make it easier to review if you added some comments though. It's really hard to follow as it currently stands. (and yes, I'm fully aware that I don't add enough comments either, I'm not saying it won't be merged without them, just that it might speed things up)

Golmote:
I'll make my best to find some time to review this properly by the end of the week.

zeitgeist87:
> it would make it easier to review if you added some comments though. It's really hard to follow as it currently stands.

To be honest I find it hard to understand my own code after all this time :smile:. I will review it again and add more comments.

LeaVerou:
> To be honest I find it hard to understand my own code after all this time :smile:

Yeah, this is exactly why comments are useful :P

zeitgeist87:
> I'll make my best to find some time to review this properly by the end of the week.

Thanks! It's really appreciated. I'll add more comments to make the intended functionality clear.

components/prism-core.js
```diff
@@ -294,7 +295,46 @@ var _ = _self.Prism = {
...
+						delNum = 3;
+
+						if (to <= len) {
+							delNum = 2;
+							combStr = combStr.slice(0, len);
+						}
+						str = combStr;
+					}
 
 					if (match) {
```
Golmote: For consistency with line 315, I would have probably changed this condition to `if (!match) { continue; }`

Golmote:
Ok so apart from my tiny comment related to the condition, this looks good to me. I still feel quite frustrated by the fact we aren't actually completely solving the issue, but I guess this is better than nothing.

Maybe you could add a few tests for the C-like and the JS components, so that we can ensure it works and it will continue to work in the future. ;)

zeitgeist87:
> Ok so apart from my tiny comment related to the condition, this looks good to me.

Nice! Thanks for the review.

> I still feel quite frustrated by the fact we aren't actually completely solving the issue, but I guess this is better than nothing.

I tend to see it more like the perfect compromise. Prism is supposed to be lightweight.

> Maybe you could add a few tests for the C-like and the JS components, so that we can ensure it works and it will continue to work in the future. ;)

I'll get right on to it.

LeaVerou:
Is this completely backwards compatible? Can we think of any cases that would be highlighted correctly in the past but not after this is merged?

components/prism-core.js
```diff
@@ -294,7 +295,46 @@ var _ = _self.Prism = {
...
+						// To be a valid candidate, the new match has to start
+						// inside of str
+						if (from >= str.length) {
+							continue;
+						}
+						var to = match.index + match[0].length,
+						    len = str.length + nextToken.length;
+
+						// Number of tokens to delete and replace with the new
+						// match
```
LeaVerou: Small remark: Readability is more important than strict adherence to a certain line length. Please don't break comments in multiple lines unless the second line is actually long enough to warrant this.

zeitgeist87:
> Is this completely backwards compatible? Can we think of any cases that would be highlighted correctly in the past but not after this is merged?

I can't think of any situation where it would make things worse. It also passes all of the tests in our test-suite.

zeitgeist87:
I am really excited about this. I've fixed the small issues you pointed out. Can we merge it?

zeitgeist87:
While writing the documentation for the greedy-property, I had an idea for a better name. What do you think of the name `backtrack: true`?

LeaVerou:
Yay! Congrats team!! This is fantastic!
Neither `greedy` nor `backtrack` describe what it does, just how it works. I think it's worth it to bikeshed the name a bit, as we can't change it easily once people start using it.

zeitgeist87:
How about `ignoreOtherMatches` or something shorter along those lines.

BTW, I love the expression "to bikeshed something". I had never heard it before. I had to look it up on Wikipedia.

zeitgeist87:
How about `important: true` or `privileged: true`

LeaVerou:
Haha, we use that expression all the time in the CSS WG to mean "endless discussions about the name of a new feature". :)

`important` or `privileged` are on the right track in the sense that they attempt to describe the purpose instead of the way, but are still confusing.

Just throwing some ideas in the brainstorming: What about something centered around nesting? You would use this flag in tokens that can be nested inside other tokens, right? If not, can we find a general description about which tokens this will be used in? It will help us a lot in coming up with a good name.

LeaVerou:
>  this property should be set on every token, that may contain other tokens.

Well, every token that can contain other tokens which could also contain the same token. Otherwise one could just nest it in the grammar. 

I think we're on to something here :)

zeitgeist87:
Here is my description:

It is for tokens, that match arbitrary content between start and end markers, because the content can contain stuff that could be matched by other tokens. The Problem is, that Prism has to execute its tokens in order. It the token in the content matches first, the outer token cannot match any more.

LeaVerou:
> It is for tokens, that match arbitrary content between start and end markers, because the content can contain stuff that could be matched by other tokens. The Problem is, that Prism has to execute its tokens in order. It the token in the content matches first, the outer token cannot match any more.

I have no idea what this is saying, and I've even seen the code.
I think you're too close to the implementation right now to write good docs. This is not a flaw, it happens to almost everyone.
Ideally, someone else should write the docs for this, and you could review to correct factual mistakes.
I might give it a shot later on, if nobody is available, but I would be very grateful to anyone who offers.

LeaVerou:
How about:

Enable this flag to cancel out tokens that come earlier in the language definition if they appear inside this token. For example, comments are parsed first, but if the string `/* foo */` appears inside a string, you would not want it to be highlighted as a comment. Use this flag with restraint, as it incurs a small performance overhead.

Just an idea to get us started, I'm not saying it's perfect!

zeitgeist87:
> I have no idea what this is saying, and I've even seen the code.

It's probably because of my bad English. At least in part...

> Ideally, someone else should write the docs for this

Probably a good idea, but the above text, was not my proposal for the documentation. This is my draft:

```html
<dt>greedy</dt>
		<dd>This is a boolean attribute. It is intended to solve a common problem with
		patterns that match long strings like comments, regex or string literals. For
		example a comment enclosed by a string should be highlighted as a string.
		However, if the pattern for comments is executed before the pattern for strings, then
		the comment is highlighted and the string cannot match anymore.
		<pre><code class="language-javascript">"my /* comment */ string"</code></pre>
		The greedy-property allows a pattern to ignore previous matches of other patterns, and
		overwrite them when necessary. The following example demonstrates its usage:
		<pre><code class="language-javascript">'string': {
	pattern: /(["'])(\\(?:\r\n|[\s\S])|(?!\1)[^\\\r\n])*\1/,
	greedy: true
}</code></pre>
```

LeaVerou:
The English was correct, the problem was it read like code written in natural language.
The docs you're suggesting are much better though, so I retract my statement. :) 

Now if only we can decide on a good name :)

Golmote:
container? Cause it may contain other tokens?

zeitgeist87:
`nonesting` because we want to prevent the nested tokens from being highlighted...

zeitgeist87:
I still think `important` is quite good, because it kind of has the same meaning in CSS. The order of the rules in CSS is significant. The rules further down overwrite earlier properties. With `!important` the rule cannot be overwritten any more by following rules.

It is kind of the same with Prism, only in reverse. The rules that come first are the most important ones. With the property `important: true` we can increase the importance of rules lower down in the chain.

zeitgeist87:
We could even make it an integer `important: 5`, so that a token can only replace less important tokens.

LeaVerou:
I think `important` is a bit confusing. 
`nonesting` is hard to parse (I first read it as "none sting" and was scratching my head as to what you meant)

zeitgeist87:
What about `global: true`? Instead of matching against local string snippets between previous matches, it matches globally on the whole text. In the process it overwrites previous matches that happen to conflict with the new match.

zeitgeist87:
I've just found a bug related to this. It is quite limited, because the second slash has to be followed by `[,.;]`:

```js
var test = 1 / 4 + "/,  bug";
```

This `/ 4 + "/` is incorrectly highlighted as a regex. A possible fix would be, to prevent greedy patterns from deleting tokens created by other greedy patterns. I'll work on it.

PR #780, opened 2015-09-26
Crystal support

MakeNowJust:
I added [Crystal](http://crystal-lang.org/) language definition, tests and examples.


components/prism-crystal.js
```diff
@@ -0,0 +1,54 @@
+(function(Prism) {
+	Prism.languages.crystal = Prism.languages.extend('ruby', {
+		number: /\b(?:0b[01_]*[01]|0o[0-7_]*[0-7]|0x[0-9a-fA-F_]*[0-9a-fA-F]|(?:[0-9][0-9_]*[0-9]|[0-9])(?:\.[0-9_]*[0-9])?(?:[eE][+-]?[0-9_]*[0-9])?)(?:_[uif](?:8|16|32|64|))?\b/,
```
Golmote: This part `(?:[0-9][0-9_]*[0-9]|[0-9])` could probably be rewritten as `[0-9](?:[0-9_]*[0-9])?`.
Also, the last group `(?:8|16|32|64|)` might be more explicit written as `(?:8|16|32|64)?`.

components/prism-crystal.js
```diff
@@ -0,0 +1,54 @@
...
+	});
+
+	Prism.languages.insertBefore('crystal', 'operator', {
```
Golmote: Why do you use insertBefore to handle keywords instead of just overriding it in the `extend` call?

components/prism-crystal.js
```diff
@@ -0,0 +1,54 @@
...
+		keyword: [
+			/\b(?:abstract|alias|as|asm|begin|break|case|class|def|do|else|elsif|end|ensure|enum|extend|for|fun|if|ifdef|include|instance_sizeof|lib|macro|module|next|of|out|pointerof|private|protected|rescue|return|require|self|sizeof|struct|super|then|type|typeof|undef|union|unless|until|when|while|with|yield|__DIR__|__FILE__|__LINE__)\b/,
+			{
+				pattern: /(\.\s*)(is_a|responds_to)\?/,
```
Golmote: Because you took care of making all others groups non-capturing where not needed, may I suggest you add `?:` to the second group here too?

examples/prism-crystal.html
```diff
@@ -0,0 +1,29 @@
...
+	end
+{% end %}</code></pre>
+
+<h2>Known failures</h2>
+<p>There are certain edge cases where Prism will fail.
+	There are always such cases in every regex-based syntax highlighter.
+	However, Prism dares to be open and honest about them.
+	If a failure is listed here, it doesn’t mean it will never be fixed. This is more of a “known bugs” list, just with a certain type of bug.
+</p>
+
```
Golmote: Aren't there also failures when strings contain attribute-like or expansion-like substrings?

Golmote:
Hi! Thanks for contributing. With example and tests, this is a great PR!

Would you mind taking a look at the few comments I made?

MakeNowJust:
@Golmote Thanks for reviewing my code.  I fixed them.

components/prism-crystal.js
```diff
@@ -0,0 +1,53 @@
+(function(Prism) {
+	Prism.languages.crystal = Prism.languages.extend('ruby', {
+		keyword: [
+			/\b(?:abstract|alias|as|asm|begin|break|case|class|def|do|else|elsif|end|ensure|enum|extend|for|fun|if|ifdef|include|instance_sizeof|lib|macro|module|next|of|out|pointerof|private|protected|rescue|return|require|self|sizeof|struct|super|then|type|typeof|undef|union|unless|until|when|while|with|yield|__DIR__|__FILE__|__LINE__)\b/,
+			{
+        pattern: /(\.\s*)(?:is_a|responds_to)\?/,
```
Golmote: Can you fix the indentation of this line? It needs 4 tabs.

Golmote:
Thanks. For some reason, the indentation of one line went crazy, can you check it out? Once this is fixed, I can merge the PR.

components/prism-crystal.js
MakeNowJust: Sorry. I'll fix it.  My editor uses spaces as indent. Why doesn't this project use linter or `.editorconfig`?

Golmote: Well, there is no reason. And this is a good idea. I'll definitely think about adding it. ;)

Golmote: I just added a basic .editorconfig file. Thanks for suggesting this.

PR #783, opened 2015-09-30
Plugin: gradient previewer

Golmote:
Previewer for gradients. It should support all gradients syntaxes but the old webkit one (`-webkit-gradient()`). Old syntaxes are converted to the W3C-valid one.

The plugin will work only in browsers supporting the W3C syntax, unprefixed. 

prism-previewer-gradient
prism-previewer-gradient-repeating


nauzilus:
Should this also work alongside with previewer-colour? It seems which ever comes first out of gradient and colour will take precedence:
* [Colour before gradient](http://jsfiddle.net/nauzilus/75nLeojy/)
* [Gradient before colour](http://jsfiddle.net/nauzilus/75nLeojy/1/)

It might just be one of those edge cases like strings inside comments that can't really be helped, just thought I'd bring it up :)

Golmote:
Thanks for reporting this behaviour. I think gradient should always take precedence if included (no matter the order). I'll see if I can fix this.

Golmote:
@nauzilus The issue should be fixed. Now the gradient previewer should take precedence over the color previewer if both are included, even if the color previewer is included first.

PR #784, opened 2015-09-30
Show Language: fix the problem of language label scrolling out of view with code.

chaosshen:
Fix this problem by moving the label to a previous sibling of 'pre'.

I've made tests with Line Number plugin and it works fine.

### DIFF
--- a/plugins/show-language/prism-show-language.css
+++ b/plugins/show-language/prism-show-language.css
@@ -1,8 +1,8 @@
-pre[class*='language-'] {
+div.prism-show-language {
 	position: relative;
 }
-pre[class*='language-'][data-language]::before {
-	content: attr(data-language);
+
+div.prism-show-language > div.prism-show-language-label[data-language] {
 	color: black;
 	background-color: #CFCFCF;
 	display: inline-block;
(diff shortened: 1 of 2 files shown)


Golmote:
Hi! Thanks for contributing!

Your idea is really smart, I like it! I noticed that when the hook executes twice, the extra divs are inserted twice per block too. Could you make some check to prevent this?

(You can reproduce the bug on the Show Language plugin page, because the File Highlight plugin runs the highlighting twice per block)

chaosshen:
Yeah, this bug can be fixed by checking the extra divs' existence and I'll add it in the next commit.

On the Show language plugin page I found another bug in my code. The label doesn't move with a nonzero `margin-right` of pre. What's more, if I limit the height of pre, the label will cover a part of vertical scroll bar... I still can't solve them. Do you have any good idea?

chaosshen:
I've found the solution to this not using the extra divs! Add the following code to `prism-show-language.css`:

```CSS
pre[class*='language-'][data-language] {
	padding: 0;
	position: relative;
	/* hide pre's scroll bar */
	overflow: hidden;
}

code[class*="language"] {
	padding: 1em;
	display: block;
	/* use code's scroll bar */
	overflow: auto;
}
```

Then `pre::before` won't be affected by scrolling of `code`. The solution is from theme coy.

However, part of the vertical scroll bar may still be covered by the label. One way is to add a `margin-right: 1em;` to `pre::before` and set `overflow-y: scroll;` to `code`, which will always show the vertical scroll bar even if no need.

In addition, it's best to set `code[class*="language-"]` the same `max-height` as `pre` in `style.css` when applying `display: block;` like above.

Please close this PR and I'll open another. Or you do it yourself?

Golmote:
Sorry it didn't work.
So right now, we're back with the first solution and its two issues right? I took some time to think about them but I can't find a way to solve them.

PR #787, opened 2015-10-01
Add support for Here-Documents for the Bash language

zeitgeist87:
This patch adds support for Here-Documents for the Bash language.
A Here-Document should be highlighted as a multiline string
and looks like this:

```bash
cat << EOF
foo bar
EOF
```

I have also cleaned up the Bash language a little bit. I have put those changes into separate commits to make the review easier.


Golmote:
Thank you for working on this.

Can you rename the test file from `important_feature.test` to `shebang_feature.test`?

zeitgeist87:
> Can you rename the test file from important_feature.test to shebang_feature.test?

Sure, I just updated the PR.

Golmote:
Thanks!

sunknudsen:
@zeitgeist87 

Thanks for this PR! Do you know how to fix here-document support in the shell-session language?

Currently, everything after `cat` is registered as the output.

here-document

PR #789, opened 2015-10-02
Remove Clike dependency from Bash

zeitgeist87:
The Bash language is actually very different from Clike. Only
'boolean', 'operator' and 'punctuation' are shared between them.
'operator' and 'punctuation' should be changed in the future,
because Bash actually has slightly different operators.

Since they share almost nothing, it makes sense to remove the
dependency. This patch copies 'boolean', 'operator' and
'punctuation' from Clike, but doesn't change them.

This patch also fixes a typo and adds the missing built-in keyword
`let`. `let` is not in the POSIX-standard, but Bash and some other
shells support it.


Golmote:
Thanks!

PR #790, opened 2015-10-05
Plugin: time previewer

Golmote:
Previewer for times (e.g. durations).

The plugin does not work in IE.

prism-previewer-time


nauzilus:
These are brilliant :+1:

PR #792, opened 2015-10-05
Fix regression in Bash strings

zeitgeist87:
In a recent patch the regex for single-quoted and double-quoted
strings of the Bash language was split into two separate regexes
to prevent the highlighting of variables inside of single-quoted
strings.

This causes a bug whenever a double-quoted string apears inside a
single-quoted string, because the double-quoted string is matched
first.

```bash
'"foo"'
```

The same problem exists with the newly introduced Here-Documents.
This patch fixes the problem by matching Here-Documents first and
merging the regexes for single-quoted and double-quoted
strings again.

This patch also adds testcases for this, to prevent future
regressions.


zeitgeist87:
The regression was introduced by my latest patch series for the Bash language. Sorry for the mess :(

Golmote:
Thanks for fixing it

Issue #796, opened 2015-10-07
Most languages not working (e.g. Python)


fladd:
I selected all languages on the download page, then downloaded the corresponding js and css files.

Only the "built-in" languages are highlighted (e.g. language-clike), but not the extra ones (e.g. language-python does not work!), even though they are part of the js file.

I tried both, development and minified version.

I use the default theme.

nauzilus:
Woah :(

Bison extends the language **c**, but Prism is loads bison first (since `components.js` is in alphabetical order), and the **c** grammar hasn't yet been defined, so fails for everything **bison**.

I think **crystal** is also going to be a problem, because it loads **ruby**.

I'm at work so can't fix this at the moment. I think Prism (specifically `download.js`) is going to need to be smarter about handling dependencies. To make sure they're loaded in an appropriate manner.

Golmote:
This is fixed now #797 is merged.
Thanks for reporting!

PR #797, opened 2015-10-07
Fixed components order.

podgorskiy:
Bug: If support of **Bison** is selected, highlightening stops working. And again, this mean  if I check "all" checker, highlightening does not work either, because of **Bison**.

It happens due to that **Bison** requires **C** and goes before **C**.

In case if one component requires other, the required component should go first. 
In this fix components are sorted regarding their requirements.

In other case, as it happens with **Bison**, that goes before **C**, an exception is thrown at these lines:

	'comment': Prism.languages.c.comment,
	'string': Prism.languages.c.string,

As *Prism.languages.c* is not defined yet.

Limitations: The fix does not take into account case of circular requirement, that doesn't make much sense.

### DIFF
--- a/download.js
+++ b/download.js
@@ -374,6 +374,35 @@ function delayedGenerateCode(){
 	timerId = setTimeout(generateCode, 500);
 }
 
+function getSortedComponentsByRequirements(components){
+	var sorted = [];
+	for (var component in components) {
+		sorted.push(component);
+	}
+	var i = 0;
+	while (i < sorted.length) {
+		var id = sorted[i];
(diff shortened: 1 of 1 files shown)


nauzilus:
Nice :+1: 

This is the cause of #796, and also fixes the **Crytsal** dependency on **Ruby**.

podgorskiy:
Yes, it also fixes issue with **Crystal**, I've just checked.

Golmote:
Thank you very much for investigating this @podgorskiy!

podgorskiy:
Was glad to help!

PR #799, opened 2015-10-08
Provide a way to register languages being added [WIP] [BC break]

Golmote:
Ok, so let me explain this PR a bit.

After the recent bug that occurred on the download page due to the order of the components (#796, #797), I've been thinking and came up with a idea that could both solve this issue more elegantly and put Prism on its path to become properly asynchronous some day.

I added two methods on the `Prism.languages` object:

* `add(id, dependencies, callback)`

  This method will register the component `id` as being added to Prism. The `callback` function will contain everything needed to define the component. The important thing to note here is that the `callback` function will only be called once the `dependencies` (if any) are registered too (that is, if they already exist or if they are registered in the future).
* `whenLoaded(languages, callback)`

  This method runs the `callback` function once the `languages` are registered. It's used by `add()` but can also be used on its own.

With these functions, we could update all existing components like the following:
* Wrap the component's code in an `Prism.languages.add` call, specifying the required dependencies (the same that are defined in the `components.js` file) if any.
* Use `Prism.languages.whenLoaded(some_language)` instead of `if(Prism.languages[some_language])` for the optional dependencies (like filters in Jade, or Markup in JavaScript, etc.)

Applying these tiny changes will make the order of the components irrelevant. At some point in the future, it might even allow us to go asynchronous by (re)triggering the highlighting when some dependency is loaded.

So... what do you think about it? :)


zeitgeist87:
I think this is a good idea. I only have one comment. Couldn't you reuse the existing hook system for that?

Some pseudo-code to give you an idea what i mean (NOT tested):

```js
function add(id, dependencies, callback) {
     // wait for dependencies

    // call callback

    _.hooks.run('new-language', id);
}

function whenLoaded(id, cb) {
    // ....
    var langCb = function(langId) {
        if (id === langId) {
            cb();
            _.hooks.remove('new-language', langCb);
        }
    }
    _.hooks.add('new-language', langCb);
    // ...
}
```

The method `_.hooks.remove()` does not exist, but can be implemented easily.

Golmote:
That's a great idea. I'll give it a try tonight!

Golmote:
It took me a moment to make the `remove()` method work properly when it's called during the `run()` method, but I think I found a way of handling it that is not too ugly. What do you think @zeitgeist87?

LeaVerou:
Why not use promises? Callbacks and `whenLoaded`-type methods are a bit outdated.

Golmote:
You really wanna add a promise library in Prism?

mAAdhaTTah:
Promises aren't supported in any version of IE... go figure.

LeaVerou:
Promises are supported in Edge, so polyfills will soon become irrelevant, whereas we can't easily change our API.
How large is a basic promise polyfill?

Golmote:
Well, [the one we use](https://github.com/jakearchibald/es6-promise) on Prism website is around 5Kb minified.
There are some others, like [promiscuous](https://github.com/RubenVerborgh/promiscuous) or the less known [plite](https://github.com/chrisdavies/plite), that are around 1Kb minified. I never had the occasion of testing any of them, though.
I may give them a try.

LeaVerou:
promiscuous looks pretty good. We could include it as an addon ("IE9-11 support") :P

Golmote:
Well... I played around with Promises for several hours now... The biggest issue I encouter, ironically, is that they are *really* asynchronous. ^^' The consequence is that changes in grammars may happen at any time, including once Prism is done highlighting.

The only solution I've found so far would be to add another hook `updated-language` that should be triggered by any component/plugin after a grammar has been updated.

Consider the filters in the jade component as an example:

```javascript
filters.forEach(function (filter) {
    /* [...] build an object o with additional tokens */
    Prism.languages.insertBefore('jade', 'filter', o);
    Prism.hooks.run('updated-language', 'jade');
});
```

and in `highlightElement()`, we would have something like:

```javascript
// Whenever the language is loaded
_.languages.getPromise(language).then(function (grammar) {
    env.grammar = grammar;
    var doHighlight = function () {
        /* [...] */
        env.element.innerHTML = env.highlightedCode;
        /* [...] */
    };
    doHighlight();
    _.hooks.add('updated-language', function (langId) {
        if (language === langId) {
            // The language for this code block has been updated somehow:
            // refresh the grammar and highlight again!
            env.grammar = grammar = _.languages[language];
            hl();
        }
    });
});
```

Would that be ok?

Golmote:
Also, because of the asynchronicity, note that this change is not backward compatible and will lead to the release of a major version.

LeaVerou:
That does seem like a pretty huge change. Can you hold off making these changes for a few more days please in case I find a moment to think about it? (I don't want to stall progress, and there's high probability I won't find the time, but give me a chance to :)

Golmote:
Sure, that's why I changed the title of this PR this morning. ^^
It is indeed a huge change and we need to carefully think about all of this.

PR #801, opened 2015-10-10
Improve highlight.

chaosshen:
Add highlight for directives in C/C++ macros and C#/F# preprocessor lines;
Add some predefined constants in C;
Remove `true` and `false` from keyword list in JavaScript, they should be considered as boolean constant.


Golmote:
Thanks!

PR #806, opened 2015-10-13
Fixed detection of float numbers.

vertax:
Before not work correctly detection of numbers like: 0.3f.


Golmote:
Thanks for contributing!

PR #816, opened 2015-10-28
Revert white-space to normal when code is inline instead of in a pre

darobin:
All of the themes have in common that they set `white-space: pre` on `code`, even if `code` is not inside a `pre`. That is problematic for two reasons:

1. `code` normally does not work that way, so you can normally safely go to the next line inside of an inline `code`; with `pre` that white space becomes visible. When we turned Prism on for our site (which has a lot of inline `code`) it immediately turned into a train wreck.
2. The lack of wrapping means that it easily overflows on small screens. For `pre` we are protected by wrappers that scroll the overflow, but you don't expect that from inline elements.

This PR simply returns inline `code` to its expected `normal` value.

### DIFF
--- a/themes/prism-coy.css
+++ b/themes/prism-coy.css
@@ -77,6 +77,7 @@ pre[class*="language-"] {
 	color: #c92c2c;
 	border: 1px solid rgba(0, 0, 0, 0.1);
 	display: inline;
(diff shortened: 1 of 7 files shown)


Golmote:
Hi! Thanks for contributing!

PR #825, opened 2015-11-16
Trim plugin

telic:
Trims blank lines from the beginning and end of code blocks, as well as removing indents and end-of-line whitespace.

This makes the "remove-initial-line-feed" plugin redundant.

This plugin needs to run before those that attach to the "complete" hook (e.g. line-numbers), or strange things may happen. There didn't appear to be any way to ensure this when running on "language-none" blocks, so I added a new "wont-highlight" hook for this use.

### DIFF
--- a/components.js
+++ b/components.js
@@ -594,6 +594,11 @@ var components = {
 			"owner": "Golmote",
 			"noCSS": true
 		},
+		"trim": {
+			"title": "Whitespace trimmer",
+			"owner": "telic",
+			"noCSS": true
+		},
 		"keep-markup": {
 			"title": "Keep Markup",
 			"owner": "Golmote",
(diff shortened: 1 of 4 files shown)


LeaVerou:
I like the idea.
FWIW, to match `\s` without `[\r\n]` you could do something like `/(?![\r\n])\s/`
More info on this technique: http://lea.verou.me/2012/05/hacking-lookahead-to-mimic-intersection-subtraction-and-negation/

telic:
Yes, that would arguably make it more readable. The negative lookahead would need to be repeated for every character though (i.e. `/(?:(?![\r\n])\s)+$/`), which I imagine could be quite slow…

Here's a JSPerf test: http://jsperf.com/negated-vs-modified-class. Safari does especially poorly with the lookahead version for me.

LeaVerou:
Closing this, since the whitespace normalizer plugin already does this. Thanks for your efforts, I hope you contribute again in the future!

PR #826, opened 2015-11-18
Eval

jandockx:
Hello there.

Thx for creating Prism.

Here is a proposal for a new plugin, "eval", which evaluates the JavaScript highlighted by Prism, and shows evaluation results and output.

It is intended for explaining and demonstrating JavaScript code, and literate programming.


LeaVerou:
I like the idea, though tbh this seems like too much code for a simple feature like this. I understand that debugging is important, but here over 50% of the code is error handling! It’s a bit overkill for a tiny plugin, IMO. @Golmote what do you think?

Btw any chance you could temporarily add these commits to gh-pages so we could see a live demo of this on http://jandockx.github.io/prism/#plugins ?

jandockx:
merged eval into gh-pages in the clone.
See http://jandockx.github.io/prism/plugins/eval/index.html.

jandockx:
Hi, Lea.

If you can detail which parts you think are overkill, we can no doubt resolve that. But you might want to take a second look. I believe "over 50% of the code is error handling" is a bit of a misunderstanding.

There are some (6) debug/trace-outputs, and we can remove them, if you like (although I normally don't think that is a good idea - they were put there during development for a reason, but anyway - potato, potato). The conditional structures they are in are functional, however, so the only thing that can be removed are the 6 output statements (and the debug function and "debug log level", if it is no longer used).

The other output is either 1) the point of the plugin - giving output, or 2a) info and warning about probable tag / attribute misconfiguration and 2b) problems during execution of the subject code. This was added during use.

I found myself hunting for "why doesn't this output doesn't show up". A simple info or warn in the console helps me to find the tag / attribute misconfiguration quickly.

As the plugin is intended for literate programming, I also need a mechanism to kind of "debug" the subject code. If something goes wrong with the subject code, like an uncaught exception, a syntax error, etc., I would like as good and as direct as feedback as possible. It a negative point that I cannot give more clear feedback than I do now (no line numbers, no stacktrace, ...).

LeaVerou:
On a second look, since this is a plugin, I guess the amount of code doesn't matter much. I'd be in favor of merging it if the others agree.

mAAdhaTTah:
Any security / sandboxing concerns about straight eval'ing like this? Given Prism aims to be as broadly targeted as possible, I'd be a bit worried about enabling something like this without either ensuring the eval was sandboxed appropriately or big, screaming warnings were included in the documentation.

LeaVerou:
That would restrict the use cases though. I'd prefer a big warning in the docs to not use this if the code samples are user-generated.

jandockx:
@mAAdhaTTah : `eval` _an sich_ is not an issue. _Any_ code you run in a browser is a security concern. 

In the threat model, 3 parties are involved: the user, the author of the web site visited, and potentially 3rd parties. The browser has measures to protect the user from malicious intent by the author, by sandboxing, and to protect the code of the author as much as possible from malicious 3rd parties. The user trusts the author in the functionality between the user and the author. The author must protect its code from malicious 3rd parties (such as other users) within the leeway left by the browser. The latter comes down to 'do not execute code in your site that you have not authored yourself'. And here `eval` is a potential vector to watch out for. But so is, e.g., loading a Javascript file from a user supplied URL.

[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval) says:

> Do not ever use eval!
>
> eval() is a dangerous function, which executes the code it's passed **with the privileges of the caller.** If you run eval() with a string that could be affected by a malicious party, you may end up running malicious code on the user's machine **with the permissions of your webpage / extension**. More importantly, a third-party code can see the scope in which eval() was invoked, which can lead to  possible attacks in ways to which the similar Function is not susceptible.

A more important caveat might be that `eval` is _very_ slow, and it also slows down other code in the page.

There is nothing against adding 'screaming warnings' that this feature executes code. As this is not the primary goal of prism, that might indeed by cautious. But, as @LeaVerou  already noted, that's kind of the point of this feature. And when you author code, any code, the  issues mentioned above arise.

Apart from that, I am sure the code can be ameliorated in a thousand ways …

mAAdhaTTah:
I appreciate the contribution and work that went into this, but I'm inclined to think this should live outside Prism core, as no core changes are required and this does more than just highlight existing code. Happy to revisit this, but going to close for now.

Thank you for the contribution! I'd definitely be willing to add a link to the site if you create an external repository for this.

PR #827, opened 2015-11-24
Eiffel: Added basic types as keywords

carlfriess:


Conaclos:
Hi!
https://github.com/PrismJS/prism/pull/471 could handle the highligthing of the basic types since theyr are also classes in Eiffel :)

carlfriess:
Jep! True, I didn't know about your PR when I made this one. It might be nice to add an alias like `builtin` so that themes highlight class names by default.

PR #828, opened 2015-11-25
Highlight incomplete block comments

ashkitten:


components/prism-clike.js
```diff
@@ -1,7 +1,7 @@
 Prism.languages.clike = {
 	'comment': [
 		{
-			pattern: /(^|[^\\])\/\*[\w\W]*?\*\//,
+			pattern: /(\/\*[\w\W]*?\*\/)|(\/\*[\w\W]*)/,
```
LeaVerou: Why not just make the closing bit optional then, with `(\/\*)?`? This is exactly what this does, but with more duplication.

LeaVerou:
Thanks! See comment.

Golmote:
I made the change manually. @LeaVerou actually just making the closing bit optional would not have worked because the lack of greediness would have stopped the highlighting far too soon. I found a way by adding a `$` anchor as an alternative to the closing bit.

ashkitten:
Oh sorry, forgot about this. Thanks @Golmote for implementing it, when I put my new website up I'll be sure to update it with this change.

PR #840, opened 2015-12-17
Remove unnecessary data-language attribute from show-language

nauzilus:
It was only applicable when using CSS pseudo elements to show the language name, which is no longer the case.

People _may_ have been targeting blocks by language using `[data-language='C#']` for example, but it's more likely they'd have used `.language-csharp`, since `data-language` is only ever there because of a plugin.


LeaVerou:
Thanks!
Btw, is there any particular reason for using `innerHTML` instead of `textContent`?

nauzilus:
Nope, no reason, it was already there from earlier commit; I just rearranged the code and didn't notice it :( I shall fix later this evening.

Issue #818, opened 2015-10-31
Uncaught RangeError: Maximum call stack size exceeded


gytdau:
This error occurs in my console when the script is included, even if I didn't add any language classes anywhere.

This is my body:
```
<body>
<pre><code>
Code goes here
</code></pre>
<script src="/prism/prism.js"></script>
</body>
```

I've included most of the languages available in my installation, and the line numbers plugin.

This is the full error, as seen on my Chrome dev tools:
```
Uncaught RangeError: Maximum call stack size exceeded
match @ (program):128
_self.Prism.util.type @ prism.js:2
_self.Prism.languages.DFS @ prism.js:2
_self.Prism.languages.DFS @ prism.js:2
_self.Prism.languages.DFS @ prism.js:2
_self.Prism.languages.DFS @ prism.js:2
_self.Prism.languages.DFS @ prism.js:2
_self.Prism.languages.DFS @ prism.js:2
_self.Prism.languages.DFS @ prism.js:2
_self.Prism.languages.DFS @ prism.js:2
```

The highlighting still works - I just can't get the line numbers to work when I add a 'line-number' class to the `<pre>`.

Golmote:
Hi, thanks for reporting.

This issue will be a tough one to fix... The DFS function is not able anymore to walk through all the languages. And the DFS function is called anytime insertBefore is called, which means... many times.

The DFS function is mainly needed because Prism relies on the order of the keys in the grammars. insertBefore creates a new object each time it's called (it can't be done "in place") and it needs to walk through all references of the previous object to update each one to the new reference.

I think #799 could help fixing this. I need to get back working on it.

For now, the only solution I can suggest is that you reduce the number of languages you included in your installation, preferably to match only the ones you actually need.

gytdau:
After some testing, I can indeed confirm that the problem stems from too many languages. 

Thanks for your support!

bebraw:
FYI, I hit the same problem at [prism-languages](https://github.com/survivejs/prism-languages) when trying to get it up to date. Given I just concat the language definitions, this happens.

bebraw:
I "resolved" this problem at **prism-languages**. I simply hacked together a dependency declaration of my own. Now it's possible to consume the definitions from Node easily. This makes it ideal for backend processing. You probably don't want to use this on frontend, though, due to the sheer amount of definitions.

[...]

PR #841, opened 2015-12-21
Prevent infinite recursion in DFS

zeitgeist87:
This PR should fix issue #818 but it should also increase the performance of DFS in general. A lot of languages have direct references to each other and DFS is called for all languages every time insertBefore() is called. So it makes sense to skip objects that have been traversed already.

The only problem is, that I don't know if this modification breaks any of the plugins or languages.


LeaVerou:
Looks great! Merging.

LeaVerou:
Btw, you've got mail :)

PR #843, opened 2015-12-29
adds remove-final-line-feed plugin

cormacmccarthy:
Add plugin to remove final blank line.

### DIFF
--- a/components.js
+++ b/components.js
@@ -560,6 +560,11 @@ var components = {
 			"owner": "Golmote",
 			"noCSS": true
 		},
+		"remove-final-line-feed": {
+			"title": "Remove final line feed",
+			"owner": "cormacmccarthy",
+			"noCSS": true
+		},
 		"previewer-base": {
 			"title": "Previewer: Base",
 			"owner": "Golmote"
--- a/plugins/remove-final-line-feed/index.html
+++ b/plugins/remove-final-line-feed/index.html
@@ -1,0 +1,58 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+
+	<meta charset="utf-8" />
+	<link rel="shortcut icon" href="favicon.png" />
+	<title>Remove final line feed ▲ Prism plugins</title>
+	<base href="../.." />
+	<link rel="stylesheet" href="style.css" />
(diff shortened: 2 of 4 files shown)


zeitgeist87:
Thanks for the PR. It seems like you've included the file `components/prism-jsx.min.js` by accident.

I don't know if this was discussed already somewhere, but I find the plugin `remove-initial-line-feed` and `remove-final-line-feed` quite odd. Why don't we just create a simple `trim` plugin that does both? The user could add the class `no-left-trim` and `no-right-trim` to disable the respective trim operation. @Golmote Is there a particular reason why the plugin `remove-initial-line-feed` removes only a single line feed and not all leading whitespace?

cormacmccarthy:
I'd love to see remove-initial-line-feed remove all leading whitespace as well.

And, yeah, a simple trim plugin that does both would probably be better, but, remove-initial already existed, so...

LeaVerou:
Thanks for contributing @cormacmccarthy!
Don't we have a general plugin for whitespace removal? If not, we should. We can't have one plugin for extra indent removal, one for last line, one for first line etc, it’s getting ridiculous. We can always allow options (not on the element, as these options are likely intended to be global).

zeitgeist87:
> Don't we have a general plugin for whitespace removal? If not, we should.

I don't think so. There is only the `remove-initial-line-feed` plugin, which removes exactly one line from the start.

cormacmccarthy:
@LeaVerou if you know of such a plugin, please let me know! I'm about to write one for extra indent removal today if I can't find it, because that's the next thing I need. I agree it would totally be better for them all to be the same plugin. If I write one that does extra indent removal, I'll just throw the trim stuff in there as well to consolidate it all down to one.

zeitgeist87:
@cormacmccarthy I have just submitted a PR #847 with a general purpose whitespace plugin, that also does extra indent removal, but it needs more testing and I probably have to change the way it gets its global configuration, since that is a new feature to Prism and subject to change.

cormacmccarthy:
Nice, thanks! OK, I'll pull that into my project to help test. I'll comment on #847 with findings.

Golmote:
As PR #847 was merged, I consider the issue fixed.

PR #844, opened 2015-12-29
Defer File Highlight until the full DOM has loaded.

chriswells0:
This allows the File Highlight plugin to work properly even when the script is loaded before the elements that should have a file loaded into them.

### DIFF
--- a/plugins/file-highlight/prism-file-highlight.js
+++ b/plugins/file-highlight/prism-file-highlight.js
@@ -71,6 +71,10 @@
 
 	};
 
-	self.Prism.fileHighlight();
+	if (document.addEventListener) {
+		document.addEventListener('DOMContentLoaded', self.Prism.fileHighlight);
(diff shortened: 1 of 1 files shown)


zeitgeist87:
Thanks!

LeaVerou:
Thanks @chriswells0!
However, no browser we’re targeting doesn't support `addEventListener`. This is a pointless check that adds pointless extra code for everyone. Can we please remove it?

zeitgeist87:
> However, no browser we’re targeting doesn't support addEventListener. This is a pointless check that adds pointless extra code for everyone. Can we please remove it?

Good point. I've removed the check.

chriswells0:
> no browser we’re targeting doesn't support `addEventListener`.

I thought that was the case, but I'm old school and struggle with telling browsers "no."  ;)  I'll get used to it one day.

PR #845, opened 2015-12-30
Remove the need for an "output" class.

chriswells0:
@zeitgeist87: These are the changes you suggested in PR #831.


plugins/command-line/prism-command-line.js
```diff
@@ -63,7 +63,10 @@ Prism.hooks.add('complete', function (env) {
 
 		if (!isNaN(outputStart) && !isNaN(outputEnd)) {
 			for (var j = outputStart; j <= outputEnd && j <= prompt.children.length; j++) {
-				prompt.children[j - 1].className = 'output';
+				var node = prompt.children[j - 1];
+				node.removeAttribute('data-user');
+				node.removeAttribute('data-host');
+				node.removeAttribute('data-prompt');
```
LeaVerou: How about something like:
```js
["user", "host", "prompt"].forEach(function(attr) { node.removeAttribute("data-" + attr); });
```

LeaVerou:
Thanks! LGTM, see comment :)

plugins/command-line/prism-command-line.js
chriswells0: In my opinion, the original code is cleaner and easier to read.  It's "nice" that it could be done in 1 line, but saving those 18 characters effectively adds 4 function calls and 3 string concatenations.

LeaVerou: It’s not about doing it in one line, it's about removing repetition. Actually, I wouldn't do it in one line, I would format it as 3, for readability. Anyhow, no strong opinion here.

PR #846, opened 2015-12-30
Fix broken heading links.

d10:


zeitgeist87:
I see, the heading links on all of the plugin pages are wrong because of the `<base href="../.." />` tag. This fix looks good to me.

zeitgeist87:
Thanks!

Issue #848, opened 2016-01-02
Line Highlighter bad positioning in COY theme with line-number


mdolbin:
It seems to be a known & [fixed issue](https://github.com/PrismJS/prism/blob/9652d75222e1791f26b97ac3c6aec11c8e55b079/plugins/line-highlight/prism-line-highlight.js#L68) but highlighted area still fails to position itself properly in COY theme.
Here is a fiddle: http://codepen.io/anon/pen/RrKGLw
Fix appeared to be fairly simple, I used this fiddle to test it: http://codepen.io/anon/pen/obBzMK
![image](https://cloud.githubusercontent.com/assets/3534827/12073908/1ede663c-b140-11e5-899c-168802e7657c.png)

mdolbin:
@LeaVerou what do you think about always putting ```div.line-highlight``` inside ```code``` and updating styles for each theme? I know it's a global modification that can lead to new issues but nesting looks like the best and the most convenient & correct solution if we want to position one block inside another.

So, just to be sure that we're on the same page, I assume that, 
* ```pre``` without ```code``` is not enough for prism to work
* in order for line-number or line-highlighter plugins to works, there must be both ```pre``` and ```code``` elements on the page. 
* there is no point in enumerating \ highlighting of 'inline' code - when ```code``` is used without ```pre```

Please let me know if there are any objections or I can try to do my best to improve highlighting :)

LeaVerou:
Then it would be wiped on rehighlighting and it would need to be recreated and repositioned. 

Regarding your assumptions:
- Correct. However `code` without `pre` works.
- Correct
- Inline code *is* highlighted. (see first bullet point)

mdolbin:
@LeaVerou 
> Then it would be wiped on rehighlighting and it would need to be recreated and repositioned.
What's the use case for rehighlighting of already highlighted code? Are you worried about performance? At the end of the day, line-numbers plugin is placed inside the ```code``` the same way I propose to put line-highlighting and it seems to work fine.

> Inline code is highlighted. (see first bullet point)
I mean line-numbering and line-highlighting plugins should work if and only if both ```pre``` and ```code``` are present and there is no point in line-highlighting or line-numbering of a single line of code - when there is a ```code``` but not ```pre```. I already saw the docs so I just wanted to make sure.

LeaVerou:
> What's the use case for rehighlighting of already highlighted code?

Any changes to the code, such as those made in dabblet or CodeFlask. 

[...]

PR #849, opened 2016-01-02
#848 highlight positioning in coy theme fix - changed overflow for pre and code

mdolbin:
Here is a fiddle to reproduce the issue: http://codepen.io/anon/pen/RrKGLw
I used this fiddle to test this fix: http://codepen.io/anon/pen/obBzMK


LeaVerou:
Thanks & happy new year :)
This breaks the shadows under the code area in this theme :(

PR #856, opened 2016-01-17
Skip highlighting for output lines.

chriswells0:
Lines marked as output are skipped by the highlighting process to prevent false positives.


LeaVerou:
Thanks! Wouldn't it be more usable if there was a character in the beginning of such lines to indicate them instead of having to figure out line numbers for `data-output`?

chriswells0:
Do you mean some sort of non-code text at the beginning of every line of output?  My instinct says that's at least as cumbersome as identifying the output lines in an attribute, but then I haven't tried it.  :)

There are 2 other concerns that come to mind:

1. It seems like a kludge (for lack of a better word) to add fake data when the role of markup is to describe data.
2. How would we choose character(s) to identify output in any given language and know it's never part of the code/output?

I may have misunderstood what you were suggesting, though.  Please correct me if I'm misinterpreting it.

plugins/command-line/prism-command-line.js
```diff
@@ -35,41 +96,30 @@ Prism.hooks.add('complete', function (env) {
...
+		for (var i = range.start; i <= range.end; i++) {
+			var node = prompt.children[i];
+			node.removeAttribute('data-user');
+			node.removeAttribute('data-host');
+			node.removeAttribute('data-prompt');
 		}
-	}
+	});
 
 	env.element.innerHTML = prompt.outerHTML + env.element.innerHTML;
```
zeitgeist87: I know this is not part of this PR, but I've just noticed this line. How about something like this:

```js
pre.insertBefore(prompt, env.element);
```

zeitgeist87: Or even better:

```js
env.element.insertBefore(prompt, env.element.firstChild);
```

plugins/command-line/prism-command-line.js
```diff
@@ -4,6 +4,67 @@ if (typeof self === 'undefined' || !self.Prism || !self.document) {
 	return;
 }
 
+// Scoped to this IIFE so they can be accessed by both hooks. -- cwells
+var outputLines = [];
```
zeitgeist87: I think you should store your state in the `env` object. That's what it's for.

plugins/command-line/prism-command-line.js
```diff
@@ -4,6 +4,67 @@ if (typeof self === 'undefined' || !self.Prism || !self.document) {
...
+var outputRanges = [];
+
+Prism.hooks.add('before-highlight', function (env) {
+	var pre = env.element.parentNode;
+	if (!pre || !/pre/i.test(pre.nodeName)) {
+		return;
+	}
+
```
zeitgeist87: You have to do the test for the `command-line` class here as well.

You could move all the tests from the `complete`-hook here and add a flag to the `env` object, so you don't have to test it again for the other hooks.

plugins/command-line/prism-command-line.js
```diff
@@ -35,41 +96,30 @@ Prism.hooks.add('complete', function (env) {
...
-		if (!isNaN(outputStart) && !isNaN(outputEnd)) {
-			for (var j = outputStart; j <= outputEnd && j <= prompt.children.length; j++) {
-				var node = prompt.children[j - 1];
-				node.removeAttribute('data-user');
-				node.removeAttribute('data-host');
-				node.removeAttribute('data-prompt');
-			}
+	// Remove the prompt from the output lines. -- cwells
+	outputRanges.forEach(function (range) {
+		for (var i = range.start; i <= range.end; i++) {
```
zeitgeist87: You could use `for (var i in outputLines) {` here, like you do in the `before-insert`-hook. Then you wouldn't need to store the output ranges and then you could simplify the `before-highlight`-hook.

plugins/command-line/prism-command-line.js
```diff
@@ -35,41 +96,30 @@ Prism.hooks.add('complete', function (env) {
 	}
 
 	// Create the "rows" that will become the command-line prompts. -- cwells
-	var lines = new Array(1 + env.code.split('\n').length);
+	var promptLines = new Array(1 + env.code.split('\n').length);
```
zeitgeist87: It seems a bit wasteful to split `env.code` again, just to get to the number of lines. You could store that information in the `env` object and reuse it here.

zeitgeist87:
It is a bit strange, that user has to specify the output lines with `data-output`. Wouldn't it be more natural to specify the lines that have prompts? I think that could also simplify the code. What do you think @chriswells0 ?

plugins/command-line/prism-command-line.js
chriswells0: I had the same thought and considered sending a note to ask about adding a way to compartmentalize variables in the `env` object, but I didn't want to rock the boat too much for my own needs.  I would have done a few things differently if there was an obvious way to set variables without concern for stepping on anyone else's code.

I suggest adding a property on the `env` object specifically for plugin usage such as:

```javascript
env.pluginVars["command-line"].outputLines = [];
```

With this approach, every plugin (being uniquely named) could have its own variables in the `env.pluginVars` array using any names the developer chooses without regard to other plugins' variables.  Thoughts?

plugins/command-line/prism-command-line.js
chriswells0: I'm in a rush and will need to come back to revisit this.

plugins/command-line/prism-command-line.js
chriswells0: I chose `forEach()` here because it made the code easier to read for anyone coming behind me.  My initial code was actually a nested `for` loop with `i` and `j` variables.  I rewrote it using `forEach()` and it looked cleaner and easier to read.

chriswells0:
> I guess it depends on the type of content. If you have more commands or more output lines...

Exactly.  If the purpose of the plugin is to highlight commands, then the output lines would be the outliers.  That's the way I see it, anyway.

plugins/command-line/prism-command-line.js
zeitgeist87: The `forEach()` is fine. I was just suggesting something like this:

```js
for (var i in outputLines) {
	var node = prompt.children[i];
	node.removeAttribute('data-user');
	node.removeAttribute('data-host');
	node.removeAttribute('data-prompt');
}
```

Then you would not need the variable `outputRanges` at all.

LeaVerou:
> It seems like a kludge (for lack of a better word) to add fake data when the role of markup is to describe data.

Your argument seems like a philosophical concern, whereas the problem I pointed out is very real: Specifying line numbers is not only difficult (how do you even know if you're not using the line number plugin?), but also fragile (all numbers change if new lines are added).
Philosophical purity is NOT above usability. At least not in this project.

> How would we choose character(s) to identify output in any given language and know it's never part of the code/output?

Same way as any metacharacter or keyword in any language, ever: You provide an escape sequence to escape it in the odd chance it's actually used as code. We're not treading new ground here :P

plugins/command-line/prism-command-line.js
zeitgeist87: At the moment plugins and languages use the `env` object directly for their variables. For example `php` creates the property `env.backupCode` and the `keep-markup` plugin uses `env.keepMarkup`. As you rightly point out this practice could lead to conflicts. I like your proposed solution, but maybe we should use a shorter name like:

```js
env.vars['command-line'].outputLines = [];
```

plugins/command-line/prism-command-line.js
chriswells0: > Then you would not need the variable `outputRanges` at all.

Indeed.  I'll wipe out that var.

chriswells0:
> Do you count them, one by one, with your finger?

That seems easier than manually adding text before every line of output, but that could be solely because I've only done it a handful of times using the current approach and no other.  From a usability perspective, would you rather count the lines or paste special chars before each line?  To simplify that, we could instead insert beginning and ending markers around the output lines, but then we're effectively resorting to marking it up.

> You provide an escape sequence to escape it in the odd chance it's actually used as code.

Obviously, we could make it very unlikely that the characters would match the output (e.g., `^\[OUTPUT\]`) or go with something simple (e.g., `^_{5}`).  Then we're asking users to insert those chars in front of every line and also escape the ones that do exist in their output.

LeaVerou:
From a usability perspective, I would definitely prefer to add a prefix before these lines than to have to find their line numbers and keep them in sync with the code.
The main issue I see with that approach is that it spoils the fallback in case Prism fails to load/run for whatever reason. But we could make the prefix readable, e.g. something like `(Output) `.
Of course, both approaches can co-exist so that you can still use your line numbers and everyone else can use a prefix.

chriswells0:
> it spoils the fallback in case Prism fails to load/run for whatever reason

If that's a concern, then it might make sense to add extra data.  I'd appreciate retaining the attribute's functionality so I wouldn't have to add the data in my case, but maybe I'll change my mind once I see it in action.

1. What would you like the prefix to be?
2. Do you want that in a separate PR?  Seems like it should be.
3. Should I add the same fallback functionality to the Line Highlight plugin?

zeitgeist87:
I think it is a good idea to also support a marker character for the output lines. You already have the code lines in the variable `codeLines`. It would be a simple matter to iterate through them and check if they start with a certain character. A simple loop and regex should do it. You can still support your `data-output` attribute at the same time:

```js
// using > as line marker and \> to escape it
for (var i = 0; i < codeLines.length; ++i) {
	codeLines[i] = codeLines[i].replace(/(^\s*)(\\?>)(.*)/, function(match, p1, p2, p3) {
		if (p2 == '>') {
			outputLines[i] = p1 + p3;
			return '';
		} else {
			return p1 + '>' + p3;
		}
	});
}
```

chriswells0:
Is `>` specifically the character we want to use?

If it's being used to mark the output lines, is there any reason to allow whitespace in front of the marker?  I would think it should come at the beginning of the line when used as a marker.

zeitgeist87:
> Is > specifically the character we want to use?

No, that was just an example. Feel free to use something different. I chose it, because it reminded me of the multiline string marker in bash:
```
[andreas@terok ~]$ echo "multiline
> 
> 
> string"
```

> If it's being used to mark the output lines, is there any reason to allow whitespace in front of the marker? I would think it should come at the beginning of the line when used as a marker.

You are right, there is no good reason. If we disallow whitespace, then we don't need a regex and we can simplify the code.

LeaVerou:
I think `>` is not a good option exactly because it's like the multiline character, so multiline commands and output would get confused.
Anything works for me, but please choose something that is short to type and works as a fallback too.
I suggested `(Out)` earlier, but I'm not married to it.

chriswells0:
@zeitgeist87:  thoughts on `(Out)` or other suggestions?

I'd lean toward all lowercase or all uppercase, but we might as well make it case-insensitive anyway.

chriswells0:
@LeaVerou, @zeitgeist87:  Would you expect the `data-output` attribute and the line prefix to be allowed on the same code block or would you prefer to skip the prefix check if the attribute is provided?  I could go either way on that.

zeitgeist87:
> Would you expect the data-output attribute and the line prefix to be allowed on the same code block or would you prefer to skip the prefix check if the attribute is provided?

I think it's better not to allow them at the same time. As you pointed out, there is no need to force the user to escape code that only uses the `data-output` attribute.

> thoughts on (Out) or other suggestions?

We could do it like the `diff` format used by `git diff`. Instead of using a long marker like `(Out)` and escaping it, we could simply add a marker to every line. For example a `+` for a prompt and a space ` ` for an output line. For example this code

```
ls -al
total 44
drwxr-xr-x 11 root root 4096 Jan 23 23:44 .
drwxr-xr-x  5 root root 4096 Dec 31  2014 ..
```

would be escaped like this:

```
+ls -al
 total 44
 drwxr-xr-x 11 root root 4096 Jan 23 23:44 .
 drwxr-xr-x  5 root root 4096 Dec 31  2014 ..
```
Notice that the output lines also have a space added in front of them. We don't have to use `+` and ` `. We can use any single character and don't have to worry about escaping things. What do you think about that?

chriswells0:
The only issue I see with such an approach is that would mean a little more effort to prefix every line--even when there are no output lines.

chriswells0:
Allowed specifying output prefix using `data-filter-output` attribute, so users can choose their own prefix for the output lines.
+ @Golmote for review.

plugins/command-line/prism-command-line.js
```diff
@@ -4,33 +4,99 @@ if (typeof self === 'undefined' || !self.Prism || !self.document) {
...
-		// Remove the class "command-line" from the <code>
+	var codeLines = env.code.split('\n');
+	env.vars['command-line'].numberOfLines = codeLines.length;
+	env.vars['command-line'].outputLines = [];
+
+	var outputSections = pre.getAttribute('data-output');
+	var outputFilter = pre.getAttribute('data-filter-output');
+	if (outputSections || outputSections === '') { // The user specified the output lines. -- cwells
+		outputSections = outputSections.split(',');
+		for (var i in outputSections) { // Parse the output sections into start/end ranges. -- cwells
```
Golmote: I'm not a big fan of iterating through Arrays with `for/in` loops. Could you change this to a regular `for` loop please?

plugins/command-line/prism-command-line.js
```diff
@@ -4,33 +4,99 @@ if (typeof self === 'undefined' || !self.Prism || !self.document) {
...
+			var range = outputSections[i].split('-');
+			var outputStart = parseInt(range[0]);
```
Golmote: You might want to specify `parseInt`'s second parameter to `10`, just to make `012-013` work the same as `12-13`.

plugins/command-line/prism-command-line.js
```diff
@@ -4,33 +4,99 @@ if (typeof self === 'undefined' || !self.Prism || !self.document) {
...
+				outputEnd--;
+				// Save the output line in an array and clear it in the code so it's not highlighted. -- cwells
+				for (var j = outputStart; j <= outputEnd; j++) {
+					env.vars['command-line'].outputLines[j] = codeLines[j];
+					codeLines[j] = '';
+				}
+			}
+		}
+	} else if (outputFilter) { // Treat lines beginning with this string as output. -- cwells
+		for (var i in codeLines) {
```
Golmote: Same here, please use a regular `for` loop.

plugins/command-line/prism-command-line.js
```diff
@@ -4,33 +4,99 @@ if (typeof self === 'undefined' || !self.Prism || !self.document) {
...
+Prism.hooks.add('before-insert', function (env) {
+	env.vars = env.vars || {};
+	env.vars['command-line'] = env.vars['command-line'] || {};
+	if (env.vars['command-line'].complete) {
+		return;
+	}
+
+	// Reinsert the output lines into the highlighted code. -- cwells
+	var codeLines = env.highlightedCode.split('\n');
+	for (var i in env.vars['command-line'].outputLines) {
```
Golmote: Same here.

plugins/command-line/prism-command-line.js
```diff
@@ -4,33 +4,99 @@ if (typeof self === 'undefined' || !self.Prism || !self.document) {
...
+	env.vars = env.vars || {};
+	env.vars['command-line'] = env.vars['command-line'] || {};
+	if (env.vars['command-line'].complete) {
+		return;
+	}
+
+	var pre = env.element.parentNode;
+	var clsReg = /\s*\bcommand-line\b\s*/;
+	if (clsReg.test(env.element.className)) { // Remove the class "command-line" from the <code>
 		env.element.className = env.element.className.replace(clsReg, '');
```
Golmote: I'm wondering: if the class of the `<code>` is something like `language-shell command-line foobar`, won't it be replaced with `language-shellfoobar` here? The replacement string should probably be a space `' '`.

plugins/command-line/prism-command-line.js
```diff
@@ -4,33 +4,99 @@ if (typeof self === 'undefined' || !self.Prism || !self.document) {
...
+
+Prism.hooks.add('complete', function (env) {
+	env.vars = env.vars || {};
+	env.vars['command-line'] = env.vars['command-line'] || {};
+	if (env.vars['command-line'].complete) {
+		return;
+	}
+
+	var pre = env.element.parentNode;
+	var clsReg = /\s*\bcommand-line\b\s*/;
```
Golmote: Maybe you could move that declaration outside of the hooks definitions, since you use it twice.

plugins/command-line/prism-command-line.js
```diff
@@ -39,43 +105,31 @@ Prism.hooks.add('complete', function (env) {
...
-		if (!isNaN(outputStart) && !isNaN(outputEnd)) {
-			for (var j = outputStart; j <= outputEnd && j <= prompt.children.length; j++) {
-				var node = prompt.children[j - 1];
-				node.removeAttribute('data-user');
-				node.removeAttribute('data-host');
-				node.removeAttribute('data-prompt');
-			}
-		}
+	// Remove the prompt from the output lines. -- cwells
+	for (var i in env.vars['command-line'].outputLines) {
```
Golmote: Same here, please use a regular `for` loop.

plugins/command-line/prism-command-line.js
chriswells0: I can't believe I missed that.  :(  Thanks for catching it.

plugins/command-line/prism-command-line.js
chriswells0: `outputLines` should only have an `i` for lines that are output, so I leveraged that in each loop through its properties.  I switched the loops anyway and guarded with `hasOwnProperty()`.

chriswells0:
I've made the updates and squashed it all into 1 commit.

Golmote:
Alright, looks good to me!

PR #857, opened 2016-01-18
Fixing to initial copyright year

pra85:
According to http://www.copyright.gov/circs/circ01.pdf , listing the first year of publication in the copyright is enough


LeaVerou:
Thanks!

PR #863, opened 2016-01-22
No background for <code> in <pre>

jordandanielewski:
Avoid things like:

screen shot 2016-01-22 at 14 34 39

screen shot 2016-01-22 at 14 40 32

screen shot 2016-01-22 at 14 43 20

screen shot 2016-01-22 at 14 44 06

when we have something like

```css
pre, code {
  background-color: #F7FAFB;
}
```

in our CSS before prism

Not needed in funky theme

### DIFF
--- a/themes/prism-coy.css
+++ b/themes/prism-coy.css
@@ -7,6 +7,7 @@
 code[class*="language-"],
 pre[class*="language-"] {
 	color: black;
+	background: none;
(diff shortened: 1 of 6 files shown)


LeaVerou:
Why a new rule? There's already a rule styling `<code>` inside a Prism-highlighted `pre`.

jordandanielewski:
Yes you are right, I moved it to the existing rule

screen shot 2016-01-22 at 16 42 00

LeaVerou:
Merged, thanks!

PR #864, opened 2016-01-24
Add prefix for Firefox selection to Twilight theme

jordandanielewski:
The two `pre[class*="language-"]::selection ` are basically the same rules
Firefox rule should have a prefix and be moved before the unprefixed rule


zeitgeist87:
Thanks!

PR #865, opened 2016-01-24
Fix coy theme shadows

zeitgeist87:
This patch removes max-height from the pre tag in the coy theme,
because this height limitation is inconsistent with the other
themes.

The above change causes a bug in the shadows of the coy theme.
Adding max-height: 13em to the ::before and ::after selector fixes
the shadow bug.


LeaVerou:
Cool!

PR #868, opened 2016-01-26
Allow users to use their own element selector

zeitgeist87:
Add the parameter 'data-element-selector' to the script tag, to
allow users to set the selector used to query for code elements.


LeaVerou:
Why not `data-selector`? It's not like it's used for another setting.

Btw, we should use `document.currentScript` to get the script, and only use `getElementsByTagName` as a fallback.

zeitgeist87:
> Why not data-selector? It's not like it's used for another setting.

Good point. It's `data-selector` now.

> Btw (tangentially related), we should use document.currentScript to get the script, and only use getElementsByTagName as a fallback.

Good idea. I've created another PR for that #870.

LeaVerou:
Merged, but keep in mind we need docs too.
Thanks!

Issue #867, opened 2016-01-25
support for <kbd> and <samp>


garretwilson:
Hey, I really like Prism---you've done a good job and making it simple and "just work" (for the most part, which is must more than I can say about most software out there nowadays).

I had a thought, and I'm not 100% of my opinion on this, but... currently you only support `<code>`, which is completely correct for source code blocks. I was already using `<pre><code>` in fact.

But to be pedantic, HTML5 says we should use `<kbd>` for user input (such as the command-line stuff) and `<samp>` for program output. So for for programming language source code, we're good, but for the command-line or sample output... HTML5 wants us to use other tags. See:
- https://www.w3.org/TR/html5/text-level-semantics.html
- http://www.iandevlin.com/blog/2012/07/html5/using-html-kbd-samp-and-pre-elements

I know that `<kbd>` and `<samp>` aren't widespread, and that everybody tries to cram everything like this into `<code>` (as I used to do), but what are your thoughts on supporting `<kbd>` and `<samp>` when it's semantically appropriate?

Cheers,

Garret

LeaVerou:
To collaborators: We should have a config option on the `<script>` element for the selector used. This is a variation of a very frequent request.

zeitgeist87:
> To collaborators: We should have a config option on the script element for the selector used. This is a variation of a very frequent request.

@LeaVerou I've created to versions of this. A simple attribute to provide the selector directly #868 and another version where the user can specify the tag names of the code blocks #869.

In any case the user has to modify the prism theme to support the selected elements.

garretwilson:
Ack, no---that's hideous! We don't need more `data-*` attributes. Plus it defeats the whole purpose of semantic tags! The whole point of HTML5's recommendations is for people to use `<code>` for source code, `<kbd>` for keyboard input, and `<samp>` for sample output. That's what they say, not me.

But if you're going to add all these weird extra kludge things, I withdraw my suggestion. I'd rather just continue using `<code>` everywhere the way it is now that to dirty up my markup with this stuff.

Are you forgetting what the Prism page itself [says](http://prismjs.com/#basic-usage)? "Prism does its best to encourage good authoring practices. Therefore, it only works with `<code>` elements, since marking up code without a `<code>` element is semantically invalid." And that's the way it should be!

I was only suggesting that things like `language-bash` and the command-line plugin should work with `<kbd>`---automatically, not with an ugly hack. And maybe things like `language-git` should apply to `<samp>`, since it applies to sample output from Git.

[...]

PR #869, opened 2016-01-26
Support for additional tags like <kbd> or <samp>

zeitgeist87:
This patch adds the attribute 'data-tags' to the script tag, to
allow the user to use additional tags like <kbd> or <samp>.


components/prism-core.js
```diff
@@ -417,6 +424,7 @@ script = script[script.length - 1];
 
 if (script) {
 	_.filename = script.src;
+	_.tags = (script.getAttribute('data-tags') || 'code').split(/\s*,\s*/g);
```
LeaVerou: If they're just tag names why comma separate them? Why not just space-separated?

components/prism-core.js
```diff
@@ -146,7 +146,14 @@ var _ = _self.Prism = {
 	plugins: {},
 	
 	highlightAll: function(async, callback) {
-		var elements = document.querySelectorAll('code[class*="language-"], [class*="language-"] code, code[class*="lang-"], [class*="lang-"] code');
+		var selector = [];
+		for (var i = 0, tag; tag = _.tags[i++];) {
+			selector.push(tag + '[class*="language-"]');
+			selector.push('[class*="language-"] ' + tag);
+			selector.push(tag + '[class*="lang-"]');
+			selector.push('[class*="lang-"] ' + tag);
```
LeaVerou: You don't need multiple push calls. push accepts multiple arguments.

LeaVerou:
Interesting idea, thanks! See comments. Also, merge conflicts.

Also, it might produce cleaner code to do something like:

```js
var selector = _.tags.map(function(tag) {
	return [
		tag + '[class*="language-"]',
		'[class*="language-"] ' + tag,
		tag + '[class*="lang-"]',
		'[class*="lang-"] ' + tag
	].join(", ");
}).join(", ");
```

Just an idea, you don't have to follow it, but the current version of the code is a bit verbose and repetitive.

components/prism-core.js
zeitgeist87: Good point.

zeitgeist87:
I have just updated the PR according to your suggestions. I changed the map function a little bit.

components/prism-core.js
```diff
@@ -415,6 +420,8 @@ var script = document.currentScript || [].slice.call(document.getElementsByTagNa
 
 if (script) {
 	_.filename = script.src;
+	_.tags = (script.getAttribute('data-tags') || 'code').split(/\s+/g)
+				.filter(function(tag) {return !!tag;});
```
LeaVerou: Why the filtering?

LeaVerou:
Interesting idea.

However, I'm a bit worried that we're jumping to add features way too early, which is a slippery slope and can result in bloat over time. I actually like this feature and how elegantly we implemented it. However, the guy who actually opened #867 did not and I'm a bit hesitant to merge a feature we have no real use cases for. What do you think? :/

components/prism-core.js
zeitgeist87: To filter empty strings. For example the following code
```js
 '   code kbd   '.split(/\s+/g);
```
returns this array:
```js
[ '', 'code', 'kbd', '' ]
```
after the filter I get:
```js
[ 'code', 'kbd' ]
```

zeitgeist87:
I totally agree. If nobody uses it, we shouldn't add it. It is only really useful with a custom theme anyway.

components/prism-core.js
LeaVerou: Just call `.trim()` before splitting then :)

LeaVerou:
I actually needed something similar yesterday, and neither this, nor selector cut it: I was using `<script type="text/code">` to be able to have unescaped markup. Since it wasn't highlighted by default, I wrote a small script:

```js
$$("script[type='text/code']").forEach(script=>{
	var pre = document.createElement("pre");
	var code = document.createElement("code");
	pre.appendChild(code);
	script.parentNode.insertBefore(pre, script);
	code.appendChild(script);
});
```

It was better for me than replicating the entire selector just to add `script[type="text/code"]` to it.

PR #870, opened 2016-01-26
Use document.currentScript

zeitgeist87:
Use document.currentScript instead of document.getElementsByTagName()

### DIFF
--- a/components/prism-core.js
+++ b/components/prism-core.js
@@ -410,10 +410,8 @@ if (!_self.document) {
 	return _self.Prism;
 }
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
I was thinking more something like:

```js
var script = document.currentScript || [].slice.call(document.getElementsByTagName("script")).pop();
```

zeitgeist87:
> I was thinking more something like:
> 
> var script = document.currentScript || [].slice.call(document.getElementsByTagName("script")).pop();

Brilliant I love it!

PR #872, opened 2016-01-28
Fix Solarized Light theme

zeitgeist87:
The official color palette is included as a CSS comment and every
color used is named with a comment.

### DIFF
--- a/themes/prism-solarizedlight.css
+++ b/themes/prism-solarizedlight.css
@@ -7,27 +7,58 @@
  Twitter Handle: https://twitter.com/allonsykraken)
 */
 
+/*
+SOLARIZED HEX
+--------- -------
+base03    #002b36
+base02    #073642
+base01    #586e75
+base00    #657b83
+base0     #839496
+base1     #93a1a1
+base2     #eee8d5
+base3     #fdf6e3
+yellow    #b58900
+orange    #cb4b16
+red       #dc322f
+magenta   #d33682
+violet    #6c71c4
+blue      #268bd2
+cyan      #2aa198
+green     #859900
+*/
+
 code[class*="language-"],
 pre[class*="language-"] {
-	color: #2176c7;
-	font-family: Consolas, Monaco, 'Andale Mono', monospace;
+	color: #657b83; /* base00 */
+	font-family: Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace;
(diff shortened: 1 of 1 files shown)


zeitgeist87:
It definitely needs a bit of tweaking. I am not very good at themes, so I'm open for suggestions on how to improve it.

zeitgeist87:
The result looks like this:

solarized

LeaVerou:
Thanks!

The main problem I see is that the default color is too light. We should use the darkest color there.

LeaVerou:
For anyone interested in helping, here is a comparison. First image is Solarized Light in my editor, second is Prism from the website.

image
image

Ours looks a bit more colorless, but overall, not too bad actually!!
Thanks again @zeitgeist87!!

mAAdhaTTah:
That may be a function of the syntax highlighting rather than the theme. You'll notice in the editor `tokens.content` is two different colors, whereas they're both the same in Prism. Also, in the editor, `tokens.map` (e.g. {object}.{method}) is highlighted differently than  `tokens.replace`. Not sure more can be done without changing Prism itself. 

Also, wondering if a new version of Prism could be pubbed to npm. Would love to get this theme into my plugin ☺.

zeitgeist87:
> Not sure more can be done without changing Prism itself.

We can quite easily make the punctuation red. But I agree, that some of it is not possible with Prism.

> Also, wondering if a new version of Prism could be pubbed to npm. Would love to get this theme into my plugin ☺.

@mAAdhaTTah I will try to create a new release this weekend. There is a bit of work involved, because the CHANGELOG has to updated.

themes/prism-solarizedlight.css
```diff
@@ -7,29 +7,60 @@
...
 }
 
+pre[class*="language-"]::-moz-selection, pre[class*="language-"] ::-moz-selection,
+code[class*="language-"]::-moz-selection, code[class*="language-"] ::-moz-selection {
+	background: #073642; /* base02 */
+}
+
+pre[class*="language-"]::selection, pre[class*="language-"] ::selection,
+code[class*="language-"]::selection, code[class*="language-"] ::selection {
+	background: #073642; /* base02 */
```
zeitgeist87: I would appreciate it, if someone could check if the selection looks right. I am not sure if `base02` is the right color for that.

PR #874, opened 2016-01-30
Ensure show-invisibles compat with autoloader

mAAdhaTTah:
When using the autoloader, the language isn't loaded
at the time the show-invisibles plugin attaches its
extra tokens. This defers attaching the extra tokens
to just before highlighting, so we ensure we're able
to attach to the language grammar.


zeitgeist87:
Your solution is correct. In the future we should add some kind of `language-loaded` hook.

zeitgeist87:
Thanks!

PR #875, opened 2016-02-01
Colorful Solarized Light theme

zeitgeist87:
This patch uses more colors for the Solarized Light theme. All
punctuation is red, operators are green and variables are blue.

The result looks like this:

solarized-js
solarized-html

### DIFF
--- a/components.js
+++ b/components.js
@@ -32,7 +32,7 @@ var components = {
 		},
 		"prism-solarizedlight": {
 			"title": "Solarized Light",
-			"owner": "hectormatos2011 "
+			"owner": "hectormatos2011"
 		}
 	},
 	"languages": {
--- a/themes/prism-solarizedlight.css
+++ b/themes/prism-solarizedlight.css
@@ -88,7 +88,7 @@ pre[class*="language-"] {
 }
 
 .token.punctuation {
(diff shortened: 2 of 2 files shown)


LeaVerou:
Thanks for working on this!
However, I disagree with such a bright color for punctuation. In all our themes, punctuation has a color that will make it less noticeable, not more. The reason is that you don't want it to stand out when you scan the code, you want it to fade to the background so you can scan what actually matters.

LeaVerou:
This is how Solarized Light looks in my editor (Atom) btw, if it helps:

image

zeitgeist87:
> However, I disagree with such a bright color for punctuation.

On PR #872 you posted this picture:

751365da-c60c-11e5-9de1-9d6d813ed759

It has red punctuation. I just tried to match it as closely as possible.

zeitgeist87:
I could easily make an exception for `markup`:

```css
.language-markup .token.punctuation {
	color: #586e75; /* base01 */
}
```

LeaVerou:
Sorry :(
I think it might be good to collect samples for our core languages from various editors, compare them and try to do something similar. But it's kind of a chore to do that.

PR #876, opened 2016-02-01
Add support for the space character for the show-invisibles plugin

zeitgeist87:
This patch adds a bit of position magic, so that the invisibles
are displayed on top of the original character. This allows for
the inclusion of the space character.

The result looks like this:

invisibles

### DIFF
--- a/plugins/show-invisibles/prism-show-invisibles.css
+++ b/plugins/show-invisibles/prism-show-invisibles.css
@@ -1,7 +1,16 @@
+.token.tab:not(:empty),
+.token.cr,
+.token.lf,
+.token.space {
+	position: relative;
+}
+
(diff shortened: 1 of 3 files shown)


mAAdhaTTah:
Nice!

mAAdhaTTah:
I dig it. I think the saturation could be turned down a bit. Compared to how I have it in my editor, I think they're a bit dark and hard to distinguish from the characters surrounding them. I changed it to this:

```css
.token.space:before {
  color: hsl(24, 20%, 35%);
  content: '\00B7';
}
```

mAAdhaTTah:
Actually, that only looks better on the darker themes (which is what I tend to use). Not sure what the balance is here.

LeaVerou:
Nice!

PR #877, opened 2016-02-03
Fix DFS bug

zeitgeist87:
The DFS function uses an object as a key into another object. This
doesn't work, because the key gets stringified to "[object
Object]" and is therefore useless.


Golmote:
Looks like an quite efficient fix to me. You can start the id at 0 since you use pre-incrementation. 
Thanks for reacting so quickly!

LeaVerou:
Thanks from me too!

Hey, @Golmote, nice to see you again around here! 👋

PR #880, opened 2016-02-06
Fix Keep Markup plugin incorrect highlighting

### LINKED ISSUE #879: allow HTML <mark> tags
I love Prism, but I'm a little confused about why it is mutually exclusive with my own highlighting. Let's say I want to highlight an annotation in Java:

````html
<pre ><code class="language-java">package com.example;

public class Point
{
 
  …
 
  <mark>@Deprecated</mark>  //NOTE: annotation
  public double calculateSlope(final int x2, final int y2)
````

If I use just plain HTML5, the `<mark>` tag makes a nice highlighted section, e.g. with a yellow background. But Prism makes it as if I had never even used the `<mark>` tag. Why can't I use both? It would _sure_ be useful to use Prism, yet maintain the ability to highlight certain passages.

### PR DESCRIPTION
**Pull request by montogeek**
Hi!

This change fixes #879, it looks like that the Regex replace is messing the correct positions of the markers in the markup, that's why it looked _moved_ one position:
screen shot 2016-02-05 at 22 29 42

After this fix it looks like this:
screen shot 2016-02-06 at 00 05 33

I am not completely sure about this fix because I do not have experience with Prism, probably @Golmote or @LeaVerou can find the real issue.

Thanks!

### DIFF
--- a/plugins/keep-markup/prism-keep-markup.js
+++ b/plugins/keep-markup/prism-keep-markup.js
@@ -23,7 +23,7 @@
 				} else if(child.nodeType === 3) { // text
 					if(!firstWhiteSpaces) {
 						// We need to ignore the first white spaces in the code block
-						child.data = child.data.replace(/^(?:\r?\n|\r)/, '');
+						child.data = child.data.replace(/^(?:\r\n|\r)/, '');
 						firstWhiteSpaces = true;
 					}
 					pos += child.data.length;
--- a/plugins/keep-markup/prism-keep-markup.js
+++ b/plugins/keep-markup/prism-keep-markup.js
@@ -5,7 +5,6 @@
 	}
 
 	Prism.hooks.add('before-highlight', function (env) {
-		var firstWhiteSpaces = false;
 		var pos = 0;
 		var data = [];
 		var f = function (elt, baseNode) {
@@ -21,11 +20,6 @@
 				if (child.nodeType === 1) { // element
 					f(child);
 				} else if(child.nodeType === 3) { // text
-					if(!firstWhiteSpaces) {
-						// We need to ignore the first white spaces in the code block
-						child.data = child.data.replace(/^(?:\r\n|\r)/, '');
-						firstWhiteSpaces = true;
-					}
 					pos += child.data.length;
 				}
 			}

### CODE REVIEW

**Comment by LeaVerou** [comment]
Hey, thanks for looking into this! It's a nice little plugin and it's sad that it had that bug.

I’m very surprised this fix works. The new regex won't match newlines in many (if not all) modern systems, since they tend to be `\n` these days. Perhaps commenting out that line has the same effect?

**Comment by zeitgeist87** [comment]
> Perhaps commenting out that line has the same effect?

Yes it does. it works perfectly in Firefox and Chrome if you comment out the whole firstWhiteSpaces stuff.

**Comment by Golmote** [comment]
I guess I wrote it at a time where the initial line feeds were trimmed in the core.
Not sure how it would behave with the Remove initial line feed plugin...

**Comment by zeitgeist87** [comment]
> Not sure how it would behave with the Remove initial line feed plugin...

It would probably stop working in combination with any plugin, that trims whitespace or changes the length of the text in any way. Like my proposed whitespace normalizer plugin #847.

**Comment by montogeek** [comment]
Hey! thanks for answer :)

> Perhaps commenting out that line has the same effect?
Yes, by removing the `replace` line or removing the `firstWhiteSpaces` stuff like @zeitgeist87 said, it completely works.

**Comment by LeaVerou** [comment]
Good! Should I close this PR so you can send a new one or do you prefer to update it?

> Not sure how it would behave with the Remove initial line feed plugin...

We can make sure to include that plugin before this one.

**Comment by montogeek** [comment]
Hi @LeaVerou I just updated this PR with the latest changes.

**Comment by LeaVerou** [comment]
Awesome, thanks!

**Comment by montogeek** [comment]
Thanks to you!

**Comment by montogeek** [comment]
Hi @LeaVerou 
I noticed that the minified version of the plugin was not updated, it should be updated manually?

**Comment by LeaVerou** [comment]
Ah yeah, you need to run gulp.

**Comment by montogeek** [comment]
So, I should send another PR with the minified version updated to the `gh-pages` branch?
Or it is automatically updated when the `master` branch is synced and a new version is released?

I am just trying to understand how is the branch workflow :)

**Comment by zeitgeist87** [comment]
> So, I should send another PR with the minified version updated to the gh-pages branch?

It is usually a good idea, to include the minified version in your PR. You don't have to create a new PR for `keep-markup`, because I have updated the minified version for it a few minutes ago.

We periodically run gulp on the `gh-pages` branch to update all minified files, but there is no fully automatic update process.

**Comment by felixsanz** [comment]
@montogeek I used the example from website:

```
<pre><code class="language-css">
@media <mark>screen</mark> {
	div {
		<mark>text</mark>-decoration: <mark><mark>under</mark>line</mark>;
		back<mark>ground: url</mark>('foo.png');
	}
}</code></pre>
```

and the bug still happens, its not aligned

PR #884, opened 2016-02-10
Preserve Markup in Normalize-Whitespace plugin

zeitgeist87:
Since the Normalize-Whitespace plugin is only concerned with
whitespace, it is easy to preserve the markup for other plugins
like keep-markup.


LeaVerou:
Merged, thanks!

Btw, it's your plugin, you can commit without PRs. Unless you want feedback, in which case PRs are ok :)

LeaVerou:
Btw I used this yesterday. Works really well. The documentation needs some updating, if I didn't know I would assume it requires the Parse Settings plugin (which isn't even merged AFAIK!)

zeitgeist87:
> Btw, it's your plugin, you can commit without PRs. Unless you want feedback, in which case PRs are ok :)

Ok, I'll keep that in mind. In fact I wanted some feedback on this PR. This change is a bit intrusive:

```js
var html = before + env.element.innerHTML + after;
env.element.innerHTML = Normalizer.normalize(html, env.settings);
env.code = env.element.textContent;
```

It basically requires, that `normalize-whitespace` is loaded before any other plugin that manipulates `env.code`. It shouldn't be a problem, because no other plugin that I am aware of seems to change `env.code` directly.

> Btw I used this yesterday. Works really well.

Glad to hear it :smile: 

> The documentation needs some updating, if I didn't know I would assume it requires the Parse Settings plugin (which isn't even merged AFAIK!)

Yeah it is designed to work with Parse Settings, but it is not a hard requirement. I'll update the documentation to make this clear.

LeaVerou:
Oh, sorry, I had skimmed this very quickly.

Yes, that would cause issues with plugins that also modify env.code before yours runs, I think.
You could at least check if the keep markup plugin is actually loaded before doing all this.

Also, you're telling people to load the keep markup plugin after this one, but most people make their own build through our download page, so we should make sure this happens (we can introduce an `after` key in `components.js` and try to follow the order (try because it will not be possible if there are cycles).

zeitgeist87:
> You could at least check if the keep markup plugin is actually loaded before doing all this.

Agreed.

> (we can introduce an after key in components.js and try to follow the order (try because it will not be possible if there are cycles).

How about we add a new function to `Prism.hooks`. Something like this:

```js
unshift: function (name, callback) {
    var hooks = _.hooks.all;

    hooks[name] = hooks[name] || [];

    hooks[name].unshift(callback);
},
```

It would add the callback at the start of the list instead of the end. This way the order in which the plugins are loaded is no longer relevant. Any plugin that needs to run before all others uses `Prism.hooks.unshift()` instead of `Prism.hooks.add()`.

We could also add a third parameter to the `Prism.hooks.add()` function:

```js
add: function (name, callback, onTop) {
	var hooks = _.hooks.all;

	hooks[name] = hooks[name] || [];

	hooks[name][(onTop) ? 'unshift' : 'push'](callback);
},
```

LeaVerou:
This introduces latency during runtime though, whereas making sure the order is correct in the bundle is something that has zero performance impact when Prism is actually running.

zeitgeist87:
> This introduces latency during runtime though, whereas making sure the order is correct in the bundle is something that has zero performance impact when Prism is actually running.

Well, in case of the extra parameter for `Prism.hooks.add()` it is literally one extra if statement `(onTop) ? 'unshift' : 'push'` and it is only ever executed once or twice per plugin on startup. I doubt it would even be measurable.

Nevertheless, it is a good idea to implement a mechanism to get the right order in the bundle. I will work on an `after`-key in `components.js`.

zeitgeist87:
I think one problem is, that a lot of people, including myself, don't use the download page on http://prismjs.com. I for example get Prism through `bower`. So it would be useful if the order of the plugins  is also enforced at runtime.

LeaVerou:
Heh, this is why I don't like bower much. Is there even a way to specify which plugins you want with it or do you just get all of them?

The hooks data structure is public. Whoever wants to do fancy stuff with it can always do it manually, we don't need to add methods for every possible kind of array manipulation.

zeitgeist87:
> Heh, this is why I don't like bower much. Is there even a way to specify which plugins you want with it or do you just get all of them?

I think you can only specify the version of Prism you want. You basically get the whole git repo into the directory `bower_components/prism`. I then use `gulp` to pick the stuff I need from  `bower_components/prism/components` and merge it into a big minified javascript file. The nice thing about it is, that everything is automated.

> The hooks data structure is public. Whoever wants to do fancy stuff with it can always do it manually, we don't need to add methods for every possible kind of array manipulation.

Ah I see. With `Prism.hooks.all` I can access the whole data structure.

PR #885, opened 2016-02-10
Corrected instructions in command-line plugin

sbrl:
It instructes the user to add the `command-line` class to the `<pre>` tag, but it actually needs adding to the `<code>` tag in order to get it to work.

The other tags are fine on the `<pre>` tag.

### DIFF
--- a/plugins/command-line/index.html
+++ b/plugins/command-line/index.html
@@ -28,7 +28,7 @@
 
 	<p>This is intended for code blocks (<code class="language-markup">&lt;pre>&lt;code></code>) and not for inline code.</p>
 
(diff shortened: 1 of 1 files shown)


zeitgeist87:
Hi @sbrl,

Thank you for contributing!

It should work on both tags and the examples in `plugins/command-line/index.html` have it on the `<pre>`-tag and it seems to work. Rather than updating the documentation, we should fix it. Could you provide an online example that reproduces this bug?

sbrl:
@zeitgeist87 Never mind, it is working as expected now that I get it isolated on a test page. Here's a link to the test page on my website:

https://starbeamrainbowlabs.com/libraries/prism-test.html

I'm sure it wasn't working before...!

PR #889, opened 2016-02-12
Fix catastrophic backtracking regex issues

zeitgeist87:
How to reproduce the bug:

* Go to http://prismjs.com/download.html
* Click on `Select/unselect all`

Then one CPU core will run at 100% forever and the highlighting will never finish. I tracked down the problem to the `'template-string'`-Regexp in the JavaScript language. It is supposed to match a multiline string, so if it doesn't find a match it will scan to the end of the file and then backtrack. Every backslash doubles the number of steps needed https://regex101.com/r/jV4nX7/1 That is exponential growth!

The regex is also invalid as demonstrated here https://regex101.com/r/oY7mE5/1

My fix is not only correct, but it also performs well enough so that the download page works https://regex101.com/r/rW9dL9/1

This PR also contains a fix for a bug in the multiline string regexp for Bash.


zeitgeist87:
I created a PR for this, because I would like someone to double-check my regexps.

Golmote:
Looks good to me. Thanks for working on this!

LeaVerou:
Awesome, thanks for catching this and for fixing it!

Issue #887, opened 2016-02-11
HTML Code Block Issue For Better Solution


hlaporthein:
Hello All, 

I found prismjs yesterday when googling. It is really awesome. It's features meet all my requirements. But I have a little problems in html markup.

For example if I would like to create below html code section

Code No.1
```
<pre>
    <code class="language-markup">
        <p class="2020" id="ok2o">
            one
        </p>
    </code>
</pre>
```

The above code only give me result **one**, I have no idea why paragraph tag is missing. After that I found out the problems, I would need to do HTML Escape to work like below code (No.2)

Code No.2
```
<pre>
    <code class="language-markup">
        &lt;p class="2020" id="ok2o"&gt;
            one
        &lt;/p&gt;
    </code>
</pre>
```

Let's say, if the above code.2 method is correct, how should I solve over **escaping fifty html tags**? I know there are some html escape convert tools. In facts, I know that html escape tools is good if I need to escape more html tags. 

So, My Question is 

Is using html escape convert tools  correct method?
Does prismjs have better solution for that?
Or... Should I create javascript function (like code No.3) to search and escape? ( I think that is not good approach.). Please give me advice and Thanks.

Code No.3
```
String.prototype.escape = function() {
    var tagsToReplace = {
        '<': '&lt;',
        '>': '&gt;'
    };
    return this.replace(/[&<>]/g, function(tag) {
        return tagsToReplace[tag] || tag;
    });
};

jQuery(document).ready(function($){

    $('code').each(function(){
    var  org_val = $(this).html(),
         esc_val = org_val.escape();
         $(this).html(esc_val);
    });
});
```

Golmote:
You are right. HTML code needs to be escaped to be displayed properly in Prism. Note that this is not related to Prism but to the way the browsers interpret content inside pre and code.

LeaVerou:
I’ve had this problem myself and I use `<script type="text/plain">...</script>` elements.
Then I use a small script to convert them to `<pre><code>`:

```js
$$("script[type='text/plain']").forEach(script=>{
	var pre = document.createElement("pre");
	var code = document.createElement("code");
	code.className = script.className;
	pre.appendChild(code);
	script.parentNode.insertBefore(pre, script);
	code.appendChild(script);
});
```

Note that this *depends* on Prism’s behavior of removing markup. If using the keep markup plugin, it will need a few adjustments.

I should clean up the code and add it as a plugin, since that's such a common request.

hlaporthein:
[...]

PR #890, opened 2016-02-13
Add before-highlightall hook

LeaVerou:
Needed for unescaped markup plugin, as discussed in #887.
Alternatively, we could reinstate @zeitgeist87’s PR about making the selector a config option.
Thoughts?

### DIFF
--- a/components/prism-core.js
+++ b/components/prism-core.js
@@ -154,10 +154,17 @@ var _ = _self.Prism = {
 	plugins: {},
 	
 	highlightAll: function(async, callback) {
-		var elements = document.querySelectorAll('code[class*="language-"], [class*="language-"] code, code[class*="lang-"], [class*="lang-"] code');
+		var env = {
+			callback: callback,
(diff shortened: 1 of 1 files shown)


mAAdhaTTah:
If we're adding a `before-highlightall` hook, can we get an `after-highlightall` hook as well?

LeaVerou:
For parity or because there are use cases?

mAAdhaTTah:
A place to resolve a promise was what I was thinking originally, though in retrospect, it wouldn't actually handle what I was trying to do at the time, which was resolve a promise when all the async highlighting was finished. I ended up not going that route, so maybe it's not necessary, but I thought it might still be useful to include.

zeitgeist87:
> Alternatively, we could reinstate @zeitgeist87’s PR about making the selector a config option.

The hook `before-highlightall` is definitely a better solution than the config option. How about this small addition:

```js
var elements = env.elements || document.querySelectorAll(env.selector);
```

This would optionally enable a plugin to select/filter the elements directly. Could be useful in the future...

LeaVerou:
I like that idea!

Issue #286, opened 2014-07-05
Prism toolbar


vdecree:
Hi there,

Does prism still use prism toolbar plugin so that you can easily get the 'copy to clipboard' button to appear? I've been trying but it doesn't seem to work.

Regards
M

geoffreymcgill:
The following `Copy to Clipboard` sample should be helpful to anyone wanting to implement this functionality within their Prism code sample.

https://gist.github.com/geoffreymcgill/d118939d1d358d2ed5e8

This sample is adapted from a sample originally posted by [Philip Lawrence](http://dev.misterphilip.com/prism/plugins/toolbar/).

Hope this helps.

vkbansal:
@geoffreymcgill Thanks for the link! I was working on something similar. 

@LeaVerou Will it make sense to have a plugin that depends on third party tools in the official repo?

LeaVerou:
Sure, that’s what plugins are for. (of course it’s always better to not depend on 3rd party tools, but we’re not going to reinvent the wheel for no reason — we reinvent the wheel only when it results in shorter/more elegant code :) ).

mAAdhaTTah:
I'm definitely in need of that, @vkbansal; was going to look at tackling it myself but probably not for a while.

vkbansal:
@mAAdhaTTah can you tell what features are commonly required in a toolbar?

mAAdhaTTah:
I mostly need "copy to clipboard," but leaving it open for hooking into would also be great, as I also will need to add an edit button that will take you to the backend (if possible).

LeaVerou:
Is this still only solvable with Flash or does the Clipboard API support this now?

Golmote:
AFAIK, still only Flash. And ZeroClipboard (used in the above example) is the reference for this.

santo007:
@Golmote @mAAdhaTTah @vkbansal <br/>Get the plugin from the public repository i have created <a href="https://github.com/santo007/prism-toolbar">Prism-toolbar</a>

vkbansal:
@santo007 nice!

uranusjr:
For this particular use-case only *copy to clipboard* is needed, and surprisingly (at least to me) support for it [seems pretty decent](http://caniuse.com/#feat=clipboard). It’s mostly only unsupported on mobile browsers, so maybe a Flash-free solution is reasonable?

Golmote:
[...]

Issue #280, opened 2014-06-20
Copy code snippets


rlavoie:
Is there a feature to copy code snippets rather than highlighting all.

Golmote:
This showed up recently: https://github.com/zenorocha/clipboard.js.
Just thought I would let you know ;)

zenorocha:
+1 for clipboard.js ;)

Golmote:
This issue is fixed with those PRs merged.

PR #891, opened 2016-02-13
Plugins: Toolbar & Copy to Clipboard

mAAdhaTTah:
I'm using Prism as the syntax highlighter for my WordPress plugin, [WP-Gistpen](https://github.com/mAAdhaTTah/WP-Gistpen). I've built a toolbar plugin that I'd like to contribute to Prism. I've got a first pass at extracting the plugin from its CommonJS origins,, but I need some feedback on how to change it to fit the Prism way of doing things, e.g. the HTML API. I also see some possible synergy with the [Parse Settings](https://github.com/PrismJS/prism/pull/851) plugin.

The Toolbar exposes two methods: a `hook` method, which is registered with Prism's `complete` hook, and a `registerButton` method, which takes a callback that creates the buttons. The callback will be passed the `env` and should return an `a` or `span` element, which other plugins can use to register themselves with the Toolbar to get their text or link to display.

I've also created a [Clipboard](https://github.com/mAAdhaTTah/WP-Gistpen/blob/7b26f5027eba6d65fc54432401459811f98c0168/src/js/prism/clipboard.js) plugin is an example of this callback, which creates an anchor tag that, when clicked, copies the code to the clipboard. I plan to extract that into a Prism plugin as well, but I'm not sure how Prism wants to handle dependencies like [Clipboard.js](https://clipboardjs.com/), or how the download page will work with internal dependencies like this. In the interest of user friendliness, what's the best approach to take here?

Prior art:
- https://github.com/PrismJS/prism/issues/280
- https://github.com/PrismJS/prism/issues/286
- http://dev.misterphilip.com/prism/plugins/toolbar/
- https://github.com/PrismJS/prism/pull/142/files

Open Questions:
- [x] Should the `registerButton` method be polymorphic?
- [x] How can the CSS styles be improved?
- [x] How should we handle 3rd party dependencies, like [Clipboard.js](https://clipboardjs.com/)?
- [x] What should the HTML API be? (See: #910)
- [x] How do we deal with this plugin & filename plugin conflicts? (See: #910, #914)


mAAdhaTTah:
@LeaVerou @Golmote @zeitgeist87 I'm on vacation this coming week, so I should have some time to tweak this and extract the other plugin to your preferences, if you have guys have a chance to take a look and let me know what you think.

Thanks!

zeitgeist87:
Hi @mAAdhaTTah,

I've tested your plugin and it works fine. I have a few comments on the plugin itself, but I can't really help you with your question about third-party dependencies.

* Why do you need to expose the `hook` method? Is there a need to execute the hook manually?
* You could additionally to `registerButton` provide a higher level method to create a button. Something like: `createButton(text, onClick)`
* I think the helloworld button should do something. Maybe something like `alert('Hello World!');`
* You could also create a useful demo button, that selects all of the code in the box: http://stackoverflow.com/questions/11128130/select-text-in-javascript
* The CSS styling looks a bit boring, but that is just my opinion

Thanks for contributing to Prism! Sorry for the late response.

mAAdhaTTah:
> Why do you need to expose the `hook` method? Is there a need to execute the hook manually?

It was mostly to expose the function in case a plugin/developer wanted to unhook and possibly run manually. Not sure if it's necessary, but figured the extra flexibility would be helpful down the line.

> You could additionally to `registerButton` provide a higher level method to create a button. Something like: `createButton(text, onClick)`

This is actually a good idea. What about making the `registerButton` method polymorphic:

```javascript
function registerButton(opts) {
  var callback;

  if (typeof opts === 'function') {
    callback = opts;
  } else {
    callback = function(env) {
      // create element from opts.text & optional opts.onClick
    };
  }
  callbacks.push(callback);
}
```

I would probably set this up to ensure the `env` is passed into the `onClick` method as well.

> I think the helloworld button should do something. Maybe something like `alert('Hello World!');`

Can do. And if we like the polymorphic idea, we can use the helloworld button to demonstrate how to do it that way.

> You could also create a useful demo button, that selects all of the code in the box: http://stackoverflow.com/questions/11128130/select-text-in-javascript

Interesting. Might be more useful as an example than the helloworld button. The Clipboard.js implementation uses "select-all" as a fallback when the actual copying fails (primarily in Safari, maybe older browsers).

> The CSS styling looks a bit boring, but that is just my opinion

I mostly copied it from the [original implementation](https://github.com/PrismJS/prism/pull/142/files) with some tweaks (I preferred having each button appear visually distinct), but I am not really a designer, so if anyone has any improvements, I'm all ears.

> Thanks for contributing to Prism! Sorry for the late response.

Not a problem! Figured it was worth following up, since you guys are pretty active. I've added checkboxes for the open questions remaining on this PR. If the polymorphic idea is acceptable, I should be able to make those changes tonight.

zeitgeist87:
> This is actually a good idea. What about making the registerButton method polymorphic:

I like that idea. It also allows later additions to the interface. Maybe someone would like to implement buttons with icons or something else...

Golmote:
Regarding third party scripts, I think the plugin page should simply mention explicitly that script is required. And that's it. ^^ I think Prism should not distribute directly the third party script.

mAAdhaTTah:
@golmote Any issues/considerations for Node/Browserify users? I don't know if that matters on the Node side, but Browserify users, I believe, will have to explicitly assign the Clipboard var as a global. I'll have to take a closer look myself, but would a `require` check and call be acceptable? And would we be ok including it as a dependency in `package.json`?

Golmote:
I really don't feel like including it as a dependency of Prism globally. 
Don't know much about Browserify so I really can't tell.

LeaVerou:
Yeah, I would be opposed to adding global dependencies to support a plugin.

However, I'm not sure about just mentioning that Clipboard.js is needed. Part of why Prism is so easy to use is that you download a copy and it just works. I wonder if we can use gulp to pull in the latest version from clipboard.js’ repo and bundle it with the plugin.

mAAdhaTTah:
There is an `optionalDependencies` key, but it doesn't appear to function differently from regular dependencies. I'll investigate some options and report back.

mAAdhaTTah:
So [`optionalDependencies`](https://docs.npmjs.com/files/package.json#optionaldependencies) would mean that the dependency gets installed when installed from `npm`, but it's not required, so if it's not found or something, the install would still work. I don't know if that's acceptable, but one option is to add Clipboard.js as an `optionalDependency`. Presumably, if you're installing from `npm`, you have a build process (likely Browserify/Webpack), which should work with the `require` check I added (though I still need to fully test that).

Alternatively:

> I wonder if we can use gulp to pull in the latest version from clipboard.js’ repo and bundle it with the plugin.

This should be possible; maybe we could do something similar to the way the autoloader's language dependencies map is built. It would mean that the Clipboard.js library would be inlined into the Copy to Clipboard plugin.

> Alternatively, the plugin could load the CDN version of clipboard.js dynamically.

The only issue with this is we can't be sure the CDN asset will be fully loaded before the highlighting starts, unless `DOMContentLoaded` will only fire after all assets (including async ones) are loaded.

The only other idea I had was to modify the downloads page and ensure Clipboard.js is loaded first there when selecting the Copy to Clipboard plugin, but I'm not sure how to accomplish that at first glance of the `downloads.js` file.

I've pushed a first pass at Copy to Clipboard plugin. The demo page has Clipboard.js loaded from the CDN. Definitely interested in your feedback.

plugins/toolbar/prism-toolbar.js
```diff
@@ -0,0 +1,82 @@
...
+		if (typeof opts === 'function') {
+			callback = opts;
+		} else {
+			callback = function (env) {
+				var element;
+
+				if (typeof opts.onClick === 'function') {
+					element = document.createElement('a');
+					element.addEventListener('click', function () {
+						opts.onClick(env);
```
zeitgeist87: I would also pass the `this` pointer like this:

```js
opts.onClick.call(this, env, opts);
```

That way the callback function gets access to the element that was clicked.

plugins/toolbar/prism-toolbar.js
```diff
@@ -0,0 +1,82 @@
...
+					});
+				} else {
+					element = document.createElement('span');
+				}
+
+				element.innerHTML = opts.text;
```
zeitgeist87: Is this use of `innerHTML` intentional or would `textContent` be enough?

zeitgeist87:
The design on the https://clipboardjs.com/ homepage looks nice. They have nice copy buttons for their source code samples. Nice fade-in and fade-out animations, nice icon, colored box-shadows and small transparent speech bubbles with status messages. We could use that design as an inspiration.

plugins/toolbar/prism-toolbar.js
mAAdhaTTah: I don't know how important it is, but I didn't want to change the `this` value of the passed in callback in case the developer is relying on it.

plugins/toolbar/prism-toolbar.js
mAAdhaTTah: Should be, actually. I don't expect (or even want) other buttons to provide more complicated HTML content.

plugins/toolbar/prism-toolbar.js
zeitgeist87: Well without it, the `this` pointer will point to the `opts` object and the developer has no access to the created DOM element. I think it is standard practice to set the `this` pointer to something sensible, like the source of the click event. In what specific use case would the developer want `this` to be the `opts` object?

mAAdhaTTah: I guess it was more about not messing with developer expectations, but I think you're right. Will update.

mAAdhaTTah:
Added a hover animation and a drop shadown. Does that make it pop a little bit more?

Also, I added `Clipboard.js` to `optionalDependencies`, but if that's not something you want, I can remove that commit. Even with that added, the download page doesn't bundle with `Clipboard.js`. Should it?

mAAdhaTTah:
Bump. Thoughts on handling the dependency?

LeaVerou:
What about checking if Clipboard.js is loaded (e.g. via some global it defines) and if not, loading it dynamically from the CDN? That way we don't have the issues of bundling a possibly outdated version or the hassle of requiring users to load it themselves.

mAAdhaTTah:
The issue I had with that is you can't be sure the CDN dependency is loaded before highlighting begins. How would you suggest resolving that?

LeaVerou:
You can listen to the script's load event, then add the toolbar to already highlighted elements, and use the hook for elements that are going to be highlighted in the future.

mAAdhaTTah:
@LeaVerou Thanks! I think I got it. This should cover all use cases. You can load it yourself, it will get compiled automatically if you're using browserify/webpack, and if you don't do either, it'll load automatically from a CDN.

mAAdhaTTah:
I added two open questions to the PR as well, related to the discussions around the filename plugin and/or a potential HTML API for this plugin to automatically add labels to the toolbar from data-attributes.

mAAdhaTTah:
Bump. Any thoughts on the remaining open issues?

Golmote:
Regarding the HTML API, isn't registerButton enough?  Or am I missing the exact expectation here?

Golmote:
Regarding the Show Language plugin (or anything it might evolve into), couldn't it require this plugin and use it just like the Copy to clipboard plugin does?

mAAdhaTTah:
> Regarding the HTML API, isn't registerButton enough? Or am I missing the exact expectation here?

There was a suggestion to add something like `data-label` to read from and add a label to highlighted code.

> Regarding the Show Language plugin (or anything it might evolve into), couldn't it require this plugin and use it just like the Copy to clipboard plugin does?

It can; the question is whether we subsume the current plugin into an extension for the toolbar, how that impacts BC, etc.

mAAdhaTTah:
Bumping this PR again. I've seen @LeaVerou mention a few times that expanding the Show Language plugin to be more generic would be a good idea. I've been spending a lot of time lately on my plugin that uses PrismJS, so I'm wondering how feasible it is to get some of this work merged back upstream. The `toolbar` plugin would be useful for #1030, which was opened recently.

My proposal here is to make the show-language plugin dependent on the toolbar plugin, using the same API in the HTML to provide the language. The major difference is the styles changed in order for me to get it work. The `position: absolute` of the `show-language` styles made it difficult to include several buttons without positioning each individually.

The first question I have is whether this is an acceptable approach, as I'm not sure how to deal with BC, if that's even an issue here. I'm honestly not entirely certain how you think about BC; are you assuming people have to download the latest version from the site? If so, then as long as the downloadable work, it should be BC.

The other issue I had last time I played around with this was getting the download page to automatically check the `toolbar` plugin w/ `show-language`. Any thoughts on handling that?

Thanks!

Golmote:
> The first question I have is whether this is an acceptable approach, as I'm not sure how to deal with BC, if that's even an issue here. I'm honestly not entirely certain how you think about BC; are you assuming people have to download the latest version from the site? If so, then as long as the downloadable work, it should be BC.

I think @LeaVerou should decide about this. I don't know if CSS changes can be considered BC breaking in our case.

> The other issue I had last time I played around with this was getting the download page to automatically check the toolbar plugin w/ show-language. Any thoughts on handling that?

I had to deal with this in the Previewers plugins (even though it's on my todo list to merge them into a single plugin... as suggested in #913). If I remember properly, you just have to add the `require` key in the `components.js` file, just like you would do for a language.

LeaVerou:
I'm not super concerned with BC on plugins, especially when the improvement is quite substantial. It's a concern, but a low priority one.

mAAdhaTTah:
Ok, I'll update this PR to update the show-language plugin to depend on the toolbar plugin.

LeaVerou:
Wait, what? How did we get to such a dependency? What I was agreeing on was refactoring the Show Language plugin to be more generic…

mAAdhaTTah:
The Toolbar plugin is the "generic" version, with which you can register a callback to generate a label on a prism code snippet. The Show Language plugin will turn into a plugin that registers a callback with the Toolbar to show the language.

LeaVerou:
I'm not sure I understand why we need two plugins where one's functionality is a subset of the other. In similar cases when this happened in the past, we just kept one (e.g. whitespace normalizer vs trim first line plugin). We can just have a setting that shows the language.
Also, "Toolbar" is a rather confusing name for a plugin that shows labels on code snippets. What about "Show labels" or something?

Golmote:
The Toolbar plugin was meant to be the hook on which other plugins could be attached, like Show language, Copy to clipboard, Download source file, you name it.

mAAdhaTTah:
> I'm not sure I understand why we need two plugins where one's functionality is a subset of the other. In similar cases when this happened in the past, we just kept one (e.g. whitespace normalizer vs trim first line plugin). We can just have a setting that shows the language.

Toolbar doesn't do anything on its own. It just exposes an API to allow other plugins to register labels to it, which the Toolbar plugin will add during highlighting. The Show Language plugin will be one of these plugins; all it will do is register the callback to create the language label. So there isn't really any overlap. Toolbar manages & adds all of the label; Show Language creates one of these labels for the Toolbar to add.

> Also, "Toolbar" is a rather confusing name for a plugin that shows labels on code snippets. What about "Show labels" or something?

Happy to oblige. Should get a chance to work on this over the weekend.

Golmote:
I personally don't think the name should be changed. The name "Toolbar" reflects the plugin ability to display labels and / or button to perform actions.

The plugins should allow custom labels and / or action buttons to be added in the simplest possible way. And it should be used as a "container" for the Show Language plugin and the hypothetic plugins Copy to clipboard (issue #280) and Download source (issue #1030).

mAAdhaTTah:
@Golmote Copy to Clipboard is included in this PR, so not so "hypothetic" :).

Golmote:
Oh, sorry about that! You're right! ^^

LeaVerou:
And what if people want to display custom labels? They write their own plugin? This goes against our design goal of being able to do things with HTML alone.

Golmote:
Having an HTML API has been mentioned as a good idea in this very thread : https://github.com/PrismJS/prism/pull/891#issuecomment-214761557

LeaVerou:
Here's a different idea, not sure if it's a good one, but I'm throwing it on the table anyway:

What about, the label plugin sets a CSS variable with the language name (e.g. `--language`). Then, in the CSS for the label, it uses `content: var(--label, var(--language))`. This means that:
a) If `--label` is not defined, it will fall back to `--language`
b) If `--language` is not defined, there will be no pseudo element

Sure, `style="--label: 'foo'"` is a bit more syntax than `data-label="foo" but not a crazy amount. And you get inheritance out of the box!

Golmote:
Should we really use CSS variables that early? I would be in favor of waiting for better support first.

LeaVerou:
They are supported everywhere except Edge (and they're in development on Edge) and this is a plugin.

Golmote:
Still. IE11 and Edge represent a large part of global usage. Our plugins should not be that less compatible than Prism's core.

LeaVerou:
Global usage doesn't matter in this case: Prism is used on sites that *developers* are reading, not random people. Developers tend to use modern browsers. When Prism itself came out, every other library had IE8 support, and Prism did not, for the same reason. Just over the years, Prism's browser support got better as people were upgrading. That shouldn't make us conservative today though.

Golmote:
IMO we should still stay compatible with at least the last version of every major browser, including edge. We should not restrict the usage of Prism just because we think it's used only on tech oriented websites read by people not using IE. We cannot make that assumption just like this. 

Moreover, an HTML API would be perfectly usable, as much easy to implement and more coherent with all the other plugins.

I can't see any valuable reason to insist on using CSS variables here.

mAAdhaTTah:
If the design goal is to be able to do things with HTML alone, why would we use CSS variables?

Golmote:
I guess she meant "without JS" instead of "with HTML alone".

LeaVerou:
> I also don't know CSS variables (although I feel like I read there are differences between them and say, SASS variables), so I would have to spend some time learning them in order to implement this myself.

They are more like custom properties than Sass variables. They work exactly like any CSS inheritable property (including setting/getting with JS), and they start with `--`. You call them with `var()`, e.g. `var(--foo)`. The second argument to `var()` is a fallback if the variable is not set. There, now that you know CSS variables, perhaps we can move on? :)

mAAdhaTTah:
Ok, but I'm still not sure of the benefit over `data-label`. Wouldn't the use of `content` be bad for accessibility?

LeaVerou:
Screen readers are supposed to read generated content, per the specification.

Anyway, as I said, I was just throwing an idea out there, it's not necessarily a good one. But browser support or accessibility are not the reasons why it may not be good, I'm more concerned about usability for our users.

Golmote:
Hence the HTML API. Which might even allow to add multiple labels at once, set the order and configure other stuff.

mAAdhaTTah:
> Anyway, as I said, I was just throwing an idea out there, it's not necessarily a good one.

No worries. I guess I'm just not sure what the benefits are of CSS variables over an HTML data-attribute API.

If I'm following correctly, the Show Language plugin would inject the CSS `--language` variable into an inline-style on the `pre` element. The CSS `content` uses the `--label` first, which would be set by the Prism user on the HTML element, falling back to the injected `--language` if `--label` isn't found. Is this correct?

I think how the `--language` variable gets injected is where I'm not sure, unless that's set in the CSS, in which case, we'd need to create `language-{languageKey}` classes in the CSS for every language to set it.

That would decouple things, which would be nice, though wouldn't it mean the user could have either a label or a language, but not both?

> I'm more concerned about usability for our users.

That's part of why I pointed out my ignorance of the feature. I'm not sure its widespread/common enough to be in the toolset of an average developer (yet). Also, the data-attribute method is already in use in other plugins, so it would be more consistent for users to continue that pattern rather than introduce a new pattern here.

All of that said, I'll play around with this idea when I get chance to work on this so we can see what it looks like and how we feel about it.

> Hence the HTML API. Which might even allow to add multiple labels at once, set the order and configure other stuff.

Do you have other ideas about the HTML API besides just a simple `data-label`?

LeaVerou:
> No worries. I guess I'm just not sure what the benefits are of CSS variables over an HTML data-attribute API.

1. Inheritance out of the box, i.e. users can set a default `--label` on `<body>` or `<html>` and use that for every code snippet.
2. People can set it in their CSS, it can use other variables etc. It's more flexible.

> If I'm following correctly, the Show Language plugin would inject the CSS --language variable into an inline-style on the pre element. The CSS content uses the --label first, which would be set by the Prism user on the HTML element, falling back to the injected --language if --label isn't found. Is this correct?

Yes.

> I think how the --language variable gets injected is where I'm not sure

Not sure why you're confused, I said you set it and get it exactly like any other CSS property. 

```js
// Get variable from inline style
element.style.getPropertyValue("--foo");

// Get variable from wherever
getComputedStyle(element).getPropertyValue("--foo");

// Set variable on inline style
element.style.setProperty("--foo", 38 + 4);
```

mAAdhaTTah:
> Not sure why you're confused, I said you set it and get it exactly like any other CSS property.

Well, yes, I assumed you _can_ do it w/ JavaScript. But for `--language` specifically, is that how you want to inject it, with JavaScript, for the show-language plugin? I also mentioned we could do it in CSS, with a class for each language setting the custom property. I think this is more for our implementation than userland.

Also, the current implementation is intended to allow multiple labels, e.g. the plugin I'm building has a copy button, an edit button (for logged in users to easily edit the snippet, so that's a link) and the language. Right now, that's handled by registering an element-returning callback with the toolbar plugin. How do you envision that working?

Right now, the Toolbar just accepts callback or configuration objects to register a label or button. Is this label functionality with the CSS variables going into the Toolbar itself or the label registered by the Show Language plugin?

LeaVerou:
If there are buttons etc that's not gonna work. `data-` attributes it is then!

mAAdhaTTah:
Updated the Show Languages plugin & added a basic HTML API using `data-label` to add labels to the code snippet. Any ideas for anything more complicated/extensive?

plugins/toolbar/prism-toolbar.js
```diff
@@ -0,0 +1,100 @@
...
+		var callback;
+
+		if (typeof opts === 'function') {
+			callback = opts;
+		} else {
+			callback = function (env) {
+				var element;
+
+				if (typeof opts.onClick === 'function') {
+					element = document.createElement('a');
```
Golmote: We should probably use a `<button type="button">` element here, instead of an `<a>`.

Golmote:
I think there should be a way to reorder the labels/actions in the toolbar, especially since some of them may be added automatically by plugins, and so depend on the order in which the plugins executed.

There could be a way to add buttons with the HTML API by providing a `data-label-onclick` attribute that takes the name of a global function?

There should also be a way to add multiple labels, maybe with numbered attributes, like data-label-1, data-label-2, etc.

```html
<script>
function version_info() { alert('Some info about the version'); };
</script>
<pre data-label-1="v2.8" data-label-1-onclick="version_info" data-label-2="Download lib" data-label-2-href="http://example.com/lib.js">
```

I'm still very bad at naming things, so there are really just suggestions...

mAAdhaTTah:
That sounds interesting, but how do we coordinate between the HTML API & the JS API? Copy to Clipboard isn't using the HTML API; should it just go second, all of the HTML-based labels/buttons have been added?

Regarding the `data-label-onclick`, if you're going to have to write JavaScript anyway, I think I'd rather push the user to use the JS API than reference a global function. Providing a url via HTML attributes would be useful though.

Golmote:
You're right the data-label-onclick is quite useless. 

Regarding the coordination, I think there should be a way to position the buttons that were added through plugins. Maybe this could be achieved with other data attributes

mAAdhaTTah:
I fixed the `a` -> `button` and added support for `data-url` to add a link instead of just a plain label. I was looking into managing multiple labels, and I'm beginning to feel that API might get ugly & unwieldy. I think if you need multiple labels/buttons, you should be doing them in JS.

The HTML API provides a simple API for adding a single label. If you need multiple labels, or onClick handlers, or anything else more complex, I think we should push users to use the JS API.

Golmote:
I'm fine with it. If there is a real need for multiple labels through HTML API, it might be added later anyway. 

What about the order, when using plugins? Can you think of a way to handle this using HTML?

mAAdhaTTah:
We could ask plugins to register by name, then have users declare on a the `html` or `body` the order they should appear, e.g.:

```
<html data-toolbar-order="['label', 'clipboard', 'language']">
```

We can fall back to the order they were registered if the data attribute doesn't exist.

Golmote:
It seems good. Though I don't think there's a need for the brackets and single quotes. A comma separated list of aliases should be enough.

Would it be possible to control the position of the label added through the data-label attribute this way? Maybe by giving it an arbitrary alias?

mAAdhaTTah:
> Would it be possible to control the position of the label added through the data-label attribute this way? Maybe by giving it an arbitrary alias?

In the above example, `label` was supposed to refer to the position of the `data-label` attribute. Maybe `toolbar` is more explicit?

Golmote:
Oh no, sorry, "label" is fine! I missed it somehow. 

The plugins should probably register their exact name for consistency, i.e. "show-language".

LeaVerou:
Idea: What about enabling `data-label` to link to a `<template>` element containing arbitrary HTML? Then people can have custom buttons and everything in that, with their own event handlers.
Of course it shouldn't be the only way to specify a label, because having to include a separate `<template>` element if all you want is a textual label is kind of annoying.

mAAdhaTTah:
Just updated to control the order via data-attribute.

@LeaVerou I'll take a look at that later today. Would be useful if someone wanted to include an icon or something of that nature. How would someone declare their event handlers in a `<template>`? Unless you just mean inline JS, e.g:

```
<template id="my-custom-button">
    <button onclick="alert('Hi')">Hello!</button>
</template>
```

mAAdhaTTah:
If we wanted to go crazy w/ that idea, we could even implement a mini-DSL and allow people to slot in the code snippet's language and such into the label:

```html
<template id="my-custom-button">
    <button onclick="alert('Hi')">Hello {{language}}!</button>
</template>
```

mAAdhaTTah:
@LeaVerou PR updated to include that suggestion.

I think that addresses all of the questions I had when I opened the PR originally. I know this is a big one, and I just want to say thank you for everyone's patience and feedback in working through this.

mAAdhaTTah:
Will have some time this weekend to make edits as needed. Bumping this PR; maybe get a merge?

Golmote:
Sorry for the late answer. I'm merging now.

mAAdhaTTah:
No worries! Happy to contribute.

PR #893, opened 2016-02-15
Use a HTML-comment to escape code

zeitgeist87:
If an element does not contain any code, but a single HTML-comment, then the comment is used as code.

```html
<code class="language-markup"><!-- <p>Example</p> --><code>
```

It works very well, but I feel silly for even proposing this. It seems semantically wrong to abuse a comment like this. I created this to contribute to the discussion around #887.


LeaVerou:
No matter what we do for this, it will be a hack. Which is ok, as long as we keep it out of the core and in the unescaped markup plugin. Wanna move the commit to that? :)

LeaVerou:
Just please amend the documentation to mention this, and make sure to only apply it for block code (it's much more likely that inline code will actually contain a single comment, for block code I think it's kinda rare).
Nice hack!

zeitgeist87:
I've updated the PR. Unfortunately some small changes to `prism-core.js` are still necessary. I had to move the check for empty code below the `before-highlight` hook.

LeaVerou:
I’m a bit worried that this could be a breaking change for third party plugins. Perhaps we should add a new hook before the check?

zeitgeist87:
> I’m a bit worried that this could be a breaking change for third party plugins.

I think the chances of that happening are quite low, because `env.code` contains an empty string at that point and not something unexpected like `undefined` or `null`. Any plugin should be able to handle that. It seems to me, the if-statement is mainly a performance improvement that allows Prism to return early if there is nothing to highlight. In my opinion there is no reason to add another hook.

zeitgeist87:
On the other hand, a new hook would guarantee, that the `unescape-markup` plugin runs before any other plugin that needs to change `env.code`.

zeitgeist87:
@LeaVerou I've resolved the merge conflicts for this PR. Do you think it is ready to be merged?

LeaVerou:
Sure, ok! Thanks!

PR #904, opened 2016-02-26
Add after-property to allow ordering of plugins

zeitgeist87:
Some plugins do not depend on each other and can be used
independently, but if they are used at the same time, they have to
be executed in a specific order.

The "after"-property in components.js, allows the download page to
concatenate the plugins in the right order, without enforcing a
hard dependency.


LeaVerou:
Awesome, thanks!!!

PR #906, opened 2016-03-09
Update linear-gradient syntax from `left` to `to right`

### LINKED ISSUE #881: linear-gradient syntax in CSS
In these two lines:
https://github.com/PrismJS/prism/blob/gh-pages/themes/prism-twilight.css#L166
https://github.com/PrismJS/prism/blob/gh-pages/plugins/line-highlight/prism-line-highlight.css#L17
```linear-gradient``` attribute should be set according to this guidelines: http://www.w3schools.com/css/css3_gradients.asp#grad3

davewasmer:
Fixes #881


LeaVerou:
Nice, thanks! We also don't need the prefixes anymore.

PR #919, opened 2016-03-26
Add @ operator for clike languages

tomatau:
This operator is used in decorators which are at stage 1 for ECMAScript https://github.com/tc39/ecma262


LeaVerou:
Hi there,
Thanks for this, but I think we should wait a bit. Stage 1 is too early. If you need it for your blog etc, you can add a line to overwrite `Prism.languages.clike.operator`. I will leave this open for when the proposal has advanced a bit. Please ping us then? Thanks!

tomatau:
Already doing the customisation line. But felt this was worth adding just because decorators are already being adopted all over the place. The @ operator has not been in question, only the finer details of the implementation.

LeaVerou:
But it doesn't have any browser support yet, right? Also, if it's JS only, it should be in javascript, not clike.

tomatau:
Both true! But JSX doesn't work in a browser either so I don't think that makes a difference :) And you can make decorators work in a browser if you use babel's browser transform.

Would you rather duplicate the whole operator regex for JavaScript just to add that operator when it goes into the language? I was considering whether to do that in this PR but felt like it wasn't worth duplicating the regex value just to avoid adding support for @ in other c-like languages.

LeaVerou:
Well, there are many other JS-specific new operators we could add, e.g. the spread operator (...) or the exponentiation operator (**)

tomatau:
Happy to move them into this PR, or would you prefer separate requests?

Golmote:
I think we'll wait for, at least, an official specification. I couldn't find them in the [ES2016 spec](http://www.ecma-international.org/ecma-262/7.0/), so I'm closing this PR for now.

PR #922, opened 2016-03-30
Add missing keywords for SAS.

chucknelson:
Two changes:
(1) PROC keyword now also includes its associated name (e.g., PROC TABULATE, PROC SQL).
(2) QUIT now included, as it is used to close/complete various procedures.

### DIFF
--- a/components/prism-sas.js
+++ b/components/prism-sas.js
@@ -23,7 +23,7 @@ Prism.languages.sas = {
 		alias: 'number'
 	},
 	'string': /(["'])(?:\1\1|(?!\1)[\s\S])*\1/,
-	'keyword': /\b(?:data|else|format|if|input|proc|run|then)\b/i,
+	'keyword': /\b(?:data|else|format|if|input|proc\s\S+|quit|run|then)\b/i,
 	// Decimal (1.2e23), hexadecimal (0c1x)
 	'number': /(?:\B-|\b)(?:[\da-f]+x|\d+(?:\.\d+)?(?:e[+-]?\d+)?)/i,
 	'operator': /\*\*?|\|\|?|!!?|¦¦?|<[>=]?|>[<=]?|[-+\/=&]|[~¬^]=?|\b(?:eq|ne|gt|lt|ge|le|in|not)\b/i,
--- a/components/prism-sas.js
+++ b/components/prism-sas.js
@@ -23,7 +23,7 @@ Prism.languages.sas = {
 		alias: 'number'
 	},
 	'string': /(["'])(?:\1\1|(?!\1)[\s\S])*\1/,
-	'keyword': /\b(?:data|else|format|if|input|proc\s\S+|quit|run|then)\b/i,
+	'keyword': /\b(?:data|else|format|if|input|proc\s\w+|quit|run|then)\b/i,
 	// Decimal (1.2e23), hexadecimal (0c1x)
 	'number': /(?:\B-|\b)(?:[\da-f]+x|\d+(?:\.\d+)?(?:e[+-]?\d+)?)/i,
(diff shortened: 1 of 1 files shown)


zeitgeist87:
Hi @chucknelson 

Thanks for your contribution.

First of all I have never used SAS, so please correct me if I am wrong on the language specific details. As I understand it, the `proc` keyword calls some kind of library function. If that is true, wouldn't it be better to highlight the function name as a `function` instead of as a `keyword`:

```js
	'function': {
		pattern: /(\bproc\s+)\w+\b/i,
		lookbehind: true
	},
	'keyword': /\b(?:data|else|format|if|input|proc|quit|run|then)\b/i,
```

chucknelson:
Hi @zeitgeist87,

`PROC` is actually a "special" boundary/step in SAS, similar to the `DATA` step. For more info, [SAS documentation](https://support.sas.com/documentation/cdl/en/lrcon/68089/HTML/default/viewer.htm#n0k19ydenc62jtn14mpv4fhzjmyg.htm) is a good reference for how they refer to some of these concepts (the PROC step in this case).

[Functions are also a concept in SAS](https://support.sas.com/documentation/cdl/en/lefunctionsref/67960/HTML/default/viewer.htm#p1u7ioka3nk1azn1acbxzqz4n3ii.htm), but more of what you expect, passing parameters within parentheses (e.g., `x = SUM(a, b);`).

This was just a little "warm up" contribution, but the eventual goal I want to hit is to get SAS support much more complete in Prism, eventually trying to mimic what users are familiar with in the Base SAS editor when it comes to syntax highlighting, as seen below:

image

Any thoughts on how we can work towards that goal? Thanks!

zeitgeist87:
> PROC is actually a "special" boundary/step in SAS, similar to the DATA step. For more info, SAS documentation is a good reference for how they refer to some of these concepts (the PROC step in this case).

I see. In that case ignore my previous comments.

> This was just a little "warm up" contribution, but the eventual goal I want to hit is to get SAS support much more complete in Prism, eventually trying to mimic what users are familiar with in the Base SAS editor when it comes to syntax highlighting, as seen below:

That would be great. I am happy to help if you have any questions concerning the implementation. Unfortunately I know next to nothing about SAS.

The regex `proc\s\S+` only allows a single space between `proc` and the procedure name. Is this intentional? `\S+` matches anything that is not a space, which is probably too much. For example your regex matches `proc te;st`. I would suggest something like `proc\s+\w+`

chucknelson:
The single space was intentional due to how `PROC` is always followed by a single space and then a single word name, but I definitely agree with your change to a word match versus the way-too-open `\S+` I had in there.

I'm going to push a new commit with the regex change for `PROC`. I think for this small pull request we should keep things in `keywords` for now, and I'll play around in my own fork on how I can get the SAS language implementation more complete in general (which may involve custom token names, not sure yet).

I also have an issue open on how to do language-specific styling for included languages like SAS (#923), in an effort to get to that Base SAS editor styling I mentioned. If you have any input on that I'd really appreciate it! Thanks!

zeitgeist87:
> I also have an issue open on how to do language-specific styling for included languages like SAS (#923), in an effort to get to that Base SAS editor styling I mentioned. If you have any input on that I'd really appreciate it! Thanks!

There is currently no support for themes that are only used for specific languages. To get the colors you want, you have to create your own theme. The language definition can only define the tokens. The themes set the colors.

However the `alias` property is very useful to get the colors you want out of a general purpose theme.

```js
'semantically-correct-token-name': {
    pattern: /something/,
    // use the same color as 'number'
    alias: 'number'
}
```

If your theme looks good with SAS it'll probably look good with other languages too. Then you can submit it to https://github.com/PrismJS/prism-themes

chucknelson:
Great, thanks so much! I'll do what I can to help build out some of the SAS support :)

PR #927, opened 2016-04-04
Override function highlighting.

benembery:
Introduce new highlighting to support methods with generic arguments and attributes that have no arguments 

https://regex101.com/r/iB3xW2/1

### DIFF
--- a/components/prism-csharp.js
+++ b/components/prism-csharp.js
@@ -1,5 +1,5 @@
 Prism.languages.csharp = Prism.languages.extend('clike', {
-	'function': /([a-z0-9_]+(?=(\(|\<))|(?!\[)([a-z0-9_]+)(?=\])|(?!\<)([a-z0-9_]+)(?=\>))/i,
+	'function': /([a-z0-9_]+(?=(\<[a-z0-9_,]+\>)*\()|(^\[[a-z0-9_,\(\)"\/.]+\]))/i,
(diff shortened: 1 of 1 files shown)


zeitgeist87:
Hi @benembery 

Thanks for contributing. The C# functions indeed need to be updated.

Here are some things I've noticed:

* You use negative lookahead `(?!\[)`  and `(?!\<)` but that doesn't seem to work. `Test]` is highlighted as a function the same as `[Test]`. I think you need lookbehind, which is not supported by JavaScript. Prism has a workaround for it though.

* What about the array access operator. That shouldn't be highlighted as a function: `myarray[0] = 7;` You could avoid that by limiting attributes to the start of the line.

* What about the following example. `test` is highlighted as a function: `if (test<3) { /* do something */ }` 

* Generics can have multiple parameters:
 ```csharp
Dictionary<string,string> example = new Dictionary<string,string>();

example.Add("hello","world");

var index = "hello";

Console.Writeline(example[index]);
```
* You could use non-capturing groups in a few places.

* You could also support whitespace between the generics and the function name (optional)

I think you have to match the whole function call including the generics and use the `inside` property to highlight punctuation. But I have not tested that and maybe there is a better way...

benembery:
Yeah, a bit gung-ho of me there, sorry about that. 

I feel bad enough shoe-horning attributes into functions, couple that with the full functionality of generics and it feels like they should all be separate.

https://regex101.com/r/iB3xW2/5

I've stripped the functionality of Generics and just gone for attributes and methods, instead of the look behind I've highlighted the entire wrapped attribute.

zeitgeist87:
Thank you for reporting this bug. I've decided to solve this a little bit differently [[`6f75735`](https://github.com/PrismJS/prism/commit/6f75735)].

Also I think attributes look just fine. There is no need to include them in the functions pattern.

PR #928, opened 2016-04-06
Ignore tests directory in bower.json

flowlo:
Hi!

Your  `tests/` is quite large, and contains rather strange files (having [`!` in the name](https://github.com/PrismJS/prism/tree/gh-pages/tests/languages/css!%2Bcss-extras), for example). When consuming this package (`bower install`and off you go) the tests and associated files are not needed, thus they can be ignored in `bower.json`.

This would save a little bandwidth and also prevent issues with the file names.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -16,6 +16,7 @@
     "**/.*",
     "img",
     "templates",
+    "tests",
     "CNAME",
     "*.html",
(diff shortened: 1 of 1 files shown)


zeitgeist87:
Thanks!

PR #931, opened 2016-04-10
Add Node usage documentation to Basic Usage

clarle:
I added documentation for using Prism in Node to the Basic Usage docs, based on the simple explanation by @geraintluff in #179.

For context, I was trying to build an AMP site that supports syntax highlighting, where no external browser-side JavaScript can be used, and I didn't realize Prism could be used in Node until I started digging through the old issues / PRs.  

I think it's worth having this under Basic Usage, since this might be something that more people might be looking for as time goes on, but let me know if there might be a better place for it too!


LeaVerou:
Awesome, thanks!

PR #938, opened 2016-04-25
add protobuf support

just-boris:
[Protobuf](https://developers.google.com/protocol-buffers/docs/proto) is a binary format for data transfer.

This is an example how [test file](https://github.com/google/protobuf/blob/master/examples/addressbook.proto) will look:
image


zeitgeist87:
Thanks!

zeitgeist87:
I renamed the language to `protobuf`. It would be nice if you could add an examples page `examples/prism-protobuf.html` and/or some test cases. We have a tutorial on how to create test cases: http://prismjs.com/test-suite.html

PR #941, opened 2016-04-28
HTTPS links to twitter.com and w3.org + Google Analytics to HTTPS

coliff:
HTTPS links to twitter.com and w3.org (they are HTTPS anyway so it saves redirect)
Updated Google Analytics URL to be HTTPS (recommended by Google)


zeitgeist87:
Thanks!

Issue #649, opened 2015-07-28
YAML Highlighting Incorrect w/Multiple Blocks


grayghostvisuals:
Currently YAML highlighting trips up when used in multiple blocks like so…

![screenshot 2015-07-28 16 01 57](https://cloud.githubusercontent.com/assets/934322/8942228/47eb1cd4-3542-11e5-9e75-4d57e74e3240.png)

This is correct YAML AFAIK or am I incorrect?

nauzilus:
It looks like the leading `---` is important. It highlights correctly with that included.

I can't see to upload a screen shot, but you can probably test easily enough your self :smile: 

```
---
a:
  version: a
  modifierclass: active
  state: checked
b:
  version: b
  modifierclass: hidden
```

grayghostvisuals:
Thanks! This does fix the issue although using a leading ```---``` is typically for YAML Front Matter and this is a data file not requiring any trailing slashes.

mAAdhaTTah:
Yes and no; from the [spec](http://yaml.org/spec/current.html#document boundary marker/):

> A line beginning with “---” may be used to explicitly denote the beginning of a new YAML document.

It's not required, as you say, but it can be used to indicate the start of the document. Most of the YAML files I've seen do start with `---`, so it does seem to be a convention. Not sure what that mean for Prism as a highlighter, but just thought I'd throw that out there.

grayghostvisuals:
wow! @mAAdhaTTah That is totally enlightening. Thanks for pointing this out. Will follow the spec from now on and continue with this convention you and @nauzilus mention :beers:

PR #943, opened 2016-04-29
fix hilighting of YAML keys on first line of code block

### LINKED ISSUE #942: Yaml ignores first line of code blocks
If I have this:

````
<pre class="language-yaml"><code>somekey: somevalue
anotherkey: anothervalue</code></pre>
````

Then prism will only highlight the second line.  I can do this:

````
<pre class="language-yaml"><code>
somkey: somevalue
anotherkey: anothervalue
</code></pre>
````

...and now prism will correctly highlight all the code, but I have a blank line at the start of the block (which impacts display and layout, because I am trying to top-align the code blocks with documentation in a multi-column layout).

larsks:
The YAML highlighter would fail to highlight a key if it occurred on
the first line of a `<code>` block, as in:

    <pre><code>somekey: somevalue
    anotherkey: anothervalue</pre></code>

This adds "beginning of file" as an acceptable marker for the start of
a key.

Resolves #942 and, it turns out, #649.

### DIFF
--- a/components/prism-yaml.js
+++ b/components/prism-yaml.js
@@ -6,7 +6,7 @@ Prism.languages.yaml = {
 	},
 	'comment': /#.*/,
 	'key': {
-		pattern: /(\s*[:\-,[{\r\n?][ \t]*(![^\s]+)?[ \t]*)[^\r\n{[\]},#]+?(?=\s*:\s)/,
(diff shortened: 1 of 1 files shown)


LeaVerou:
Thanks, merged!

Issue #947, opened 2016-05-04
Plugin to use custom class name


dvkndn:
**Idea:** I want to customize those classes generated by Prism for each token (for example, `var` in JS will be generated to `<span class="token keyword">var</span>`. I want to change the string "token keyword".

**Reason:** Some situation require me to use locally scoped classes, like in a CSS Modules project or one that follow BEM Style.

**Solution:** I wrote a small plugin, based on [Highlight Keyword](http://prismjs.com/plugins/highlight-keywords/), to customize those classes using a style-map object. The object looks like this:

``` javascript
var customStyleMap = {
  tokenType: className
}
```

with `tokenType` is type of token I want to customize (eg: `'keyword'`, `'string'`, `'operator'`...) and `className` is the class string I want to use for that token (eg: `'my-special-class-for-token'` or a hashed one like `'_3ka93h'`

**Example Usage:**

``` javascript
var customStyleMap = {
  keyword: 'special-keyword',
  string: 'special-string'
};
customizePrismClasses(customStyleMap);
```

Or we can use the object returned by CSS Modules:

``` javascript
var customStyleMap = require('my-custom-style-map.css');
customizePrismClasses(customStyleMap);
```

Then, the string `var` will become `<span class="special-keyword">var</span>`.

The source code is very simple, but I think it is useful, especially for project which we want to scope the class name locally, like CSS Modules projects.

Please tell me what do you guys think about this. I'll make a pull request if you think it is useful enough.

LeaVerou:
The functionality is useful. I would suggest you also allow nested objects, in case someone wants to customize these tokens differently per language.
Also, do not use a global function, instead hang it on Prism.plugins.yourpluginname

dvkndn:
Thank you, I'll make the pull request

dvkndn:
@LeaVerou Hi LeaVerou, can you suggest which hooks should I use? I'm currently use 'wrap' but env.language is undefined

```javascript
{
  "type": "punctuation",
  "content": "&lt;",
  "tag": "span",
  "classes": [
    "token",
    "punctuation"
  ],
  "attributes": {},
  "parent": [
    //...
  ]
}
```

LeaVerou:
Damn. No, wrap is the perfect hook for this. @zeitgeist87 do you think it would be easy to add language to env there?

dvkndn:
[...]

PR #948, opened 2016-05-04
Auto provide language if undefined in highlight function

dvkndn:
As discussed in #947, I suggest a solution to provide the `language` parameter if it is undefined (mostly because we call it directly, as listed in API documentation).

This `language` parameter should be provided in order to be passed to `Token.stringify` below and available in hooks.


LeaVerou:
This seems very hacky to me. Also not sure it's needed (see my comment in the other thread).

dvkndn:
Yes I see. Let close it.

PR #949, opened 2016-05-05
Add Ada support

Lucretia:
I've included support and examples for Ada83, Ada95, Ada 2005 and Ada 2012.


zeitgeist87:
Hi @Lucretia 

Every self-respecting syntax highlighter needs support for Ada. Thanks for your contribution!

The only difference between the various Ada versions is the keyword token. Would it be possible to create one language definition  simply called `prism-ada.js` to fit them all? That is the approach we used for other languages with multiple versions.

Lucretia:
I had these as one language file (with each language defined as they are now) originally, but due to the way the examples.html/js worked ending up splitting them up. I would be possible to have one definition with all keywords included for all, I suppose, but this would then mean that the interface keyword would be highlighted as a keyword even when it's not, prior to Ada 2005.

zeitgeist87:
> I had these as one language file (with each language defined as they are now) originally, but due to the way the examples.html/js worked ending up splitting them up.

Every language has to be in a separate file that is correct.

> I would be possible to have one definition with all keywords included for all, I suppose, but this would then mean that the interface keyword would be highlighted as a keyword even when it's not, prior to Ada 2005.

Would that be so bad? The Prism language definitions don't have to be perfect. They should be small and fast. How likely is it, that somebody has some Ada-83 code with a variable named `interface`.

But I can see your point. Let's leave it as separate languages for now. I'll do a proper review this weekend.

Golmote:
I'm also in favor of merging all those variants into one single language (or maybe two if there is a real need). That's what we did for Pascal for example, because having too many languages would be tedious for both the maintainers and the users.

LeaVerou:
Thanks for this PR @Lucretia! 
I'm also in favour of merging them. Prism is not a linter and doesn't need to be precise.
If anything, this will be harder to use, as people will be using language-ada and expecting it to just work.

Golmote:
I merged them manually.

PR #950, opened 2016-05-06
add plugin custom class name

dvkndn:
This plugin allows to replace Prism default class names (like `.comment`, `.string`, `.property`, etc) with your defined ones (like `.editor__comment` (BEM style) or `._7sh3a` (CSS Modules hashed class)).

For more information, please see the plugin page: http://quoinex.github.io/prism/plugins/custom-class-name/

Issue: #947


dvkndn:
@LeaVerou @zeitgeist87 I have some time to write the plugin as discussed previously with you two. Please take a look at it.

I also have one problem here: I create several separate files in my plugins folder to demo how this plugin works, but it seems that make uglify failed to run. Do you have any advice to fix this pull request?

zeitgeist87:
Sorry for the late response. I was quite busy last month. Looking over your source code I had the following idea. Your plugin could also offer the option of a prefix to all classes. Something like a namespace:

```js
Prism.plugins.customClassName.prefix("mynamespace-");
```

That way the user would not have to provide a map for every possible class. You can of course still support the map. What do you think?

dvkndn:
Oh right, I should have that. Actually that is my first solution when trying to isolate Prism's class (before I meet CSS Modules)

zeitgeist87:
Your example JavaScript files like `example-2.js` cause the minifier to crash. You cannot have any extra `*.js` files in that directory.

Also your examples are very small. It is probably best if you include them inline in `index.html`.

plugins/custom-class-name/prism-custom-class-name.js
```diff
@@ -0,0 +1,29 @@
+(function(){
+
+if (
+	typeof self !== 'undefined' && !self.Prism ||
+	typeof global !== 'undefined' && !global.Prism
+) {
```
zeitgeist87: I think you have this wrong. What if both `self` and `global` are undefined? How about this:
```js
(typeof self === 'undefined' || !self.Prism) && (typeof global === 'undefined' || !global.Prism)
```

zeitgeist87:
Your plugin currently only maps a single token type to a single class, which is a bit limiting. Prism supports aliases, so `env.classes` could look like this:

```js
env.classes = ['token', 'equation', 'string', 'important'];
```

Wouldn't it be better to map classes to classes?

```js
env.classes = env.classes.map(function(c){
	return classMap[c] || c;
});
```

LeaVerou:
> Sorry for the late response. I was quite busy last month. Looking over your source code I had the following idea. Your plugin could also offer the option of a prefix to all classes. Something like a namespace:

> ```js
Prism.plugins.customClassName.prefix("mynamespace-");
```

> That way the user would not have to provide a map for every possible class. You can of course still support the map. What do you think?

Accepting a function as a parameter that takes a class and returns the namespaced class would also be good for maximum customization (though I'm a bit worried it might be going towards feature creep :P)

**Comment by LeaVerou** [comment]
Also, customClass would be a better I think, customClassName is a bit too verbose.

**Comment by dvkndn** [comment]
Hi @zeitgeist87 and @LeaVerou , I'm very sorry for the late response. I finally have time to enhance the pull request as your suggestions. Please take a look and tell me anything that I need to fix.

Summary changes:
- Rename `customClassName` to `className` (https://github.com/PrismJS/prism/pull/950#issuecomment-224289704)
- Inline example code (https://github.com/PrismJS/prism/pull/950#issuecomment-223747175)
- Support prefix namespace (https://github.com/PrismJS/prism/pull/950#issuecomment-223519639)
- Map classes to classes (instead of single class) (https://github.com/PrismJS/prism/pull/950#issuecomment-223748063)
- Update docs (docs for prefix feature, add “Motivation”, simplify examples)

**Comment by dvkndn** [comment]
Hi @Golmote , would you mind taking a look at this? It seems @zeitgeist87 and @LeaVerou don't have time for this :(

**Comment by Golmote** [comment]
@dvkndn I only had time to look through the code right now. It looks quite good to me. I'll find a moment this week to actually test your plugin.

I think I spotted some grammar errors in your plugin HTML page. I'm not an native English speaker myself, though, so I'm not really the best candidate to help fixing this.

File: plugins/custom-class/index.html [hunk 2]
```diff
@@ -0,0 +1,133 @@
...
+	<div class="intro" data-src="templates/header-plugins.html" data-type="text/html"></div>
+
+	<h2>Custom Class</h2>
+	<p>This plugin allows you to prefix Prism default classes (<code>.comment</code> will become <code>.namespace--comment</code>) or replace them with your defined ones (like <code>.editor__comment</code> or <code>.comment_7sh3a</code>).</p>  
+</header>
+
+<section>
+	<h1>Motivation</h1>
+
+	<p>Prism default classes are sensible but fixed and too generic. This plugin provide some ways to customize those classes to suite your needs. Example usages:</p>
```
Golmote: Shouldn't it be "suit" instead of "suite"?

plugins/custom-class/index.html
```diff
@@ -0,0 +1,133 @@
...
+
+	<ul>
+		<li>
+			 You want to add namespace for all of them (like <code>.prism--comment</code>) to avoid conflict with your existed classes.
```
Golmote: I should be "existing" instead of "existed".

plugins/custom-class/index.html
```diff
@@ -0,0 +1,133 @@
...
+
+	<h2>2. Replace some Prism classes with your defined ones via an object</h2>
+
+	<pre class="language-js"><code>Prism.plugins.customClass.map({
+	keyword: 'special-keyword',
+	string: 'string_ch29s',
+	comment: 'comment_93jsa'
+})</code></pre>
+
+	<p>Object's keys are the tokens you want to replace (eg: <code>comment</code>), with its value is the classes you want to use (eg: <code>my-comment</code>). Tokens which are not specified will stay the same.</p>
```
Golmote: "with its value is the" should probably be replaced with something like... "with their values being the", I guess...

plugins/custom-class/index.html
```diff
@@ -0,0 +1,133 @@
...
+
+		</li>
+
+		<li>In most cases, using 1 feature is enough. However, it is possible to use both of them together if you want (Result will be like <code>.my-namespace--comment_93jsa</code>).</li>
+
+		</ul>
+
+		<h2>CSS Modules Usage:</h2>
+
+		<p>The initial purpose of this plugin is to use with CSS Modules. It works perfectly with the class map object returned by CSS Modules. For example:</p>
```
Golmote: "is to use" should probably be replaced by "is to be used".

dvkndn:
Thank you @Golmote :D I fixed those grammar errors as you spotted. Sorry for my bad English :(

Golmote:
Ok, it looks good to me.

One last thing... why is components/prism-markup.min.js listed in the changed file? Can you revert the changes on this file?

dvkndn:
Sure I reverted the file. That is an accident, sorry

Golmote:
Great! Merging, then!

jGleitz:
Why was this meged into `gh-pages`? Will it be merged into `master` and published to `npm` any time soon?

Golmote:
Because everything is merged to `gh-pages` first. I'll try to make a release on `master` this month if nobody does it before.

jGleitz:
I see. Thanks!

PR #957, opened 2016-05-17
Use toLowerCase on language names

andeersg:
In Jekyll the language names have to be written with capitalized letter at the beginning (Javascript, CSS, SCSS), this makes Prism not detect them correctly.

This small change runs the `toLowerCase` on the language after extracting it from the class.

I'm not 100% sure I did this correctly, but I pulled, made my change to components folder and ran `gulp`.


Golmote:
I'm not sure I understand what the issue is in the first place...?

andeersg:
If the class names on code-element is:
```
<code class="language-Javascript"
```
prism.js wont actually work on it. The initial check detects it, since it only checks for lang(uage)-*
```
var lang = /\blang(?:uage)?-(\w+)\b/i;
```

But this part of `highlightElement` saves it as "Javascript":
```
language = (parent.className.match(lang) || [,''])[1];
```
while the rest of the code uses "javascript" (with only lower case). This seems to be the same for the rest of the languages as well. By running toLowerCase the only difference is that "language-JavaScript", "language-JAVASCRIPT" or for example "language-javaSCRIPT" would be parsed to "javascript".

It's not a major issue, but as far as I have found out it's not something you can control when using Jekyll with Github Pages.

Golmote:
OK. I'm still not sure I understand why Jekyll would force caps in languages names, but nevermind. Your change looks good to me.

andeersg:
I'm not sure of that either, makes no sense to me. But when I try to use "javascript" it ignores it.

Golmote:
I checked quickly and couldn't find native support for Prism in Jekyll. Are you using a plugin or something?

andeersg:
I’m loading prism on the pages I want it and define my codeblocks in Markdown with three backticks and language name, if that's what you are thinking on.

Plugin support is very limited in Jekyll and specially on Github Pages.

PR #959, opened 2016-05-26
Allow for asynchronous loading of prism.js

rupl:
I found #75 while trying to implement Prism on my personal site and figured the issue could use a PR. It uses the [fix within the comments](https://github.com/PrismJS/prism/issues/75#issuecomment-13744815) of the issue.


LeaVerou:
Oh wow, I thought we fixed that ages ago. Thanks for spotting that!
I would suggest `requestAnimationFrame` instead of `setTimeout`.
Otherwise, LGTM, but I want someone else on the team to weigh in too before merging.

rupl:
No problem, branch is updated with rAF.

zeitgeist87:
I've reproduced the bug and confirmed that the solution works. Thanks!

rupl:
woohoo! 🎉

WickyNilliams:
Thanks for this! My little hacky solution from years ago finally made it in :)

brendanfalkowski:
Great, thanks!

Issue #963, opened 2016-06-01
JavaScript highlighting should be improved


felixsanz:
JS highlighting fails too much, even at simple use cases. Example:

![Examle](http://i.imgur.com/x0vxPHf.png)

Most of it it's wrong. I wish i could provide a patch but i can just report it.

Golmote:
Please provide the entirety of the code you're trying to highlight. It seems to me that the initial highlighting problem occurs before this snippet.

felixsanz:
You're right. That single piece works better as standalone.

```
import cluster from 'cluster'
import os from 'os'
 
const workers = os.cpus().length
 
cluster.setupMaster({ exec: 'app.js' })
 
function log(msg) {
  console.log(`[SERVER] ${new Date().toISOString().replace(/T/, ' ').replace(/\..+/, '')} ${msg}`)
}
 
log(`Master with pid ${process.pid} starting...`)
 
for (let i = 0; i < workers; i++) {
  cluster.fork()
}
 
cluster.on('exit', (worker, code, signal) => {
  log(`worker with pid ${worker.process.pid} died. Restarting...`)
  cluster.fork()
})
 
cluster.on('online', worker => {
  log(`Worker with pid ${worker.process.pid} started`)
})
```

Golmote:
@LeaVerou @zeitgeist87 Does it look safe to you to move up the template string pattern in the language definition of JS so that it's processed before normal strings? IMO it is much more common to encounter quotes inside template strings than backticks inside normal strings. What do you think?

LeaVerou:
Agreed.

zeitgeist87:
If we move the `template-string` before `string`, then one of the tests for `coffeescript` fails, because `coffeescript` uses backticks for inline javascript. However this is easily fixed by adding the following line to the `coffeescript` language:

```js
delete Prism.languages.coffeescript['template-string'];
```

Shall I commit my changes?

LeaVerou:
Sounds good to me!

Golmote:
Yup, agreed!

zeitgeist87:
Should be fixed with [[`1158e46`](https://github.com/PrismJS/prism/commit/1158e46)]

felixsanz:
I was about to test this (even if it's already a closed issue...) to report the status and how that code renders now, but i don't know how to install this with npm since this versions are published into gh-pages branch (uh?), and that is not released into npm, right?

How i do install it then?

zeitgeist87:
[...]

PR #967, opened 2016-06-04
Improvement to greedy-flag

zeitgeist87:
This patch expands the idea of the greedy flag and significantly
improves it, by matching against the whole text instead of just
the next couple of tokens. This does not only improve the
results, but it should also slightly improve performance.


zeitgeist87:
The issue #963 inspired me to look into improving the greedy-flag feature and I am quite excited about the result.

The idea is essentially the same as before, but instead of matching a greedy-pattern against the concatenated string of the next few tokens, it matches against the whole text. With this patch the following code matches correctly:

```js
/ '1' '2' '3' '4' '5' /
"foo `a` `b` `c` `d` bar"
```

Golmote:
Sounds awesome! I'll try to get some time to review this later this week.

LeaVerou:
Thank you @zeitgeist87! 
I'm rally swamped with preparing talks right now, so if @Golmote feels it's ok, feel free to merge!
From a quick look it looks good!

Golmote:
This looks good to me, @zeitgeist87. It seems to works really well, kudos! Merge it whenever you want.

components/prism-core.js
```diff
@@ -280,9 +280,15 @@ var _ = _self.Prism = {
 					lookbehindLength = 0,
 					alias = pattern.alias;
 
+				if (greedy && !pattern.pattern.global) {
+					// Without the global flag, lastIndex won't work
+					pattern.pattern = RegExp(pattern.pattern.source, pattern.pattern.flags + "g");
```
Golmote: @zeitgeist87: How recent is the `flags` property? It [looks like](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/flags) it's not supported very well... I just ran into errors running `npm test` and I think they are due to this.

Maybe we could refactor like so:

```javascript
var flags = pattern.pattern.toString().match(/[imuy]*$/)[0];
pattern.pattern = RegExp(pattern.pattern.source, flags + "g");
```

What do you think?

zeitgeist87: @Golmote Yes you are right it is not supported well at all. Good catch. Your solution is perfect and we should add it.

zeitgeist87: @Golmote Shall I add this fix or do you?

Golmote: @zeitgeist87 I [just did](9893489c04bcea81f3e14d6a00b738a3ce0c4dac), thanks. ;)

LeaVerou: I would rather we added a separate polyfill for flags instead of refactoring our code not to use it. That way, once support is more broad, we can just ditch the polyfill.

PR #976, opened 2016-06-15
Fix line numbers not displaying correctly

anirudhb:
This patch will fix line numbers not displaying correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,12 +19,13 @@
   "readmeFilename": "README.md",
   "devDependencies": {
     "chai": "^2.3.0",
+    "graceful-fs": "^4.1.4",
     "gulp": "^3.8.6",
     "gulp-concat": "^2.3.4",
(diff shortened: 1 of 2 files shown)


zeitgeist87:
It seems to work fine here http://prismjs.com/plugins/line-numbers/. Could you provide an online example that reproduces this bug?

anirudhb:
Download Prism itself and use it on a local page. Then try. The line numbers do not display correctly. The reason they display correctly at http://prismjs.com/plugins/line-numbers/ is because it includes the **correct** CSS file, but the downloaded copy **[JS only]** does not contain the patch I've submitted.

Golmote:
The line-numbers plugin is meant to be used with [its associated CSS file](https://github.com/PrismJS/prism/blob/gh-pages/plugins/line-numbers/prism-line-numbers.css), which already includes this padding.

PR #977, opened 2016-06-15
Fix highlighting of empty comments in PowerShell

rbricheno:
A single # on a line by itself with no other characters following it is a valid comment in powershell.


LeaVerou:
Thanks!

Issue #960, opened 2016-05-28
Dynamic language-definitions load - documentation needed


soar:
[Download page](http://prismjs.com/download.html) offers to download All-In-One JS file. With all available languages (what if I need them all?) this file weights about 250 KB and it will slow page loading and processing.

But [here](http://prismjs.com/plugins/autoloader/) is some information about "Autoloader" plugin, which should take care about language processing. This sounds good, but where is documentation about this feature? Where should I download "core" version of Prism? Where should I download language-definitions? Where should I place them?

LeaVerou:
I think if you're using the autoloader the easiest way is to just clone the repo and take the components folder from there. You can then download Prism core & any plugins from the Download page, without checking any languages (or just check the languages you want to load by default, e.g. if you're using a language a lot, then you probably want to save the extra HTTP request).

@Golmote, perhaps we can add something like the above to the plugin's documentation? Is there an easy way to provide a download of the components folder as a zip without having to manually update it?

Golmote:
Hi! Sorry for the late answer. I updated the plugin documentation to include info on how to get the grammars.

We could also add a "Download all languages grammars" button on the plugin page, that would generate a zip file automatically in JS. I gave it a try and it seems to work pretty well, using [JSZip](https://github.com/Stuk/jszip) and [FileSaver](https://github.com/eligrey/FileSaver.js). I will make a PR so you can take a look.

PR #981, opened 2016-06-18
Download all grammars as a Zip directly from the Autoloader plugin page

Golmote:
See #960 for details.


zeitgeist87:
I've tested it in multiple browsers and it works really great! Amazing what you can do with JavaScript.

I've noticed one minor thing: You added `<script src="components.js"></script>` twice. Other than that it is ready to be merged.

zeitgeist87:
We could also add all the third party libs to `package.json` and copy them over with `gulp`. For example add this to `package.json`:

```json
    "jszip": "^3.0.0",
    "file-saver": "^1.3.2"
```

and this to `gulp`:

```js
gulp.task('vendor', function () {
	return gulp.src([
			'./node_modules/jszip/dist/jszip.min.js',
			'./node_modules/file-saver/FileSaver.min.js'
		])
		.pipe(gulp.dest('vendor/'));
});
```

What do you think?

Golmote:
> I've noticed one minor thing: You added `<script src="components.js"></script>` twice.

Woops! Nice catch.

> We could also add all the third party libs to `package.json` and copy them over with `gulp`.

We probably could. But we did not do it for es6-promises. Should we? I don't see the real benefits here.

zeitgeist87:
> We probably could. But we did not do it for es6-promises. Should we? I don't see the real benefits here.

On the one hand, it would make it easier to keep these libs up to date. On the other hand, every second PR would probably contain changes to these files, because people don't know how to use git 😄.

You're right it's probably not a good idea.

Golmote:
Anyway, we'll be able to change this anytime in the future if we feel the need ;). I'm merging.

Issue #987, opened 2016-06-30
requestAnimationFrame


isglass:
Hi,

Some of our visitors seem to use browsers lacking support for requestAnimationFrame. We highlight on demand so it is not needed for them. Propose the following fix:

`requestAnimationFrame(_.highlightAll, 0);`

Changed to:

`if (window.requestAnimationFrame) window.requestAnimationFrame(_.highlightAll, 0);`

Regards,
Johan

Golmote:
I'd rather fallback on setTimeout since we're supposed to support IE9.

zeitgeist87:
Do we really have to support IE9? We could also simply add a polyfill for `requestAnimationFrame` to the `ie8` plugin, or create an `ie9` plugin with the polyfill...

PR #988, opened 2016-06-30
Add setTimeout fallback for requestAnimationFrame

brentertz:
Address issue #987 - Add `setTimeout` fallback for `requestAnimationFrame`

### DIFF
--- a/components/prism-core.js
+++ b/components/prism-core.js
@@ -479,7 +479,11 @@ if (script) {
 
 	if (document.addEventListener && !script.hasAttribute('data-manual')) {
 		if(document.readyState !== "loading") {
-			requestAnimationFrame(_.highlightAll, 0);
+			if (window.requestAnimationFrame) {
+				window.requestAnimationFrame(_.highlightAll);
+			} else {
+				window.setTimeout(_.highlightAll, 16);
+			}
 		}
 		else {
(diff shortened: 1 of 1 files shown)


valtlait:
`requestAnimationFrame` doesn’t have the second attribute. Can you remove it, please?

brentertz:
Yep, removed.  Not sure why that was in there in the first place or why I didn't notice it.

Golmote:
Indeed. It was there because it originally was a setTimeout and the person who changed it forgot to remove it. 
It went unnoticed because that change was part of an unrelated PR...

Golmote:
Why 16, by the way?

brentertz:
60 frames per second (fps) is the target for smooth animations

1s or 1000ms / 60fps = ~16ms

There are other ways to handle this, mind you.  This is the simplest implementation.  You however may be interested in calculating the value more precisely.  https://gist.github.com/paulirish/1579671

zeitgeist87:
You could also do it with a one-liner:

```js
(requestAnimationFrame || setTimeout)(_.highlightAll, 0);
```

But it would be a bit of a hack and your solution is probably better.

Golmote:
Just noticed you created the PR against the master branch. We currently work on gh-pages. The master is only updated on new release.
I merged this PR manually.

PR #990, opened 2016-06-30
Optimize Prism logo

valtlait:
Optimizations to Prism SVG logo:
– remove XML prolog
– remove `version` attribute (doesn’t have any effect)
– remove unused `xlink` namespace
– remove `viewBox` attribute (doesn’t have any effect in this context)
– remove `px` units (plain numbers are pixels)
– optimize path data


Golmote:
Looks good to me ;)

LeaVerou:
Please add viewBox back. Every SVG should have a viewBox. @svgeesus can explain why.

valtlai:
@LeaVerou, see #1002.

valtlai:
> Every SVG should have a viewBox. @svgeesus can explain why.

@svgeesus, I’m very interested to know why.

svgeesus:
The width and height attributes make an SVG image non-scalable. Unless it is desired to always force a specific size, they should not be present.
A viewbox specified the range of world coordinates used. In this case, the range started at 0,0 (which is the default if viewBox is not specified) and extends to the same numbers as the width and height attributes (whichis why removing the viewBox was successful in this instance). However, it still forces a fixed size rather than being, well, *scalable*.
Using a viewBox (and no width or height) allows the logo to be used at different sizes. See https://github.com/PrismJS/prism/pull/1005

Also see
https://www.w3.org/TR/SVG2/coords.html#ViewBoxAttribute
for the more complex cases, in which width/height and viewBox are not providing essentially the same info.

PR #993, opened 2016-07-03
Add an API for plugin settings to Prism core

zeitgeist87:
This patch adds a new utility function called getSettings() to
Prism. It uses Object.defineProperty() to create a special
object. Every property of this object has a getter function,
which will search the DOM for the requested setting on demand.
Only the first access triggers a search of the DOM. Any
subsequent access is served from a cached value.

Usage is very simple:

``` js
var settings = Prism.util.getSettings(env.element, {
    'remove-trailing': { type: 'bool', 'default': true },
    'break-lines': { type: 'int', 'default': 80 },
});

console.log(settings['break-lines'];
```


zeitgeist87:
A brand new approach for Prism to parse settings out of the DOM tree. It is a nice alternative to the ParseSettings plugin. The PR also demonstrates the use on the `normalize-whitespace` plugin.

Golmote:
Can settings still be set in a programmatic way, rather than using attributes?

zeitgeist87:
> Can settings still be set in a programmatic way, rather than using attributes?

No, I removed that. I didn't like the interface. I probably should put it back in for backwards compatibility...

Golmote:
We do need a way to set defaults / global options. This PR is great for per-block options, but harder to use for global options. I think we need both mechanisms.
And yes, we might need to question about BC too...

LeaVerou:
Agreed with Golmote, we need both Imperative and declarative ways to set settings.

LeaVerou:
Things like `data-foo="true"` are a bit of an abomination. They are against the design of HTML in which boolean properties are true by presence and their value has nothing to do with it. Also, boolean settings should/could be set by a class (`foo` or `no-foo`, which is consistent with current practice).

components/prism-core.js
```diff
@@ -41,6 +41,46 @@ var _ = _self.Prism = {
...
+					    el = element;
+					while (el && el.getAttribute) {
+						var match = pat && pat.exec(el.className);
+						if (match)
+							return value = !match[1];
+
+						var attr = el.getAttribute('data-' + name);
+						if (attr !== null) {
+							switch(spec.type) {
+							case 'bool':
```
LeaVerou: `bool` is not a thing in JS. `boolean` please.

LeaVerou:
In general, keep in mind that the priority of constituencies is us < plugin authors < Prism users < their end users. Requiring people to use attributes for everything (no classes, no JS API) is against this: It makes things harder for Prism users, for the benefit of plugin authors (and us?).

I would suggest boolean settings be classes by default, with a switch to make them attributes.

zeitgeist87:
> Edit: Ah, turns out this already supports class names! Nice! I'll leave the above because it's useful as a general principle.

It also already supports properties that are true by presence:

```html
<div data-true-by-presence></div>
```

mAAdhaTTah:
This would also be a pretty good solution for #1438. Two birds w/ one stone.

components/prism-core.js
LeaVerou: Yes

LeaVerou:
I see that one of the main concerns was that there was no JS API to set these settings. Has this been resolved?
Also, what about inheritance? It looks like this supports inheritance by manually traversing the DOM. Why not construct an appropriate selector and just use `element.closest(selector)`?

Golmote:
@LeaVerou Beware of the browser compatibility for [`closest()`](https://caniuse.com/#feat=element-closest).

LeaVerou:
@Golmote Browser compatibility is > 88% of users, which seems pretty good to me. Anyone wanting IE11 support can use polyfill.io... What do you think?

Golmote:
@LeaVerou As long as we document our compatibility limitations (and solutions) explicitely, like maybe on the website *and* in the README, I'm fine with the idea.

mAAdhaTTah:
I'm going to close this, as I don't see a pressing need for a settings API at this time, but happy to revisit this if we decide we need it.

Issue #485, opened 2015-01-21
URL highlighting


LeaVerou:
This issue is more of a note to self, as I'd love to write this myself when I get some time.

The lang definition would go through all existing languages and replace url tokens with itself as `inside`. Then it would highlight data URIs (e.g. SVG in data URIs, like [this](https://github.com/LeaVerou/prism/issues/457#issuecomment-70612483)), it could highlight hashes and query params and many other things.

Thoughts? Do you guys think it would be useful?

Golmote:
This may not be easy, since data URIs can contain pretty much anything, right?

LeaVerou:
Of course it wouldn’t be in Core, I was planning to have it as a language definition.

Re:data URIs, I meant the urlencoded ones (which these days can have no urlencoded characters at all, as few characters really need escaping in them)

Golmote:
If it is as a language definition, then could it be used in combination with another language definition?
In the example you gave, with the data URI containing SVG inside CSS, how will it be possible to highlight both the CSS and the content of the data URI (as markup) if it's implemented as a language definition?

What I meant about the contents of the data URI is that it may need to be highlighted as markup (like in you example, again) or as pretty much anything else depending on the mime type used.

PR #996, opened 2016-07-10
Data-URI Highlight plugin

Golmote:
This PR is a proposal for #485. It is inspired by the Autolinker plugin.

f851ed7b73

A small change was made to the Autolinker plugin so that both are compatible.


zeitgeist87:
Very nice plugin! It works very well.

Maybe we should remove `'comment'` from the `candidates` array. It is very unlikely that a data-uri is inside a comment and even if there is one it probably shouldn't be highlighted, because it looks a bit weird:

comment

What do you think?

Golmote:
Haha indeed this is ridiculous! ^^ This is fixed, thanks!

PR #998, opened 2016-07-12
Don't restore env.code from backupCode in Handlebars, PHP and Smarty …

### LINKED ISSUE #935: Remove initial line feed plugin + php component = bad line numbering
The "remove initial line feed" plugin gets triggered on "before-highlight" and changes env.code (removes the redundant initial line feed).

The php component has a hook on "before-insert", which restores env.code to env.backupCode.

The line numbering plugin gets triggered on "complete" and uses env.code to figure out how many lines there are so it counts an extra line.

Possible fix (works fine for me): env.backupCode = env.code; in the Remove Initial Line Feed plugin.

### LINKED ISSUE #997: PHP + Line Numbering results in extra line
Hey,

using the line numbering-plugin and PHP I get an extra line at the end of the block that gets numbered. I am not entirely sure how to show that off in a better way, but here's a screenshot: http://imgur.com/e846OPW

I tried both javascript and css, both of these languages show the correct number of lines.

Trying to look for a solution in some (updated) files here or other issues I only found unrelated and resolved stuff.

Hope you can help me out,
Flynx

Golmote:
…components.

The PHP component used to store `env.code` in a `backupCode` property in the `before-highlight` hook and restore it in the `before-insert` hook. Both the Handlebars component and the Smarty component copied this behaviour.

I don't see any reason to do this currently, though. It seems it was made to work properly with the Line numbers plugin but it actually breaks it when used in conjunction with the Remove initial line feed plugin or the Normalize whitespace plugin...

The Line numbers plugins runs on the `complete` hook, which gets executed after the `after-highlight` hook, so the code used to compute the line numbers is the correct one.

This PR should fix #997 and probably fix #935.


Flynxify:
Can confirm that this does fix #997.

PR #1005, opened 2016-07-24
remove width and height, so logo becomes scalable. 

svgeesus:
Aspect ratio and starting coordinates already established by viewBox.


LeaVerou:
Do we set the size with CSS? If not, we should before merging this.

svgeesus:
Yes, you should

Golmote:
I merged this manually. Thanks for contributing.

PR #1011, opened 2016-08-11
Create show-filename plugin. And fix line-numbers and unescaped-markup plugins

dfitiskin:
Show a name of file from data-src or data-filename attributes
Store data attributes in unescaped markup plugin. For example data-start for line-numbers plugin.
Fix removing line-numbers class form code. It removes space between other classes in code.


LeaVerou:
First off, thanks for contributing!

This has come up before as a request, basically people need the ability to add custom labels to their code snippets beyond simple language. I suggest we modify show-language to be sufficiently broad and customizable to include these use cases instead of having a dozen plugins that do the same thing with minor differences.

PR #1012, opened 2016-08-13
Plugins fixes (unescaped-markup, line-numbers)

dfitiskin:
Store data attributes in unescaped-markup plugin. For example data-start for line-numbers plugin.
Fix removing line-numbers class form code. It removed space between other classes in code.


Golmote:
Sorry for the long time waiting. Thank you for contributing!

Note that I did not include the changes in the `index.html` file, because I consider your change is not a feature that needs to be shown, it's just the way it should have worked all along ;)

PR #1014, opened 2016-08-19
Add support for the Jolie language

thesave:


components/prism-jolie.js
```diff
@@ -0,0 +1,16 @@
+Prism.languages.jolie = Prism.languages.extend('java', {
+	'keyword': /(<-|=>)|\b(is_defined|undef|include|main|outputPort|inputPort|Location|Protocol|RequestResponse|throw|OneWay|interface|any|long|type|void|sequential|raw|scope|forward|install|execution|single|concurrent|Interfaces|cset|csets|double|global|linkIn|linkOut|string|bool|int|synchronized|courier|extender|throws|this|new|Interfaces|nullprocess|Redirects|embedded|extender|Aggregates|spawn|constants|with|foreach|instanceof|<-|over|define)\b/g,
```
Golmote: I think the arrow `<-` near the end of the regexp was not intended to be there anymore.

components/prism-jolie.js
```diff
@@ -0,0 +1,16 @@
...
+	'builtin': /\b(string|int|long|Byte|bool|double|float|char|any)\b/g,
+	'number': /\b0x[\da-f]*\.?[\da-f\-]+\b|\b\d*\.?\d+[e]?[\d]*[dfl]?\b/gi,
+	'operator': /[-+]{1,2}|!|<=?|>=?|={1,3}|&{1,2}|\|\||\*|\//g,
```
Golmote: Are `-+` and `+-` valid operators in Jolie? If not, you should write something like `--?|\+\+?` instead of `[-+]{1,2}`.

Also, you could make the regexp more clear by combining all single symbols into a character class : ` [!*\/]`

components/prism-jolie.js
```diff
@@ -0,0 +1,16 @@
+Prism.languages.jolie = Prism.languages.extend('java', {
+	'keyword': /(<-|=>)|\b(is_defined|undef|include|main|outputPort|inputPort|Location|Protocol|RequestResponse|throw|OneWay|interface|any|long|type|void|sequential|raw|scope|forward|install|execution|single|concurrent|Interfaces|cset|csets|double|global|linkIn|linkOut|string|bool|int|synchronized|courier|extender|throws|this|new|Interfaces|nullprocess|Redirects|embedded|extender|Aggregates|spawn|constants|with|foreach|instanceof|<-|over|define)\b/g,
+	'builtin': /\b(string|int|long|Byte|bool|double|float|char|any)\b/g,
+	'number': /\b0x[\da-f]*\.?[\da-f\-]+\b|\b\d*\.?\d+[e]?[\d]*[dfl]?\b/gi,
```
Golmote: There is no need for the square brackets around `[e]` or `[\d]`. This would be fine: `e?\d*`

components/prism-jolie.js
```diff
@@ -0,0 +1,16 @@
...
+	'operator': /[-+]{1,2}|!|<=?|>=?|={1,3}|&{1,2}|\|\||\*|\//g,
+	'punctuation': /[{}[\]()\.:]/g,
+	'operation': /[a-z][A-Za-z0-9_]+(?=\@)/gm,
+	'service': {
+		pattern: /((?:(?:\@\s*)))[A-Z][A-Za-z0-9_]+/ig,
+		lookbehind: true,
+	},
+	'symbol': /\||;|\@/g,
```
Golmote: Here, a character class would make the regexp more clear, imo: `[|;@]`

Golmote:
Thanks for contributing. Please take a look at the few comments I made on the code.

Also, it would be truly awesome if you could:

- add some code examples, like the [existing ones here](https://github.com/PrismJS/prism/tree/gh-pages/examples)
- add some tests, like the [existing ones here](https://github.com/PrismJS/prism/tree/gh-pages/tests/languages) (more info regarding tests on [the Test Suite page](http://prismjs.com/test-suite.html))
- run `gulp` on the command line to generate the minified version of your component.

components/prism-jolie.js
```diff
@@ -0,0 +1,16 @@
+Prism.languages.jolie = Prism.languages.extend('java', {
+	'keyword': /(<-|=>)|\b(is_defined|undef|include|main|outputPort|inputPort|Location|Protocol|RequestResponse|throw|OneWay|interface|any|long|type|void|sequential|raw|scope|forward|install|execution|single|concurrent|Interfaces|cset|csets|double|global|linkIn|linkOut|string|bool|int|synchronized|courier|extender|throws|this|new|Interfaces|nullprocess|Redirects|embedded|extender|Aggregates|spawn|constants|with|foreach|instanceof|<-|over|define)\b/g,
+	'builtin': /\b(string|int|long|Byte|bool|double|float|char|any)\b/g,
+	'number': /\b0x[\da-f]*\.?[\da-f\-]+\b|\b\d*\.?\d+[e]?[\d]*[dfl]?\b/gi,
+	'operator': /[-+]{1,2}|!|<=?|>=?|={1,3}|&{1,2}|\|\||\*|\//g,
+	'punctuation': /[{}[\]()\.:]/g,
```
Golmote: The dot `.` does not need escaping inside a character class.

components/prism-jolie.js
```diff
@@ -0,0 +1,16 @@
...
+	'operation': /[a-z][A-Za-z0-9_]+(?=\@)/gm,
```
Golmote: The `@` character does not need escaping in regexps in JS. (This comment applies to the regexp for `'services'`, below, too)

Also, why the `m` flag?

components/prism-jolie.js
```diff
@@ -0,0 +1,16 @@
...
+	'service': {
+		pattern: /((?:(?:\@\s*)))[A-Z][A-Za-z0-9_]+/ig,
+		lookbehind: true,
+	},
+	'symbol': /\||;|\@/g,
+	'string': /(""")[\W\w]*?\1|("|\/)[\W\w]*?\2|('.')/g
```
Golmote: Finally, could you add the non-capturing part `?:` inside all parenthesized group that do not need capture? Like the last one in this regexp, and also in the words list in `'keyword'` and `'builtin'`, at the top. ;)

thesave:
Hi, thanks for the very useful comments :)
I took the time to extend the support for the Jolie language (also following you advises).
As suggested, I added the examples, tests, and built with `gulp`.

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
+Prism.languages.jolie = Prism.languages.extend('clike', {
+	'keyword': /\b(?:include|define|is_defined|undef|main|init|outputPort|inputPort|Location|Protocol|Interfaces|RequestResponse|OneWay|type|interface|extender|throws|cset|csets|forward|Aggregates|Redirects|embedded|courier|extender|execution|sequential|concurrent|single|scope|install|throw|comp|cH|default|global|linkIn|linkOut|synchronized|this|new|for|if|else|while|in|Jolie|Java|Javascript|nullProcess|spawn|constants|with|provide|until|exit|foreach|instanceof|over|service)\b/g,
+	'builtin': /\b(?:undefined|string|int|void|long|Byte|bool|double|float|char|any)\b/,
+	'number': /\b(?:\b\d*\.?\d*(e[+-]?\d*)?[l]?)\b/i,
```
Golmote: There is not need for the square brackets around `[l]`. Also, the second `\b` seems redundant to me. Finally, it looks like your regexp may match nothing at all, since everything is optional. Please replace one of the two first `\d*` occurrences with `\d+`. I also suspect this part `e[+-]?\d*` should be `e[+-]?\d+` since numbers after the exponent symbol are usually not optional.

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+	'operator': /->|<<|=<|=>|[!+->=]?=|<|>|!|&&|\|\||\?|\*|\/|%|--?|\+\+?|\^/g,
```
Golmote: Couldn't you combine all single operators into one character class `[<>!?*\/%^]`?

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+	'symbol': /[|;@?:]/,
```
Golmote: `?` is already listed in the `operator` regexp, so I guess it will never match as a `symbol`. It should be removed, then.

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+	'punctuation': /[{}[\]().]/,
+	'string': /(""")[\W\w]*?\1|("|\/)[\W\w]*?\2|('.')/g
```
Golmote: The last pair of parentheses looks unneeded to me. There is nothing to group or capture here ;)

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+});
+
+delete Prism.languages.jolie['class-name'];
+delete Prism.languages.jolie['function'];
+
+Prism.languages.insertBefore( 'jolie', 'keyword', {
+	'function':
+	{
+		pattern: /(?:(?:\b(outputPort|@|inputPort|in|service|courier)\b))[A-Za-z0-9_]+/g,
```
Golmote: There are also unneeded parentheses here... I think only the one pair surrounding the keywords is needed. And it should be non-capturing ;)

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+		lookbehind: true
+	},
+	'aggregates': {
+		pattern: /(?:(?:(\bAggregates\b[\n\r\s]*:[\n\r\s]*)))([A-Za-z0-9_]+[\n\r\s]*(with[\n\r\s]*[A-Za-z0-9_]+)?,[\n\r\s]*)*[A-Za-z0-9_]+([\n\r\s]*with[\n\r\s]*[A-Za-z0-9_]+)?/,
```
Golmote: Here, the first 3 pairs of parentheses look unneeded to me.
The `\b` after `Aggregates` can be removed as well.
`\r` and `\n` are already part of the `\s` class, so all instances of `[\n\r\s]*` can be replaced with `\s*`.
The occurrences of `[A-Za-z0-9_]+` should probably be replaced with `\w+` to improve readibility.
Finally, all remaining parenthesized groups should be non-capturing by adding `?:`

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+		lookbehind: true,
+		inside: {
+			'withExtension': {
+				pattern: /\bwith\b[\n\r\s]+[A-Za-z0-9_]+/,
```
Golmote: In your first regexp, the blank character after `with` was optional, here it's mandatory. Is it intended?
If it is indeed mandatory, the `\b` is not needed.

Please take my previous comments into account in this regexp, regarding `\s` and `\w+`.

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+				inside: {
+					'keyword' : /\bwith\b/
+				}
+			},
+			'function': {
+				pattern: /[A-Za-z0-9_]+/
```
Golmote: Same comment regarding `\w+`

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+			}
+		}
+	},
+	'redirects': {
+		pattern: /(?:(?:(\bRedirects\b[\n\r\s]*:[\n\r\s]*)))([A-Za-z0-9_]+[\n\r\s]*=>[\n\r\s]*[A-Za-z0-9_]+[\n\r\s]*,[\n\r\s]*)*[A-Za-z0-9_]+[\n\r\s]*=>[\n\r\s]*[A-Za-z0-9_]+/,
```
Golmote: All comments about `aggregates` also apply to this one ;)

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+		lookbehind: true,
+		inside: {
+			'function': {
+				pattern: /[A-Za-z0-9_]+/g
```
Golmote: `\w+` here

Golmote:
I reviewed your additions and made a bunch of comments about them. Please take a look when you can.

Those additions would also need test files ;)

components/prism-jolie.js
Golmote: `[A-Za-z0-9_]+` can be replaced by `\w+` ;)

thesave:
Hi, thank you very much for your review, I really appreciate it :)
I tried to follow all your suggestions and added tests for the new features.

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
+Prism.languages.jolie = Prism.languages.extend('clike', {
+	'keyword': /\b(?:include|define|is_defined|undef|main|init|outputPort|inputPort|Location|Protocol|Interfaces|RequestResponse|OneWay|type|interface|extender|throws|cset|csets|forward|Aggregates|Redirects|embedded|courier|extender|execution|sequential|concurrent|single|scope|install|throw|comp|cH|default|global|linkIn|linkOut|synchronized|this|new|for|if|else|while|in|Jolie|Java|Javascript|nullProcess|spawn|constants|with|provide|until|exit|foreach|instanceof|over|service)\b/g,
+	'builtin': /\b(?:undefined|string|int|void|long|Byte|bool|double|float|char|any)\b/,
+	'number': /\b(?:\d*\.?\d+(e[+-]?\d+)?l?)\b/i,
```
Golmote: The outer parentheses are not needed here. And the inner ones should be made non-capturing.

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+	'operator': /->|<<|=<|=>|[!+->=]?=|[:<>!?*\/%^]|&&|\|\||--?|\+\+?/g,
```
Golmote: I can't see any support for `<=` here. Supposing it exists, it probably should be added to `[!+->=]?=`.
On the opposite, I couldn't find any reference to the `=<` operator from a quick look at the docs. Please confirm this isn't a typo.
I think Jolie also supports `*=` and `%=` (but not `/=`... ?). Could you check?

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+	'symbol': /[|;@]/,
+	'punctuation': /[{}[\].]/,
+	'string': /(""")[\W\w]*?\1|("|\/)[\W\w]*?\2|('.')/g
```
Golmote: I still don't see any use for the last pair of parentheses. They should be removed.

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+});
+
+delete Prism.languages.jolie['class-name'];
+delete Prism.languages.jolie['function'];
+
+Prism.languages.insertBefore( 'jolie', 'keyword', {
+	'function':
+	{
+		pattern: /(?:(\b(outputPort|inputPort|in|service|courier)\b|@)\s*)\w+/,
```
Golmote: Here, your parentheses still have problems. I think what we expect is this:

`/((?:\b(?:outputPort|inputPort|in|service|courier)\b|@)\s*)\w+/`

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+		lookbehind: true
+	},
+	'aggregates': {
+		pattern: /(\bAggregates\s*:\s*)(?:\w+(\s+with\s+\w+)?\s*,\s*)*\w+(?:\s+with\s+\w+)?/,
```
Golmote: There is still a group here `(\s+with\s+\w+)?` that should be made non-capturing.

components/prism-jolie.js
```diff
@@ -0,0 +1,47 @@
...
+					'keyword' : /\bwith\b/
+				}
+			},
+			'function': {
+				pattern: /\w+/
+			}
+		}
+	},
+	'redirects': {
+		pattern: /(Redirects\s*:\s*)(?:\w+\s*=>\s*\w+\s*,\s*)*(?:\w+\s*=>\s*\w+)/,
```
Golmote: I notice you've removed the initial `\b`? Is it intended? This one actually made sense to me.

Golmote:
Great! Almost there!

thesave:
Did the required fixes.
You are right about `<=` and `=<`, I added also the support for `*=` (whilst `%=` is not supported)
I also moved `=>` from operators to symbols.

Golmote:
The code looks quite good to me now.

I pulled your branch to test it a bit, and I noticed your examples for strings and numbers aren't exactly working well, though. Most of those cases don't appear to be handled by the current regexps. Either we need to handle them or they should not appear in the examples. Can you take a look?

prism-jolie-broken-examples

Also, in the full example, I'm curious about the color of the parentheses and commas. They probably should go inside the `punctuation` rule, shouldn't they?

thesave:
Sorry for the late reply. Work swamped me :)
I fixed the example file and the errors on strings and punctuations.

Golmote:
Sorry for the late answer too. It's merged now.

thesave:
Thank you very much for your time :)

PR #1019, opened 2016-08-30
Default margin

SandroMiguel:
If the sample code is directly within the `<body>` tag it will have a margin of 1em.

### DIFF
--- a/themes/prism.css
+++ b/themes/prism.css
@@ -4,6 +4,10 @@
  * @author Lea Verou
  */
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
Hi there Sandro, thank you for contributing to Prism!
However, I think this is a stylistic decision that should be left up to the authors, not something we should bake in to the default stylesheet. Therefore, I will close this PR. I hope this does not put off you off from future contributions!

PR #1035, opened 2016-10-15
Add yarn.lock

idleberg:
Since many people are like going to use `yarn` over `npm`, it's probably a good idea to add this


LeaVerou:
Thx!

PR #1042, opened 2016-10-27
Update TypeScript language component.

sebald:
Hey there,

super small fix, but was bugging me a lot! :)

This Patch
- adds missing `from` keyword to Typescript language declaration
- sets `ts` as alias.


LeaVerou:
Thank you!

PR #1047, opened 2016-10-31
Remove important token in ini definition

### LINKED ISSUE #1044: language-ini uses "important" class
For INI sections such as `[foo]`, the `language-ini` plugin adds things like:

``` html
<span class="token important">[foo]</span>
```

The problem is that I'm already using a stylesheet that defines something for the `important` CSS class, so my INI file listings now have icons before the sections such as `[foo]`.

I could fiddle around with my CSS stylesheets to work around this issue, but I'm curious --- why was the CSS class "important" used? How is "important" semantically related to an INI section? Wouldn't the CSS class "ini-section" have made more sense (and have less chance of clashing with user stylesheet classes)?

I'm asking more than being critical, to see if there is a good reason. Is this something we could improve?

This is one of the last things on my list of things today, so I apologize for any errors or oversights from my sleepy mind.

aviaryan:
Fixes #1044

### DIFF
--- a/components/prism-ini.js
+++ b/components/prism-ini.js
@@ -1,6 +1,6 @@
 Prism.languages.ini= {
 	'comment': /^[ \t]*;.*$/m,
-	'important': /\[.*?\]/,
+	'selector': /^\s*\[.*?\]/m,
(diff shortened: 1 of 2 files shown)


LeaVerou:
Thanks!

PR #1051, opened 2016-11-09
Reduce risk of XSS

Rob--W:
Prism and its plugins sometimes mixes potentially unsafe input with HTML. In the worst case, this may lead to a XSS vulnerabilities, especially when prism is used to highlight external code.

See the individual commits for the fixes. If you decide to merge them all, don't squash the commits as doing so may obscure the relation between the commit message and the fix.


Rob--W:
(rebased because of conflicts in toolbar-plugin.min.js)

@Golmote Can you take a look at this PR and merge it? I [tweeted](https://twitter.com/RobSinje/status/796419024406151168) @LeaVerou, but she probably overlooked it.

PR #1064, opened 2016-11-26
Update typescript keywords

andrewiggins:
Update the list of TypeScript keywords with the list of JavaScript keywords and new words added to the [TypeScript spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#221-reserved-words). Also added a link to the spec in the source.


LeaVerou:
Thanks!

PR #1072, opened 2016-12-12
Add left shoe underbar and right shoe underbar

jayfoad:
They're used by NARS2000, and left shoe underbar will be used by Dyalog APL.


Golmote:
Thanks!

PR #1079, opened 2017-01-10
Add `es6` alias for `javascript` language.

spen:
I'm currently working on a PR @ https://github.com/Automattic/wp-calypso/pull/10359 which implements Prism for the DevDocs of the [Calypso](https://github.com/Automattic/wp-calypso) project.

One issue that I've noticed is that ` ```es6` is not processed by Prism.
I may be wrong here as I am fairly new to Prism but since `es6` support is already quite good in `prism-javascript.js` it seems to make sense to just add an alias.

Alternatively we could replace all instances of ` ```es6` with ` ```js` but I feel that this may cause some issues with syntax highlighting on GitHub.com.
Another alternative might be to set the alias ourselves, but I feel other projects might benefit from having the alias set up upstream.

TIA :)

### DIFF
--- a/components/prism-javascript.js
+++ b/components/prism-javascript.js
@@ -46,3 +46,4 @@ if (Prism.languages.markup) {
 }
 
 Prism.languages.js = Prism.languages.javascript;
+Prism.languages.es6 = Prism.languages.javascript;


LeaVerou:
Thanks for contributing!

ES6 is still JS, so not sure I understand the need to distinguish. That said, I'm not opposed to this PR if the others think it's ok. 

@spen  Do you need this to be in the core or would a custom alias work for you guys?

spen:
Hi @LeaVerou - thanks for taking a look at this!

Although ES6 is JS, I was seeing an error / no processing on code blocks tagged as`es6` like so:
`  ```es6  someCode... ``` ` when using `Prism` with [`marked`](https://github.com/chjj/marked) because `es6` was not being found as a language in `Prism.languages`.

Here's the code (annotated) from the PR referenced above :

```es6
marked.setOptions( {
	// 'code' may be 'js', 'javascript' or 'es6' amongst others.
	highlight: function( code, language ) {
		// making sure that there is a language, otherwise Prism.highlight(code, undefined) errors.
		const syntax = Prism.languages[ language ];
		return syntax ? Prism.highlight( code, syntax ) : code; // skip highlighting if lang is not found.
	}
} );
```

Calypso has a lot (hundreds?) of readmes, many of which have `es6` tagged code-blocks.
We could look to convert all existing `es6` tags with `js` tags. 
That would be reasonable as I think that GitHub now renders the same highlighting for both (I feel like previously that was not the case) but I worry that `es6` tagged blocks may find their way in to the codebase over time.

By the way, the solution I've used for the related PR is to manually alias the language like so:
```js
Prism.languages.es6 = Prism.languages.js
```
That's fine for the needs of the PR I referenced but I feel others may see the same issue though it really might not be a very common issue :)

Just to summarise: I'm aiming for as much consistency between Prism's highlighting and GitHub's

zeitgeist87:
I think this PR is ok. Another alias can't hurt. 👍

LeaVerou:
> By the way, the solution I've used for the related PR is to manually alias the language like so

I think this solution is fine, if it works for you. This is a perfectly good use of Prism's API, and it was designed to make aliases easy to define on purpose. If many people need the es6 alias we can always revisit this, but I'd be opposed with adding it just for this.

spen:
No Worries @LeaVerou, thanks for taking a look! - I didn't actually realise that it was so simple until after opening this PR :)

PR #1085, opened 2017-01-18
Django/Jinja2 template language definition

romanvm:
**Jinja2** is a popular templating engine for Python web-frameworks. **Django** is a popular Python web-framework with its own templating engine. They have similar syntax so I created one definition for both.


Golmote:
Hi! Thanks for contributing!

PR #1087, opened 2017-01-24
Initial implementation of manual highlighting

### LINKED ISSUE #765: Provide a global way to disable auto highlighting
Similar to #366, this is very cumbersome when working with something like a CMS that aggregates is assets and doesn't allow extra attributes (like `data-manual`) to be set on the script tags.

The only recourse I had was to manually "hack" the HTML template:
https://github.com/unicorn-fail/drupal-bootstrap.org/blob/831f047cfb7125778364cbccd2db34256b643222/docroot/sites/all/themes/bootstrap_subtheme/templates/system/html.tpl.php#L77

I'm ok with this solution, temporarily, but it doesn't bode well in the long term for performance reasons. Adding the file in this fashion prevents the file from becoming aggregated and causes an extra HTTP request.

I really don't care either way if it's called `window.PRISM_JS_MANUAL` or something else. The fact is that there needs to be a better way for Prism to detect when it's not supposed to run automatically than checking for a data attribute.

mAAdhaTTah:
Borrow the `manual` property of whatever already existing Prism instance
there is.

Fixes #765.

In this setup, the `manual` property would have to be set before the script is included. Open to suggestions for improvement.

### DIFF
--- a/components/prism-core.js
+++ b/components/prism-core.js
@@ -12,12 +12,6 @@ var _self = (typeof window !== 'undefined')
  * @author Lea Verou http://lea.verou.me
  */
 
-var manual = false;
-
-if (_self.Prism) {
-	manual = _self.Prism.manual;
-}
-
 var Prism = (function(){
 
 // Private helper vars
@@ -25,7 +19,7 @@ var lang = /\blang(?:uage)?-(\w+)\b/i;
 var uniqueId = 0;
 
 var _ = _self.Prism = {
-	manual: manual,
+	manual: _self.Prism && _self.Prism.manual,
 	util: {
 		encode: function (tokens) {
 			if (tokens instanceof Token) {
--- a/components/prism-core.js
+++ b/components/prism-core.js
@@ -12,6 +12,12 @@ var _self = (typeof window !== 'undefined')
  * @author Lea Verou http://lea.verou.me
  */
 
+var manual = false;
+
+if (_self.Prism) {
+	manual = _self.Prism.manual;
+}
+
 var Prism = (function(){
 
 // Private helper vars
@@ -19,6 +25,7 @@ var lang = /\blang(?:uage)?-(\w+)\b/i;
 var uniqueId = 0;
 
 var _ = _self.Prism = {
+	manual: manual,
 	util: {
 		encode: function (tokens) {
 			if (tokens instanceof Token) {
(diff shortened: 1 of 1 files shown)


components/prism-core.js
```diff
@@ -12,13 +12,20 @@ var _self = (typeof window !== 'undefined')
...
+}
+
 var Prism = (function(){
 
 // Private helper vars
 var lang = /\blang(?:uage)?-(\w+)\b/i;
 var uniqueId = 0;
 
 var _ = _self.Prism = {
+	manual: manual,
```
LeaVerou: I would be in favor of minimizing the number of changes required. It would all be done here:

```js
manual: _self.Prism && _self.Prism.manual
```

That way, there's no need for the code changes above!

LeaVerou:
Thanks! See comment.

mAAdhaTTah:
Updated!

mAAdhaTTah:
Just a thought: this does require some checks in the code that's supposed to disable Prism, if as in the scenario from the original issue, the "disabling" script can't know whether it's been loaded before or after Prism, but I don't think there's anything we can do about that.

LeaVerou:
Thanks, merged!

Yeah it does, but it could be as simple as:

```js
(window.Prism = window.Prism || {}).manual = true;
```

mAAdhaTTah:
I dont think that'll work because the callback will have already been attached. So more like:

```js
if (window.Prism) {
  document.removeEventListener('DOMContentLoaded', Prism.highlightAll);
else {
  window.Prism = { manual: true };
}```

LeaVerou:
Oh, right. The check for `_.manual` should be in the callback, not before attaching the event! Is there any problem if we change it to that?

mAAdhaTTah:
We can; we would break BC for people using the old `document.removeEventListener('DOMContentLoaded', Prism.highlightAll)` method of removing auto-highlighting, as now the callback we're attaching won't be just `Prism.highlightAll`.

LeaVerou:
I think that's ok, since this is much more robust :) What do you think?

mAAdhaTTah:
Fine by me; if we do that, we should document that somewhere though. Is the CHANGELOG enough or should we include a "how to disable auto-highlighting" somewhere?

LeaVerou:
Ideally it should be in the docs, alongside the docs for `data-manual`.

k-funk:
My use case, which I think is fairly general when wanting to use manual, is to import and use the `Prism.highlight` function directly. An example of that on the docs would be really helpful.

This is what I'm currently using, with the data-manual on my html/script tag. I'm hoping with this PR I can take that html tag out.

    import Prism from 'prismjs';
    import 'prismjs/components/prism-json';

    // initialize view
    Prism.highlight(JSON.stringify(rawJSON, null, 2), Prism.languages.json)

omeid:
@mAAdhaTTah I am using PrimsJS yet again after many years and hit the same issue that had before, (#366), so thanks for fixing this.

This could use some documentation though. It is not obvious how one would set this option.

mAAdhaTTah:
@omeid Thanks for the reminder. Open #1402 to track. Would love a PR if you've figured out how to use it 😉

PR #1089, opened 2017-01-28
Add TAP component

isaacs:
I'd like to use Prism for the node-tap docs website http://node-tap.org/

Still needs styling.  I'm not sure of the best way to go about this.  Is it better to go through and add `.token.fail`, `.token.pass`, etc., or just re-use token names from another language?


mAAdhaTTah:
@Golmote This also looks useful. Thoughts?

Golmote:
@mAAdhaTTah It seems a bit like a niche usage, but why not. There is stuff to do on this PR though, there are changes related to `manual` that seem unrelevant here, and it will need examples, tests, etc. It seems we do have write access on the branch, so we _could_ update it ourselves.

mAAdhaTTah:
Superceded by  #1430.

PR #1094, opened 2017-02-07
fixes #1061 JSX spread operator break

### LINKED ISSUE #1061: JSX spread operator breaks highlighting
If the JSX spread operator is used inside a tag, highlighting breaks starting at that point.  Example code to be highlighted:

```js
class MyComponent extends Component
    render() {
        const {tabs, ...otherProps} = this.props;
        const {currentTab} = this.state;
        
        return (
            <TabBar
                {...otherProps}
                currentTab={currentTab}
                onTabClick={this.onTabClick}
                tabs={tabs}
            />
        )
    }
```

If `{...otherProps}` is there, the props after that are not highlighted.  If `{...otherProps}` is removed, they highlight correctly:

prism-jsx-broken
prism-jsx-correct

vkbansal:


components/prism-jsx.js
```diff
@@ -3,10 +3,20 @@
 var javascript = Prism.util.clone(Prism.languages.javascript);
 
 Prism.languages.jsx = Prism.languages.extend('markup', javascript);
-Prism.languages.jsx.tag.pattern= /<\/?[\w\.:-]+\s*(?:\s+[\w\.:-]+(?:=(?:("|')(\\?[\w\W])*?\1|[^\s'">=]+|(\{[\w\W]*?\})))?\s*)*\/?>/i;
+Prism.languages.jsx.tag.pattern= /<\/?[\w\.:-]+\s*(?:\s+[\w\.:-]*(?:=?(?:("|')(\\?[\w\W])*?\1|[^\s'">=]+|(\{[\w\W]*?\})))?\s*)*\/?>/i;
```
Golmote: Could you please explain that change a bit? There seems to be a large part of the regexp that would now be able to match only whitespaces... I'm afraid this could increase the amount of backtracking.

vkbansal: The change is that it was previously matching a word followed by `=` and it was compulsory. I just made the word and `=` optional by replacing `+` with `*` for the word and adding `?` after `=`.

Golmote: The consequence is that you are allowing a whole bunch of weird stuff doing this. With the new regexp, `<div "">` would match, for example.

I'd prefer a more explicit version, that allows the spread operator. Something like...

```
/<\/?[\w\.:-]+\s*(?:\s+(?:[\w\.:-]+(?:=(?:("|')(\\?[\w\W])*?\1|[^\s'">=]+|(\{[\w\W]*?\})))?|\{\.{3}\w+\})\s*)*\/?>/i
```

or if you want to combine the spread part with the existing part using curly braces, then something like:

```
/<\/?[\w\.:-]+\s*(?:\s+(?:[\w\.:-]+(?:=(?:("|')(\\?[\w\W])*?\1|[^\s'">=]+))?|(?:[\w\.:-]+=)?\{[\w\W]*?\})\s*)*\/?>/i
```
though I think the first one is a little bit more readable, and might backtrack a bit less.

What do you think?

vkbansal: I too like the first approach. I'll update the PR

vkbansal: done

Issue #1075, opened 2017-01-04
[YAML Highlight] # inside strings is treated as comment


mjrussell:
In YAML, the `#` for comments can appear inside strings which are enclosed with `"` or `'`

Github formatting:
```yaml
# comment
foo: "#bar"
foo: "bar"
```

Prism:
![image](https://cloud.githubusercontent.com/assets/1641895/21627907/2dd7f5ac-d1d0-11e6-9c6a-596432ba4dd9.png)

I understand it is because https://github.com/PrismJS/prism/blob/gh-pages/components/prism-yaml.js#L7 is too greedy. But Im not sure how to change it. Would look behind for `"` or `'` be sufficient?

mjrussell:
Ah looks like the key is to have `greedy: true` on the string. Will submit a PR

mjrussell:
Actually its pretty tricky to support all these edge cases so not going to submit a PR yet.

The current behavior supports multi line strings, but Im having trouble keeping that functionality while enabling comments inside of strings. Hoping someone more experienced with Prism might have some insights

Golmote:
@mjrussell: thanks for reporting.
`greedy: true` on the string pattern seems to be the right solution here, indeed. Could you provide the problematic edge case you tested, please?

(Poke @hason: you might want to check this issue)

mjrussell:
Mutli line's with comments i.e.:
```yaml
mutli: "mutli line
  with # inside" # and out
```

will fail because the regex (probably not new behavior actually with greedy):
`/([:\-,[{]\s*(![^\s]+)?[ \t]*)("(?:[^"\\]|\\.)*"|'(?:[^'\\]|\\.)*')(?=[ \t]*($|,|]|}))/m`

Expects the last case `$|,|]|}` to end to be treated as a string. Seems like it needs one more alternative to eat whitespace until another comment (but let it be parsed as a comment)

Golmote:
As you can see in my commit above, the issue with # inside strings should be solved. That is, `foo: "#bar"` should get highlighted properly.

But we are still facing two issues here.

The first one, related to YAML, is the abusive use of that group `$|,|]|}` everywhere. I'm pretty sure there is a better way to handle those highlightings. (If we fix it, YAML would end up with the same behaviour as the other components, descibed below.)

The second one is a flaw in the current greedy behaviour, which can be reproduced with nearly every other component:
JS:
[...]

PR #1095, opened 2017-02-08
Fix corner cases for the greedy flag

zeitgeist87:
By refactoring the tokenize() method into two separate methods, it is
possible to recursively clean up any broken tokens left by the greedy
feature.

This should fix the issue #1075.


zeitgeist87:
Actually this is not a good solution after all. It only works if the comment contains another comment.

zeitgeist87:
The same error occurs with a template string in JS:

greedy

Golmote:
Thank you for taking a look into this!
I think your idea should be able to fix even the template string example. After the part ``//" ` sdsf ` `` is invalidated, we could reprocess only the remaining fragment, that is `` ` sdsf ` ``, which should highlight as a template string, right?

zeitgeist87:
Yes I think it is possible to optimize and generalize this approach. I will try a few things and then update this PR.

zeitgeist87:
I don't think we can make the `greedy` flag perfect. It was a compromise from the beginning and there will always be a few corner cases where it fails.

At some point we have to decide if the extra effort of fixing `greedy` is worth it.

zeitgeist87:
It should be fixed now. I've also added test cases to a few languages including one test case for template strings in Javascript.

zeitgeist87:
@Golmote Do you think this can be merged?

Golmote:
@zeitgeist87 Again sorry for the long time before review. This looks good to me. Thank you so much for digging into this.

Feel free to merge when you can!

zeitgeist87:
@Golmote Thanks for the review! I've rebased it and solved the merge conflicts.

PR #1098, opened 2017-02-10
Remove yarn.lock

idleberg:
See [Lockfiles should be committed on all projects](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/)


Golmote:
Sorry for the delay. Thanks for contributing!

Issue #1097, opened 2017-02-09
Double slashes in PHP strings


ethanhowell:
If I have two backslashes in a php string, Prism will mark the rest of the line as a comment.
![example](https://cloud.githubusercontent.com/assets/21162595/22766544/57f7502e-ee33-11e6-8827-18fdfa81b012.png)

zeitgeist87:
Thanks for reporting this. This is definitely a bug.

The problem is, that the `greedy` flag is set on the `comment` token. Comments are always highlighted first, so there seems to be no reason for the frag there. I suspect, that it was put there to fix some other issue.

zeitgeist87:
This issue should be fixed by #1100

Issue #197, opened 2013-12-28
PHP Comment with <element> inside


mavimo:
There is some issue on syntax highlight in PHP
```php
/**
 * A standard comment.
 */

/**
 * Another comment with broken syntax highlight
 *
 * @author Marco Vito Moscaritolo <marco@mavimo.org>
 */
```

In [test drive](http://prismjs.com/test.html?language=php), when you enable PHP there is an issue that do not display the second item as comment. Clike syntax work fine.

After some debugging seam to be an issue related to markup, if I add a ```&& false``` at [line 40 of prism-php.js](https://github.com/LeaVerou/prism/blob/gh-pages/components/prism-php.js#L40) the syntax highlight work fine.

PR #1100, opened 2017-02-10
Fix the PHP language

zeitgeist87:
This patch is an attempt to fix the PHP language, when it is combined
with markup.

The problem is, that markup has a higher priority than all other tokens.
This leads to weird errors, where HTML tags are highlighted inside of
comments #197. One solution to this was to set the `greedy` flag for the
comment token, but this leads to far worse errors like #1097.

This patch should fix both issues #197 and #1097, by switching the
grammar to markup on the fly. One potential problem is, that it relies
on the `<?php` tag to detect if markup is present. So if a PHP file
contains only markup and no PHP code at all, the result will look
broken.


zeitgeist87:
Before (Note: The JavaScript function is erroneously highlighted by PHP. CSS is not highlighted at all. Strings inside comments do not work):
before

After (Note: The string `"px"` looks different, because CSS has custom styles for the class `token string`. JavaScript and CSS are highlighted by the correct language.) :
after

zeitgeist87:
There is one problem with this PR. If there are no `<?php` tags, then it assumes, that it is pure PHP code, which could be wrong:
wrong
correct2

If a `<?php` tag is present, then markup is also highlighted:
correct

zeitgeist87:
@LeaVerou @Golmote What do you think about this PR?

LeaVerou:
> If there are no <?php tags, then it assumes, that it is pure PHP code, which could be wrong:

I don't think that's a huge problem (they can use `markup` in those cases), as long as it supports the shortened PHP tags as well (`<?=` etc). Though I don't see why we can't check if a PHP tag is present and use the markup definition if not.

zeitgeist87:
> Though I don't see why we can't check if a PHP tag is present and use the markup definition if not.

Yes that is one option, but I think it is more intuitive to assume that it is all PHP in that case. For example someone could use small snippets of PHP code between paragraphs on a blog to explain a larger file:

```php
$test = "<div>" . $_SERVER['PHP_SELF'] . "</div>";
```

If we switch to `markup` in that scenario, we would force the users to put PHP tags around every small snippet:

```php
<?php $test = "<div>" . $_SERVER['PHP_SELF'] . "</div>"; ?>
```
If there is no PHP code at all, people should use `markup` directly.

LeaVerou:
Ah yeah, excellent point! Completely agree.

Golmote:
@zeitgeist87 Sorry for the looong time before review. This looks ok to me! ;) Feel free to merge.

zeitgeist87:
@Golmote Thanks again for the review! Merging...

PR #1102, opened 2017-02-12
Add Web IDL language support

grorg:
Hi. This is my first attempt at adding a language. I'm not sure I got everything right - the examples page has this language greyed out, for example. It seems that it tries to load content from github?

I'm happy to change anything necessary.

WebIDL features not yet supported: extended attributes and predefined error/exception names.


components/prism-webidl.js
```diff
@@ -0,0 +1,12 @@
...
+	'boolean': /\b(true|false)\b/,
+	'operator': /-|:|=|\?|\*|\//
+});
+
+Prism.languages.insertBefore('webidl', 'keyword', {
+	'class-name': {
+		pattern: /(class\s+)[a-z0-9_]+/i,
+		lookbehind: true
+	}
+});
```
LeaVerou: What's the point of calling `insertBefore` for webidl right after you've defined webidl? `insertBefore` is used to modify existing languages, if you're defining a language yourself you don't need it, just put `class-name` in the object literal. If you don't want to use the newer `['class-name']` syntax due to browser support, just pick a different token name without a hyphen and use `class-name` as an alias. @Golmote what do you think?
 
Also, your dependencies list C-like, but here you seem to be extending C. These are two different language definitions.

LeaVerou:
I've added a few comments, though I'd wait for some more review from the other maintainers before merging. 

But I'm really excited about this! I assume you're using Prism in specs then?

components/prism-webidl.js
grorg: I copied the cpp implementation, so maybe there is something in the insertBefore code that it needed. Either way, I don't understand so I'm happy to do whatever.

As for the dependencies, you're right that it's not clear. WebIDL isn't technically C, C-like, or CPP. I figured it would be ok to extend from cpp though, since it does have fairly similar syntax.

grorg:
> But I'm really excited about this! I assume you're using Prism in specs then?

I was using it in some [WebGPU](https://github.com/gpuweb/) stuff. Neither prism or highlight.js supported WebIDL. I decided prism was easier :)

components/prism-webidl.js
LeaVerou: Just add the "class-name" token before the keyword token. Due to the hyphen, you'd need to surround it in quotes, but apart from that, it's just a normal property.

As for the dependencies, I don't mind which one you inherit from, just be consistent between your code and the components declaration, otherwise people will get a download that doesn't work. Perhaps look over their examples to decide which of the three is closer to WebIDL?

zeitgeist87: @LeaVerou The `c` language actually deletes the `class-name` property from `clike`, so if you inherit from `c` and you want to put `class-name` at a certain position, you have to use `insertBefore()`. However if you inherit from `clike`, you can directly add `class-name` to the language.

LeaVerou: I see. So the main blocker here is that @grorg needs to decide whether he's extending C, C++ or C-like. I think the latter is probably a best fit, but he knows WebIDL better.

grorg: Thanks for the tips @LeaVerou and @zeitgeist87. I'll update the patch soon.

> he knows WebIDL better

Hahahahaha. If only.

mAAdhaTTah:
@grorg Are you able to make the changes suggested here? We'd love to land this!

grorg:
Yeah, thanks for poking me. I'll clean it up and post again soon.

RunDevelopment:
Closed in favor of #3107.

PR #1106, opened 2017-02-28
remove comments spellcheck for amp validation

emmanuelgautier:
spellcheck attribute cause AMP validation errors. This pull request remove this optional attribute to allow prismjs preprocess AMP pages code.


Golmote:
Hi @emmanuelgautier. Is there any way we could keep when not on AMP pages? Like, could we detect we're on an AMP page and decide to not add it then?

emmanuelgautier:
Hi @Golmote, highlighting is processing on server side so it could be possible to imagine an option for the ```highlight``` function. It also possible to detect that the tag ```<html ⚡>``` or this one ```<html amp>``` is present. I think the best way is to give function the capability to know his running context.

What do you think about this two solutions ?

Golmote:
I'd rather add it as a global option for Prism, the kind we have for `manual` highlighting. Something in the lines of:

```html
<script>
var Prism = {
    spellcheck: false
};
</script>
```

See https://github.com/PrismJS/prism/pull/1087 and https://github.com/PrismJS/prism/pull/1188

mAAdhaTTah:
Why do comments need spellcheck="true"?

Golmote:
I actually never though about this... I guess @LeaVerou tought it would be convenient. This has been around for quite some time I believe. Now that I think about it, it only makes sense with `contenteditable="true"`, so it must have been used initially in [Dabblet](http://dabblet.com/).

mAAdhaTTah:
There is a `_.hooks.run('wrap', env);` right after the `env` gets built. If this is needed by Dabblet, it should be doable to hook into the env from outside and add `spellcheck="true"` if needed by Dabblet or anyone else. I'm :+1: for going ahead and removing, unless @LeaVerou has objections due to conflicts in Dabblet.

LeaVerou:
I’m fine with removing it. :)
I don't think we need to add a pre-init option for it, as @mAAdhaTTah pointed out, there's a hook if someone needs it desperately.

oliviertassinari:
🎉

PR #1107, opened 2017-03-07
Replace [\w\W] with [\s\S] and [0-9] with \d in regexes

valtlai:
`[\w\W]` &nbsp;→&nbsp; `[\s\S]`
&nbsp; `[0-9]` &nbsp;→&nbsp; `\d`


Golmote:
Wow, I actually didn't know about the `[^]` shortcut. Although it sounds clever, I'm afraid it might be less clear than the more usual `[\s\S]` and `[\w\W]` variants.
I'm perfectly ok with replacing `[0-9]` with `\d` though.

What's your opinion about this @LeaVerou, @zeitgeist87 ?

LeaVerou:
I'm fine with it :)

zeitgeist87:
According to https://regex101.com/ the `[^]` regex works only with the Javascript flavor. It doesn't work in Python, Go or PHP. I don't know how accurate that is and I haven't actually tested it with different regex engines. It seems to be a pretty obscure trick though. Maybe we should take that into consideration...

Golmote:
@zeitgeist87, I agree. @valtlai, would you mind updating this PR to use one of `[\s\S]` or `[\w\W]` (we could definitely uniform all occurrences to one of those choices) to match any character, instead of `[^]` which might lead to confusion?

valtlai:
@Golmote Yeah, done.

Golmote:
Thanks for contributing!

PR #1122, opened 2017-03-27
Add VB.Net

Bigsby:


Golmote:
Thanks for contributing!

Bigsby:
One is glad to be os service.

PR #1126, opened 2017-04-06
Update prism-apl.js

jayfoad:
Add @ (At) and ⌺ (Stencil), two new dyadic operators coming in Dyalog APL version 16.0.


Golmote:
Thanks for contributing!

PR #1139, opened 2017-05-23
Support CSS #RRGGBBAA

valtlai:
Add support for CSS #RRGGBBAA notation.


LeaVerou:
LGTM. Note that this also allowed invalid 5 digit formats, but eh, Prism is not a linter.

PR #1143, opened 2017-06-03
Support HTML entities in attribute values

valtlai:
prism-attr-entity


LeaVerou:
Thanks for contributing!
However, we tend to avoid duplicating expressions, and instead prefer to point to them by reference. This way, they will always be in sync. You could add a line under the language definition that does this.

PS: Nice theme! Yours?

valtlai:
Done.

Yeah, it’s mine. The colors are from Chrome DevTools’ material colors palette.

LeaVerou:
Merged, thanks!

Please consider releasing the theme somewhere, it's gorgeous.

PR #1158, opened 2017-07-10
remove extra space before language class

Sibiraj-S:
remove additional space before adding append language to the element.

### DIFF
--- a/prism.js
+++ b/prism.js
@@ -188,13 +188,13 @@ var _ = _self.Prism = {
 		}
 
 		// Set language on the element, if not present
-		element.className = element.className.replace(lang, '').replace(/\s+/g, ' ') + ' language-' + language;
+		element.className = element.className.replace(lang, '').replace(/\s+/g, ' ') + 'language-' + language;
 
 		// Set language on the parent, for styling
 		parent = element.parentNode;
 
 		if (/pre/i.test(parent.nodeName)) {
-			parent.className = parent.className.replace(lang, '').replace(/\s+/g, ' ') + ' language-' + language;
+			parent.className = parent.className.replace(lang, '').replace(/\s+/g, ' ') + 'language-' + language;
 		}
 
 		var code = element.textContent;


Golmote:
Currently, you can use Prism this way:
```html
<div class="language-markup">
    <pre class="foobar"><code>&lt;div></code></pre>
</div>
```
When Prism runs, the class `language-markup` will be added on the `<pre>` tag.

I'm pretty sure your change would break this behaviour, and set the class to `foobarlanguage-markup` instead of `foobar language-markup`.

LeaVerou:
Yup, @Golmote is right, sorry!

Sibiraj-S:
No @Golmote. I made this PR because there is an extra space . let me attach the screen shot too

it is appending like this. and I checked that too 

```
<div class=" language-markup">
    <pre class="foobar"><code>&lt;div></code></pre>
</div>
```

Sibiraj-S:
This is a plunker I showing it to test http://embed.plnkr.co/F77rcio6e4Cj5T0Pk2w3/

screen shot 2017-07-12 at 11 06 38 am

I marked the extra space. ie., actually three spaces between class

untitled

Golmote:
I checked again this morning. In many other use cases, the extra space is perfectly useless and it would be fine to remove it. But I confirm your PR prevents Prism from properly adding a class when there is already a class on the `<pre>` or `<code>` tag, and breaks the example I gave previously.

Sibiraj-S:
yeah. i checked to my PR breaks the code.

Sibiraj-S:
@Golmote . I heard that having extra space may cause problems during the class string manipulations

kylekatarnls:
It would not be the responsibility of Prism but of the library that manipulate the class string. HTML coder should be able to write this freely:
```
<p class="
  foo
  bar
"></p>
```
Every browser will understand this properly for getElementByClassName API or for CSS.

So libraries too could be free to add spaces anywhere in classes if it helps.

LeaVerou:
> I heard that having extra space may cause problems during the class string manipulations

Only for extremely sloppy code. @kylekatarnls is right.

PR #1163, opened 2017-07-16
Scroll 'Coy' background with contents

nickersonm:
Fix on <code>themes/prism-coy.css</code>: attaches &lt;pre&gt; background (alternating line colors) to &lt;code&gt; contents when in limited-height &lt;pre&gt; that requires scrolling.

Example: http://prismjs.com/ with 'Coy' theme selected; most evident on smallest scroll, such as the SVG logo highlight.

### DIFF
--- a/themes/prism-coy.css
+++ b/themes/prism-coy.css
@@ -30,15 +30,18 @@ pre[class*="language-"] {
 pre[class*="language-"] {
 	position: relative;
 	margin: .5em 0;
-	box-shadow: -1px 0px 0px 0px #358ccb, 0px 0px 0px 1px #dfdfdf;
+	overflow: visible;
+	padding: 0;
+}
+pre[class*="language-"]>code {
+	position: relative;
 	border-left: 10px solid #358ccb;
+	box-shadow: -1px 0px 0px 0px #358ccb, 0px 0px 0px 1px #dfdfdf;
 	background-color: #fdfdfd;
 	background-image: linear-gradient(transparent 50%, rgba(69, 142, 209, 0.04) 50%);
 	background-size: 3em 3em;
 	background-origin: content-box;
 	background-attachment: local;
-	overflow: auto;
-	padding: 0;
 }
 
 code[class*="language"] {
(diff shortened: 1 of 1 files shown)


LeaVerou:
Hey, thanks! Is there an online demo of this anywhere?

nickersonm:
Usage: drag the scroll bar vertically by small amounts.
Main branch: https://jsfiddle.net/d3L8nvjf/
With patched coy-style.css: https://jsfiddle.net/dbksxu0c/1/

However, from reading other prism commits and comments, I realized that this removes the 'curved' dropshadow below the &lt;pre&gt; blocks (which are hidden at narrower sizes- expand the fiddle result window), inserted with :before and :after entities.  This is because the dropshadow is on the same layer as the background: the contents of the &lt;pre&gt;, and is also aligned to the &lt;pre&gt;.  I have not been able to find a good CSS-only solution to including both.

It would be possible by <a href="https://jsfiddle.net/dbksxu0c/3/">wrapping the &lt;pre&gt; in a &lt;div&gt;</a> for the positioning (explained <a href="http://webcache.googleusercontent.com/search?q=cache:MQBoyjJ-cBkJ:front-back.com/how-to-make-absolute-positioned-elements-overlap-their-overflow-hidden-parent+&cd=10&hl=en&ct=clnk&gl=us">here</a>), but that's not desired behavior.

LeaVerou:
Thanks for the demo, that was fast!

> I realized that this removes the 'curved' dropshadow

Yeah, that would be a problem I’m afraid… :/

nickersonm:
Understood.  I [committed a change](https://github.com/PrismJS/prism/pull/1163/commits/23df4685b51d7689fe0ef0a4e12e50e15741a1b3) to this pull branch that reverts to using the original <code>overflow: auto</code> on the &lt;code&gt; block and moves the background from the &lt;pre&gt; block to the &lt;code&gt;.

It seems to both scroll the background and preserve everything else, but I have not tested it as extensively.

Demo: 
Main branch: https://jsfiddle.net/d3L8nvjf/
With patched coy-style.css: https://jsfiddle.net/dbksxu0c/1/ (auto-updated as it uses the raw nickersonm/patch-2 branch CSS)

Golmote:
Demo seems to work just fine! Thank you for contributing!

PR #1168, opened 2017-07-26
Omit unnecessary space before language class

ycherniavsky:
Suddenly I have mentioned that after applying Prism to `pre` element with only one language class in producing `pre` and `code` elements class attributes contain unnecessary space:
```html
<pre class=" language-javascript"><code class=" language-javascript">
...
</code></pre>
```
which simple could be fixed by applying additional replace pattern (as it already made for contiguous spaces).

### DIFF
--- a/components/prism-core.js
+++ b/components/prism-core.js
@@ -183,13 +183,13 @@ var _ = _self.Prism = {
 		}
 
 		// Set language on the element, if not present
-		element.className = element.className.replace(lang, '').replace(/\s+/g, ' ') + ' language-' + language;
+		element.className = element.className.replace(lang, '').replace(/\s+/g, ' ').replace(/\S$/g, '$& ') + 'language-' + language;
 
 		// Set language on the parent, for styling
 		parent = element.parentNode;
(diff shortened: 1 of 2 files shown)


Golmote:
@LeaVerou Could you give a look at this PR? I know you have been quite reticent about changes regarding this extra-space in the past. Please tell me if you think this PR is worth it or not.

LeaVerou:
Is there any actual reason to remove that space? It's perfectly valid and browsers deal with it just fine.
The only argument I've seen about it is variations of "I don't like how it looks", which is not a serious reason to do code changes, especially changes that hinder code readability.

Golmote:
Yes, you're right, the only reason people want to remove it is "aesthetic".
That's what I thought you'd answer. ^^ So I'm closing this PR.

LeaVerou:
> Yes, you're right, the only reason people want to remove it is "aesthetic".

Selective aesthetics may I add. The proposed change to Prism’s code is much less aesthetically pleasing than an extra space IMO.

PR #1173, opened 2017-08-11
fixed an rendering issue for encoded urls

newgene:
If a url contains some special characters, like space or "<", it needs to be encoded in order to be recognized as a valid url. Adding a decoding step allows these urls to render correctly.

For example, a url
```
http://example.com?q=a AND b
```
will be rendered by autolinker as a wrong url which breaks before the first space.

we can use this encoded one to get around it

```
http://example.com?q=a%20AND%20b
```

The rendered url works, but does not decode the special characters.

This simple one line should fix this issue.


Golmote:
Thanks for contributing!

PR #1174, opened 2017-08-17
Add Elm (elm-lang.org) support

zwilias:


components/prism-elm.js
```diff
@@ -0,0 +1,35 @@
...
+	char: /'([^\\']|\\([abfnrtv\\"'&]|\^[A-Z@[\]\^_]|NUL|SOH|STX|ETX|EOT|ENQ|ACK|BEL|BS|HT|LF|VT|FF|CR|SO|SI|DLE|DC1|DC2|DC3|DC4|NAK|SYN|ETB|CAN|EM|SUB|ESC|FS|GS|RS|US|SP|DEL|\d+|o[0-7]+|x[0-9a-fA-F]+))'/,
+	string: {
+		pattern: /"([^\\"]|\\([abfnrtv\\"'&]|\^[A-Z@[\]\^_]|NUL|SOH|STX|ETX|EOT|ENQ|ACK|BEL|BS|HT|LF|VT|FF|CR|SO|SI|DLE|DC1|DC2|DC3|DC4|NAK|SYN|ETB|CAN|EM|SUB|ESC|FS|GS|RS|US|SP|DEL|\d+|o[0-7]+|x[0-9a-fA-F]+)|\\\s+\\)*"/,
+		greedy: true
+	},
+	import_statement: {
+		// The imported or hidden names are not included in this import
+		// statement. This is because we want to highlight those exactly like
+		// we do for the names in the program.
+		pattern: /(\r?\n|\r|^)\s*import\s+([A-Z][_a-zA-Z0-9]*)(\.[A-Z][_a-zA-Z0-9]*)*(\s+as\s+([A-Z][_a-zA-Z0-9]*)(\.[A-Z][_a-zA-Z0-9]*)*)?(\s+exposing\s+)?/m,
```
Golmote: - Do you really need to capture the line-feed at the beginning? Isn't the `^` anchor enough with the `m` flag?
- Occurrences of `[_a-zA-Z0-9]` should be simplified as `\w`.
- There is no need for the parentheses around `([A-Z][_a-zA-Z0-9]*)`.

components/prism-elm.js
```diff
@@ -0,0 +1,35 @@
+Prism.languages.elm = {
+	comment: {
+		pattern: /(^|[^-!#$%*+=?&@|~.:<>^\\\/])(--[^-!#$%*+=?&@|~.:<>^\\\/].*|{-[\s\S]*?-})/m,
+		lookbehind: true
+	},
+	char: /'([^\\']|\\([abfnrtv\\"'&]|\^[A-Z@[\]\^_]|NUL|SOH|STX|ETX|EOT|ENQ|ACK|BEL|BS|HT|LF|VT|FF|CR|SO|SI|DLE|DC1|DC2|DC3|DC4|NAK|SYN|ETB|CAN|EM|SUB|ESC|FS|GS|RS|US|SP|DEL|\d+|o[0-7]+|x[0-9a-fA-F]+))'/,
```
Golmote: This looks quite overkill to me. Remember that Prism is not a linter.
Is there any case where a single-quoted string of characters should not be highlighted as `char`? If there isn't, you should be able to drastically simplify this regexp.

Golmote: Also, you might want to add the `greedy` flag, here.

components/prism-elm.js
```diff
@@ -0,0 +1,35 @@
...
+	string: {
+		pattern: /"([^\\"]|\\([abfnrtv\\"'&]|\^[A-Z@[\]\^_]|NUL|SOH|STX|ETX|EOT|ENQ|ACK|BEL|BS|HT|LF|VT|FF|CR|SO|SI|DLE|DC1|DC2|DC3|DC4|NAK|SYN|ETB|CAN|EM|SUB|ESC|FS|GS|RS|US|SP|DEL|\d+|o[0-7]+|x[0-9a-fA-F]+)|\\\s+\\)*"/,
```
Golmote: Same comment here: this looks overcomplicated. Are there string-like patterns in Elm that are not to be highlighted as `string`?

components/prism-elm.js
```diff
@@ -0,0 +1,35 @@
...
+	// decimal integers and floating point numbers | octal integers | hexadecimal integers
+	number: /\b(\d+(\.\d+)?(e[+-]?\d+)?|0x[0-9a-f]+)\b/i,
+	// Most of this is needed because of the meaning of a single '.'.
+	// If it stands alone freely, it is the function composition.
+	// It may also be a separator between a module name and an identifier => no
+	// operator. If it comes together with other special characters it is an
+	// operator too.
+	operator: /\s\.\s|[-!#$%*+=?&@|~.:<>^\\\/]*\.[-!#$%*+=?&@|~.:<>^\\\/]+|[-!#$%*+=?&@|~.:<>^\\\/]+\.[-!#$%*+=?&@|~.:<>^\\\/]*|[-!#$%*+=?&@|~:<>^\\\/]+|`([A-Z][_a-zA-Z0-9']*\.)*[_a-z][_a-zA-Z0-9']*`/,
+	// In Elm, nearly everything is a variable, do not highlight these.
+	hvariable: /\b([A-Z][_a-zA-Z0-9]*\.)*[_a-z][_a-zA-Z0-9]*\b/,
```
Golmote: Occurrences of `[_a-zA-Z0-9]` should be replaced with `\w`.

components/prism-elm.js
```diff
@@ -0,0 +1,35 @@
...
+	constant: /\b([A-Z][_a-zA-Z0-9]*\.)*[A-Z][_a-zA-Z0-9]*\b/,
```
Golmote: Occurrences of `[_a-zA-Z0-9]` should be replaced with `\w`.

components/prism-elm.js
```diff
@@ -0,0 +1,35 @@
...
+	// These are builtin variables only. Constructors are highlighted later as a constant.
+	builtin: /\b(abs|acos|always|asin|atan|atan2|ceiling|clamp|compare|cos|curry|degrees|e|flip|floor|fromPolar|identity|isInfinite|isNaN|logBase|max|min|negate|never|not|pi|radians|rem|round|sin|sqrt|tan|toFloat|toPolar|toString|truncate|turns|uncurry|xor)\b/,
+	// decimal integers and floating point numbers | octal integers | hexadecimal integers
+	number: /\b(\d+(\.\d+)?(e[+-]?\d+)?|0x[0-9a-f]+)\b/i,
+	// Most of this is needed because of the meaning of a single '.'.
+	// If it stands alone freely, it is the function composition.
+	// It may also be a separator between a module name and an identifier => no
+	// operator. If it comes together with other special characters it is an
+	// operator too.
+	operator: /\s\.\s|[-!#$%*+=?&@|~.:<>^\\\/]*\.[-!#$%*+=?&@|~.:<>^\\\/]+|[-!#$%*+=?&@|~.:<>^\\\/]+\.[-!#$%*+=?&@|~.:<>^\\\/]*|[-!#$%*+=?&@|~:<>^\\\/]+|`([A-Z][_a-zA-Z0-9']*\.)*[_a-z][_a-zA-Z0-9']*`/,
```
Golmote: - Occurrences of `[_a-zA-Z0-9']` should be replaced with `[\w']`.
- I might be wrong, but couldn't the whole regexp be rewritten as this one?
(single dot | special char (including dot) appearing two or more times | special char alone (not including dot) | backtick stuff)
``/\s\.\s|[-!#$%*+=?&@|~:<>^\\\/.]{2,}|[-!#$%*+=?&@|~:<>^\\\/]|`([A-Z][\w']*\.)*[_a-z][\w']*`/``

kachkaev:
Thanks a lot for this PR @zwilias! Just pinging as it'd be great if you could respond to @Golmote's review. Looking forward to see elm highlighting in [markdown-preview-enhanced](https://shd101wyy.github.io/markdown-preview-enhanced/#/) atom/vscode plugin, which uses PrismJS!

zwilias:
Put it on my todo list; I'll try to get back to this asap. Thanks for the reminder, @kachkaev, and so sorry for dropping the ball here, @Golmote. <3

kachkaev:
@zwilias do you need help with anything?

components/prism-elm.js
zwilias: Yeah, upon closer inspection, I think `/'([^\\']|\\([abfnrtv']|\d+|x[0-9a-fA-F]+))'/` will do - any single character (except `'` itself) or an escape sequence which comes in a few forms. Thanks!

components/prism-elm.js
zwilias: You're right, this is way overcomplicated. I adapted it from the haskell one; and I'm now realizing that - other than being overcomplicated - it's also wrong. Elm doesn't do multiline strings that way, but rather with `"""`. Trying to figure out a good way to match either simple strings or multiline strings, which is slightly complicated by the latter being able to include single `"`.

components/prism-elm.js
zwilias: Yep. I'm gonna have a look at the characters the actual Elm parser allows, which should allow me to scrap a few. I included the backtick stuff for compatibility with Elm 0.17 but I think there's no real point to doing that, especially with 0.19 coming close to release.

I'll simplify, thanks :)

components/prism-elm.js
zwilias: 👌

zwilias:
Finally got around to cleaning up the patterns and fixing a few more bugs 😅

components/prism-elm.js
```diff
@@ -3,33 +3,45 @@ Prism.languages.elm = {
...
-	string: {
-		pattern: /"([^\\"]|\\([abfnrtv\\"'&]|\^[A-Z@[\]\^_]|NUL|SOH|STX|ETX|EOT|ENQ|ACK|BEL|BS|HT|LF|VT|FF|CR|SO|SI|DLE|DC1|DC2|DC3|DC4|NAK|SYN|ETB|CAN|EM|SUB|ESC|FS|GS|RS|US|SP|DEL|\d+|o[0-7]+|x[0-9a-fA-F]+)|\\\s+\\)*"/,
+	char: {
+		pattern: /'([^\\'\r\n]|\\([abfnrtv\\']|\d+|x[0-9a-fA-F]+))'/,
 		greedy: true
 	},
+	string: [
+		{
+			// Multiline strings are wrapped in triple ". Quotes may appear unescaped.
+			pattern: /"""((.|\s)*?(?="""))"""/,
```
Golmote: Is there an actual need for the positive look-ahead here?
Also, can you rewrite `(.|\s)` as `[\s\S]`? It's what we use everywhere else as a match-everything token.

tests/languages/elm/char_feature.test
```diff
@@ -7,6 +8,7 @@
 
 [
 	["char", "'a'"],
+  ["char", "'\\''"],
```
Golmote: Can you use a tab to fix the indentation of this line please?

tests/languages/elm/string_feature.test
```diff
@@ -1,17 +1,19 @@
...
+"""foo
+   bar"""
+"""foo -- comment
+   bar"""
 
 ----------------------------------------------------
 
 [
 	["string", "\"\""],
+  ["string", "\"regular string\""],
```
Golmote: Same here, please use a tab.

kachkaev:
@zwilias would you be happy to address @Golmote's comments! Feels like so few bits are left to fix! 🚀

kachkaev:
ping @zwilias 🙌

kachkaev:
Many thanks @zwilias! 👏  @Golmote do you think the PR can be now merged and produce a new PrismJS version?

Golmote:
Nice! Thank you both for sticking around. Expect a new version release (1.10) during the week.

Golmote:
@zwilias @kachkaev And 1.10.0 is out and published on NPM! 🎉

PR #1178, opened 2017-08-19
Added new feature + bugfixe

MattewEon:
I just updated the code to hide the first/last empty lines, it works now with the line-numbers plugin


LeaVerou:
Isn't this a subset of the [Normalize Whitespace](http://prismjs.com/plugins/normalize-whitespace/) plugin?

MattewEon:
You're right, I didn't thought about checking in an existing plugin as it seemed to be important !
Sorry for this mistake

PR #1180, opened 2017-08-29
Update the list of keywords for dockerfiles

queicherius:
This PR just adds a few missing keywords for [Dockerfiles](https://docs.docker.com/engine/reference/builder/) and reorders them to be in alphabetical order for easier future diffs.


Golmote:
Thank you for contributing!

PR #1187, opened 2017-09-13
Copy to clipboard: Get around name conflict in Chrome v61+ Fixes #1181

### LINKED ISSUE #1181: Uncaught TypeError: Illegal constructor Error
Hi,

I am getting this error:

```
Uncaught TypeError: Illegal constructor
    at n (prism.js:1)
    at prism.js:1
    at prism.js:1
    at Array.forEach (<anonymous>)
    at Prism.plugins.toolbar.hook (prism.js:1)
    at Object.run (prism.js:1)
    at Object.highlightElement (prism.js:1)
    at HTMLDocument.highlightAll (prism.js:1)
```

It seems to be showing when I add the "**Copy to Clipboard Button**" package.

My browser I am getting the error is Google Chrome Version 62.0.3200.0 (Official Build)

Here is a screenshot of the error in dev tools:

untitled

For now I am just going to live with out adding this plugin. I thought I'd let you know anyway.

p.s. Normal setup works fine no problem. It only appears with this plugin addon.

Golmote:
@mAAdhaTTah Do you think this could do the trick?


plugins/copy-to-clipboard/prism-copy-to-clipboard.js
```diff
@@ -10,9 +10,23 @@
 	}
 
 	var Clipboard = window.Clipboard || undefined;
+	var _nativeClipboard;
+
+	try {
+		// Try using something ClipboardJS-specific
+		new Clipboard('').destroy();
```
mAAdhaTTah: You probably don't need the `.destroy()` bit cuz `new Clipboard` will throw on its own, so if we `new` and it doesn't throw, we _should_ be able to assume it's not the global version? Unless `window.Clipboard` becomes instantiable.

Although we're kind of fighting against the future, trying to guess what the global Clipboard will be capable of...

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
```diff
@@ -10,9 +10,23 @@
...
+	} catch(e) {
+		// Existing Clipboard variable is not ClipboardJS.
+		// Store it so we can restore it later.
+		_nativeClipboard = Clipboard;
+		Clipboard = undefined;
+	}
 
 	if (!Clipboard && typeof require === 'function') {
 		Clipboard = require('clipboard');
+		if (_nativeClipboard) {
```
mAAdhaTTah: We shouldn't need to restore here because `require('clipboard')` shouldn't change the global instance.

Golmote: True that!

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
Golmote: Well, yeah I know... It's supposed to be a temporary fix, anyway. I hope ClipboardJS will release a non-conflicting version soon, so we can properly rollback this.

mAAdhaTTah:
tbh, the thing that bugs me about this, given Prism's target audience, is I'm not sure how many people are going to upgrade to the new version, which is part of why I'm hesitant to push in a temp fix. Users are going to have to realize the problem, trace it back to Prism (difficult on its own), then regenerate the prism.js file from the site and put it up on their site, a not-uninvolved process that could need to repeat should the browser change its behavior.

Which is why I was holding off on a fix until I got more info. I don't know if it makes more sense to push out a temp fix now and then fix it again or give it a chance to be resolved in a more permanent way.

Golmote:
Hm... That's a really good point I did not tought about. I think you're right. Let's give ClipboardJS some time to fix this on their own, and if need be this PR would still be usable. I'm closing for now. Thank you for reviewing!

PR #1194, opened 2017-09-25
Fixes to compatibility of line number and line higlight plugins

VitaliyR:
Hello.

First of all, sorry about created compatibility issues with this plugins. But 2 years passed after that PR.. I've just missed this point :) 

In this PR:
* Fixes to line numbers plugin with the different line endings
* Renamed `data-start` attribute to `data-line-numbers-start` - because line highlight plugin is using this attribute as well
* Fixes compatibility between these two plugins

I've decided to not separate bug fix to into its own branch because it is tied up to each other.

### DIFF
--- a/plugins/line-numbers/prism-line-numbers.js
+++ b/plugins/line-numbers/prism-line-numbers.js
@@ -109,13 +109,37 @@
 		lineNumbersWrapper.className = 'line-numbers-rows';
 		lineNumbersWrapper.innerHTML = lines;
 
-		if (pre.hasAttribute('data-start')) {
-			pre.style.counterReset = 'linenumber ' + (parseInt(pre.getAttribute('data-start'), 10) - 1);
+		if (pre.hasAttribute('data-line-numbers-start')) {
+			pre.style.counterReset = 'linenumber ' + (parseInt(pre.getAttribute('data-line-numbers-start'), 10) - 1);
 		}
 
 		env.element.appendChild(lineNumbersWrapper);
 
 		_resizeElement(pre);
 	});
+	
+	/**
+	 * Global exports
+	 */
+	Prism.plugins.lineNumbers = {
+		/**
+		 * Get node for provided line number
+		 * @param {Element} element pre element
+		 * @param {Number} number line number
+		 * @return {Element|undefined}
+		 */
+		getLine: function (element, number) {
+			if (element.tagName !== 'PRE' || !element.classList.contains(PLUGIN_CLASS)) {
+				return;
+			}
+			
+			var lineNumberStart = element.getAttribute('data-line-numbers-start') || 0;
+			var lineNumberRows = element.querySelector('.line-numbers-rows');
+			
+			number = Math.abs(lineNumberStart) + number;
+			
+			return lineNumberRows.children[number];
+		}
+	};
 
 }());
--- a/plugins/line-numbers/prism-line-numbers.js
+++ b/plugins/line-numbers/prism-line-numbers.js
@@ -22,14 +22,13 @@
 	 */
 	var _resizeElement = function (element) {
 		var codeStyles = getStyles(element);
-		var code = element.querySelector('code');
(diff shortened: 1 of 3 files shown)


plugins/line-numbers/prism-line-numbers.js
```diff
@@ -9,20 +9,27 @@
...
+	var NEW_LINE_EXP = /\n(?!$)/g;
 
 	/**
 	 * Resizes line numbers spans according to height of line of code
-	 * @param  {Element} element <pre> element
+	 * @param {Element} element <pre> element
 	 */
 	var _resizeElement = function (element) {
 		var codeStyles = getStyles(element);
+		var code = element.querySelector('code');
```
Golmote: This is redundant with `codeElement` declared a few lines below.

plugins/line-numbers/prism-line-numbers.js
```diff
@@ -102,13 +109,37 @@
 		lineNumbersWrapper.className = 'line-numbers-rows';
 		lineNumbersWrapper.innerHTML = lines;
 
-		if (pre.hasAttribute('data-start')) {
-			pre.style.counterReset = 'linenumber ' + (parseInt(pre.getAttribute('data-start'), 10) - 1);
+		if (pre.hasAttribute('data-line-numbers-start')) {
```
Golmote: I understand why you want to rename this data attribute... but it looks quite BC-breaking to me. We would need to update the docs, etc.

I would suggest you do the exact opposite: rename the data attribute used in line-highlight. AFAIK it's only used for the CSS ;)

Golmote:
No need to be sorry at all. And a huge thanks for diving into this!

VitaliyR:
@Golmote Thanks! PR fixed. I've revert rename and, also, found that we can even don't rename it at all - anyway if these plugins used together, line highlight won't need `data-start` attribute at all. 

And thanks for showing me that silly mistake, removed it.

VitaliyR:
So, guys, what do you say? Should we wait two years for this one? :D

plugins/line-numbers/prism-line-numbers.js
```diff
@@ -110,5 +116,29 @@
...
+			if (element.tagName !== 'PRE' || !element.classList.contains(PLUGIN_CLASS)) {
+				return;
+			}
+			
+			var lineNumberStart = element.getAttribute('data-start') || 0;
+			var lineNumberRows = element.querySelector('.line-numbers-rows');
+			
+			number = Math.abs(lineNumberStart) + number;
+			
+			return lineNumberRows.children[number];
```
Golmote: @VitaliyR I think there is something wrong here. `children` is zero-indexed, while `number` (a value from `data-line`) is one-indexed. I tested in local and the highlighted line in indeed shifted down by one when line-numbers is used.

VitaliyR:
@Golmote thanks for finding! Indeed, I've encountered a bug. So, I've pushed fixes for it, which are also includes other bugfixes related to compatibility issues between this two plugins. 

Also, I've created a hook for line-numbers - in the case, when line highlight is loaded before line numbers. It could be if user manually connects <script> tag of line highlight before line numbers, or if the user has downloaded Prism from the website builder, where line numbers are added after line highlight :)

Golmote:
> P.S. Sorry for long response

I can't blame you, can I? ^^

Is the hook required? Not that I'm particularly against it, but since the plugin already runs in the `complete` hook and the highlighting occurs when the DOM is ready... do you have a use case where the `complete` hook can run before both plugins are loaded?

VitaliyR:
@Golmote haha :) 
Yes, pay attention at http://prismjs.com/download.html when you're selecting both plugins - line highlight is first in the JS source code textarea. So, it will be loaded & parsed & added to complete hook firstly - before line numbers. So it will ask line numbers plugin for getting line number before the plugin will do the job - and here is a problem.

Golmote:
Looks good to me! Thanks for working on this!

PR #1201, opened 2017-10-10
Rename Jade -> Pug

jjcm:
Jade was renamed Pug a couple years back. 
https://github.com/pugjs/pug#rename-from-jade

This is just to update that so you can use language-pug to identify pug, instead of language-jade which is the deprecated name.


Golmote:
Thank you for contributing!

PR #1205, opened 2017-10-12
Add Python builtin function highlighting

erinwild:
List of builtin functions comes from: https://docs.python.org/2.7/library/functions.html


Golmote:
Hi @erinwild! Thanks for contributing.
Would you mind adding/updating the tests for this (see http://prismjs.com/test-suite.html and existing tests: https://github.com/PrismJS/prism/tree/gh-pages/tests/languages/python).
Also, please run `gulp` on your local version to generate the minified version of the component.

Golmote:
While you're at it, could you also add the `nonlocal` keyword, requested in #1208 ?

erinwild:
@Golmote Done!

PR #1206, opened 2017-10-13
Reset Clipboard to undefined if native code

### LINKED ISSUE #1181: Uncaught TypeError: Illegal constructor Error

mAAdhaTTah:
Chrome now ships with a window.Clipboard built in, which breaks
this check. This ensures we're not grabbing the built-in
Clipboard and definitely have the correct package version.

Fixes #1181.


nghuuquyen:
i waiting this PR, i have error on my product because this problem

mAAdhaTTah:
@Golmote How do you feel about this approach?

nghuuquyen:
@mAAdhaTTah Hi, Currently my solution for my product is remove your copy-to-clipboard plugin, so sad :( . But i hope it will work well later

Golmote:
Well, I can't say it's pretty. If it fixes the issue, that's ok for me, though. Hopefully this will be a temporary fix.

mAAdhaTTah:
I'm inclined to doubt it'll be temporary, as I haven't seen any movement from the developer on the Clipboard.js repository. I do think it'll be stable, so I don't think we'll need to remove it.

Issue #1223, opened 2017-11-17
Coy theme: height CSS tag creates design issues.


FlipperPA:
The `height=100%` CSS tag in the coy theme - which doesn't seem to exist in other theme CSS definitions - is causing the code block to render as the browser's height:

Broken:

![image](https://user-images.githubusercontent.com/68164/32969259-5fdfb972-cbb2-11e7-9fad-ea1bd349eb78.png)

Fixed:

![image](https://user-images.githubusercontent.com/68164/32969299-8974d8ee-cbb2-11e7-88c1-068ebf30d5c6.png)

Since it is such a minor change, I will propose a code change via GitHub to remove this single line from the Coy theme. Thanks for your efforts, they are widely appreciated!

Golmote:
I'm not sure what it was for... But I couldn't find any explicit browser issue without the `height` property, so I guess it's ok to remove it. Thanks!

PR #1224, opened 2017-11-17
The 'height' element makes code blocks the height of the browser canvas.

FlipperPA:
Removing the height element - which does not seem to be in other themes - fixes this issue. See https://github.com/PrismJS/prism/issues/1223 for details and screenshots. Thanks!


Golmote:
Thanks for contributing!

PR #1227, opened 2017-11-20
Unicode characters in CSS properties

valtlai:
Highlight Unicode characters in CSS properties correctly, for example `.foo { --bär: 3 }`.

Optimized the “function” regex in the JavaScript definition: replaced `_a-zA-Z0-9` with `\w`, removed uppercase `A-Z` because the `i` flag is set.

Used that regex for CSS but removed `$` and added `-`.


Golmote:
Hi! Thanks for contributing. Could you run `gulp` to generate the minified versions (and the prism.js file) please?

Also, can you update the [test file for CSS properties](https://github.com/PrismJS/prism/blob/gh-pages/tests/languages/css/property_feature.test) with an example using unicode characters, and make sure it passes? You can use `npm test` to run the tests.

valtlai:
Do I need to install Gulp and NPM, eh?

valtlai:
Added a test (and fixed indentation).

Golmote:
> Do I need to install Gulp and NPM, eh?

Yes, indeed.

valtlai:
I have not use them ever. When I have installed both do I just `cd` to the local directory and run the commands?

Golmote:
You need to `cd` to the local directory and run `npm install`.
Once it's done, you should be able to run `gulp` and `npm test`.

valtlai:
Added minified versions, and tests passed.

Golmote:
Awesome thanks! Can you add the `prism.js` to the PR? It should have been modified when you ran `gulp`.

Golmote:
Great! Looks good to me. Thanks again for contributing!

PR #1229, opened 2017-11-20
Optimize more Unicode regexes

valtlai:
In the previous commit, #1227, a “function” regex JavaScript definition was updated. The same regex is used at few other points, and this PR optimizes them too.

Tests are ran and passed.


Golmote:
Perfect! Thanks!

PR #1230, opened 2017-11-22
Allow any element being root instead of document.

tanakahisateru:
Just like GitHub Markdown preview, some web apps would use a certain element that acts like as root `document`. They rebuild DOM which contains some source codes into the root-like element without reloading whole document.

This way looks better than #397 for compatibility. Remember this pull-request created 3 yeas ago, please.


Golmote:
Yes! Thanks for contributing!

PR #1242, opened 2017-12-06
upgrade autoloader

bazooka07:
Hi Golmote,

Two points for missing components with **autoloader** plugin : 
* the data-autoloader-path attribute for the <script> tag is required in some cases. Documentation already exists at http://golmote.free.fr/prism-autoloader/plugins/autoloader/
* It's better if the languages_path in config object is relative to the src attribute of the calling script than to the document root of a web site.


plugins/autoloader/prism-autoloader.js
```diff
@@ -10,8 +10,19 @@
...
+	var script = document.getElementsByTagName('script');
+	script = script[script.length - 1];
+	var languages_path;
+	if(script.hasAttribute('data-autoloader-path')) {
+		var path = script.getAttribute('data-autoloader-path').trim();
+		if(path != '') {
+			languages_path = path;
+		}
+	} else {
+		path = script.src.replace(/[\w\-]+\.js$/, 'components/');
```
Golmote: Shouldn't you set `languages_path` here instead?
Also, you don't need to escape `\-` in that case.

plugins/autoloader/prism-autoloader.js
```diff
@@ -10,8 +10,19 @@
...
+	}
 	var config = Prism.plugins.autoloader = {
-		languages_path: 'components/',
+		languages_path: path,
```
Golmote: Here too, I think you meants `languages_path`.

plugins/autoloader/prism-autoloader.js
```diff
@@ -10,8 +10,19 @@
...
 	var ignored_language = 'none';
 
+	var script = document.getElementsByTagName('script');
+	script = script[script.length - 1];
+	var languages_path;
+	if(script.hasAttribute('data-autoloader-path')) {
+		var path = script.getAttribute('data-autoloader-path').trim();
+		if(path != '') {
+			languages_path = path;
+		}
```
mAAdhaTTah: If this condition isn't met, we'd end up with `languages_path` as `undefined`. Should we have a fall back / default for `languages_path`?

bazooka07:
Yes, you are right. I fix my PR
```
    var script = document.getElementsByTagName('script');
    script = script[script.length - 1];
    var languages_path = 'components/';
    if(script.hasAttribute('data-autoloader-path')) {
        var path = script.getAttribute('data-autoloader-path').trim();
        if(path.length > 0 && !/^[a-z]+:\/\//i.test(script.src)) {
            languages_path = path.replace(/\/?$/, '/');
        }
    } else if (/[\w-]+\.js$/.test(script.src)) {
        languages_path = script.src.replace(/[\w-]+\.js$/, 'components/');
    }
    var config = Prism.plugins.autoloader = {
        languages_path: languages_path,
        use_minified: true
    };
```
@mAAdhaTTah 
If the test is broken, we fallback into the initial value.

@Golmote 
We rejected data-autoloader-path beginning with "http://", ... for security reason.
a '/' is added at the end if missing.

Golmote:
You have conflicts to resolve now @bazooka07 
Apart from this, the PR looks to me. I'll wait for your approval too, @mAAdhaTTah, since you're around.

mAAdhaTTah:
LGTM. Just need to rebuild the `autoloader.min.js`.

bazooka07:
I'm not an expert about Git and I don't know how to resolve this conflict.
Somebody updates the gh-pages branch just an hour before I pushed my pull-request.

So I do the following commands :
```
git checkout gh-pages
git pull upstream gh-pages
git checkout autoloader-path1
git rebase gh-pages
```
But at last, I have a conflict with prism-autoloader.min.js. Even, if I launch gulp.

Can somebody help me ?
If not, I have to create a new branch.

Golmote:
Don't worry it's ok. I solved the conflict locally. Thanks for contributing!

PR #1244, opened 2017-12-07
Previewers: New plugin combining previous plugins

### LINKED ISSUE #913: Previewer plugins should be merged into one
CC @Golmote 

Right now there are too many separate Previewer plugins. This is disruptive, both for people who want them, and for people who don't want them.
For the former, they need to check each one individually and if a new one is added, they do not get the update because it's a separate plugin. I think it's more common for people to want more than one previewer than for them to only want one.
For the latter, they have to scroll through a long list of previewers to find the plugins they actually need. It's a lot of visual clutter.
My proposal is: Merge them into one “CSS Previewers" plugin, with a way to turn certain previewers off globally. That way, you might even be able to reduce the amount of total code, as some of it will be shared across previewers. Thoughts?

Golmote:
Combines all those plugins into one single plugin: Previewer: Base, Previewer: Angle, Previewer: Color, Previewer: Easing, Previewer: Gradient and Previewer: Time.

Fix #913


LeaVerou:
Yay!! 🎉 
You rock!

PR #1245, opened 2017-12-11
Add support for 6502 assembly

kzurawel:
This PR adds basic support for 6502 assembly highlighting as `asm6502`.


components/prism-asm6502.js
```diff
@@ -0,0 +1,28 @@
+Prism.languages.asm6502 = {
+	'comment': /;[^\n]*/,
```
Golmote: Is there a reason to explicitely allow `\r` here? Can't you use `;.*`?

components/prism-asm6502.js
```diff
@@ -0,0 +1,28 @@
...
+	'directive': {
+		pattern: /\.\w+ /,
```
Golmote: Is it intended to consume the final space here? If not, you might want to consider using a positive look-ahead assertion instead `(?= )`.

components/prism-asm6502.js
```diff
@@ -0,0 +1,28 @@
...
+		alias: 'keyword'
+	},
+	'string': /(["'`])(?:\\.|(?!\1)[^\\\r\n])*\1/,
+	'opcode': {
+		pattern: /\b(adc|and|asl|bcc|bcs|beq|bit|bmi|bne|bpl|brk|bvc|bvs|clc|cld|cli|clv|cmp|cpx|cpy|dec|dex|dey|eor|inc|inx|iny|jmp|jsr|lda|ldx|ldy|lsr|nop|ora|pha|php|pla|plp|rol|ror|rti|rts|sbc|sec|sed|sei|sta|stx|sty|tax|tay|tsx|txa|txs|tya|ADC|AND|ASL|BCC|BCS|BEQ|BIT|BMI|BNE|BPL|BRK|BVC|BVS|CLC|CLD|CLI|CLV|CMP|CPX|CPY|DEC|DEX|DEY|EOR|INC|INX|INY|JMP|JSR|LDA|LDX|LDY|LSR|NOP|ORA|PHA|PHP|PLA|PLP|ROL|ROR|RTI|RTS|SBC|SEC|SED|SEI|STA|STX|STY|TAX|TAY|TSX|TXA|TXS|TYA)\b/,
```
Golmote: Please make the group non-capturing her, since you don't need the capture.

components/prism-asm6502.js
```diff
@@ -0,0 +1,28 @@
...
+	'hexnumber': {
+		pattern: /#?\$[\da-fA-F]{2,4}/,
+		alias: 'string'
+	},
+	'binarynumber': {
+		pattern: /#?%[01]+/,
+		alias: 'string'
+	},
+	'decimalnumber': {
+		pattern: /#?[\d]+/,
```
Golmote: `[\d]` can be simplified to `\d`.

kzurawel:
Thanks for the feedback, I think I've addressed everything.

Golmote:
LGTM, thanks!

PR #1246, opened 2017-12-11
Add support for IchigoJam BASIC

BlueCocoa:
Hi. This PR adds support for [IchigoJam](https://ichigojam.net/), which is a board that using its own BASIC language, so I was not extending this from `basic`, but submit as a standalone language.

If there're anything necessary to change, please reply to me and I'll respond to you ASAP.


components.js
```diff
@@ -261,6 +261,10 @@ var components = {
 			"title": "HTTP",
 			"owner": "danielgtaylor"
 		},
+		"ichigojam": {
+			"title": "IchigoJam",
+			"owner": "Cocoa Oikawa"
```
Golmote: Please set `owner` to your exact Github username, so that we can link to your page from the Prism download page.

components/prism-ichigojam.js
```diff
@@ -0,0 +1,15 @@
+// according to the offical reference https://ichigojam.net/IchigoJam-BASIC-reference.pdf
```
Golmote: Is there an english version of this reference? If there is we should add it to this comment. If there is not, could you just add `(JP)` somewhere in the comment?

components/prism-ichigojam.js
```diff
@@ -0,0 +1,15 @@
...
+Prism.languages.ichigojam = {
```
Golmote: The only differences between IchigoJam BASIC and standard BASIC, AFAICT, lie in the `keyword` and `function` patterns. You should really extend the original `basic` component here.

components/prism-ichigojam.js
```diff
@@ -0,0 +1,15 @@
...
+	"function": /\b(?:ABS|ACCESS|ACOS|ANA|ANGLE|AREA|ARITHMETIC|ARRAY|ASC|ASIN|ASK|AT|ATN|BASE|BEGIN|BIN|BREAK|BTN|CAUSE|CEIL|CHR|CLIP|COLLATE|COLOR|CON|COS|COSH|COT|CSC|DATE|DATUM|DEBUG|DEC|DECIMAL|DEF|DEG|DEGREES|DELETE|DET|DEVICE|DISPLAY|DOT|ELAPSED|END|EPS|ERASABLE|EXLINE|EXP|EXTERNAL|EXTYPE|FILETYPE|FIXED|FP|FREE|GO|GRAPH|HANDLER|HELP|HEX|I2CR|I2CW|IDN|IMAGE|IN|INKEY|INT|INTERNAL|IP|IS|KEYED|LANG|LBOUND|LCASE|LEFT|LEN|LENGTH|LINE|LINES|LOG|LOG10|LOG2|LTRIM|MARGIN|MAT|MAX|MAXNUM|MID|MIN|MISSING|MOD|NATIVE|NUL|NUMERIC|OF|OPTION|ORD|ORGANIZATION|OUTIN|OUTPUT|PEEK|PI|POINT|POINTER|POINTS|POS|PROGRAM|PROMPT|RAD|RADIANS|RANDOMIZE|RECORD|RECSIZE|RECTYPE|RELATIVE|REMAINDER|REPEAT|REST|RETRY|REWRITE|RIGHT|RND|ROUND|RTRIM|SAME|SEC|SELECT|SEQUENTIAL|SCR|SET|SETTER|SGN|SIN|SINH|SIZE|SKIP|SOUND|SQR|STANDARD|STATUS|STR|STREAM|STYLE|TAB|TAN|TANH|TEMPLATE|TEXT|THERE|TICK|TIME|TIMEOUT|TRACE|TRANSFORM|TRUNCATE|UBOUND|UCASE|USE|USR|VAL|VARIABLE|VER|VIEWPORT|VPEEK|WHEN|WINDOW|WITH|ZER|ZONEWIDTH)(?:\$|\b)/i,
```
Golmote: Please use single quotes around `'function'`, only for consistency.

BlueCocoa:
Hi, I've made some modifications and I decided to keep this as a standalone language after digging into its [docs](https://ichigojam.net/IchigoJam-en.html). Because IchigoJam uses a very small set of ```basic``` and introduces its own markers.

components/prism-ichigojam.js
```diff
@@ -0,0 +1,11 @@
+// according to the offical reference (EN)
+// https://ichigojam.net/IchigoJam-en.html
+Prism.languages.ichigojam = {
+	'string': /"(?:""|[!#$%&'()*,\/:;<=>?^_ +\-.A-Z\d])*"/i,
+	'comment': /\b(?:'|REM)(?:.*)?/i,
```
Golmote: I'm a bit confused here.
I kinda get that you want the have a word boundary _before_ the `REM`, but are you sure you want to remove the one that was _after_? This would make `REMABC` be highlighted as a comment, is it intended?

Also, `.*` is already a pattern that can match nothing, you don't need to make it optional with `?`.

Please double check this pattern, and see if something like `/(?:'|\bREM\b).*/i` wouldn't be more appropriate.

components/prism-ichigojam.js
```diff
@@ -0,0 +1,11 @@
...
+	'number': /(?:(?:\b|\B[.-])(?:\d+\.?\d*)(?:E[+-]?\d+)?|(?:(?:\B`)(?:0|1)+))/i,
```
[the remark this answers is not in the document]
Golmote: I see the documentation also refers to hexadecimal numbers. Don't you want to highlight them too?

components/prism-ichigojam.js
```diff
@@ -0,0 +1,11 @@
...
+	'keyword': /\b(?:BEEP|BPS|CASE|CLEAR|CLK|CLO|CLP|CLS|CLT|CLV|CONT|COPY|ELSE|END|FILE|FILES|FOR|GOSUB|GSB|GOTO|IF|INPUT|KBD|LED|LET|LIST|LOAD|LOCATE|LRUN|NEW|NEXT|OUT|RIGHT|PLAY|POKE|PRINT|PWM|REM|RENUM|RESET|RETURN|RTN|RUN|SAVE|SCROLL|SLEEP|SRND|STEP|STOP|SUB|TEMPO|THEN|TO|UART|VIDEO|WAIT)(?:\$|\b)/i,
+	'function': /\b(?:ABS|ANA|ASC|BIN|BTN|DEC|END|FREE|HELP|HEX|I2CR|I2CW|IN|INKEY|LEN|LINE|PEEK|RND|SCR|SOUND|STR|TICK|USR|VER|VPEEK|ZER)(?:\$|\b)/i,
+	'operator': /<[=>]?|>=?|[+\-*\/^=&]|\b(?:AND|EQV|IMP|NOT|OR|XOR)\b/i,
```
Golmote: The reference you linked seems to enumerate existing operators. Since you make a full component for IchigoJam, you might want to make a specific list and not just copy-paste the one from `basic`.
For example, the reference mentions `~` but does not seem to have `EQV` or `IMP`. Please double check this.

components/prism-ichigojam.js
```diff
@@ -0,0 +1,11 @@
...
+	'punctuation': /[,;:()]/
```
Golmote: The reference you linked mentions array variables using `[]` notation. You might want to add those chars to the `punctuation` pattern.

components/prism-ichigojam.js
BlueCocoa: Yes, based on its docs and real code, commands starting with ```REM```  is all comments, no space required.

And the regex ```(?:.*)?``` is my mistake, I'll correct it in the next commit.

components/prism-ichigojam.js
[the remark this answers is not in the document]
BlueCocoa: I'm so sorry for the regexp, I'm not quite familiar with it. And thank you for your advices, I'm going to change it to the pattern below.
```/(?:\b|\B[.-])(?:[0-9A-F]+\.?[0-9A-F]*)(?:E[+-]?[0-9A-F]+)?|(?:\B`[01]+)/i```

components/prism-ichigojam.js
BlueCocoa: Yes, I forget the punctuation part and I'll just fix it in the next commit.

Thank you for your time on checking all these mistakes and providing the detailed suggestions.

components/prism-ichigojam.js
BlueCocoa: Okay, I'm going to do a double check for all the parts before the next commit.

Golmote:
@BlueCocoa Nice! Looks good enough to me, I'm merging. Thanks for contributing!

PR #1251, opened 2017-12-23
add Io syntax

AlesTsurko:


components/prism-io.js
```diff
@@ -0,0 +1,31 @@
...
+        {
+            pattern: /(^|[^\\])\/\*[\s\S]*?(?:\*\/|$)/,
+            lookbehind: true
+        },
+        {
+            pattern: /(^|[^\\:])\/\/.*/,
+            lookbehind: true
+        },
+        {
+            pattern: /(^|[^\\:])#.*/,
```
Golmote: I think you can remove the `:` in the lookbehind as it is (was?) used to prevent URLs to match `//` comments (not even sure it's useful anymore).

components/prism-io.js
```diff
@@ -0,0 +1,31 @@
...
+            lookbehind: true
+        }
+    ],
+    'triple-quoted-string': {
+        pattern: /"""(?:\\.|(?!""")[^\\])*"""/,
```
Golmote: Are you sure line terminators can't be escaped? Cause if they are you need to replace `\\.` with `\\[\s\S]`. Please double check.

components/prism-io.js
```diff
@@ -0,0 +1,31 @@
...
+        greedy: true,
+        alias: 'string'
+    },
+    'string': {
+        pattern: /"(?:\\.|(?!")[^\\\r\n])*"/,
```
Golmote: You can remove the negative look-ahead if you include `"` in the negated character class.

components/prism-io.js
```diff
@@ -0,0 +1,31 @@
...
+    'builtin':/\b(?:Array|AudioDevice|AudioMixer|Block|Box|Buffer|CFunction|CGI|Color|Curses|DBM|DNSResolver|DOConnection|DOProxy|DOServer|Date|Directory|Duration|DynLib|Error|Exception|FFT|File|Fnmatch|Font|Future|GL|GLE|GLScissor|GLU|GLUCylinder|GLUQuadric|GLUSphere|GLUT|Host|Image|Importer|LinkList|List|Lobby|Locals|MD5|MP3Decoder|MP3Encoder|Map|Message|Movie|Notification|Number|Object|OpenGL|Point|Protos|Regex|SGML|SGMLElement|SGMLParser|SQLite|Server|Sequence|ShowMessage|SleepyCat|SleepyCatCursor|Socket|SocketManager|Sound|Soup|Store|String|Tree|UDPSender|UPDReceiver|URL|User|Warning|WeakLink|Random|BigNum|Sequence)\b/,
+    'boolean': /\b(?:true|false|nil)\b/,
+    'number': /\b-?(?:0x[\da-f]+|\d*\.?\d+(?:e-?\d+)?)\b/i,
+    'operator': /==|\!=|\*=|\/=|%=|\+=|-=|\^=|>>?=?|<<?=?|&=|\|=|:?:?=|\+\+?|--?|\*\*?|\/\/?|%|\|\|?|&&?|return|and|or|not|@@?|\?\??|\.\./,
```
Golmote: I think some of those could be combined. Please try to group them by starting character (like `\/[=/]?` for example).
Also you should probably add word boundaries around the keyword operators (`\b(?:return|and|or|not)\b`)

.gitignore
```diff
@@ -2,3 +2,4 @@ hide-*.js
 node_modules
 .idea/
 .DS_Store
+package-lock.json
```
Golmote: Please don't include this change in this PR. It is irrelevant to the addition of the support for Io.

components/prism-io.js
```diff
@@ -0,0 +1,31 @@
+Prism.languages.io = {
+    'comment': [
```
Golmote: I'm reviewing from my phone so I can't actually check, but can you make sure you're using tabs for indentation in this file please?

package.json
```diff
@@ -43,4 +43,4 @@
       "prism.js"
     ]
   }
-}
\ No newline at end of file
+}
```
Golmote: Please don't include this change in this PR. It is irrelevant to the addition of the support for Io.

Golmote:
LGTM, thanks!

Issue #1253, opened 2017-12-27
C++ highlighting doesn't handle raw string literal correctly


lshamis:
For example:
```c++
std::string Func() {
  return R"raw(
# 0: Thing 0
# 1: Thing 0
)raw";
}
```

zeitgeist87:
I created a PR #1254 that fixes this issue.

zeitgeist87:
@lshamis Thank you for reporting this issue. It should be fixed now.

PR #1254, opened 2017-12-27
Add the C++11 raw string feature to the cpp language

zeitgeist87:


zeitgeist87:
This should fix issue #1253

components/prism-cpp.js
```diff
@@ -9,4 +9,12 @@ Prism.languages.insertBefore('cpp', 'keyword', {
 		pattern: /(class\s+)\w+/i,
 		lookbehind: true
 	}
+});
+
+Prism.languages.insertBefore('cpp', 'string', {
+	'raw-string': {
+		pattern: /R"([^()\ ]{0,16})\([\s\S]*\)\1"/,
```
Golmote: There is no need to escape the space char in the char class, is there?
Also, the part in the parentheses looks greedy: does this mean `R"(foo)"R"(bar)"` is a single string?

zeitgeist87: Hi @Golmote, thanks for the review. Yes you are correct. My intention was not to escape the space char, but to exclude the backslash. It should be `\\` instead of a single `\`. You are also correct in that `R"(foo)"R"(bar)"` should not be a single string. I guess my regex skills are getting a bit rusty 😄 I'll fix it...

Golmote:
@zeitgeist87 Great! Looks good to me now. 👍

PR #1256, opened 2017-12-30
pkg: add attribute `style`

ZhouHansen:


mAAdhaTTah:
What's the purpose / need for this

ZhouHansen:
I'm using a browserify style transform [sheetify](https://github.com/stackcss/sheetify), now:
```js
var css = require('sheetify')
css('prismjs/themes/prism.css')
```
after add attribute `style`:
```js
var css = require('sheetify')
css('prismjs')
```
[sheetify](https://github.com/stackcss/sheetify) "resolves" module that targets CSS instead of JavaScript by the `style` attribute.

Golmote:
The change looks quite harmless. Even if it does not address the different themes, it can be a nice shortcut to access the default theme.

Issue #892, opened 2016-02-15
Unescaped markup plugin


LeaVerou:
As discussed in #887 .

I just pushed https://github.com/PrismJS/prism/commit/07d77e53a4b78dd2c26c744d13c9f0cefa302728 but it’s still hidden (i.e. not in components.js)
URL: http://prismjs.com/plugins/unescaped-markup/
I didn’t do a PR so that it has a public URL and it’s hidden anyway if I don’t add it to components.js (and because I was being a bit lazy).
Does it look good to everyone? Would “raw markup” be a better name? Unescaped markup is kinda long, albeit descriptive.

ping @zeitgeist87 @Golmote

zeitgeist87:
Hi @LeaVerou,

Great plugin! Here are my thoughts on it. First of all there seem to be a few minor whitespace issues in the `before-highlight` hook, where you mixed tabs and spaces a bit inconsistently. Secondly, Is there any reason to limit the plugin to the `markup` language? Lastly, the title of `index.html` is still `Keep Markup`.

zeitgeist87:
What do you think about #893. I know it is wrong to abuse HTML-comments like this, but it works very well.

zeitgeist87:
`innerHTML` seems to work with `<template>`-tags. No need for cloning.

LeaVerou:
Huh, you’re right! How did I miss that? `innerHTML` does indeed work! I feel so silly now.
We should still provide the other options (script and the comment method), since `template` only works in modern browsers.

LeaVerou:
What syntax would you suggest for `<template>`? We cannot be processing every template tag on the page, but also, we don't want people to have to add a language class to every template they use for this.
What about `<template>` inside `<pre>`?

zeitgeist87:
> What syntax would you suggest for `<template>`?

Puh, I don't know.

> What about `<template>` inside `<pre>`? But that is kinda verbose. Alternatively, template with either a language class or a code class?

We could allow both...

Golmote:
I agree that this should not be limited to the Markup component.

andrewmur:
Hi, guys restriction of the plugin to markup language only is not enough for me as well.
So i just use plugin script with edited language check statement in the following way and it works for me:

```
Prism.hooks.add('before-sanity-check', function (env) {
		if (!env.language == "markup" && !env.element.className.includes("unescaped")) {
			return;
		}
                ....
```

So the code embedded code block in html looks like 
```
[...]

PR #1265, opened 2018-01-16
Unescaped markup plugin: Make it work with any language

### LINKED ISSUE #1262: Best way to to get unescaped HTML working?
Hello,

I am looking for a good way to use the `class="language-php"` so that I can use HTML snippets without having to escape the characters `<` and `>`.

I have read already https://github.com/PrismJS/prism/issues/572 and understand that most people here think the replacement should be done serverside and not be part of the plugin. I respect that, but I would still like to find a solution.

I found that for the class `language-markup` there exists already a plugin (http://prismjs.com/plugins/unescaped-markup/) but I found that this is only working for the class "language-markup" and unfortunately not for "language-php".

I found another solution  which suggests to put a script tag inside the code tag to show the actual HTML: https://stackoverflow.com/a/27495949/2311074

Is that the recommended way?

And as a suggestion: Wouldn't it be nice if the "unescaped markup" plugin also works for `language-php` since you often have mixed PHP and HTML code?

Golmote:
This PR fixes #1262.
It also follows with what was initially said in #892.
@LeaVerou and @zeitgeist87, could you have a quick look? I can't think of any real issue here, but just to be sure.


zeitgeist87:
Looks good to me. I can't think of any situation where it would be harmful to allow it for other languages.

LeaVerou:
Can't see why it's useful for other languages, but can't see any harm either.
I guess other languages exist that have a lot of angle brackets?

Golmote:
@LeaVerou, @zeitgeist87, thanks!
Yes, there are definitely languages using angle brackets out there. First of all, languages that work with HTML, such as PHP, ASP, Handlebars, Smarty, Twig, JSX, Markdown, Textile, etc. Also, Wiki markup, syntaxes like `vector<int>` in C++ that are quite common, or even `s<foo><bar>` in Perl... and probably other use cases I can't think of. There is a real need for this. :)

iwasherefirst2:
Thank you! That is a great feature!

LeaVerou:
@Golmote Thanks for doing this then!!

PR #1271, opened 2018-01-27
Add C++ platform-independent types

isc30:
* int8_t / uint8_t
* int16_t / uint16_t
* int32_t / uint32_t
* int64_t / uint64_t


Golmote:
Hi! Thanks for contributing. Could you please keep the keywords in alphabetical order?
Also, please run `gulp` to generate the minified version of this file. 
Finally, could you update [the test](https://github.com/PrismJS/prism/blob/gh-pages/tests/languages/cpp/keyword_feature.test) to reflect those additions? You should run `npm test` to make sure they still pass.

isc30:
Hi!
I thought that minifier was running on merge. I'll update the tests + min versions.

Golmote:
Thanks! Please do the same for your other PRs 😉

isc30:
this one is ready

isc30:
I don't understand why `gulp` isn't updating `prism.js` in this PR but it does for the other one: 
https://github.com/PrismJS/prism/pull/1272

please check it before merging

Golmote:
Thanks!
(The `prism.js` file is used on the website, it's a bundle containing only a few components. It does use C-Like but does not use C++, that's why it wasn't changed here.)

PR #1108, opened 2017-03-09
.npmignore: Unignore components.js

papandreou:
For the trick suggested by @apfelbox in https://github.com/PrismJS/prism/issues/593#issuecomment-219856317 to work, we need both of these files to be in the NPM package:

* /components.js
* /tests/helper/components.js

PR #1274, opened 2018-01-29
Update .npmignore

evanplaice:
Trim the fat on the prism.js npm package.

Ignore:
tests/ - not necessary for production
.tgz - used while testing packages w/ `npm pack`
bower.json - not necessary in a NPM package
composer.json - not necessary in a NPM package
components.js - used internally
examples.js - used internally
gulpfile.js - used for dev only
.editorconfig - used for dev only
.gitattributes - used for dev only
.travis.yml - used for dev only

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -4,13 +4,23 @@
 hide-*.js
 
 CNAME
-img
-templates
+img/
+templates/
+tests/
+*.tgz
 *.html
 style.css
 favicon.png
 logo.svg
+bower.json
+composer.json
+components.js
 download.js
+examples.js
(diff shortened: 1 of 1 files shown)


Golmote:
Hi! Thanks for contributing.

The only one I'm "worried" about is `components.js`. It is the only file that currently holds the information on which component depends on which other component. And we do know that some users might rely on it to get the list of dependencies.

There was an explicit request for this in https://github.com/PrismJS/prism/pull/1108.

evanplaice:
OK. I'll update the PR as soon as I get back to my computer.

PR #1016, opened 2016-08-24
Add support for HTTP security headers.

ScottHelme:
I often find myself writing configurations for CSP, HSTS and HPKP. These changes add support for these configurations to Prism.

PR #1275, opened 2018-01-29
Add support for highlighting CSP, HSTS, and HPKP

papandreou:
Adds the missing pieces to https://github.com/PrismJS/prism/pull/1016. I couldn't push to the existing branch, so I'm opening a fresh PR instead.


tests/run-child.js
```diff
@@ -15,7 +15,10 @@ if (argv.language) {
 				}
 				process.send({success: true});
 			} catch (e) {
-				process.send({error: JSON.stringify(e)});
+				process.send({error: JSON.stringify({
+					message: e.message,
+					stack: e.stack
+				})});
```
Golmote: Please don't include this change here and rather make another PR for it.

Golmote:
Hi @papandreou! Thanks for working on this.
From a quick look, this seems quite complete to me. I'll have to give it a try before merging though.

I think running `gulp` should make changes to the Show Language plugin. Could you include them in the PR?

@ScottHelme Let me know if you're ok with this being merged instead of #1016. Note that @papandreou referenced you as the "owner" of the languages. This means we expect you to do some kind of maintenance if future issues occur on these components. Please discuss it if @papandreou should have this responsability instead.

papandreou:
@Golmote, thanks for taking a look! I moved the error propagation thing to a separate PR (https://github.com/PrismJS/prism/pull/1276) and force pushed my branch here.

Also pushed commits that build the minified components and the "Show Language" plugin.

ScottHelme:
All good on my side!

papandreou:
I can help out as well, but @ScottHelme is certainly riding the bleeding edge more than I am :)

Golmote:
Thank you both for contributing!

PR #1276, opened 2018-01-30
Fix passing of non-enumerable Error properties from the child test runner

papandreou:
Before I got an error like this when I had added tests for a language that hadn't yet been added to `components.js`:

```
     Error: the object {
  "uncaught": true
} was thrown, throw an Error :)
```

After this change it turned into:

```
Uncaught Error: Language 'csp' not found.
```


Golmote:
Looks good to me!

PR #1277, opened 2018-01-30
improve typescript with builtins and add tsx

fyrkant:
Improved typescript with some builtins and added tsx.


components.js
```diff
@@ -578,6 +578,10 @@ var components = {
 			"require": "markup",
 			"owner": "Golmote"
 		},
+		"tsx": {
+			"title": "React Tsx", 
+			"require": ["markup", "javascript", "jsx", "typescript"]
```
Golmote: I think there is no need to require `markup` and `javascript` here, since they are already requirements for `jsx`.

Golmote: Also, could you please move the component up so that the alphabetical order is based on the `title` value instead of the key. I know this is kinda counter-intuitive, but it's the way we did with JSX, so let's keep it consistent.

fyrkant:
Ok, I'll remove the tsx stuff from this one and a separate PR for that! 👍

fyrkant:
Removed tsx from this one and opened up a new PR with only tsx!

Golmote:
Thanks for contributing!

PR #1280, opened 2018-02-02
added tsx support

fyrkant:
Added support for tsx (typescript jsx).


fyrkant:
Hmm, should I not include the show-language file?

Golmote:
You should include it, yes.

Golmote:
@fyrkant Could you add an example file for TSX, following the format used in [existing examples](https://github.com/PrismJS/prism/tree/gh-pages/examples)?

Also, could you please move the component up in the `components.js` file, so that the alphabetical order is based on the `title` value instead of the key. I know this is kinda counter-intuitive, but it's the way we did with JSX, so let's keep it consistent.

Finally, you need to merge/rebase the `gh-pages` branch to fix the conflict. Your branch is a few commits behind, and the Show Language plugin has been modified in b08cae5603ac0a594814e7c88e9fa60f001cf00c.

fyrkant:
Ah sorry about the example and components order, you already told me that
once! 😅

fyrkant:
Rebased, fixed `components.js` order and add example file!

components.js
```diff
@@ -506,6 +506,10 @@ var components = {
 			"require": ["markup", "javascript"],
 			"owner": "vkbansal"
 		},
+		"tsx": {
+			"title": "React Tsx",
```
Golmote: Can you write "TSX" in uppercase here?

examples/prism-tsx.html
```diff
@@ -0,0 +1,34 @@
...
+<p>To use this language, use the class "language-tsx".</p>
+
+<h2>Full example</h2>
+<pre><code>import * as React from 'react';
+
+interface IState {
+	clicks: number;
+}
+
+export class Clicker extends React.Component<any, IState> {
```
Golmote: I think you must escape all `<` characters in this file using `&lt;`.

You can ensure the example displays properly by updating your branch, and then changing [this line](https://github.com/PrismJS/prism/blob/gh-pages/examples.js#L9) to test your [Examples](http://prismjs.com/examples.html) page locally.

fyrkant:
Fixed it!

reyronald:
Just wanted to drop by to thank @fyrkant and @Golmote for working on this, and for having the updated version of PrismJS up and runing so quickly!

My user base and myself were waiting for this here https://github.com/refined-bitbucket/refined-bitbucket/issues/129 and we are satisfied with the results! 🙌

PR #1297, opened 2018-02-16
add emacs

JuanCaicedo:
This adds the plugin developed by [akirak/prism-emacs-lisp](https://github.com/akirak/prism-emacs-lisp) for the emacs lisp language. The extension is `emacs`. 

I would like to support also `lisp` and `elisp` since both of those names are used, but I'm not sure what would be the best way to do that.

Let me know if there's anything missing to be able to add this 😀


JuanCaicedo:
I'll try to post an example soon of this working

JuanCaicedo:
You can see some of the highlighting at https://prism-emacs-ogfsnjvwlg.now.sh

components/prism-emacs.js
```diff
@@ -0,0 +1,179 @@
...
+		'boolean': {
+			pattern: primitive("(?:t|nil)"),
+			lookbehind: true
+		},
+		'number': [
+			{
+				pattern: primitive("[-+]?\\d+(?:\\.\\d*)?"),
+				lookbehind: true
+			}
+		],
```
Golmote: Since the array contains only one pattern, it could be be removed.

components/prism-emacs.js
```diff
@@ -0,0 +1,179 @@
...
+		'defvar': {
+			pattern: new RegExp(par + 'def(?:var|const|custom|group)\\s+' + symbol),
+			lookbehind: true,
+			inside: {
+				'keyword': /def[a-z]+/,
```
Golmote: I wonder what the highlighting would be here with a code piece like `(defvar defabc`? Maybe a `^` anchor should be added at the beginning of this regexp.

components/prism-emacs.js
```diff
@@ -0,0 +1,179 @@
...
+				'variable': new RegExp(symbol)
+			}
+		},
+		'defun': {
+			pattern: new RegExp(par + `(?:cl-)?(?:defun\\*?|defmacro)\\s+${symbol}\\s+\\([\\s\\S]*\\)`),
+			lookbehind: true,
+			inside: {
+				'keyword': /(?:cl-)?def\S+/,
```
Golmote: A `^` anchor might be needed here too.

components/prism-emacs.js
```diff
@@ -0,0 +1,179 @@
...
+					lookbehind: true
+				},
+				'punctuation': /[()]/
+			}
+		},
+		'lambda': {
+			pattern: new RegExp(par + `lambda\\s+\\((?:&?${symbol}\\s*)*\\)`),
+			lookbehind: true,
+			inside: {
+				'keyword': /lambda/,
```
Golmote: A ^ anchor might be needed here too.

JuanCaicedo:
@Golmote Thanks for the feedback! I'm having some trouble understanding the tests. I'll upload here with some comments on what I'm having trouble with.

I think the renamings I made will cause the previous comments to appear outdated. I'm adding this here as a reminder that they still need to be addressed

components.js
```diff
@@ -197,6 +197,10 @@ var components = {
 			"title": "Eiffel",
 			"owner": "Conaclos"
 		},
+		"elisp": {
+			"title": "Elisp",
+			"owner": "JuanCaicedo"
```
JuanCaicedo: I'm not sure if I should be the owner, since I didn't develop this. Totally willing to defer to @akirak if they would like to take it over 😀

tests/languages/elisp/string_feature.test
```diff
@@ -0,0 +1,14 @@
...
+----------------------------------------------------
+
+[
+	["string", ["\"\""]],
+	["string", "\"foo\\\r\nbar\""]
+]
+
+----------------------------------------------------
+
+Checks for all keywords.
```
JuanCaicedo: This comment needs to be updated

tests/languages/elisp/string_feature.test
```diff
@@ -0,0 +1,14 @@
+""
+"foo\
+bar"
+
+----------------------------------------------------
+
+[
+	["string", ["\"\""]],
+	["string", "\"foo\\\r\nbar\""]
```
JuanCaicedo: This leads to a failure which I don't understand
```
Token Stream:
[["string",["\"\""]],"\r\n\"foo\\\r\nbar\""]
-----------------------------------------
Expected Token Stream:
[["string",["\"\""]],["string","\"foo\\\r\nbar\""]]
-----------------------------------------
```

JuanCaicedo: My impression is that the second case is not recognized as a string. I'm not sure if it's a problem I'm how I'm writing the test cases or in the implementation

JuanCaicedo: I'm inclined to assume the implementation is right and my testing is wrong. If so, maybe the problem is that listing strings like this isn't valid lisp. Maybe the test tokens could me changed to 
```lisp
(
""
"foo
bar"
)
```

Golmote: Currently, your string pattern in the component allows only unescaped line feeds. https://github.com/PrismJS/prism/pull/1297#discussion_r168692063

components.js
akirak: Thanks for this pull request. 

I don't mind your being owner of the module, but this plugin is experimental, and there are some corner cases where it doesn't render well, like in [this example](https://jingsi.space/post/2017/04/05/organizing-a-complex-directory-for-emacs-org-mode-and-deft/). It's doing complex things, and I don't know how to fix this bug. 

If you still want to accept this patch, you or whoever can be the owner. I will mention this merge in the README of my repo, but I won't be responsible for the plugin.

akirak:
Also note, you may have to add some classes to your theme (CSS) to fully take advantage of it. I have tried it with Dracula theme included in the repo, which is a dedicated effort for the language, but I haven't tested it with any other themes for Prism.

akirak:
@Golmote I'm the original author of the plugin, and I'm sorry for this tedious work. That code was not meant to be merged into the code base of PrismJS. It should be [transpiled using babel](https://github.com/akirak/prism-emacs-lisp/blob/master/package.json), but apparently its resulting code will not be suitable for maintenance by human. Therefore I didn't send a merge request to you. 

I am currently working on another project, and I'm not sure if I will adapt this stuff to the proper plugin format. Thank you for your comments.

akirak:
The name `emacs` relates to a limitation of a blog engine I am using. My blog source contains code blocks in `emacs-lisp` language, but my blog engine somehow splits the name by hyphen, and the resulting class name becomes `emacs`. The plugin was named `emacs` so that it can highlight Emacs Lisp code in my blog, and I didn't consider supporting other Lisp dialects. Perhaps it should have been first named `emacs-lisp`, but I would still have to add an alias `emacs` for use in my blog.

components.js
Golmote: @akirak has a point, here. In Prism, we do expect the component's "owner" to do some kind of maintenance if issues regarding this component arise.
So you should probably keep it with your ownership, @JuanCaicedo.

Golmote:
@akirak No worries. I noticed that some patterns are not handled by the official themes. It's not the first component doing so though, and it's ok if some parts are tagged but not styled. I'll just check the appearance once the PR is ready to be merged and maybe we'll make some adjustments.  We might also want to add a comment with a link to your original repo, to credit your original work and in case some users are interested in using the Dracula theme you made.

The transpilation to ES5 should not be that tedious in this case: you're really just using tiny bits of syntactic sugar. Transforming `const` to `var`, fat arrows to regular functions, and template strings to good old string concatenation should be pretty straightforward.

I understand the naming limitation you had. Maybe we can still add `emacs` as an alias, since it seems to be the main scripting language used by this editor.

Thanks for showing up on the PR anyway! I hope @JuanCaicedo can take it to a mergeable state. :)

components/prism-emacs.js
```diff
@@ -0,0 +1,179 @@
...
+	var language = {
+		// Three or four semicolons are considered a heading.
+		// See https://www.gnu.org/software/emacs/manual/html_node/elisp/Comment-Tips.html
+		'heading': {
+			pattern: /;;;.*/,
+			alias: ['comment', 'title']
+		},
+		'comment': /;.*/,
+		'string': {
+			pattern: /"(?:[^"\\]*|\\.)*"/,
```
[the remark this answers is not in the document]
JuanCaicedo: I think multiline feeds are supported, I say that because this is a valid function definition

```elisp
(defun multiply-by-seven (number)
       "Multiply NUMBER by seven,
second line"
       (* 7 number))
```

components.js
```diff
@@ -44,7 +44,12 @@ var components = {
 		},
 		"markup": {
 			"title": "Markup",
-			"aliasTitles": { "html": "HTML", "xml": "XML", "svg": "SVG", "mathml": "MathML" },
+			"aliasTitles": {
+				"html": "HTML",
+				"xml": "XML",
+				"svg": "SVG",
+				"mathml": "MathML"
+			},
```
JuanCaicedo: Lots of changes in this file were made by prettier in my editor. I tired to undo most of them, but I left in these last ones that I think make the file more consistent. Let me know though if you disagree and I can fi these as well

tests/languages/elisp/quoted-symbol_feature.test
```diff
@@ -0,0 +1,17 @@
+(
```
JuanCaicedo: I was unsure about how to name these files where the feature is hyphenated. Let me know if it should be changed

JuanCaicedo:
@Golmote I've made some updates in this PR to address your comments, let me know what you think!

components/prism-elisp.js
```diff
@@ -0,0 +1,192 @@
+// Functions to construct regular expressions
```
Golmote: Could you please wrap the whole file in `(function (Prism) {` and `}(Prism));` please ? That way, we won't pollute the global scope with vars and stuff.

components.js
```diff
@@ -341,6 +350,10 @@ var components = {
 			"require": "css",
 			"owner": "Golmote"
 		},
+		"lisp": {
+			"title": "Lisp",
+			"owner": "JuanCaicedo"
+		},
```
Golmote: You added `elisp` and `lisp` in this file, but not `emacs`.
The general rule here is to only add one entry, the _main_ one. So I guess here it should be `lisp` since it's the most generic. Please remove the `elisp` entry.

That also means the component should be renamed to `prism-lisp`, along with the test folder and so on.

components/prism-elisp.js
```diff
@@ -0,0 +1,192 @@
...
+// simple form
+// e.g. (interactive ... or (interactive)
+function simple_form(name) {
+	return new RegExp("(\\()" + name + "(?=[\\s\\)])")
```
Golmote: Please add semicolons, everywhere needed, in this file. We don't do semicolon-less JS in this repo.

components/prism-elisp.js
```diff
@@ -0,0 +1,192 @@
...
+					space
+			),
+			lookbehind: true,
+		},
+		{
+			pattern: new RegExp(
+				par + "(?:for|do|collect|return|finally|append|concat|in|by)" + space
+			),
+			lookbehind: true,
+		},
```
Golmote: Please remove all trailing commas in the file.

components/prism-elisp.js
```diff
@@ -0,0 +1,192 @@
...
+	],
```
Golmote: Couldn't those two patterns be merged into one?

components/prism-elisp.js
```diff
@@ -0,0 +1,192 @@
...
+		pattern: new RegExp(
+			par +
+				"(?:cl-)?(?:defun\\*?|defmacro)\\s+" +
+				symbol +
+				"\\s+\\([\\s\\S]*?\\)"
+		),
+		lookbehind: true,
+		inside: {
+			keyword: /^(?:cl-)?def\S+/,
+			arguments: null,
```
Golmote: Could you add a comment here, something like `// See below` so that it's obvious this will be filled later on.

components/prism-elisp.js
```diff
@@ -0,0 +1,192 @@
...
+			},
+			punctuation: /[()]/,
+		},
+	},
+	lambda: {
+		pattern: new RegExp(par + "lambda\\s+\\((?:&?" + symbol + "\\s*)*\\)"),
+		lookbehind: true,
+		inside: {
+			keyword: /^lambda/,
+			arguments: null,
```
Golmote: Same here.

components/prism-elisp.js
```diff
@@ -0,0 +1,192 @@
...
+	},
+	car: {
+		pattern: new RegExp(par + symbol),
+		lookbehind: true,
+	},
+	punctuation: [
+		// open paren
+		/['`,]?\(/,
+		// brackets and close paren
+		/[)\[\]]/,
```
Golmote: Those two patterns could definitely be merged into a single disjunction: ``/['`,]?\(|[)\[\]]/``

components/prism-elisp.js
```diff
@@ -0,0 +1,192 @@
...
+	},
+}
+
+language["lambda"].inside.arguments = arglist
+language["defun"].inside.arguments = Prism.util.clone(arglist)
+language["defun"].inside.arguments.inside.sublist = arglist
+
+Prism.languages.elisp = language
+Prism.languages.emacs = language
+Prism.languages.lisp = language
```
Golmote: I would suggest move `lisp` in first position, so that it's made clear it is the main language here.

examples/prism-elisp.html
```diff
@@ -0,0 +1,54 @@
+<h1>Elisp</h1>
+<p>To use this language, use one of the following classes:</p>
+<ul>
+	<li>"language-elisp"</li>
+	<li>"language-lisp"</li>
+	<li>"language-emacs"</li>
+</ul>
```
Golmote: Same thing here: let's use `Lisp` as the main language. And move the class `"language-lisp"` in first position.

tests/languages/elisp/quoted-symbol_feature.test
Golmote: This is perfectly fine. :)

Golmote:
@JuanCaicedo Note that I've just updated the `components.js` file to fix the code style. You'll have to merge `gh-pages` and fix the conflicts.

components.js
paladox: You could add alias for emacs and elisp

paladox:
bump @JuanCaicedo :)

JuanCaicedo:
@paladox Would love to do some work to get this wrapped up, but I've been busy with some other coding project over the past few weeks. I plan to get to it over the next few days 😃

JuanCaicedo:
@Golmote sorry for the delay, I'm working through these changes now 😀

components/prism-elisp.js
JuanCaicedo: @Golmote is this a required change for you or a nice to have? I ask because the tests fail when I switch to the regex you suggested. I think it probably just requires a little tweaking, but working through regexes is challenging for me 😅It seems to me like having it splitting up is clearer, but if you disagree I can work through it

components/prism-elisp.js
JuanCaicedo: I can try to combine these, but I'm finding it challenging. If it's a required change, I'll definitely work through it, but I'm going to push an update without it in case it isn't

components/index.min.js
```diff
@@ -0,0 +1 @@
+function loadLanguages(e){e||(e=Object.keys(components.languages).filter(function(e){return"meta"!==e})),Array.isArray(e)||(e=[e]),e.forEach(function(e){components.languages[e]&&components.languages[e].require&&loadLanguages(components.languages[e].require),require("./prism-"+e)})}var components=require("../components.js");module.exports=loadLanguages;
```
JuanCaicedo: This file was generated when I ran `gulp`. Not sure if it should be included. If not, I can revert the commit that added it

tests/languages/lisp/string_feature.test
```diff
@@ -0,0 +1,29 @@
+(
+""
+"foo
+bar"
+"\"foo\""
+"foo\tbar"
+"foo\tbar"
```
Golmote: Looks like there's a duplicate here. I don't think there's a need to test it twice.

themes/prism.css
```diff
@@ -104,7 +104,7 @@ pre[class*="language-"] {
 .token.url,
 .language-css .token.string,
 .style .token.string {
-	color: #9a6e3a;
+	color: #a67f59;
```
Golmote: This change should not be here. This file was changed 10 days ago in https://github.com/PrismJS/prism/commit/8aea939a6287f0147d1d64d3f0a238b6cadbc9c6. `#9a6e3a` is the correct value.

JuanCaicedo:
@Golmote I've updated based on your last comments 😀

Golmote:
@JuanCaicedo Yay! :tada:

PR #1305, opened 2018-02-22
Prevent scrolling toolbar with content

arxeiss:
If content is wider than screen, during scrolling toolbar is not moving anymore.

image


mAAdhaTTah:
I dig it, and I don't see any complications.

mAAdhaTTah:
@arxeiss Thank you for the contribution!

Golmote:
@mAAdhaTTah @arxeiss Shouldn't we check for the existence of the wrapper so that calling Prism highlighting multiple times on the same block does not add multiple toolbars? (Condition on line 62 suggests this case was handled before)

arxeiss:
@Golmote You are right! I did not see it..
And I found maybe another problem. That if another plugin will need wrapper as well, there can be conflict. Maybe adding some wrapping method to the core? That only 1 wrapper can be present?
Any ideas?

Golmote:
I think it's OK if we don't handle this in a generic way for now, cause we don't need other wrappers.

mAAdhaTTah:
@Golmote Nice catch, and agreed. If we need a wrapper for other plugins, we can develop a generic method then.

PR #1307, opened 2018-02-24
Allow text change to other languages. Plugin: copy to clipboard

AlexTorresSk:
This allows the value of the text to be changed to other languages, css icon, image, etc.

captura

### DIFF
--- a/plugins/copy-to-clipboard/prism-copy-to-clipboard.js
+++ b/plugins/copy-to-clipboard/prism-copy-to-clipboard.js
@@ -1,6 +1,6 @@
-if(typeof textCopy == 'undefined') textCopy = 'Copy';
-if(typeof textCopied == 'undefined') textCopied = 'Copied!';
-if(typeof errorCopied == 'undefined') textCopy = 'Press Ctrl+C to copy';
+if(typeof textCopy == 'undefined') var textCopy = 'Copy';
+if(typeof textCopied == 'undefined') var textCopied = 'Copied!';
+if(typeof errorCopied == 'undefined') var errorCopied = 'Press Ctrl+C to copy';
 
 (function(){
 	if (typeof self === 'undefined' || !self.Prism || !self.document) {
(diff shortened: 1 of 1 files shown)


mAAdhaTTah:
Thanks! This is definitely something I was planning on implementing as I i18n a plugin I'm working on. However, I don't think we should use global variable to handle this but data-attributes instead.

@Golmote @LeaVerou I don't think we want the translation on every `code` or `pre` instance, although maybe that's fine. What about putting it on the `script` tag itself work? Body tag? Where else? Thoughts?

Golmote:
@mAAdhaTTah Could it be inherited, like the `language-XXX` class is? Therefore any ancestor of a `pre` or `code` could provide the translations attributes.

LeaVerou:
Yes, definitely not globals! Imagine having 200 globals in the future, for 200 phrases.
However, the functionality is definitely useful.
Probably the ideal thing would be to inherit data attributes, like @Golmote suggests.
Such attributes should probably have a prefix, like `data-prism-phrase-copy` or `data-prism-phrases` with a list of key-value pairs separated by colons and commas.

AlexTorresSk:
@LeaVerou and @Golmote Yes. No matter how it is implemented, what I was looking for was the same, that can be translated or written in the language that the user decides.

mAAdhaTTah:
@AlexTorresSk Do you have enough information to take a shot at an implementation here? You can see how inheritance is handled in core [here](https://github.com/PrismJS/prism/blob/gh-pages/components/prism-core.js#L178-L187).

mAAdhaTTah:
Hi @AlexTorresSk, thank you for the contribution. We're definitely not going to go with globals for this, so I'm going to close this PR, but we'd love to merge a fix for this, if you'd be interested in taking it on. If you do, please branch from `master`, as we're in the process of deprecating `gh-pages` and moving everything to `master`.

I've opened #1438 to capture this issue, as it's something we definitely want to address. Please let us know if you have any questions if you do pick this up.

Thanks again!

PR #1121, opened 2017-03-27
Add =begin =end comments

Bigsby:

PR #1309, opened 2018-02-27
Remove multi-line comments from Ruby edge case failure examples

benkimball:
In #1121 the failing edge case for Ruby multi-line comments was fixed, but the listing for multi-line comments was not removed from the "Known failures" section of the examples/prism-ruby.html documentation file. This patch does so, since it's no longer a known failure.


Golmote:
Yes indeed! Thanks!

PR #1214, opened 2017-11-01
Add Clojure

troglotit:
Closes #1147 

Copied definitions from https://github.com/jeluard/prism-clojure, example copied from https://learnxinyminutes.com/docs/clojure/

Tests are not so thorough, but they pass, and syntax highlighting looks decent.

PR #1311, opened 2018-02-28
Add Clojure

### LINKED ISSUE #1147: Request for support Clojure
Please support the syntax highlight for [Clojure](https://clojure.org/) language. Thank you.

tkjone:
Closes #1147

Implements the updates requested in #1214 and adds syntax highlighting for:

- thread-first (->) macro
- thread-last (->>) macro
- nil

Regarding the updates, you will notice that each of my commits addresses the comments made by @Golmote in #1214.  Further, each commit should be in the order of the requested fixes.  I apologize for not piggybacking directly off the other PR.

Let me know if there is anything else I can improve upon.


components/prism-clojure.js
```diff
@@ -0,0 +1,13 @@
+// Copied from https://github.com/jeluard/prism-clojure
+Prism.languages.clojure = {
+	comment: /;+.*/,
+	string: /"(?:\\.|[^\\"\r\n])*"/,
+	operator: /(::|[:|'])\b[a-zA-Z][a-zA-Z0-9*+!-_?]*\b/g, //used for symbols and keywords
```
Golmote: Can you make the group non-capturing here? And also remove the `g` flag. You could add the `i` flag to shorten the regexp a bit, and use `\w` in the second character class. Finally, I think the last `-` is supposed to be escaped (or moved at the end).
The whole pattern could be something like:
`/(?:::|[:|'])\b[a-z][\w*+!?-]*\b/i`

paladox:
cc @Golmote  wondering if you could review this please? :)

tests/languages/clojure/boolean_feature.test
```diff
@@ -0,0 +1,13 @@
+false
+true
+
```
Golmote: Could you add a test for `nil` here? Since it's handled in the `boolean` pattern.

tests/languages/clojure/keyword_feature.test
```diff
@@ -0,0 +1,37 @@
+(def)
+(defmacro)
+(if)
+(do)
+(let)
+(var)
+(fn)
+
```
Golmote: Ideally, we should add checks for each and every keyword here.

tests/languages/clojure/keyword_feature.test
```diff
@@ -0,0 +1,37 @@
...
+----------------------------------------------------
+
+[
+	["punctuation", "("],
+  ["keyword", "def"],
```
Golmote: Could you indent those lines with a tab, please?

components/prism-clojure.js
tkjone: Resolved

tests/languages/clojure/boolean_feature.test
tkjone: Resolved

tests/languages/clojure/keyword_feature.test
tkjone: I added about 70 new keywords here - a lot of the ones that were posing issues when I was testing e.g. `<=` and `>=`.  I am happy to fill out the rest, but I have a high degree of confidence that the ones I did not include will pass ;)

tests/languages/clojure/keyword_feature.test
tkjone: Finished 😌

tkjone:
@Golmote Thanks for the review!  Let me know what you think 💪

Golmote:
@tkjone Yay! LGTM!

PR #1314, opened 2018-03-01
Bump ClipboardJS to 2.0.0 and remove hack

mAAdhaTTah:
This is no longer needed, since it no longer conflicts w/ the
built-in global. See https://github.com/zenorocha/clipboard.js/issues/468.


Golmote:
Nice! Finally! LGTM :+1:

PR #1323, opened 2018-03-02
Add shell alias for bash to components.js

paladox:


paladox:
cc @Golmote :)

Golmote:
AFAICT this is not needed. Titles are used by the Show Language plugin, but there are only needed when the title is not the capitalize version of the component's alias.

It already works as intended without this addition. https://codepen.io/anon/pen/Jpzaqm

paladox:
ah ok. thanks.

PR #1345, opened 2018-03-11
Core: Allow cloning of circular structures.

### LINKED ISSUE #1342: Nested script inside jsx tag is not parsed
```javascript
<CodeView title={<span title={123}>2</span>}>for of as</CodeView>
```
image

Golmote:
With this, it will be possible to remove most of the cloning occurrences in language definitions (since it was used mainly to prevent infinite recursion when extending).

It should even be possible to fix #1342, unexpectedly.


Golmote:
@mAAdhaTTah You're right! That's much more clear now.

Issue #972, opened 2016-06-11
Importing all languages in node.js


felixsanz:
This is related to: https://github.com/PrismJS/prism/issues/593

The components are injected when you require the module, so can we have (for the node users) an `index.js` inside `components` folder, that does this?

``` javascript
require './prism-abap.js';
require './prism-actionscript.js';
require './prism-apacheconf.js';
// ...
```

So then we only have to use:

``` javascript
import prismjs from 'prismjs';
import 'prismjs/components';
```

---

Or even something like:

``` javascript
function loadComponents(arr) {
  arr.foreach(function(language) {
    require('./prism-' + language);
  });
}

module.exports = loadComponents;
```

``` javascript
import prismjs from 'prismjs';
import loadComponents from 'prismjs/components';
loadComponents(['php', 'ruby']);
```

This won't work in the browser i think, but why not have this simple helper for the node users?

LeaVerou:
It would be nice to have this indeed, and it can be automatically generated via gulp…

Golmote:
I digged into this last night, and the loadComponent() option works great.

It might even be able to handle dependencies by reading the `components.js` file. But here is the problem: the `components.js` file can not be `require()`'d in Node.js without a `module.exports` at the end. This would break any script that tried to parse the `components.js` file (like we currently do in the `gulpfile`).
Is that a BC break?
Should we add the `module.exports` line to the `components.js` to `require()` it or parse the file?
Should we handle dependencies at all in this helper?

LeaVerou:
Would it help if we split components.js to a components.json which can be easily parsed and then create components.js + module.exports with gulp?

Golmote:
@LeaVerou Indeed that would be nice! But shouldn't it be considered a BC break anyway, since people relying on parsing the `components.js` file would have to change to `components.json`?

LeaVerou:
No, because components.js would still exist, it would just be generated with gulp! :)

Also I don't think that many people are relying on components.js :P

mAAdhaTTah:
@LeaVerou I'm working on an integration that might require access to a `component.js{on}`. I'm bundling with webpack and am looking to use its async loading to load up the languages & their dependencies, rather than using the autoloader, as this would simplify my build setup and allow me to run everything through webpack (including the CSS, which is pretty useful).

[...]

PR #1354, opened 2018-03-12
Core: Generate components.js from components.json and make it exportable to nodeJS

Golmote:
This should allow us to consider fixing #972


mAAdhaTTah:
Does this include the `loadComponents` part of that issue? I'm assuming not, but the file is minified, so I can't see the results.

Golmote:
@mAAdhaTTah No it is not included in this PR. It would probably live in a `components/index.js` file.

PR #1357, opened 2018-03-14
JSX: Add support for plain text inside tags

Golmote:
This is a second attempt at fixing #1294, based on a great idea from @supnate (https://github.com/PrismJS/prism/issues/1342#issuecomment-372570134).

It uses a new hook to post-process the tokenized code and drop tokens that were wrongly highlighted.

Unlike #1304, it should provide full support for plain text, here, since we keep count of the opening tags and braces.

Please give me your thoughts about this. @m-allanson, @LeaVerou, @mAAdhaTTah, @vkbansal


vkbansal:
LGTM. Nice work!

LeaVerou:
Me too! Thank you!

supnate:
Looks good, quite neat code! Thanks @Golmote ! Glad to see my approach could help. Just one comment: could you make jsx text a special token type (like jsx-text) rather than plain text? So that we could differentiate it from other text like variable definition, then they can have different colors.

PR #1358, opened 2018-03-15
Auto-detect command lines if 'data-filter' attribute present

bradhowes:
This change essentially performs the same operation as the existing code which processes the 'data-output' atrribute values, but in an automatic fashion.

When the 'data-filter' attribute is found in the <pre> element with a non-empty value, visit each line of the <code> content. If the line starts with the value found in the 'data-prompt' attribute, treat it as a command and wrap the text in <span class="command-line-command">. Otherwise, treat the line as command output and strip off any 'data-user', 'data-host', or 'data-prompt' attributes, just like the code which performs filter based on line numbers and ranges.


bradhowes:
Pinging @chriswells0 as he is the author of the command-line plugin.

plugins/command-line/prism-command-line.js
```diff
@@ -54,6 +55,26 @@ Prism.hooks.add('complete', function (env) {
 	prompt.className = 'command-line-prompt';
 	prompt.innerHTML = lines;
 
+	var filterContent = pre.getAttribute('data-filter', '');
+	if (filterContent.length > 0) {
```
Golmote: You could use the `hasAttribute` method instead, to allow for empty values or absence of value too.

plugins/command-line/prism-command-line.js
```diff
@@ -54,6 +55,26 @@ Prism.hooks.add('complete', function (env) {
...
+		for (var i = 0; i < content.length; i++) {
+			var line = content[i];
+			if (line.slice(0, promptText.length) == promptText) {
+				// We have a command -- strip off the prompt from the source text and wrap in <span>
+				content[i] = '<span class="command-line-command">' + line.slice(promptText.length + 1) + 
+					'</span>';
```
Golmote: If I understand this correctly, it means users who don't use the `data-filter` attribute won't benefit from the styled `<span>`. Am I right? Couldn't this behaviour be added in every cases?

plugins/command-line/prism-command-line.js
bradhowes: Agreed.

plugins/command-line/prism-command-line.js
bradhowes: I did not want to cause any possibly breaking change, hence the clear separation. I am more than willing to rework it if desired.

chriswells0:
I like the concept for this functionality.  I wonder if a more descriptive name could be used for the `data-filter` attribute to hint at what would be filtered (e.g., `data-filter-prompt`).

I have an open PR #856 from 2 years ago that automatically treats lines beginning with `>` as output, but we never really settled on the char to use as the indicator.  I see that I also have local changes that were never merged into that PR.  Perhaps if this PR used `data-filter-prompt`, we could then use `data-filter-output=">"` to do the reverse: provide a way for users to specify the leading output character or character sequence for automatic output treatment.  Thoughts?

plugins/command-line/prism-command-line.js
Golmote: I think that would be better to have a predictible behaviour when the span is always present, indeed. But yes we need to make sure nothing breaks.

Golmote:
Hi @chriswells0! This sounds like a good idea. If you start working again on #856, feel free to @ me so I can review it.

chriswells0:
My apologies for the delay--been swamped at work during the week.  I just updated #856 with my old changes modified to use `data-filter-output` to specify the prefix for output lines and rebased it.

@bradhowes, I realize this will require changes for your PR, so I was hoping to hear your thoughts before implementing the changes in mine.  I just didn't expect to have more time soon, so I changed it while I could.  I do believe the new approach will make it easier to accomplish what you're doing, though.  Search for `outputFilter` to see that I've split prompt lines / code into one array with blank placeholders where output lines belong and stored the actual output lines in another array.  After highlighting is done, the arrays get merged back together.

bradhowes:
@chriswells0 That sounds good to me. Thanks! I'm fine with closing this as a dupe of your work then.

chriswells0:
@bradhowes, I didn't mean it was a dupe.  It's the reverse, which I think is a useful feature.  My approach specifies the prefix for output lines, so you could use `data-filter-output="out"` (or any value) and it would treat lines beginning with "out" as output.  I see both as useful depending upon whether you have more commands or output in a given code block.

Golmote:
@bradhowes, I merged #856, so you'll have to merge or rebase the `gh-pages` branch in order to address my remaining comment, and eventually get this PR merged.

mAAdhaTTah:
I'm still into this if we can get the conflicts merged.

chriswells0:
Sorry, I had a busy work week and just now had a minute to return to this. @bradhowes, like I said, I can see the benefit of having both options, so it's up to you if you want to merge it.  I appreciate the contribution.

@mAAdhaTTah, in the meantime, you can also look at the `data-filter-output` attribute, which does the reverse of this PR.

PR #1359, opened 2018-03-15
Add a loadComponents function for easy component loading on NodeJS

Golmote:
This PR adds the `loadComponents` function mentionned in #972.

Usage:

```js
var Prism = require('./components/prism-core.js');
var loadComponents = require('./components/index.js');
loadComponents(['c', 'javascript']);
```

### DIFF
--- a/components/index.js
+++ b/components/index.js
@@ -1,0 +1,24 @@
+var components = require('../components.js');
+
+function loadLanguages(arr) {
+	// If no argument is passed, load all components
+	if (!arr) {
+		arr = Object.keys(components.languages).filter(function (lang) {
+			return lang !== 'meta';
+		});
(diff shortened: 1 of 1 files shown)


Golmote:
@mAAdhaTTah Do you wish to do a quick review on this?

mAAdhaTTah:
Maybe `loadLanguages` would be more appropriate as a name, since it can't load other components, like themes or plugins?

Also, we get slightly less code if we just call `load` within the `if` block, e.g.:

```js
if (components.languages[language] && components.languages[language].require) {
	loadComponents(components.languages[language].require);
}
```

Golmote:
Yes ! Both valid points. I updated the PR.

PR #1370, opened 2018-03-26
Highlight .token.generics as the same colour as .token.class-name

paladox:


Golmote:
Hi @paladox. This change should not be needed, because the `generics` token is aliased as `function`, leading to a `<span class="token generics function">`. You should not need to change anything in your theme to get the proper highlighting.

PR #1375, opened 2018-03-31
Fix Swift 'protocol' keyword

ileitch:
Change `Protocol` to `protocol`.

### DIFF
--- a/components/prism-swift.js
+++ b/components/prism-swift.js
@@ -16,7 +16,7 @@ Prism.languages.swift = Prism.languages.extend('clike', {
 			}
 		}
 	},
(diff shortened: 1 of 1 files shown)


Golmote:
Hi and thanks for contributing!
Can you please run `gulp` to generate the minified version of the component and include it in this PR.
Also, you'll have to update the [test file for keywords in Switch](https://github.com/PrismJS/prism/blob/gh-pages/tests/languages/swift/keyword_feature.test#L47).

mAAdhaTTah:
@ileitch Would you be able to make the change suggested by @Golmote? If not, I'd be happy to pick this up. Let me know.

ileitch:
@mAAdhaTTah You're welcome to take it. Prism is also missing the `extension` keyword.

PR #1388, opened 2018-04-12
HTTP: Absolute path is a valid request uri

tstapleton:
The [spec](https://tools.ietf.org/html/rfc2616#section-5.1.2) says an absolute path is valid for the request uri.


Golmote:
Hi! Thanks for contributing! Could you please run `gulp` so that the minified version of the component is generated and include it in this PR please?

tstapleton:
@Golmote done!

Golmote:
Thanks!

PR #1391, opened 2018-04-14
Use rel="icon" instead of rel="shortcut icon"

valtlai:
`<link rel="shortcut icon"` is deprecated and `<link rel="icon"` should be used.

(Also, remove unneeded `(?!\*)` from the `lang` regex in `prism-file-highlight.js`.)

### DIFF
--- a/download.html
+++ b/download.html
@@ -3,7 +3,7 @@
 <head>
 
 <meta charset="utf-8" />
-<link rel="shortcut icon" href="favicon.png" />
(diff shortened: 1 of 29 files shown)


Golmote:
Hi! Thanks for contributing again.
But please don't include both changes in a single PR, instead make two PRs, one for each issue.

valtlai:
New PRs: #1398 and #1399.

PR #1396, opened 2018-04-17
Change reserved keyword reference

aaronfay:
This changes the `bash` component's notation for the `boolean` property to behave with YUI compressor, related to https://github.com/yui/yuicompressor/issues/186.


Golmote:
Alright then! Looks good to me. Thanks for contributing!

Issue #853, opened 2016-01-13
What’s the point of `(?!\*)`?


valtlait:
I’m trying to understand the Prism source code. What’s the point of `(?!\*)` in the following line:

```js
var lang = /\blang(?:uage)?-(?!\*)(\w+)\b/i;
```

I know it’s a negative lookahead. In this regex it means that an asterisk doesn’t follow the hyphen.

But where is it needed?

kylekatarnls:
`foo(?!bar)` mean "take foo if it's not followed with bar". \* is the * symbol. So it takes - if it's not followed with * then a word. So lang-php, language-html, but not lang-*php.

valtlait:
But how that regex differs from one without `(?!\*)`?

```js
/\blang(?:uage)?-(?!\*)(\w+)\b/i.test('lang-*php');  // returns `false`
/\blang(?:uage)?-(\w+)\b/i.test('lang-*php');  // returns `false`
```

kylekatarnls:
Ho, indeed, \w exclude *, you're right it's redundant. What's the line and file? With them we could know who wrote this and ask him/her if there is a reason.

valtlait:
It’s here: [/components/prism-core.js:18](https://github.com/PrismJS/prism/blob/gh-pages/components/prism-core.js#L18)

kylekatarnls:
It seems @LeaVerou added this piece of the RegExp specifically: cfaf681
But there are no differences when I test with examples.html classes.

LeaVerou:
Yeah, I have no idea why I added that, since as you pointed out, `\w` doesn't include `*` anyway. It can be safely removed.

valtlait:
Thanks!

PR #1399, opened 2018-04-18
Simplify lang regex on File Highlight plugin

valtlai:
Similar to #854.


Golmote:
This is a weird one. It seems [I originally introduced](https://github.com/PrismJS/prism/commit/8030db914a1e7331d245db0b9b787b4364994440#diff-c096d2934bc4bc31364a77d74b1a00f1R23) this regexp in this file. My guess is that I copy-pasted it from the file `prism-core` [where it's also used currently](https://github.com/PrismJS/prism/blob/5f133c800fdd178f6899c1ffa8a38ffbdc7a746f/components/prism-core.js#L18).
@LeaVerou It appears [you did this change](https://github.com/PrismJS/prism/commit/cfaf681db9018bd1d6e296b526db3e1467986fb3#diff-be553e8a1b66b66f387cb98326dbae83R10) a long time ago. Would you by any chance remember what this negative look-ahead assertion was for?

valtlai:
@Golmote See https://github.com/PrismJS/prism/issues/853#issuecomment-172139753.

PR #1410, opened 2018-04-27
Don't process language if block language not set

jwworth:
Adding this guard to the function prevents an error if the language is undefined. `env.language.substring(0, 1).toUpperCase()` can fail with `Uncaught TypeError: Cannot read property 'substring' of undefined` when `env.language` is undefined.


mAAdhaTTah:
This shouldn't go there. Right under it, you can see a conditional that checks a number of other things before it falls back to `env.language`. The check should go before that conditional.

jwworth:
Great point! I moved the conditional down, and added a check for `language` to ensure undefined isn't put into the `span`.

Golmote:
No, LGTM too.

PR #1412, opened 2018-04-29
Support for Template Toolkit

gflohr:
This PR adds support for the [Template Toolkit](http://www.template-toolkit.org/).  Template Toolkit is arguably the most popular template language for Perl.

I have successfully used the new component for some time on http://www.qgoda.net (for example on http://www.qgoda.net/en/docs/related-documents/).


components.json
```diff
@@ -651,6 +651,14 @@
...
 			"owner": "Golmote"
 		},
+		"tt2": {
+			"title": "Template Toolkit",
+			"alias": ["tt"],
+			"aliasTitles": {
+				"tt": "Template Toolkit"
+			},
+			"owner": "gflohr"
+		},
```
Golmote: I would suggest using the name `tt` as the main name, since it seems to be the more generic (if I understand correctly, `tt2` would be the second version?). If you do so, you'll need to rename the component file too.

Golmote: Also, starting with Prism 1.14, we made it more simple to craft templating languages using the placeholders `___...___`.
To benefit from this, your component should `"require": "markup-templating"` here. That way, you'll have access to utilities functions to automate the placeholders creation/replacement process.

Have a look at the [ERB](https://github.com/PrismJS/prism/blob/gh-pages/components/prism-erb.js), [Handlebars](https://github.com/PrismJS/prism/blob/gh-pages/components/prism-handlebars.js), [PHP](https://github.com/PrismJS/prism/blob/gh-pages/components/prism-php.js) or [Smarty](https://github.com/PrismJS/prism/blob/gh-pages/components/prism-smarty.js) components for examples. You can also take a look at the functions in the [Markup-templating](https://github.com/PrismJS/prism/blob/gh-pages/components/prism-markup-templating.js) component.

The way to create/replace the placeholders changed a bit, but from your component perspective, it should only make it simpler, really.

components/prism-tt2.js
```diff
@@ -0,0 +1,78 @@
+(function(Prism) {
+
+	var tt2_pattern = /\[%[\s\S]+?%\]|^%%*?/;
+
+	Prism.languages.tt2 = Prism.languages.extend('markup', {
```
Golmote: If you use the new markup-templating component (see my previous comment), you won't have to extend `markup` anymore.

components/prism-tt2.js
```diff
@@ -0,0 +1,78 @@
...
+					lookbehind: true
+				},
+				'delimiter': {
+					pattern: /^(?:\[%|%%)-?|-?%]$/i,
+					alias: 'punctuation'
+				},
+				'string': /(["'])(?:\\.|(?!\1)[^\\\r\n])*\1/,
+				'number': /\b0x[\dA-Fa-f]+\b|(?:\b\d+\.?\d*|\B\.\d+)(?:[Ee][+-]?\d+)?/,
+				'keyword': /\b(?:GET|CALL|SET|DEFAULT|INSERT|INCLUDE|PROCESS|WRAPPER|BLOCK|IF|UNLESS|ELSIF|ELSE|SWITCH|CASE|FOREACH|WHILE|FILTER|USE|MACRO|RAWPERL|PERL|TRY|THROW|CATCH|FINAL|NEXT|LAST|RETURN|STOP|CLEAR|META|TAGS|DEBUG)\b/,
+				'operator': /\b(?:=>|==|!=|<|<=|>|>=|&&\|\|!|and|or|not)\b/,
```
Golmote: Something looks weird here: `&&\|\|!`. Is it intended? Wasn't it supposed to be something like... `&&|\|\||!` instead, maybe?

Also, I'm not sure the word-boundaries are a good idea for the operators other than words. Your regexp should probably look like:
`/=>|==|...|\b(?:and|or|not)\b/`

components/prism-tt2.js
```diff
@@ -0,0 +1,78 @@
...
+				'punctuation': /[[\]{},]/,
+				'variable': /[a-zA-Z][\.\/\$a-zA-Z0-9]*/
```
Golmote: You don't need those three escapes in the caracter class. Also, you could benefit from the `i` flag here to remove the `A-Z` occurrences, and could use `\d` instead of `0-9`.

components/prism-tt2.js
```diff
@@ -0,0 +1,78 @@
...
+			}
+		}
+	});
+
+	// Tokenize all inline TT2 expressions that are wrapped in [% %]
+	// This allows for easy TT2 + markup highlighting
+	Prism.hooks.add('before-highlight', function(env) {
```
Golmote: As previously expressed, this whole code can now be drastically simplified by using the functions available in the Markup-templating component. Plus it will make the component fully compatible with NodeJS.

components/prism-tt2.js
gflohr: Cut and paste errors from my side.

components/prism-tt2.js
gflohr: That's my personal style but I can follow your convention.  Off-topic, but this is my rationale:

- The extraneous escapes help both syntax highlighters and inexperienced readers.
- `/[a-zA-Z]/` is slightly more efficient than `/[a-z]/i`.  Since the `i` modifier is unicode aware, case-insensitive matching is potentially expensive.
- Does `\d` match "۴۲" or not? It does not.  But why make people figure, when you can be explicit.

Yes, rather academic, and I have no problem following your style instead.

components.json
Golmote: Sounds good to me. No alias at all, use `tt2` as the identifier. Maybe just simplify the title to "Template Toolkit 2" :wink:

gflohr: Agreed, I will just use "Template Toolkit 2".

components/prism-tt2.js
Golmote: Haha, you're right that I usually go for the shortest form instead, at the cost of a tiny bit of readability. But hey, regexps _are_ quite hard to read anyway. ^^

(Note that regarding your unicode-related points, I'm pretty sure they do not apply unless the `u` flag is used, if they even do.)

gflohr: True.  The `u` makes a difference in this case:

```javascript
console.log("\u212a".toLowerCase());
console.log("\u212a".match(/[a-z]/iu));
```

Although the behavior without `/u` could also be qualified as a bug.

But, honestly, do you know whether "۲" (an Arabic two) matches `/d` or whether "\u3000" (ideographic space) matches `\s` (without `/u`)?  The Arabic digit does not match `\d` but ideographic space matches `\s`.

I think the usage in Prism is correct everywhere so this is rather academic and it was really not meant as criticism.

Golmote: No worries.

components.json
gflohr: I am rewriting the component from scratch now but I am not sure that I am completely on the right track.

A typical example for a TT2 template is https://github.com/gflohr/qgoda-site/blob/master/_views/components/listing.html.

My current component is here: https://github.com/gflohr/prism/blob/gh-pages/components/prism-tt2.js

Two features are causing me a headache:

```
[%# This is a multi-line comment
    because the "#" is immediately following
    the opening tag.
%]
[% # And this is a normal comment.
    x = 42 # So is this.
%]
```

I tried to implement that morelike this:

```javascript
Prism.languages.insertBefore('tt2', 'delimiter', {
	comment: /\[%#[^]*?\%\]/
});
```

But at no avail.  Only the first line is recognized as comment.

And as an alternative to `[% ... %]` you can use [outline tags](http://www.template-toolkit.org/docs/manual/Syntax.html#section_Outline_Tags):

```
%% IF foo = "bar"
<br />
%% END
```

The "closing delimiter" is a newline here.  How can I specify that?

Both features are a little bit esoteric, and could also be left as a known limitation if there is no solution.

gflohr:
My fork is messed up. :( I shouldn't have committed the generated files.

Do you mind if I create a new fork and pull request?

Golmote:
Absolutely not.

mAAdhaTTah:
We actually do commit the generated files though.

mAAdhaTTah:
That said, it does look like your fork is mess up. Maybe a botched rebase?

gflohr:
@mAAdhaTTah Ummh ... I actually wanted to get rid of the conflicts with the rebase.  Looks like that went slightly wrong .;)

Other than that I think that I'm done except for one TT2 feature:

```
<div>
[% variable %]
%% variable
   %% variable
</div>
```

The first two lines are equivalent (when the non-default feature "outline tags" are activated) and will both render the contents of variable `variable`.  The third line is *not* considered template code because the "outline tag" `%%` is not in column 1.  I don't see how this could be implemented with any justifiable effort or did I miss something?

Golmote:
You could "look behind" for the line break, using a regexp like `/(\r|\n)%%.../` and setting `lookbehind: true`.
You could even add `^` inside the parentheses to match this feature at the beginning of the code block, but note that this has flaws: it will also match the pattern if another pattern immediately precedes it, because it will be considered the beginning of the string. (`,%% foo` would match, if you highlight the `,` as punctuation, for example).

Golmote:
> My fork is messed up. :(

I might be able to force push a cleaned up version on your branch, since you gave us write permission, if need be (https://github.com/Golmote/prism/tree/gflohr-fix).

gflohr:
As for the "outline tags": There are not enabled by default and I suggest to ignore them.  This is also documented.

I can really just create a new fork and new PR.  I will then not commit any generated files.  That's  less work, especially for you.  So feel free to close this.

Golmote:
>  I will then not commit any generated files.

I'll still ask you to commit them, but it can be done after the first review I guess. I'm closing this PR and expect to see a new one soon, then.

PR #1413, opened 2018-05-01
Make plugins Command Line and Autoloader compatible

xmuyoucai:
Command Line rendered before Autoloader, so, this is the problem. Code tags's content will be replaced when autoloader loaded. So, you maybe see Command Line appearing for a very short time, then disappeared. I just do some change in prism.js, using replace instead of '='. But i'm not sure that's the best solution.

### DIFF
--- a/prism.js
+++ b/prism.js
@@ -267,7 +267,7 @@ var _ = _self.Prism = {
 
 			_.hooks.run('before-insert', env);
 
-			env.element.innerHTML = env.highlightedCode;
+			env.element.innerHTML = env.element.innerHTML.replace(code, env.highlightedCode);
 
 			callback && callback.call(element);
 


mAAdhaTTah:
First, don't modify the the `prism.js` file. That's the one served on the website and is generated by the gulp file. You'd want to modify `components/prism-core.js`.

We're going to have to dig into this and make sure it doesn't conflict w/ other plugins & usages before we decide is this works or not though.

Golmote:
Hi! Thanks for contributing to Prism.

I think the solution to this issue could be implemented directly in the Command line plugin. Instead of aborting the process when the `.command-line-prompt` element is found, we should probably update its content instead. This is usually what we expect from calling `highlightElement` again on an already highlighted code block.

So, I think you should update this PR to make the following changes:

- Remove [the piece of code that aborts the process](https://github.com/PrismJS/prism/blob/gh-pages/plugins/command-line/prism-command-line.js#L26-L29).
- Before [creating the `.command-line-prompt` element](https://github.com/PrismJS/prism/blob/gh-pages/plugins/command-line/prism-command-line.js#L121), first check for its existence. It needs to be created only if it does not already exists.
- Move [the node insertion](https://github.com/PrismJS/prism/blob/gh-pages/plugins/command-line/prism-command-line.js#L135) in the condition created above. The node needs to be inserted only if it wasn't already there.

xmuyoucai:
Thanks for your reply！For mAAdhaTTah, I don't want to modify this core file `prism.js`, because i known this maybe cause some other issues. I just want to make this issue known by Prism. I expect the new version will fix this problem. For Golmote, your solution is just my first way that i thought of! But i really don’t have so much time to read all of these source code, so,  hah, i just leave these works for prism!

PR #1430, opened 2018-05-27
Add TAP component

mAAdhaTTah:
Still need to add tests, but I squashed the commits from #1089.


mAAdhaTTah:
@Golmote Not great at regex, so I'm struggling with getting yamlish working. Got an idea?

Golmote:
I would say you need something like

```js
	yamlish: {
		pattern: /(^[^\S\r\n]*)---(?:\r\n?|\n)(?:.*(?:\r\n?|\n))*?[^\S\r\n]*\.\.\.$/m,
		lookbehind: true,
		inside: Prism.languages.yaml,
		alias: 'language-yaml'
	}
```

The lookbehind (the first capturing group) matches the optional spaces before `---`. Then we match multiple lines `(?:.*(?:\r\n?|\n))*?`, then we end up with the optional spaces and the final `...`.

This should match:

```
  ---
  but: this one
  has:
    - lots
    - of
    - stuff
  ...
```

mAAdhaTTah:
@Golmote I think we're looking good here. Let me know if you have suggestions.

components/prism-tap.js
```diff
@@ -0,0 +1,20 @@
+Prism.languages.tap = {
+	fail: /(    )*not ok[^#\{\n]*/,
+	pass: /(    )*ok[^#\{\n]*/,
```
Golmote: You do not need to escape the `{` in the character class in those two regexps.
Also, we should probably include `\r` in the character class too, to handle the different line endings.

components/prism-tap.js
```diff
@@ -0,0 +1,20 @@
...
+	pragma: /(    )*pragma ([\+-])([a-z]+)/,
```
Golmote: You do not need to escape the `+` in the character class.
Also, the last two pairs of parentheses are not needed.

components/prism-tap.js
```diff
@@ -0,0 +1,20 @@
...
+	bailout: /(    )*bail out!(.*)/i,
+	version: /(    )*TAP version ([0-9]+)/i,
+	plan: /(    )*([0-9]+)\.\.([0-9]+)( +#[^\n]*)?/m,
```
Golmote: You can replace all occurrences of `([0-9]+)` with `\d+` (no parentheses needed here) in those two regexps.
Also, it seems the `m` could be removed, since the regexp does not use `^` or `$` assertions.
Finally, the `[^\n]*` part could probably be replaced with `.*`.

components/prism-tap.js
```diff
@@ -0,0 +1,20 @@
...
+	subtest: {
+		pattern: /(    )*# Subtest(?:: (.*))?/,
```
Golmote: The parentheses around `(.*)` are not needed.

tests/languages/tap/pass_fail_feature.test
```diff
@@ -0,0 +1,17 @@
...
+[
+	["fail", "not ok" ],
+	["fail", "not ok 42 this is the description of the test"],
+	["pass", "ok" ],
+	["pass", "ok 42 this is the description of the test"]
+]
+
+----------------------------------------------------
+
+Checks test pass & fail together correctly
```
Golmote: Considering you added this test, I think it can be overkill to also have the `pass_feature` and `fail_feature` tests separately.

tests/languages/tap/plan_feature.test
```diff
@@ -0,0 +1,11 @@
+1..10
```
Golmote: The `plan` pattern in the language definition seems to also include an optional directive (e.g. `# foobar`) after the numbers part. If this is what is intended, we should add a test for it here.

mAAdhaTTah:
@Golmote Updated.

Golmote:
Looks quite good to me. There are still a few nitpicks, but I won't mind if this is merged as is.

You did not give me feedback regarding this comment. Did you give it a try? Are those needed?
> Since you removed the `(^|\n)` parts that were checking for the beginning of the string in several regexps, I'm wondering if the `(    )*` parts are still needed. You might want to try removing them and see how the tests go.

Also I can still see some unneeded parentheses:
- around `(.*)` in `bailout`
- around `(\d+)` in `version`
- around `(\d+)` in `plan`.

Finally, all the remaining groups, expect in the `yamlish` pattern that uses lookbehind, could be made non-capturing using `?:`.

mAAdhaTTah:
I missed that comment the first time around. Made the changes suggested; tests still pass. Also added the minified file.

PR #1434, opened 2018-06-01
Make line-numbers styles more specific

mAAdhaTTah:
The padding added to `pre.line-numbers` in the line-number's CSS is
overwritten by the theme's CSS if the theme is loaded second. This
ensures the CSS can come in any order without specificity issues.

I ran into this problem in the settings page for WP-Gistpen, where the
user can toggle the theme and enabled plugins. Switching between themes
removes the CSS for the previous theme from the head and injects the new
theme into the head, so the line-numbers CSS actually appears in the
head _first_. The theme CSS that sets the padding uses `pre` while the
line-numbers `pre.line-numbers`, and who "wins" is whomever
comes second.

There's also some whitespace cleanup here as well.

I think this is unlikely to be a breaking change, but interested in feedback on this.

### DIFF
--- a/plugins/line-numbers/prism-line-numbers.css
+++ b/plugins/line-numbers/prism-line-numbers.css
@@ -1,12 +1,12 @@
-pre.line-numbers {
+pre[class*="language-"].line-numbers {
 	position: relative;
 	padding-left: 3.8em;
 	counter-reset: linenumber;
 }
 
-pre.line-numbers > code {
+pre[class*="language-"].line-numbers > code {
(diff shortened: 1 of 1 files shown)


Golmote:
Wow, good catch. It does not happen on the website because we switch the theme "in place" so the order is kept. Looks good to me!

mAAdhaTTah:
So this introduces a bug in the Coy theme unless the CSS is loaded in reverse, because Coy moves the padding to the `code` element.

PR #1443, opened 2018-06-16
Bash improvements

alice-mm:
First and foremost, sorry for the rather large pull request. Most of these edits were made while writing some kind of Bash tutorial and I initially had no ambition of contributing. I didn't even have a non-pro GitHub account.

I was using Prism to write [this thing](http://www.alicem.net/shell/). I had tons of fun, but I also noticed that lots of things were missing. If I recall correctly, even the `esac` keyword (used to close “switch” statements) was nowhere to be found despite `case` being here.

To sum most of these things up, here's a HTML document that can be used to do some kind of “before / after” comparison:

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8" />
  <link href="prism.css" rel="stylesheet" />
  <style>
    .token.variable {
      color: #9ec6d9;
    }

    .token.keyword {
      color: #4e9ce6;
    }

    .token.punctuation {
      color: #eaeaea;
    }

    .token.builtin {
      color: #fd96b3;
    }

    .token.operator {
      color: #d7cf71;
    }

    .token.string {
      color: #9bea81;
    }

    .token.function,
    .token.function-name {
      color: #ff934f;
    }

    .token.number {
      color: #0acc13;
    }

    .token.constant {
      color: #ee3928;
    }

    .token.comment,
    .token.shebang {
      font-weight: normal;
    }

    .token.entity {
      cursor: initial;
    }
  </style>
</head>
<body>
  <pre><code class="language-bash">#! /usr/bin/env bash
# More permissive shebang ↑

# Commands
column apt tac sh shellcheck shuf zenity zsh

# Keywords
esac

# Added builtin category; moreover, some builtins were neither in keywords nor in functions
unset local

# Missing special variables
$$ $#

# Here-documents:
# * support for “&lt;&lt;-” (also added in operators)
# * no expansion for version with quoted beginning tag (so no colours for dollared stuff)
&lt;&lt;- TXT
a $b $(echo)
TXT

&lt;&lt; 'TXT'
a $b $(echo)
TXT

# “+=” in operators outside arithmetic environments (append to string or array)
t+=('a')
s+='b'

# Allowed “&lt;(” and “&gt;(” (process substitution) before keywords, functions, builtins, booleans…
&lt;(yes) &lt;(if true; then :; fi) &lt;(echo) &lt;(true)
&gt;(yes) &gt;(if true; then :; fi) &gt;(echo) &gt;(true)

# Added operators within brace expansion
${x:1:2} ${x:-a} ${x:=a} ${x:?a} ${x:+a}
${!x} ${x/a/b}
${x#*a} ${x##*a} ${x%a*} ${x%%a*}
${x^a} ${x^^a} ${x,a} ${x,,a}

# Prevented “$!!!!” and “$????” from being read as long variable names
$!!!! $????

# Allowed comments in command substitution
$(
    # This is an echo
    echo
)

# Added highlighting of “\n” and such within strings
echo '1\a2\b3\c4\e5\f6\n7\r8\t9\v'
# + support for bytes with octal values (1–3 digits in 0–7)
echo '1234\056789'
# + same here but in hex (1 or 2 digits in 0–F)
echo 'abc\xdef'
# + similar things from printf's manual
printf '123\456789'
printf '\uABCDEFG'
# + escaped double quotes
"a\"b"

# Added “\” to punctuation marks, as it is used to break long lines of code
echo \
    'a'

# Highlighted some common environment variables
IFS='a'
$PS1 "$UID" ${BASH_SOURCE} "${LC_NUMERIC}"

# Highlighted square brackets as punctuation within brace expansion (array access)
${t[i]}

# Highlight names directly following “for” or “select” as variable names
for var in a b c
do :; done
select var in a b c
do :; done

# Highlight what's on the left of assignments as variable names while still detecting environment variables
var='foo'
arr+=('bar')
IFS='a'
PATH+=':a/b/c'

# Recognize function names as such in their declaration
function foo { :; }
foo() { :; }
function foo() { :; }
# Not a function:
foo { :; }

# Redirections and file descriptors
a |& b &&gt; c
a &gt;&2
a 2&lt;&
# + several exotic file descriptor manipulations from Bash's manual.

# Did my best to prevent operators that use multiple characters from being perceived as several small operators
a &lt;&lt;&lt; b &&gt;&gt; c
a && b || c
</code></pre>
</body>
```

I added a tiny bit of CSS in order to make it easier to distinguish things that used the same style in the theme I loaded (“Tomorrow night”).

I had to update / adapt tests and used the opportunity to add new test cases covering new features.

I did my best but I'm no JS expert, so there's probably room for improvement, but I have to say I like the result a lot. To check more realistic examples, you can simply take a look at the snippets from [the tutorial-ish document I was writing when customizing this language definition](http://www.alicem.net/shell/).


Golmote:
Hi! Wow this looks great! I'll try to do a proper review this week-end.

From a really quick look, I can already tell that some indentations are weird (note that we do use tabs for indentation in Prism), and that you should probably add a `builtin_feature.test` test file just like the one for the keywords.

> I also noticed that lots of things were missing.

Most of the time, components need to be actually used extensively to reveal their flaws. I'm glad you took the time to improve the component to fit your needs and even submitted a PR for it! Thanks!

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
+				'file-descriptor': {
+					pattern: /(^|[^\d])(?:\d)(?=$|[^\d])/,
+					lookbehind: true,
+					alias: 'important'
+				}
+			}
+		},
+		punctuation: /\$?\(\(?|\)\)?|\.\.|[{}[\];]|\\/,
+		number: {
+			pattern: /(^|\s)(?:([1-9]\d*|0)([.,]\d+)?)(?=$|\W)/,
```
Golmote: Could you make the groups non-capturing and remove the unneeded one around the middle part?
Also, I think the look-ahead can be replaced with a simple `\b` here.

Result would be something like `/(^|\s)(?:[1-9]\d*|0)(?:[.,]\d+)?\b/`

tests/languages/bash/environment_feature.test
```diff
@@ -0,0 +1,35 @@
...
+			["environment", "PS1"],
+			"}"
+		]],
+		"\""
+	]]
+]
+
+----------------------------------------------------
+
+Checks for environment variables.
```
Golmote: It would be great to test for the whole list from `envVarList`, at least in one of their syntaxes (probably the simplest `$XXX` form).

tests/languages/bash/keyword_feature.test
```diff
@@ -1,20 +1,30 @@
...
-	["keyword", "do"], ["keyword", "done"], ["keyword", "until"], ["keyword", "echo"], ["keyword", "exit"],
-	["keyword", "return"], ["keyword", "set"], ["keyword", "declare"],
-	["keyword", "."], ["keyword", ":"]
+	["keyword", "if"], ["keyword", "then"], ["keyword", "else"], ["keyword", 
+	"elif"], ["keyword", "fi"], ["keyword", "for"], ["for-or-select", "x"],
+	["keyword", "in"], " a\r\n", ["keyword", "while"], 
+	["keyword", "case"], ["keyword", "esac"], ["keyword", 
+	"function"], ["keyword", "select"], ["for-or-select", "x"], ["keyword",
+    "in"], " a\r\n", ["keyword", "do"], ["keyword", "done"], 
+	["keyword", "until"]
```
Golmote: This is quite hard to read, especially when an array wraps on a new line. We should rather use the same line repartition as the test code.

```
    ["keyword", "if"],
    ["keyword", "then"],
    ...
    ["keyword", "for"], ["for-or-select", "x"], ["keyword", "in"], " a\r\n",
    ...
```

tests/languages/bash/function_feature.test
```diff
@@ -1,101 +1,93 @@
...
+	["function", "uname"], ["function", "unexpand"], ["function", "uniq"], 
+	["function", "units"], ["function", "unrar"], ["function", "unshar"], 
+	["function", "uptime"], ["function", "useradd"], ["function", "userdel"], 
+	["function", "usermod"], ["function", "users"], ["function", "uuencode"], 
+	["function", "uudecode"], ["function", "v"], ["function", "vdir"], ["function", 
+	"vi"], ["function", "vmstat"], ["function", "wait"], ["function", "watch"], 
+	["function", "wc"], ["function", "wget"], ["function", "whereis"], ["function", 
+	"which"], ["function", "who"], ["function", "whoami"], ["function", "write"], 
+	["function", "xargs"], ["function", "xdg-open"], ["function", "yes"], 
+	["function", "zenity"], ["function", "zip"], ["function", "zsh"]
```
Golmote: Would you mind put each function on its own line? The readability would improve a lot.

```
apropos
apt
apt-get
aptitude
aspell
...
```

```
    ["function", "apropos"],
    ["function", "apt"],
    ["function", "apt-get"],
    ["function", "aptitude"],
    ["function", "aspell"],
    ...
```

tests/languages/bash/process_substitution_feature.test
```diff
@@ -0,0 +1,55 @@
...
+	["punctuation", ")"],
+	["operator", [">"]],
+	["punctuation", "("],
+	["boolean", "true"],
+	["punctuation", ")"]
+]
+
+----------------------------------------------------
+
+Checks for variables.
```
Golmote: I wonder if this test file is needed on its own. Maybe a file `operator_feature.test` could test all operators including process substitution?

tests/languages/bash/redirections_and_fd_feature.test
```diff
@@ -0,0 +1,31 @@
...
+	"\r\na ",
+	["operator", ["<>"]],
+	" b\r\na ",
+	["operator", ["&>>"]],
+	" b"
+]
+
+----------------------------------------------------
+
+Checks for input / output redirections and file descriptors.
```
Golmote: Same comment here, this could go in a new test file that checks all operators.

tests/languages/bash/var_assign_feature.test
```diff
@@ -0,0 +1,23 @@
...
+	["assign-left", ["foo"]],
+	["operator", ["+="]],
+	["punctuation", "("],
+	["string", ["'xyz'"]],
+	["punctuation", ")"]
+]
+
+----------------------------------------------------
+
+Checks for variable assignments.
```
Golmote: We usually try to have a correspondance between test files' names and patterns in the language definition. So you should rename this file `assign-left_feature.test` or rename the pattern, so that it's more clear what this test file does without even reading it.

tests/languages/bash/var_in_for_and_select_feature.test
```diff
@@ -0,0 +1,25 @@
...
+	["for-or-select", "var"],
+	["keyword", "in"], " a b c\r\n", ["keyword", "do"],
+	["builtin", ":"],
+	["punctuation", ";"],
+	["keyword", "done"]
+]
+
+----------------------------------------------------
+
+Checks for variable names in “for” and “select” loops.
```
Golmote: Same comment here. Can you rename the file `for-or-select_feature.test` please?

tests/languages/bash/func_def_feature.test
```diff
@@ -0,0 +1,41 @@
...
+	"\r\nbar ",
+	["punctuation", "{"],
+	["builtin", ":"],
+	["punctuation", ";"],
+	["punctuation", "}"]
+]
+
+----------------------------------------------------
+
+Checks for function definitions.
```
Golmote: Can you rename this file `function-name_feature.test` please?

tests/languages/bash/function_feature.test
Golmote: Also, it would be great to have a similar file `builtin_feature.test` to test the `builtin` pattern you added.

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
+			},
+			{
+				// b)
+				pattern: /(^|\W)(?:\w+)(?=\s*\(\s*\)\s*\{)/,
+				lookbehind: true
+			}
+		],
+		// Highlight variable names as variables in for and select beginnings.
+		'for-or-select': {
+			pattern: /((^|[\s;|&]|[<>]\()(for|select)\s+)(?:\w+)(?=\s+in\s)/,
```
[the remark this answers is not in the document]
alice-mm: Hmm… I used all this at the beginning to handle `<(for blabla…)` and so on, but I guess you're right, a `\W` might do the trick, and even more.

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
-				pattern: /((?:^|[^<])<<\s*)["']?(\w+?)["']?\s*\r?\n(?:[\s\S])*?\r?\n\2/,
+				pattern: /((?:^|[^<])<<-?\s*)(\w+?)\s*\r?\n(?:[\s\S])*?\r?\n\2/,
 				lookbehind: true,
 				greedy: true,
 				inside: insideString
 			},
+			// Here-document with quotes around the tag
+			// → No expansion (so no “inside”).
+			{
+				pattern: /((?:^|[^<])<<-?\s*)["'](\w+?)["']\s*\r?\n(?:[\s\S])*?\r?\n\2/,
```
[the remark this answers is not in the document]
alice-mm: Wow, I don't know why I didn't think about backreferences. I use them all the time in `sed` and stuff but I still feel oddly uneasy when working with JS regexs. Thanks.

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
-			pattern: /(^|[\s;|&])(?:let|:|\.|if|then|else|elif|fi|for|break|continue|while|in|case|function|select|do|done|until|echo|exit|return|set|declare)(?=$|[\s;|&])/,
+        // https://www.gnu.org/software/bash/manual/html_node/Shell-Builtin-Commands.html
+		builtin: {
+			pattern: /(^|[\s;|&]|[<>]\()(?:\.|:|break|cd|continue|eval|exec|exit|export|getopts|hash|pwd|readonly|return|shift|test|times|trap|umask|unset|alias|bind|builtin|caller|command|declare|echo|enable|help|let|local|logout|mapfile|printf|read|readarray|source|type|typeset|ulimit|unalias|set|shopt)(?=$|[\)\s;|&])/,
 			lookbehind: true
 		},
-		'boolean': {
-			pattern: /(^|[\s;|&])(?:true|false)(?=$|[\s;|&])/,
+		boolean: {
+			pattern: /(^|[\s;|&]|[<>]\()(?:true|false)(?=$|[\)\s;|&])/,
```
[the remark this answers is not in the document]
alice-mm: I was wondering about that, yeah, but still they're _used_ as booleans and have a similar semantic, I kept that as it was.

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
-		'operator': /&&?|\|\|?|==?|!=?|<<<?|>>|<=?|>=?|=~/,
-		'punctuation': /\$?\(\(?|\)\)?|\.\.|[{}[\];]/
+		'file-descriptor': {
+			pattern: /(^|\W)(?:&[0-9])(?=$|\W)/,
+			lookbehind: true,
+			alias: 'important'
+		},
+		operator: {
+			// Lots of redirections here, but not just that.
+			pattern: /\d?<>|>\||\+=|==?|!=?|=~|<<[<-]?|[&\d]?>>|\d?[<>]&?|&>|&&?|\|&|\|\|?|<=?|>=?/,
```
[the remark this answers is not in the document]
alice-mm: I actually changed this regexp tons of times and noticed that the slightest change, even the most innocent one, could break multiple-character operators into several occurrences of shorter operators, like a `<>` being interpreted as `<` followed by `>` rather than as a single operator. So I have to admit I grew kind of scared of this part of the code. :fearful: I'll give it a try, I guess, but I think readability should (alas) become secondary for this particular line. But it's too soon to give up altogether.

tests/languages/bash/environment_feature.test
alice-mm: I was wondering if those tests with whole lists were overkill things or the standard way to go for Prism, haha. I'll do as you wish.

Golmote: They probably are a bit overkill, but they proved themselves useful in the past on a few occasions. 
And they are quick to write with a few "search and replace" action.

Golmote:
Re: #1457, the issue is that the `$()` match ends at the first closing parenthesis `)` encoutered, instead of the second one in this case. But don't worry if you don't want to take time for it, it can be fixed later in a separate PR.

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
+			alias: 'variable',
 			lookbehind: true
 		},
-		'string': [
-			//Support for Here-Documents https://en.wikipedia.org/wiki/Here_document
+		string: [
+			// Support for Here-documents https://en.wikipedia.org/wiki/Here_document
 			{
-				pattern: /((?:^|[^<])<<\s*)["']?(\w+?)["']?\s*\r?\n(?:[\s\S])*?\r?\n\2/,
+				pattern: /((?:^|[^<])<<-?\s*)(\w+?)\s*\r?\n(?:[\s\S])*?\r?\n\2/,
```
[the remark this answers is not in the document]
alice-mm: Isn't this group used to make sure the here document is ended with the same tag? Like, to prevent:

```
cat << _PLOP_
blabla
_PLAP_
```

(`_PLOP_` ≠ `_PLAP_`)
There's a `\2` for this at the end.
The tests go haywire if I ungroup this thing.

alice-mm:
`components/prism-kotlin.min.js` gets modified when I run Gulp, even though it does not seem related to Bash in any way… I'll refrain from committing it for now; warn me if it is required.

alice-mm:
For https://github.com/PrismJS/prism/issues/1457, note that if you allow a pair of parentheses within `$(…)`, the opening one should not be right after the `$(` as it would look like arithmetic expansion (`$((…))`).
And yeah I'm not sure it would be wise to hammer new features in this PR. :smile:

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
+			// c) function foo() {
+			// but not “foo {”
+			{
+				// a) and c)
+				pattern: /((^|\W)function\s+)(?:\w+)(?=(\s*\(\s*\))?\s*\{)/,
+				lookbehind: true
+			},
+			{
+				// b)
+				pattern: /(^|\W)(?:\w+)(?=\s*\(\s*\)\s*\{)/,
```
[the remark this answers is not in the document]
Golmote: The very first group need to stay capturing `(^|\W)`. It's used by the `lookbehind` feature. (We do need to know what was matched to be able to remove it later.)

components/prism-bash.js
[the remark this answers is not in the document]
Golmote: Woops you're right! I missed the `\2` there. Sorry about that!

components/prism-bash.js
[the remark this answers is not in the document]
Golmote: Here too, I missed the `\2`, so the quote stuff would look like `(["'])(\w+?)\2` and the backreference at the end would be `\3`.

components/prism-bash.js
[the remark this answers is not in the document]
alice-mm: OK, thanks for the explanation as it initially looked very strange to me. So I guess no `?:` here for the `a) and c)` case either.

Golmote:
> `components/prism-kotlin.min.js` gets modified when I run Gulp, even though it does not seem related to Bash in any way… I'll refrain from committing it for now; warn me if it is required.

Hm that's weird. Please don't commit it.

Are there actual differences in the git diff?
What's your version of UglifyJS?
```
D:\Documents\prism>npm list uglify-js
prismjs@1.15.0 D:\Documents\prism
`-- gulp-uglify@0.3.2
  `-- uglify-js@2.4.6
```

components/prism-bash.js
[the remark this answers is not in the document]
alice-mm: Hm, yeah, seems OK.

components/prism-bash.js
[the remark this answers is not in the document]
alice-mm: Yeah, I did that and it seems to work fine (not pushed yet).

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
+			pattern: /(^|\W)(?:&[0-9])(?=$|\W)/,
+			lookbehind: true,
+			alias: 'important'
+		},
+		operator: {
+			// Lots of redirections here, but not just that.
+			pattern: /\d?<>|>\||\+=|==?|!=?|=~|<<[<-]?|[&\d]?>>|\d?[<>]&?|&>|&&?|\|&|\|\|?|<=?|>=?/,
+			inside: {
+				'file-descriptor': {
+					pattern: /(^|[^\d])(?:\d)(?=$|[^\d])/,
```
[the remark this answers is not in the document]
alice-mm: So just `/^\d/`? :smile: It seems OK, though.

tests/languages/bash/environment_feature.test
alice-mm: You're perfectly right: creating the `builtin` test allowed me to notice that `cd` was in both functions and builtins, seconds ago. Haha.

```
$ type cd
cd is a shell builtin
```

→ Removed from functions.

alice-mm:
```
$ npm list uglify-js
prismjs@1.14.0 /home/alice/prog/prism/prism
└─┬ gulp-uglify@0.3.2
  └── uglify-js@2.4.6
```

Git diff shows a bunch of differences, but as it is a minified file it's not really readable.
Hm… It seems the only difference is that the keyword `to` is added in the “after build” version. Did someone add that in a PR and forgot to commit the minified version?

alice-mm:
I think I took most of the comments into account. I have yet to try to simplify the operator list, though, and I did nothing regarding https://github.com/PrismJS/prism/issues/1457. I'll have lunch, for now. :smile:
Thanks again for your guidance so far.

alice-mm:
Your regexp for `(…)` within `$(…)` seems to work wonders.
image
I don't see anything else that needs to be done, now, but maybe I forgot something. I'll let you check all that.

Golmote:
Found the Kotlin commit here: 41e3d6a6f6926a4273c928ae857ce375a25fbd6e It was done right after you opened this PR, so your version is a bit behind `master`. Mystery solved!

alice-mm:
Yeah, I haven't pulled from Prism's repository for a while. Does that mean that I should now, or is it OK as long as the only files I edited did not move? Travis seemed happy so I did not give it much thought.

Golmote:
Don't worry, as long as the merge can be made without conflict, this is fine.

Rychu-Pawel:
Hey Guys. Any ETA on this? @Golmote  @alice-mm

alice-mm:
> Any ETA on this?

Not sure what exactly is happening. According to the tag changes, it seems @Golmote acknowledged my changes and intends to check them or to let someone else do so. I'm on standby, should any new remark surface.

alice-mm:
Hi everyone. Wanted to see how things were going. Noticed conflicts sprouted after https://github.com/PrismJS/prism/pull/1577 was merged. I tried to compute the union of the commands listed in the two PRs, still excluding shell builtins that tend to be mistaken for external commands. Looks like the conflicts are gone, now. Pfiou. Not used to working with multiple Git remotes. :sweat_smile:

/cc @Golmote

Still not sure why this PR is stalled right at the final-ish review stage. :cry: 

A few months ago, I used my version again (for a Reveal.js slideshow to teach Bash stuff to my workmates) and didn't notice anything weird. Prism is fun.

mAAdhaTTah:
@RunDevelopment Would you mind taking a pass through this PR and providing a final sign off & merge if everything looks good? These look like good improvements, so might be nice to get them landed.

RunDevelopment:
@mAAdhaTTah I'm quite busy at the moment, so it might take a while until I get to it.

alice-mm:
No problem! Indeed, his profile seems to show that his activity plummeted around the time of his reviews of this PR.
Thanks for reacting so fast. :slightly_smiling_face:

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
+						pattern: new RegExp("(\\{)(?:" + envVarList + ")(?=$|\\}|\\W)"),
+						lookbehind: true,
+						alias: 'constant'
+					}
+				}
+			},
+			/\$(?:\w+|[#?*!@$])/
+		],
+		// man echo / printf, + \"
+		entity: /\\([abceEfnrtv\\"]|O?[0-7]{1,3}|x[0-9a-fA-F]{1,2}|u[0-9a-fA-F]{4}|U[0-9a-fA-F]{8})/
```
RunDevelopment: 1. Please make the group non-capturing.
2. `[abceEfnrtv\\"]` looks like black magic to me. Could you please explain this to someone who doesn't know bash?

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
-			//Support for Here-Documents https://en.wikipedia.org/wiki/Here_document
+		'function-name': [
+			// a) function foo {
+			// b) foo() {
+			// c) function foo() {
+			// but not “foo {”
 			{
-				pattern: /((?:^|[^<])<<\s*)["']?(\w+?)["']?\s*\r?\n(?:[\s\S])*?\r?\n\2/,
+				// a) and c)
+				pattern: /((^|\W)function\s+)\w+(?=(\s*\(\s*\))?\s*\{)/,
```
RunDevelopment: 1. Please make the non-lookbehind capturing groups non-capturing.
2. You can replace `(^|\W)` with a simple `\b`.

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
+		'number'
+	];
 	var inside = insideString.variable[1].inside;
-	inside.string = Prism.languages.bash.string;
-	inside['function'] = Prism.languages.bash['function'];
-	inside.keyword = Prism.languages.bash.keyword;
-	inside['boolean'] = Prism.languages.bash['boolean'];
-	inside.operator = Prism.languages.bash.operator;
-	inside.punctuation = Prism.languages.bash.punctuation;
+	for(var i = 0;  i < toBeCopied.length;  i++) {
```
RunDevelopment: Please format this line. One space after `for`. No double spaces.

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
 			{
-				pattern: /((?:^|[^<])<<\s*)["']?(\w+?)["']?\s*\r?\n(?:[\s\S])*?\r?\n\2/,
+				// a) and c)
+				pattern: /((^|\W)function\s+)\w+(?=(\s*\(\s*\))?\s*\{)/,
+				lookbehind: true,
+				alias: 'function'
+			},
+			{
+				// b)
+				pattern: /(^|\W)\w+(?=\s*\(\s*\)\s*\{)/,
```
RunDevelopment: You can replace the lookbehind with `\b`.

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
+				lookbehind: true,
+				alias: 'function'
+			}
+		],
+		// Highlight variable names as variables in for and select beginnings.
+		'for-or-select': {
+			pattern: /((?:^|\W)(?:for|select)\s+)\w+(?=\s+in\s)/,
```
RunDevelopment: You can replace `(?:^|\W)` with `\b`.

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
+			{
+				pattern: /((?:^|[^<])<<-?\s*)(\w+?)\s*(?:\r?\n|\r)(?:[\s\S])*?(?:\r?\n|\r)\2/,
 				lookbehind: true,
 				greedy: true,
 				inside: insideString
 			},
+			// Here-document with quotes around the tag
+			// → No expansion (so no “inside”).
+			{
+				pattern: /((?:^|[^<])<<-?\s*)(["'])(\w+?)\2\s*(?:\r?\n|\r)(?:[\s\S])*?(?:\r?\n|\r)\3/,
```
RunDevelopment: Why is `\w+?` reluctant?
It's always followed by either `"` or `'`, so why not let it be greedy?

components/prism-bash.js
```diff
@@ -19,66 +29,168 @@
...
 			lookbehind: true
 		},
-		'operator': /&&?|\|\|?|==?|!=?|<<<?|>>|<=?|>=?|=~/,
-		'punctuation': /\$?\(\(?|\)\)?|\.\.|[{}[\];]/
+		boolean: {
+			pattern: /(^|[\s;|&]|[<>]\()(?:true|false)(?=$|[)\s;|&])/,
+			lookbehind: true
+		},
+		'file-descriptor': {
+			pattern: /(^|\W)&\d(?=$|\W)/,
```
RunDevelopment: You can replace `(?=$|\W)` with `\b`.
(But not `(^|\W)`!)

components/prism-bash.js
alice-mm: Now that you mention it… These are the characters than can have special meanings when escaped within a string. Well, it ultimately depends on the command that receives the string, but as in 99.9 % of cases the meaning will indeed be special, I thought it would be a shame not to highlight such sequences. This subset is some sort of union of what `echo` and `printf` support. The `"` is for escaped quotes, because string that contain quotation marks tend to get unreadable fast in Bash.

I'll make the existing comment more… human-readable, I guess.

components/prism-bash.js
alice-mm: Woops, an old habit of mine. Never encountered anyone that liked it. :cry:

components/prism-bash.js
alice-mm: I guess you're right. Probably copypasted that from the “without quotes” version, and my lack of confidence prevented me from fiddling with the regex too much.

alice-mm:
Thanks for the comments!

> I will have to learn a little bash to fully review this PR, so if you know a good place to get started, please let me know. Also I found this [spec](https://tiswww.case.edu/php/chet/bash/bashref.html). If there is a better one please let me know.

Well, I learned Bash with a lot of trial and error. The “First do it, then do it better” spirit. Note that [`man bash`](https://linux.die.net/man/1/bash) contains far more information than people tend to think. Info on specific builtins and even stuff like `select` or `for` can be obtained via `help <name>`. The link you provide here seems to be nice, though. Maybe even overkill for this context.

components/prism-bash.js
```diff
@@ -1,9 +1,19 @@
...
+	// - some useless things.
+	var envVarList='BASH|BASHOPTS|BASH_ALIASES|BASH_ARGC|BASH_ARGV|BASH_CMDS|BASH_COMPLETION_COMPAT_DIR|BASH_LINENO|BASH_REMATCH|BASH_SOURCE|BASH_VERSINFO|BASH_VERSION|COLORTERM|COLUMNS|COMP_WORDBREAKS|DBUS_SESSION_BUS_ADDRESS|DEFAULTS_PATH|DESKTOP_SESSION|DIRSTACK|DISPLAY|EUID|GDMSESSION|GDM_LANG|GNOME_KEYRING_CONTROL|GNOME_KEYRING_PID|GPG_AGENT_INFO|GROUPS|HISTCONTROL|HISTFILE|HISTFILESIZE|HISTSIZE|HOME|HOSTNAME|HOSTTYPE|IFS|INSTANCE|JOB|LANG|LANGUAGE|LC_ADDRESS|LC_ALL|LC_IDENTIFICATION|LC_MEASUREMENT|LC_MONETARY|LC_NAME|LC_NUMERIC|LC_PAPER|LC_TELEPHONE|LC_TIME|LESSCLOSE|LESSOPEN|LINES|LOGNAME|LS_COLORS|MACHTYPE|MAILCHECK|MANDATORY_PATH|NO_AT_BRIDGE|OLDPWD|OPTERR|OPTIND|ORBIT_SOCKETDIR|OSTYPE|PAPERSIZE|PATH|PIPESTATUS|PPID|PS1|PS2|PS3|PS4|PWD|RANDOM|REPLY|SECONDS|SELINUX_INIT|SESSION|SESSIONTYPE|SESSION_MANAGER|SHELL|SHELLOPTS|SHLVL|SSH_AUTH_SOCK|TERM|UID|UPSTART_EVENTS|UPSTART_INSTANCE|UPSTART_JOB|UPSTART_SESSION|USER|WINDOWID|XAUTHORITY|XDG_CONFIG_DIRS|XDG_CURRENT_DESKTOP|XDG_DATA_DIRS|XDG_GREETER_DATA_DIR|XDG_MENU_PREFIX|XDG_RUNTIME_DIR|XDG_SEAT|XDG_SEAT_PATH|XDG_SESSION_DESKTOP|XDG_SESSION_ID|XDG_SESSION_PATH|XDG_SESSION_TYPE|XDG_VTNR|XMODIFIERS'
```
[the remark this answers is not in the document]
alice-mm: > Do you really necessitate an extra character after the environment […]?

I guess not. Probably a mistake.

alice-mm:
Here you go, I think. :slightly_smiling_face:
Thanks again!
Gosh, now I'm hungry and it's two in the morning… :woman_facepalming:

components/prism-bash.js
```diff
@@ -1,84 +1,197 @@
...
+			{
+				// a) and c)
+				'pattern': /(\bfunction\s+)\w+(?=(?:\s*\(?:\s*\))?\s*\{)/,
+				'lookbehind': true,
+				'alias': 'function'
+			},
+			{
+				// b)
+				'pattern': /\b\w+(?=\s*\(\s*\)\s*\{)/,
+				'lookbehind': true,
```
RunDevelopment: You don't need that anymore.

components/prism-bash.js
RunDevelopment: > I'll make the existing comment more… human-readable, I guess.

You don't have to. It's clear to me now. Thank you.

alice-mm:
> I didn't mean the object properties

Haha I had huge doubts, yeah, and I thought I was supposed to be consistent. But now that I fixed that, I agree that it's convenient to be able to spot tokens in a glance.

Here, two additional cleaning commits.

RunDevelopment:
Looks good!

Well, that's all I can do for now. I will get back to you when I'm a little more comfortable with bash.

alice-mm:
Rebased to solve conflicts stemming from:
* https://github.com/PrismJS/prism/pull/1731
* https://github.com/PrismJS/prism/pull/1734

(only new command names in the “function” list)
By the way, I was a bit surprised to see that “add” was added to this list because of `yarn add`. This blurs the line between command names and their arguments, and we might end up with the whole English dictionary highlighted as commands. :thinking: Well, the decision was already taken so I went with it and integrated “add” to this branch. :slightly_smiling_face:

components/prism-bash.js
```diff
@@ -1,84 +1,196 @@
...
 		},
 		'boolean': {
-			pattern: /(^|[\s;|&])(?:true|false)(?=$|[\s;|&])/,
+			pattern: /(^|[\s;|&]|[<>]\()(?:true|false)(?=$|[)\s;|&])/,
 			lookbehind: true
 		},
-		'operator': /&&?|\|\|?|==?|!=?|<<<?|>>|<=?|>=?|=~/,
-		'punctuation': /\$?\(\(?|\)\)?|\.\.|[{}[\];]/
+		'file-descriptor': {
+			pattern: /(^|\W)&\d\b/,
```
RunDevelopment: Please make this `/\B&\d\b/`. It will behave the same but doesn't need the lookbehind.

alice-mm:
> @alice-mm
> I'm really sorry for the long delay!

Hi there! :wave:

> I really have to say that this PR is really well done. Everything I threw at it, it handled correctly.

Thanks! Well, you can still nest crazy things and end up with nonsensical colours but 1) Bash is insane anyway and even Visual Studio Code, etc. do a rather poor job at handling many cases and 2) one of my Bash-mottos is that when you have stuff like this it's a sign that you should be creating a bunch of functions instead.

> There are just 2 things I found which need to be improved:
> 
> 1. The comment I left you.
> 
> 2. `builtin` styling.
>        All themes except Tomorrow Night and Twilight highlight `builtin` tokens and `string` tokens with the same color. Syntax highlighting with just one color is kind of useless, so I suggest that we give the `builtin` token a `class-name` alias to change the style.

Oh, yeah, I noticed this wasn't ideal but that was before I actually read the docs about aliases.

I'll try to do that soon during the week. I actually reinstalled the OS of this computer a couple of hours ago and Prism's file are currently being copied back to it, haha.

> Apart from that I only have to say: Good work.
> (And a rebase would be great.)

Oh, yeah. And thanks again.

RunDevelopment:
> Well, you can still nest crazy things and end up with nonsensical colours

Because of the regex nature of Prism, this is always possible, so don't worry about it.
We define regular expressions, so it's actually impossible to match a context-free language with complex nesting correctly.

> I'll try to do that soon during the week.

Take your time ^^

alice-mm:
Couldn't help myself; did that at work during my lunch break. I hope I correctly understood what was to be done. :persevere:

PR #1444, opened 2018-06-16
C# improvements

RunDevelopment:
This PR is mainly about an improved detection of type names and better support for generic parameters in C#.

It also adds support for the `nameof` operator and now highlights type parameters the same as classes.


mAAdhaTTah:
Minor note: you'll need to commit the minified file for C#. Otherwise, I'll defer to my co-maintainers to review the Regex here.

RunDevelopment:
I added support for `foreach` variable declarations and improved the number highlighting.

components/prism-csharp.js
```diff
@@ -1,79 +1,135 @@
...
+				greedy: true
+			},
+			{
+				pattern: /("|')(?:\\.|(?!\1)[^\\\r\n])*?\1/,
+				greedy: true
+			}
+		],
+		'class-name': [
+			{
+				// (Foo bar, Bar baz, Foo[,,] bay, Foo<Bar, FooBar<Bar>> bax)
```
Golmote: I see you separated the tests for class names in multiple files... But I'm having a hard time figuring out which test file applies to which pattern here. Would you mind adding a comment above each example specifying stuff like "Declaration", "Attribute", "Variable parameters", etc.?

components/prism-csharp.js
```diff
@@ -1,79 +1,135 @@
...
+				pattern: /\b[A-Z]\w*(?:\.\w+)*(?:<(?:[^>\n=;{}]+|>(?=\s*[,.>?[)]))+>)?(?:\.\w+)*(?:\[\s*(?:,\s*)*\])*(?=\s+\w+(?:\s*[=,;:{)\]]|\s+in))/,
```
Golmote: This looks complicated. Could you explain what are the different cases you want to match here? The examples listed in the comments above look more simple that what this regexp seems to achieve.

It appears to match weird stuff like `Foo.bar.baz<>,>>>>,,>.foo.bar<?[)?<<>.foo<> test {` and there seems to be some backtracking taking place here and there.

components/prism-csharp.js
```diff
@@ -1,79 +1,135 @@
...
+				lookbehind: true
+			},
+			{
+				// as Foo<A>, is Bar<B>
+				pattern: /(\b(?:is|as)\s+)[A-Z]\w*(?:\.\w+)*(?:<(?:[^>\n=;{}]+|>(?=\s*[,.>?[)]))+>)?(?:\.\w+)*(?:\[\s*(?:,\s*)*\])*/,
+				lookbehind: true,
+				inside: classNameInside
 			}
+		],
+		'number': /(?:\b0(?:x[\da-f_]*[\da-f]|b[01_]*[01])|\B\.\d+|\b\d+(?:_+\d+)*\.?\d*)(?:ul|[flu])?\b/i
```
Golmote: Am I right assuming no underscores are allowed in the decimal part of floats?

tests/languages/csharp/class-name-typeof-default_feature.test
```diff
@@ -0,0 +1,61 @@
...
+		["keyword", "int"],
+		["punctuation", ")"],
+		["punctuation", ">"]
+	]],
+	["punctuation", ")"]
+]
+
+----------------------------------------------------
+
+Checks for class names of type expressions.
```
Golmote: This file should be named `type-expression_feature.test` to reflect the pattern it's testing.

components/prism-csharp.js
```diff
@@ -1,79 +1,135 @@
...
-		{
-			// [Foo]
-			pattern: /(\[)[A-Z]\w*(?:\.\w+)*\b/,
+		'constructor-invocation': {
+			// new List<Foo<Bar[]>> { }
+			pattern: /(\bnew\s+)[A-Z]\w*(?:\.\w+)*(?:<(?:[^>\n=;{}]+|>(?=\s*[,.>?[)]))+>)?(?:\.\w+)*(?:\[\s*(?:,\s*)*\])*(?=\s*[[({])/,
 			lookbehind: true,
+			inside: classNameInside,
+			alias: 'class-name'
+		},
```
Golmote: Could you add a simple test file `constructor-invocation_feature.test` to test this pattern please?
I guess a `type-list_feature.test` would also be needed.

components/prism-csharp.js
```diff
@@ -1,79 +1,135 @@
...
+		'return-type': {
+			// Foo<Bar> ForBar(), Foo IFoo.Bar() => 0
+			pattern: /\b[A-Z]\w*(?:\.\w+)*(?:<[^\n=;{]+?>(?:\.\w+)*)?(?:\[\s*(,\s*)*\])?(?=\s+(?:[A-Z]\w*\.)?(?:\w+(?:\s*<[^\n=;{]+?>)?\s*(?:[({]|=>)|this\s*\[))/,
```
Golmote: I feel like this is redundant with the first class-name pattern defined previously. Are those look-ahead actually needed? Are there cases where a "class name-like" word (for example a capitalized word `Foo`) can appear in C# while not being an actual class name?

tests/languages/csharp/number_feature.test
```diff
@@ -1,15 +1,33 @@
...
+.1f
+1234L
+1234u
+700000000000ul
+0xFFFFFFFFFFFFFFFFul
 0xbabe
 0XBABE
+1_000_000
+0X_BA_BE
+0b_1_0000_0000_0000_0000_0000_0000_0000_0000
```
Golmote: You could add a test with a base-10 number using underscores too.

components/prism-csharp.js
RunDevelopment: You're right. Underscores are allowed everywhere. Even the exponent...

components/prism-csharp.js
RunDevelopment: Properties, some fields, compile time constants (basically fields) and methods are written in Pascal-case. So yes, there are many places in which `Foo` could not be a class. E.g. `Foo.Bar`: `Bar` could be a property, field, method or subclass. There is no way of knowing from the syntax alone.

The `return-type` is a stand-alone pattern because it has to be executed before `generic-method` (which is the look ahead). I know that it might seem complicated but generic methods (`Foo<A, B<C, D, (E, F>, (G, H)> I<J, K, L>();`), explicite implementation (`Foo<A> IBar<B, C<D>>.Bar<E, F>();` and stuff like properties and indexers make things a little more complicated...

RunDevelopment:
Maybe I should explain the main part that matches almost every class name. That is this:

```js
/[A-Z]\w*(?:\.\w+)*(?:<(?:[^>\n=;{}]+|>(?=\s*(?:\[\s*(?:,\s*)*\]\s*)*[,.>?)]))+?>)?(?:\.\w+)*(?:\[\s*(?:,\s*)*\])*/
```

Seems complicated? Things get a little easier when you take out the part responsable for generics. Then we get:

```js
/[A-Z]\w*(?:\.\w+)*(?:\[\s*(?:,\s*)*\])*/
```

Now that's friendlier. (I removed the redundant `(?:\.\w+)*`)
All this does is to match classes, subclasses and their namespace. At the end we include possible array brackets and that's it.

Ok back to the part for generics:

```js
/(?:<(?:[^>\n=;{}]+|>(?=\s*(?:\[\s*(?:,\s*)*\]\s*)*[,.>?)]))+?>)?/
```

The main problem here is that we try to accomplish the impossible. We try to match structures like `Foo<A<B<C<D>, E<F<G>, H>, I<J<K<L>>>>, M>>` which require a context free grammar with regular expressions. 
Have fun with that, because regular expressions can't count the pointy brackets!

And fun I had. What I did, was to say that there are only a limited number of characters which can follow after a `>`, if that `>` is not the one closing the initiale class name.
This yields quite good results but is neither perfect nor very fast. In my testing it's about 2~5 times slower than `(?:<[^>]+>)?` (which can't match nested generic types). In certain edge case it also matches too much but that can be avoided by using lookarounds and cleverly ordering of the different patterns.

I know that this is far from perfect but it work quite reliably in all of my tests and code examples.

RunDevelopment:
I'm also thinking of adding real support for nullables, because with C# 8.0 coming it might be important. 
Needless to say, this will make thinks even more complicated...

tests/languages/csharp/number_feature.test
Golmote: Oops, indeed it is!

components/prism-csharp.js
Golmote: Alright. As long as we have good reasons to do complicated stuff, I'm ok with that ^^

Golmote:
> The main problem here is that we try to accomplish the impossible. We try to match structures like Foo<A<B<C<D>, E<F<G>, H>, I<J<K<L>>>>, M>> which require a context free grammar with regular expressions.
Have fun with that, because regular expressions can't count the pointy brackets!

> In my testing it's on average about 2~5 times slower than (?:<[^>]+>)? (which can't match nested generic types).

Do people really do 5 nested levels of generics in real-world code? The regexp could certainly be faster if we limited Prism's support to a certain number of nested levels (like 2... or maybe 3. Above, the regexp will get really ugly).

RunDevelopment:
> Do people really do 5 nested levels of generics in real-world code?

I certainly hope not, but generics can get messy. (Hopefully not that messy...)

> like 2... or maybe 3. Above, the regexp will get really ugly

```js
/<(?:[^<>]+|<(?:[^<>]+|<(?:[^<>]+|<[^<>]+>)+>)+>)+>/
```

I settled for 4 as the maximum depth. A depth of 3 should be enough for everything I do and 4 should satisfy even hardliners. It's probably a little overkill but better safe than sorry.
I also choose to not do any other validation as before to include things like arrays, tuples and nullables. This means that it also includes trash like `Foo<.............>`, but we're not a code validator, are we?

Btw. That beauty above is 9 characters less than its predecessor.

Golmote:
> but we're not a code validator, are we?

We're definitely not!

The regexp in your last message looks ok to me. I'll review again once the PR gets updated.

RunDevelopment:
Now there is a small problem. Consider the regex for generic methods simplified for a max depth of 2:

```js
/\w+\s*<(?:[^<>]+|<[^<>]+>)+>(?=\s*\()/
```

Just try to match the text

```
for (Int32 i = 0; i < count; i++); foreach (Foo e in Elements);
```

and see how long it takes for you. It will terminate returning that no match was found, but only after **exponential** time.
To match this text it will take my computer approximately 30 minutes...

Any idea how to fix this?

RunDevelopment:
Update:

1. The solution to the backtracking problem was actually quite simple:

    ```js
    /<(?:[^<>]|<(?:[^<>]|<(?:[^<>]|<[^<>]+>)+>)+>)+>/
    ```

    Removing the `+` behind almost all of the `[^<>]` solved it.

1. Namespaces and using directives are now highlighted.
I assumes that using aliases are always used for types. They could be namespaces, but that's rarely done.

1. To write less code I made a `replace` function which inserts the source code of regular expressions into another regular expression. It's done using placeholders and simple text replacement, but it works quite well and helps to save both file size and time. Now we can easily change the definition for commonly used parts of expressions without having to worry if the same definition was also used elsewhere.
It's simple and has its flaws, but it's really useful when handling more complex language definitions. Maybe other language definitions could also profit from this (or a better version of it).

magkal:
Any updates on this PR? Would be awesome if it could get merged within a few months. I realize you seem to be swamped with work.

RunDevelopment:
I will redo this PR once #1538, #1664, and #1679 are through.

RunDevelopment:
So here's an update for this PR. Aside from explicit implementation (which I can't do without #1679) and C# 8.0 ranges, every feature of C# I know about is supported.

Added features include:

- String interpolation
- New C# 8.0 keywords
- Nullable reference types
- Named arguments.

NotoriousRebel:
any updates on this?

RunDevelopment:
It just needs someone to review.

RunDevelopment:
Thanks for the review @Golmote!

Apparently, Travis is taking the day off. I'll try to get it to run and then merge.

RunDevelopment:
Ok. Travis actually [ran and succeeded](https://travis-ci.org/github/PrismJS/prism/builds/666489306?utm_medium=notification&utm_source=github_status) but the status isn't displayed here on GH. 
Interesting.

PR #1446, opened 2018-06-19
Identify JavaScript function parameters

RexSkz:
Support parameters for these types of functions:

```javascript
// es6 class method
foo(x, y) {}
// es6 arrow function
(x, y) => x
x => x
// es5 function
function foo(x, y) {}
// es5 anonymous function
function (x, y) {}
```


components/prism-javascript.js
```diff
@@ -14,9 +14,25 @@ Prism.languages.insertBefore('javascript', 'keyword', {
 	},
 	// This must be declared before keyword because we use "function" inside the look-forward
 	'function-variable': {
-		pattern: /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*=\s*(?:function\b|(?:\([^()]*\)|[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)\s*=>))/i,
+		pattern: /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*=\s*(?:async\s*)?(?:function\b|(?:\([^()]*\)|[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)\s*=>))/i,
```
RunDevelopment: Shouldn't this be `async\s+` ?

RunDevelopment:
First of all, I like this RP. 
We could use this to fix the false positives of keywords in parameter names (like in the [API docs](https://prismjs.com/extending.html#api)). You don't currently do that, but we could.

Now for the execution:

1. I don't get why you don't just put everything you don't want (function name) in lookarounds?
1. We don't you just make one `'parameter': [ /* patterns */ ]`? This will save you from using this many aliases.

Apart from that, I like it.
(Even though this might not equate to much because I'm not a maintainer or anything...)

components/prism-javascript.js
RexSkz: If we use `async\s+`, we couldn't match the variable `foo` in `foo = async(a, b) =>`, where the `async` here is a reserve word instead of the function name.

I found a better solution `async(?:(?:\s*(?=\())|\s+)`.

components/prism-javascript.js
```diff
@@ -14,9 +14,25 @@ Prism.languages.insertBefore('javascript', 'keyword', {
...
 		alias: 'function'
 	},
+	'es5-function-parameter': {
+		pattern: /(function\s*)(?:[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)?\s*(\([^()]*\))/,
```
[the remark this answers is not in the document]
RexSkz: You are right, I'll fix my code. Thanks~

RexSkz:
@RunDevelopment 

1. Do you mean those reserve words in the regex for `es6-class-method-parameter`? I thought in a normal JS code, beside function names, only those words will appear before a `(`. If someone thinks it hard to find all possible words, I will put all JS reserve words in it.
2. I didn't realize I can use an array here, will fix my code soon.

RunDevelopment:
@RexSkz 

> Because we need to identify function name too, function name is also wrapped in parameters. This is not the best idea, but I can't find another way.

I meant that despite the name `parameter`, you not only match the parameters, but also the function name itself. This is somewhat confusing, so I wanted to suggest, that the function name (and trailing spaces) should be included into a lookbehind.
I don't know if this approach given us trouble identifying the function name itself, so maybe I'm overlooking something?

RexSkz:
@RunDevelopment

I just found another way to exclude "function name" from "parameters": exclude `()` from "parameters" at the same time so Prism can keep using `(` to recognize "function".

A new commit was just created.

components/prism-javascript.js
```diff
@@ -19,16 +19,22 @@ Prism.languages.insertBefore('javascript', 'keyword', {
 	},
 	'parameter': [
 		{
-			pattern: /(function)(?:\s+[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)?\s*(\([^()]*\))/,
+			pattern: /(function(?:\s+[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)?\s*\(\s*)\S[^()]*(?=\s*\))/,
```
RunDevelopment: 2 Problems here:

1. You probably don't want trailing spaces, so please make it `[^()]*?`. Otherwise `(?=\s*` won't match anything.
2. This also matches `function a())`. Not valid JS, I know, but still. Replace `\S` with `[^\s()]` and your good to go.

components/prism-javascript.js
```diff
@@ -19,16 +19,22 @@ Prism.languages.insertBefore('javascript', 'keyword', {
...
 			inside: Prism.languages.javascript
 		},
 		{
-			pattern: /(\([^()]*\)|[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)(?=\s*=>)/,
+			pattern: /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*=>)/,
 			inside: Prism.languages.javascript
 		},
 		{
-			pattern: /\b(?!await|delete|export|for|if|new|return|switch|throw|typeof|while|yield)(?:[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*\s*)(\([^()]*\))(?=\s*\{)/,
+			pattern: /(\()[^()]+(?=\)\s*=>)/,
```
RunDevelopment: Leading and trailing spaces might not be wanted.

RexSkz:
@RunDevelopment 

All fixed. Is there any other problems?

RunDevelopment:
@RexSkz 

Not problems, just the suggestion to replace `(?:async(?:(?:\s*(?=\())|\s+))?(function` with `(?:async\s*)?(?:\bfunction`. It's less verbose and easier to understand imo.

At last, a question:
One of your expressions uses a bunch of keywords to avoid matching stuff like `if(abc){}`. Why did you use this specific subset of JS keywords?

````js
/((?:\b|\s|^)(?!(?:await|delete|export|for|if|new|return|switch|throw|typeof|while|yield)(?![$\w\xA0-\uFFFF]))(?:[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*\s*)\(\s*)[^\s()][^()]*?(?=\s*\)\s*\{)/

**Comment by RexSkz** [comment]
@RunDevelopment 

1. Your suggestion seems good, I'll replace my regex.
2. In normal JS code, beside function names, only those words will appear before a `(`. Maybe putting all JS keywords here is a better idea? But it will make this regex much longer.

**Comment by mAAdhaTTah** [comment]
@RunDevelopment Besides the conflict, is this PR ready to go?

**Comment by mAAdhaTTah** [comment]
@RunDevelopment I'm OK with this PR in the main JS language; maybe that PR is extensive enough to break it out into a separate language?

**Comment by atomiks** [comment]
This doesn't appear to work with destructuring:

```js
function Example({ prop }) {}
```

RexSkz:
@atomantic Yes, it's my fault. :-/

Prism now does work with destructuring, but I forgot to add `i` flag to regex, so it fails when your function name (which I use `[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*` to recognize) starts with an uppercase letter. Will make a new PR to fix it.

PR #1449, opened 2018-06-22
Add Python triple-quoted strings "known failure"

brandon-rhodes:
Documenting a failure that several of our users have run into,
since I’m not sure at first glance how to repair prism.js’s
Python rules without maybe breaking normal strings.

### DIFF
--- a/examples/prism-python.html
+++ b/examples/prism-python.html
@@ -49,3 +49,16 @@ are supported.'''</code></pre>
 if __name__ == '__main__':
     import doctest
     doctest.testmod()</code></pre>
+
+<h2>Known failures</h2>
+<p>There are certain edge cases where Prism will fail.
+	There are always such cases in every regex-based syntax highlighter.
+	However, Prism dares to be open and honest about them.
+	If a failure is listed here, it doesn’t mean it will never be fixed. This is more of a “known bugs” list, just with a certain type of bug.
+</p>
+
+<h3>Triple-quoted strings with what look like strings inside</h3>
+<pre><code>def antique(string):
+    """Replace anachronistic Latin "j" with "i"."""
+    return string.replace("j", "i").replace("J", "I")
+    </code></pre>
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Let me know if there's a better way to get this documented for Python users of your library — or if there's a quick fix that I wasn't yet able to see from the examples.

examples/prism-python.html
```diff
@@ -48,4 +48,17 @@ <h2>Full example</h2>
...
+
+<h2>Known failures</h2>
+<p>There are certain edge cases where Prism will fail.
+	There are always such cases in every regex-based syntax highlighter.
+	However, Prism dares to be open and honest about them.
+	If a failure is listed here, it doesn’t mean it will never be fixed. This is more of a “known bugs” list, just with a certain type of bug.
+</p>
+
+<h3>Triple-quoted strings with what look like strings inside</h3>
+<pre><code>def antique(string):
```
mAAdhaTTah: The `pre` needs the `language-python` class.

brandon-rhodes:
There, I added it. Thanks!

mAAdhaTTah:
Thanks!

brandon-rhodes:
(By the way — none of the other `<pre>` tags in that file have the Python language specified, in case that's something you need to fix up later.)

mAAdhaTTah:
Now that I'm looking closely, it turns out the class wasn't needed because of the way the examples page works. It injects your HTML page in a `section` tag with the correct class, and Prism inherits from that. I removed it on develop for consistency.

examples/prism-python.html
Golmote: Sorry for getting there so late! Actually it does not. Examples files are included dynamically, and they are all wrapped in a `<div class="language-XXX">` element, using the language the example is for. (See https://github.com/PrismJS/prism/blob/master/examples.js#L69-L73)

mAAdhaTTah: Yeah, I caught that, see my comment below.

Golmote: Oh, sorry. I stopped reading at your "Thanks!" comment. x) Please ignore me.

PR #1450, opened 2018-06-23
Identify CSS units and variables

RexSkz:
Assume CSS unit is a word or `%` after a number, CSS variable is a word inside `var()`.

I think `%` should not be recognized as a part of number, but the leading `-` should.

When using minus operator in `calc` function, we must type a space in both sides of `-` (value like `calc(100%-5px)` is wrong), so if we met a pattern like `-[\d.]` in a CSS value (not in selector, rule or variable), it must be the start of a negative number.


RexSkz:
Hi, this PR can now recognize operators in css-extras~

dutchenkoOleg:
when it would be merged? i need it

components/prism-css-extras.js
```diff
@@ -10,7 +10,19 @@ Prism.languages.css.selector = {
 };
 
 Prism.languages.insertBefore('css', 'function', {
+	'variable': {
+		pattern: /(var\()[^)]+(?=\))/,
```
RunDevelopment: This pattern will also include [possible default values](https://www.w3.org/TR/css-variables/#using-variables).

But we can use the naming convention of [custom properties](https://www.w3.org/TR/css-variables/#defining-variables) to make a simple pattern which matches all variables accessed via `var` or not.

> A custom property [= variable] is any property whose name starts with two dashes [...].

Using that, `/(^|[^-\w\xA0-\uFFFF])--[-_a-z\xA0-\uFFFF][-\w\xA0-\uFFFF]*/i` should do the job.

Depending on whether we insert it before `property` we can also highlight the custom property declarations.

mAAdhaTTah: Gah, hit merge too quickly!

RunDevelopment:
Too late it seems.
I'll make a new PR.

components/prism-css-extras.js
RunDevelopment: It's alright, I review too late.

dutchenkoOleg:
Finaly! thx!

PR #1461, opened 2018-07-03
add Vue support

riophae:
Vue SFCs look like this:

vue-component

Actually it's just a special HTML. See: https://vuejs.org/v2/guide/single-file-components.html

### DIFF
--- a/components.json
+++ b/components.json
@@ -48,12 +48,13 @@
 		},
 		"markup": {
 			"title": "Markup",
-			"alias": ["html", "xml", "svg", "mathml"],
+			"alias": ["html", "xml", "svg", "mathml", "vue"],
 			"aliasTitles": {
 				"html": "HTML",
 				"xml": "XML",
 				"svg": "SVG",
-				"mathml": "MathML"
+				"mathml": "MathML",
+				"vue": "Vue"
 			},
 			"option": "default"
 		},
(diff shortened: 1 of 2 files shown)


components.json
```diff
@@ -48,12 +48,13 @@
...
 			"title": "Markup",
-			"alias": ["html", "xml", "svg", "mathml"],
+			"alias": ["html", "xml", "svg", "mathml", "vue"],
 			"aliasTitles": {
 				"html": "HTML",
 				"xml": "XML",
 				"svg": "SVG",
-				"mathml": "MathML"
+				"mathml": "MathML",
+				"vue": "Vue",
```
mAAdhaTTah: The comma at the end of this line is invalid JSON. That's why the build is failing.

riophae: Fixed.

Golmote:
I'm not sure I'm in favor of this addition. Vue is not a language is it? It does not even have a special syntax?

What issue is this PR supposed to fix? It it to be able to add class `language-vue` to code blocks?

riophae:
@Golmote Yeah... Sort of like that.

Golmote:
@mAAdhaTTah I'm not worried about the filesize, we have minified versions to deal with this (obtw it appears to be missing from this PR). I'm worried about the semantics...

I went through the VueJS doc, and I see it has a Mustache-like template syntax. If we were to add a component for this (maybe it could handle multiple templating languages depending on how similar they are), we could then consider it a "language". But just adding `vue` as a alias for `markup` seems not appropriate to me.

riophae:
@Golmote I agree with you.

riophae:
For example, the current `markup` definition doesn't handle this well:

```
<div>{{ look_at_me }}</div>
```

PR #1462, opened 2018-07-06
Updated index.html to fix broken instructions

Shivang44:
Requiring components without the "index.js" part and calling "loadLanguages" results in a function not found error for me. Adding index.js, as commented by @golmote, fixes that error.

Ref: https://github.com/PrismJS/prism/issues/972


Golmote:
Hi! Does `require('prismjs/components/')` (with the trailing slash) works for you? I think it's what was meant to be written.

Shivang44:
Hey, thanks for the quick response @Golmote. That does actually fix it! I've revised my PR accordingly

Golmote:
Thanks for contributing!

Issue #1451, opened 2018-06-23
scheme language issues


jcubic:
So I found few issues:
* [x] string don't start with `'` in scheme `'` is for quote (you can quote names, which become symbols or lists) (#2263)
   solution: remove `|'[^('\s]*` from string regex
* [x] lambda,define have first parameter mark as function invocation `(lambda (foo bar) (concat foo bar))` foo is mark as function, I've try to came up with regex but was not able to find one. (#2263)
* [x] function without argument is not marked as function
   solution: `/(\()[^\s()]*(?=[\s)])/,` added `(?=[\s)])` instead of `(?=\s)` (#1463)
* [x] ~numbers that don't have space before them are not marked as numbers
  solution: optional space in the beginning of number regex `(\s?|[()])`~

Wanted to create PR but don't know how to fix function with define and lambda.

Golmote:
Hi @jcubic. You can definitely submit a PR fixing 3 bugs out of 4, and you are very welcome to do so!

Regarding the `lambda` case, it might be tricky. The negative lookbehind suggested in #1445 might come in handy to solve such a case.

jcubic:
It seems that numbers are correct and string regex don't match single quote string but symbols, so only function without arguments is not marked as function, will create PR.

jcubic:
Found another issue. In scheme there are no operators only functions and if you're adding function `**` it's marked as operator and nothing (span without `*&nbsp;`). will create PR for this as well.

jcubic:
Found another 2 issues with scheme mode:
* strings in scheme can be multiline
* function name that start as keyword like define is not mark as function (to fix this the regex first need to check for function name and after if fail it need to check for keyword like define or lambda)

Example:
```scheme
(display (list->array '(1 2 3 "foo
bar
baz")))
(list 1 2 3)
```
on GitHub multiline strings works and list->array don't have list highlighted (on GitHub scheme syntax function names are black)

jcubic:
I can create PR with the fix last one was not merged, @Golmote did you notice it?

RunDevelopment:
@jcubic 
Number 2 is already addessed in #1556.
(And Prism doesn't split the code by lines before processing.)

jcubic:
@RunDevelopment does this PR fixes function names like `==` that is marked as operator?

jcubic:
Found another issue:
```scheme
(define 1+2 10)
```

[...]

PR #1463, opened 2018-07-07
fix function without arguments in scheme language

jcubic:
This is partial fix to issue #1451 the function `(foo)` is not marked as function, numer was correct and strings token was for symbols.


Golmote:
Thanks!

Could you add a test for it in https://github.com/PrismJS/prism/blob/master/tests/languages/scheme/function_feature.test please?

Golmote:
Thanks!

PR #1464, opened 2018-07-07
Fix empty triple-quoted strings and broken example

brandon-rhodes:
* Fix: triple-quoted strings can be empty strings in Python.

* Correction: the supposedly “broken” example Python I recently added
  actually gets correctly highlighted by prism.js; so here’s a correct
  reproduction of the failure mode we’ve been seeing in some files.


tests/languages/python/triple-quoted-string_feature.test
```diff
@@ -7,13 +7,19 @@ baz"""
 #bar
 baz'''
 
+Python supports empty triple quoted strings
```
Golmote: This is probably no need for the sentence here. The tests makes it obvious. ;)

brandon-rhodes:
Okay, I was able to solve the problem: only accept `#` as a comment character if any simple strings that get opened previously on the line also get closed.

brandon-rhodes:
(I didn't try to be fancy and use back-references for single-vs-double-quoted strings because then I couldn't quite read the RE and convince myself it was correct. And ­— are back-references like `\1` supported inside `[^...]`? Anyway, it seemed simpler to read this way.)

myjian:
I also encounter this issue. Can we merge this PR?

mAAdhaTTah:
I had previously labeled this needs changes, but it appears those changes were actually integrated.

@Golmote Are you still interested in looking into that "greedy" fix or are we good to merge this as-is?

RunDevelopment:
Python's triple-quoted strings and example page have been fixed and updated.

Issue #1445, opened 2018-06-16
Negative lookbehind


RunDevelopment:
It would be great the have a negative lookbehind for language definitions.

RunDevelopment:
I'd like to suggest the following system:

```js
'positive-number': {
    pattern: /(-)?\d+/,
    lookbehind: true,
    lookbehinddirection: false
}
```

The default value for `lookbehinddirection` is `true`, representing a positive lookbehind. `false` will represent a negative lookbehind.

The first group will be considered a match, if it matches anything (including the empty string), whatever `lookbehinddirection` may be. The rest of the expression will be considered a match will if `firstGroupMatches ^ !lookbehinddirection` is true. (This of course assumes that the whole expression matches in the first place.)

This is somewhat different from the current implementation, as a non-matching first group of a positive lookbehind will currently cause the whole expression to be a match.


(As a less verbose alternative for `lookbehinddirection: false`, one could use `lookbehind: -1`)

Golmote:
@RunDevelopment I like this idea. Regarding the wording, I'd prefer something like `lookbehingNegative: true` (defaults to `false`) rather than any of you propositions.
Apart from this, your implementation concept looks quite straightforward. Would you be interested in submitting a PR for this?

RunDevelopment:
Sure, I can take a shot at implementing this. 
Also: Yes, `lookbehingNegative` is better.

RunDevelopment:
I will now close this for similar reasons as #1465.

1. It will add a lot of complexity and will only ever be used by a small number of languages.
2. I can not think of any way to implement this efficiently.
3. I have written dozens of languages and never needed it because...
4. Current positive lookbehinds can already do negation. To go with my motivating example: Using only positives lookbehinds, positive numbers can be matched using: `/(^|[^-])?\d+/`. This is a trick I regularly use.

    More general: To get the Prism lookbehind of `(?<!R)` for any regular expression `R`, you have to find the complement of `^[\s\S]*R`. This isn't easy but it is always possible.
    E.g. `(?<!a)` ~= `(^|[^a])`, `(?<!#\s*)` ~= `((?:^|[^#\s])\s*)`

PR #1465, opened 2018-07-07
Negative lookbehind

RunDevelopment:
# Negative lookbehind

This PR adds negative lookbehinds (NLB), completing the set of lookarounds. 

A pattern can be declared to have an NLB by adding `negativeLookbehind: true`. The first capturing groups will then be interpreted as an NLB group.

Example:

```js
'positive-number': {
    pattern: /(-)\b\d+/,
    negativeLookbehind: true
}
```

The pattern will be considered matching if the NLB group does not match.

Examples:

```
123    <-- matches
-123   <-- does not match
```

Despite my original proposal (#1445), I choose to go with a non-optional capturing groups to be consistent with current positive lookbehinds (PLB). (The syntax and behavior of PLBs do not change.)

## Implementation

Under the hood, the capturing group of an NLB will be made optional, if that is not already the case. (This also means that these groups can be optional.)

This operation requires the source of regular expression to be understood (at least in parts) and that alone constitutes about half of the new code (namely `spanOfFirstCapturingGroup` and `makeFirstOptional`). 
The whole thing could be avoided if we were to require the NLB group to be optional. But this is inconsistent with PLBs and solely because of the current implementation.
If somebody knows how to cut down on the extra code, I'm all ears.

To match a pattern with an NLB, it will be executed until:

1. The pattern fails to match. In this case, the pattern does not match the text. Return `null`.
1. The pattern matches but the NLB group does not. Return the match.

If the NLB group matches, `lastIndex` will be used to advance the position and the process repeats. 

The new value of `lastIndex` will be `match.index + match[0].length + 1`.
To illustrate why, consider the pattern `/(if\s*)\(/` which is supposed to match all opening braces not preceded by an `if`. 
Let's assume we found `if (true)` (not a match). How do we choose the next `lastIndex`? 
It has to be after the text the NLB group matched because otherwise, we will match the opening brace in the next round. It can't be `match.index + match[0].length` for the same reason.
As a simple solution, I added `+1`. This will work in most cases.

## Known bugs

Consider the following pattern with an NLB: `/(bb)bb/`, which will be converted to `/(bb)?bb/` under the hood. 
One would expect that for the string `bbbbb`, the first two b's (`[bb]bbb`) are matched, but that is not the case. Due to the greedy nature of the RegEx engine, the NLB group will match the first two b's and the rest of the expression will match the next two b's (`[(bb)bb]b`). And because the NLB group matches something, the whole match will not be considered matching.

This might be bad enough, but it gets even worse!
Because the expression did not match due to the NLB, it will keep trying. The next value for `lastIndex` will be `match.index + match[0].length + 1` which is `3` in this case.
Now our expression will happily match the last two b's (`bbb[bb]`). There is no space for the optional group that is our NLB, so the whole expression is considered a match. Wonderful!

This example shows that my implementation is not equivalent to real NLB. (To be fair: Our PLBs aren't equivalent to the real ones either.)

The problem will only occur when any suffix of the NLB can be a prefix of the rest of the expression.
This will rarely happen with patterns for language definition, but we have to be aware of this behavior.


RunDevelopment:
I will close this now because:

1. It adds a lot of code to Prism Core
2. Only a few languages will use this feature. (Most negative-lookbehind-things are also possible to do with positives lookbehinds.)
3. The implementation is also quite slow as there are a lot of matches which get thrown away.

If anyone wants to implement this feature anyway, feel free to use this as a base.

PR #1466, opened 2018-07-08
Fix Command Line and Autoloader used together

### LINKED ISSUE #1456: command-line plugin not showing data-user and data-host
Hi all,
first of all thank you for this great product!
I'm happily using prismjs since version 1.10.0, but with 1.15.0 i'm facing some glitches on command-line plugin.
After some digging i found that command line plugin 1.13.0 is working as expected but 1.14.0 not (probably my mistake).

```
<pre class="command-line" data-title="command line" data-user="matteo" data-host="portable" data-output="2-99">
  <code class="language-bash">
  du -sh * | sort -h | tail
  6.0M	upload
  8.5M	contratti
  11M	accettazione
  12M	report
  14M	copia.tar.gz
  18M	fatturazione
  21M	admin
  22M	amministrazione
  23M	reparti
  198M	codifica
  </code>
</pre>
```
In 

that with 1.13.0 command-line plugin:
image

with 1.14.0:
image

As you can see i exaggerate  the output line value (99) to simply add lines without to worry to increment the value.
If you want to check the initialization, the syntax, etc. you can use http://matteo.contri.net/rifattorizzazione-software-php as reference (sorry is in Italian).

Golmote:
This PR fixes #1456 where, using the Autoloader, the Command Line plugin processes the code one first time with no grammar, and its "before-insert" hook never gets executed, leading to all output lines being lost in the void.

I added a call to the "before-insert" hook when there is no grammar.
Also, the Command Line plugin must not abort when it finds the `command-line-prompt` element, but instead update its content.

This will also, consequently, allow to use the Command Line plugin with `language-none`, just like before 1.14.


mAAdhaTTah:
This seems fine; there is a bit of duplication in the calls into the hooks in that area of the code now though. Is that something we should address?

components/prism-core.js
```diff
@@ -227,7 +227,9 @@ var _ = _self.Prism = {
 		if (!env.code || !env.grammar) {
 			if (env.code) {
 				_.hooks.run('before-highlight', env);
-				env.element.textContent = env.code;
+				env.highlightedCode = env.code;
+				_.hooks.run('before-insert', env);
+				env.element.textContent = env.highlightedCode;
 				_.hooks.run('after-highlight', env);
```
mAAdhaTTah: Can we dedupe the hooks code here?

tarach:
Any updates regarding this issue ? I just run into it. Can't figure out any reasonable workaround.

mAAdhaTTah:
We haven't revisited this, but this PR would need a refresh if we wanted to get it landed. You could use this as a starting point and open a new PR if you were interested in taking that on.

PR #1467, opened 2018-07-08
Mark operator like functions as functions in scheme

jcubic:
This one more issue fix found in #1451

this one mark `(** 10)` as function not operator `*` and then blank `*&nbsp;`

there is `(?=\s)|$` in operator so it match for space look ahead but also end of the string so it match operator test file. I've also added the same to functions so it's highlighted when you start typing function call like `(fun` this will be mark as function.


components/prism-scheme.js
```diff
@@ -18,12 +18,12 @@ Prism.languages.scheme = {
 	},
 	'boolean' : /#[tf]/,
 	'operator': {
-		pattern: /(\()(?:[-+*%\/]|[<>]=?|=>?)/,
+		pattern: /(\()(?:[-+*%\/]|[<>]=?|=>?)(?:(?=\s)|$)/,
```
RunDevelopment: Could you please make it `(?=\s|$)`?
It's shorter because the non-capturing group can be removed.

components/prism-scheme.js
```diff
@@ -18,12 +18,12 @@ Prism.languages.scheme = {
...
 		lookbehind: true
 	},
 	'function' : {
-		pattern : /(\()[^\s()]*(?=[\s)])/,
+		pattern : /(\()[^\s()]+(?:(?=[\s)])|$)/,
```
RunDevelopment: Same as above but `(?=[\s)]|$)` this time.

jcubic:
Updated the code as requested, tests are passing (it seems that travis is enabled only on gh-pages).

RunDevelopment:
Thank you! 
But this still needs rebasing before we can merge this.

I don't know why Travis is taking the day off but it _should_ test your branch.

RunDevelopment:
Perfect!

Thank you for contributing!

PR #1468, opened 2018-07-09
Improved javascript regex highlight

RexSkz:
Improve JS regex highlight according to references in https://regexr.com/.

Note: this PR can't recognize `capture group` because "matching bracket" is not a thing where a pure regex way can do.


components/prism-javascript.js
```diff
@@ -9,6 +9,24 @@ Prism.languages.javascript = Prism.languages.extend('clike', {
...
+				pattern: /\[(?:\\\\|\\\]|[^\]])*\]/,
+				alias: 'escape'
+			},
+			'regex-escape': {
+				pattern: /\\./,
+				alias: 'escape'
+			},
+			'regex-punctuation': {
+				pattern: /[|^$*+(){}]|(?:\?(?![:!=<]))/,
+				alias: 'operator'
```
RunDevelopment: `regex-punctuation`, alias `operator`. What?
I know you did it for the highlighting, but this is somewhat confusing...

components/prism-javascript.js
```diff
@@ -9,6 +9,24 @@ Prism.languages.javascript = Prism.languages.extend('clike', {
...
+			'regex-flags': {
+				pattern: /[gimyu]{1,5}$/,
+				alias: 'keyword'
+			},
+			'regex-group': {
+				pattern: /\[(?:\\\\|\\\]|[^\]])*\]/,
+				alias: 'escape'
+			},
+			'regex-escape': {
+				pattern: /\\./,
```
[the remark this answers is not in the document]
RexSkz: In highlight they are the same type of chars, maybe "escape" is a bad name, but I couldn't find a better one. Any ideas?

I'll modify this regex to support them. (BTW `\0123` will not be treated as one char in JS, which only supports `\00`~`\077`.)

components/prism-javascript.js
RexSkz: I admit that "punctuation" and "operator" is also a bad name, maybe "regex-token" is better?

And I found that VSCode highlights quantifiers (such as `{1,2}`) just like the other "punctuations", I'll add this situations later.

Removing the alias and changing the name is easy, but the matched characters will not have special color unless we modify the themes...

components/prism-javascript.js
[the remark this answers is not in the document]
RunDevelopment: > BTW `\0123` will not be treated as one char in JS, which only supports `\00`~`\077`

Uups, my bad.

components/prism-javascript.js
```diff
@@ -9,6 +9,28 @@ Prism.languages.javascript = Prism.languages.extend('clike', {
...
+			'regex-charset': {
+				pattern: /\[(?:\\.|[^\\\]])*\]/,
+				alias: 'escape'
+			},
+			'regex-escape': {
+				pattern: /\\(x[0-9a-f]{2}|u[0-9a-f]{4}|u\{\d+\}|0[0-7]{1,2}|c[a-z]|.)/i,
+				alias: 'escape'
+			},
+			'regex-quantifiers': {
+				pattern: /\{\d+(,\d+)?\}/,
```
RunDevelopment: This doesn't match `a{1,}` and `a{,1}`.

RunDevelopment:
Regarding the quantifiers:
I think that it's a good idea to keep them separately, but `*`, `+` and `?` (for optional) are also quantifiers, so we should include them as well. 
And when we include those, then maybe we should include `?` (for lazy) as well? I'm not really sure, what do you think?

Regarding the escapes:
I still think that we should put anchors and default character classes into their own category, as they are not really escape sequences.

components/prism-javascript.js
RunDevelopment: Yeah, editing the themes is a little overkill...

components/prism-javascript.js
```diff
@@ -9,6 +9,28 @@ Prism.languages.javascript = Prism.languages.extend('clike', {
...
+				alias: 'operator'
+			},
+			'regex-token': {
+				pattern: /[|^$*+(){}]|(?:\?(?![:!=<]))/,
```
RunDevelopment: Sorry, but I don't quite get what `token` is supposed to match.
It includes alternation `|`, anchors `^$`, quantifiers `+*?{}`, groups `()` and the lazy modifier `?`. Also, the matching for `?` is a little strange: `a?=`, `a*?!` are both valid and in neither `?` is matched.

Also, what about the modifiers of groups? E.g. `(?:)`, `(?=)`, `(?!)`.

RexSkz:
I started to think that highlighting regex is not a simple job, maybe I should treat regex as a kind of "pseudo language" rather than just improve highlight for some situations. I'll keep trying.

RexSkz:
@RunDevelopment 

Hi, I have re-written my code according to the reference in https://regexr.com.



In this reference, `?` for both optional and lazy are called "quantifier".

Now the code can recognize `flag`, `character set`, `escape`, `reference`, `anchor`, `quantifier` and `alternation`.

It's a pity that recognizing `capture group` is not a thing where pure regex way can do, because regex can't be used for matching bracket. VSCode can't do this either, it can be proved from here (the class `.mtk6` means normal character):

RexSkz:
@RunDevelopment 

Run `escape` first will seperate the regex (for example `[a\\b\\c]` will be seperated to `[a`, `b` and `c]`), so that `\\b\\` won't be wrapped in `charset`.

Highlighting stuff inside the character set was not taken into consideration at first, but if it's necessary, using `inside` (and extra patterns) is a way. The problem is: what should I name the pattern like `a-z` or the `^` right after `[`?

Other suggestions seems great, I'll make changes to my code tomorrow, it's midnight here now :(

components/prism-javascript.js
```diff
@@ -9,6 +9,28 @@ Prism.languages.javascript = Prism.languages.extend('clike', {
 Prism.languages.insertBefore('javascript', 'keyword', {
 	'regex': {
 		pattern: /((?:^|[^$\w\xA0-\uFFFF."'\])\s])\s*)\/(\[[^\]\r\n]+]|\\.|[^/\\\[\r\n])+\/[gimyu]{0,5}(?=\s*($|[\r\n,.;})\]]))/,
+		inside: {
+			'regex-flag': {
+				pattern: /(\/)[gimyu]{1,5}$/,
+				lookbehind: true
+			},
+			'regex-charset1': {
+				pattern: /(\[(\\\]|[^\]])*\])|(\\[wds])/i,
```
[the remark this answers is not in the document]
RunDevelopment:
```js
/((?:^|[^\\])(?:\\\\)*)\[(?:\\.|[^\\\]])*\]/
```

This should work with `lookbehind: true` and resolve the issue that escaped `[` were matched.
I haven't thoroughly tested it, so there might still be a problem.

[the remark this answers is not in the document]
RexSkz: Wow, it's a great idea to match the even number of leading `\`!

If a `\` is right after any characters except `\`, it means the beginning of an escape character, so I think your pattern will work correctly.

RexSkz:
@RunDevelopment 

I have pushed another commit, which fixes the bugs and provides more test cases. The most important thing is: it can recognize stuff inside charset now!

components/prism-javascript.js
```diff
@@ -10,21 +10,31 @@ Prism.languages.insertBefore('javascript', 'keyword', {
...
+			'regex-flag': /[gimyu]{1,5}$/,
+			'regex-charset': {
+				pattern: /((?:^|[^\\])(?:\\\\)*)\[(?:\\.|[^\\\]])*\]/,
+				lookbehind: true,
+				inside: {
+					'regex-range': {
+						pattern: /((?:^|[^\\])(?:\\\\)*)\\?.-\\?./,
+						lookbehind: true
+					},
+					'regex-nageted': {
```
RunDevelopment: *negated

components/prism-javascript.js
```diff
@@ -10,21 +10,31 @@ Prism.languages.insertBefore('javascript', 'keyword', {
...
-			'regex-charset2': {
-				pattern: /\./,
-				alias: 'regex-charset'
+			'regex-flag': /[gimyu]{1,5}$/,
+			'regex-charset': {
+				pattern: /((?:^|[^\\])(?:\\\\)*)\[(?:\\.|[^\\\]])*\]/,
+				lookbehind: true,
+				inside: {
+					'regex-range': {
+						pattern: /((?:^|[^\\])(?:\\\\)*)\\?.-\\?./,
```
RunDevelopment: This has problems with ranges like `\xA0-\uFFFF`.

To do this correctly one will have to look out for escape sequences, include them as well into `range` and then match them inside again as `escape`. 
Troublesome.

Maybe we want to leave that one out?
The pattern will be ridiculously long...

components/prism-javascript.js
```diff
@@ -10,21 +10,31 @@ Prism.languages.insertBefore('javascript', 'keyword', {
...
+					'regex-reference': /\\[1-9]/,
+					// things get different, some characters become normal characters
+					'regex-charclass': /\\[wds]/i,
+					'regex-anchor': /\\b/i,
+				}
 			},
-			'regex-anchor': /(\^|\$|\\b)/i,
+			'regex-escape': /\\(x[0-9A-Fa-f]{2}|u[0-9A-Fa-f]{4}|u\{\d+\}|c[A-Za-z]|0[0-7]{1,2}|[^WwDdSsBb\d]|0)/,
+			'regex-reference': /\\[1-9]/,
+			'regex-charclass': /(\\[wds])|\./i,
```
RunDevelopment: Why the group?

RexSkz:
@RunDevelopment 

All issues are fixed, `range` is removed.

RunDevelopment:
@RexSkz 

Looks good!

From a correctness point of view, this PR is ready for merge, but the maintainers might have a thing or two to say about it.

mAAdhaTTah:
Maybe this should be combined w/ #1714 and put into a `js-extras` language?

RunDevelopment:
This is already captured by #1682.

@RexSkz
Thank you for your hard work!

PR #1474, opened 2018-07-11
Java improvements

RunDevelopment:
A bunch of small improvements for the Java language.

1. Class highlighting based on [naming conventions](http://www.oracle.com/technetwork/java/javase/overview/codeconventions-135099.html). Because of these conventions, we know that every name which starts with an uppercase letter and contains some lower case letters afterward is a class, interface or enum.
2. Nested generics are now supported. The contents of a generic statement are no longer highlighted as a `function`, but as a `class-name` which is more fitting.
3. Packages will now be highlighted as `namespace`.
4. E.g. `foo::bar`, `bar` will be highlighted as a function.
5. ~Support for `var`.~ #1549
6. `null` is now a keyword. As it should be.


RunDevelopment:
7. I added support for [underscores in numberic literals](https://docs.oracle.com/javase/7/docs/technotes/guides/language/underscores-literals.html).
1. There are now tests for packages and functions (normal and `::` syntax)

PR #1479, opened 2018-07-13
[json] Fix numbers to include negative numbers

vizjerai:
- Fix numbers to include negative numbers.
- Remove hexidecimal from numbers it is not a valid value.
- Remove `)` and `;` from punctuation they are not valid punctuation.
- Include missing tests for punctuation and operators.


tests/languages/json/number_feature.test
```diff
@@ -4,6 +4,9 @@
 5.0e8
 0.2E+2
 47e-5
+-1.23
+-2.34E33
+-4.34E-33
```
RunDevelopment: What's up with the `+`?

mAAdhaTTah: That's a git diff, saying those lines were added :D

RunDevelopment: ... Ups.
My bad

mAAdhaTTah:
@LeaVerou @Golmote These regexes look like simplifications to me, and tests all pass, so I'm going to merge this. If you think there are things to change, would love to get your feedback, as I'm not great at regex.

LeaVerou:
Looks good to me. I suspect this was copied from JS, and the punctuation was never adjusted.

Btw we should probably have common tokens as re-usable variables or something (E.g. `Prism.NUMBER`). Currently every language needs to redefine numbers, and in most cases they're the same... What do you think?

mAAdhaTTah:
Opened an issue to track (#1494).

PR #1482, opened 2018-07-16
JS: Highlighting of supposed classes and functions

### LINKED ISSUE #1471: JS: Highlighting of supposed classes and/or functions?
When reading JS code, you sometimes come across a line like this:

```js
Foo.prototype.foobar = function() { ... };
Bar.constructor.apply();
foo.bind(this);
bar.apply(...);
```

We then use our common knowledge of JS and assume that `Foo` is a **class** because its `prototype` field was used to add the **function** `foobar `. Similar for `Bar`, `foo` and `bar`.

Generally: Can't we assume that the word left from `prototype` or `constructor` is a class and that the word to the right of `prototype` or the left of either `apply`, `bind` or `call` is a function?

Of course, this will cause false positives, but it will only do so one very few occasions.

RunDevelopment:
This resolves #1471.

It also extends `function-variable` to include object properties. 
E.g. 
```js
{foo: function() {}}
```


components/prism-javascript.js
```diff
@@ -1,8 +1,12 @@
 Prism.languages.javascript = Prism.languages.extend('clike', {
+	'class-name': [
+		Prism.languages.clike['class-name'],
+		/[_$A-Z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\.(?:prototype|constructor))/
+	],
 	'keyword': /\b(?:as|async|await|break|case|catch|class|const|continue|debugger|default|delete|do|else|enum|export|extends|finally|for|from|function|get|if|implements|import|in|instanceof|interface|let|new|null|of|package|private|protected|public|return|set|static|super|switch|this|throw|try|typeof|var|void|while|with|yield)\b/,
 	'number': /\b(?:0[xX][\dA-Fa-f]+|0[bB][01]+|0[oO][0-7]+|NaN|Infinity)\b|(?:\b\d+\.?\d*|\B\.\d+)(?:[Ee][+-]?\d+)?/,
 	// Allow for all non-ASCII characters (See http://stackoverflow.com/a/2008444)
-	'function': /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*\()/i,
+	'function': /[_$a-zA-Z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*\(|\.(?:apply|bind|call)\()/,
```
mAAdhaTTah: Can we get a test for the apply / bind / call cases?

components/prism-javascript.js
```diff
@@ -1,8 +1,12 @@
 Prism.languages.javascript = Prism.languages.extend('clike', {
+	'class-name': [
+		Prism.languages.clike['class-name'],
+		/[_$A-Z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\.(?:prototype|constructor))/
```
mAAdhaTTah: Tests for prototype / constructor?

mAAdhaTTah:
Test failures :(

RunDevelopment:
Fixed ;)

PR #1484, opened 2018-07-16
Download page

RunDevelopment:
This PR eliminates awkward word breaks.

It also better aligns checkboxes and text for a cleaner look.
I also disabled `hyphens` for the owners of plugins because I don't want to see these names chopped into pieces.

I have tested it on Chrome and FireFox.

## Before
image

## After
image

### DIFF
--- a/download.html
+++ b/download.html
@@ -17,7 +17,7 @@ form label {
 	-webkit-column-break-inside: avoid;
 }
 
-	form label > *:nth-child(2) {
+	form label .name {
 		margin-right: .3em;
 	}
 
--- a/download.html
+++ b/download.html
@@ -11,11 +11,25 @@
 form label {
 	display: block;
 	padding: .2em;
+	padding-left: 1.9em;
+	page-break-inside: avoid;
+	break-inside: avoid;
+	-webkit-column-break-inside: avoid;
 }
 
-form input {
-	margin-right: .7em;
-}
+	form label > *:nth-child(2) {
+		margin-right: .3em;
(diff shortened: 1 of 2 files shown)


download.html
```diff
@@ -11,11 +11,25 @@
...
+
+	form label a.owner {
+		margin-left: 0;
+		hyphens: none;
+	}
+
+	form label input {
+		margin-right: .7em;
+		margin-left: -1.7em;
+	}
```
mAAdhaTTah: The  CSS you added appears indented further than the code around it.

download.html
```diff
@@ -11,11 +11,25 @@
...
+	padding-left: 1.9em;
+	page-break-inside: avoid;
+	break-inside: avoid;
+	-webkit-column-break-inside: avoid;
 }
 
-form input {
-	margin-right: .7em;
-}
+	form label > *:nth-child(2) {
```
mAAdhaTTah: This selector is really brittle. Is there a better way of styling this?

download.html
RunDevelopment: Well, because it is about elements inside `form label`, so it gets indented.
This practice can also be seen further down below.

download.html
RunDevelopment: @mAAdhaTTah Fixed

download.html
mAAdhaTTah: Ah ok, that works.

mAAdhaTTah:
Thanks!

PR #1489, opened 2018-07-19
More variables for better code compression

RunDevelopment:
This PR adds new temporary variables to some language definitions so the minifier can better compress the code.
Usually, this only saves about 100~200 bytes or so.
In the case of `textile`, we actually save about 0.6kb or about 16% of the minified file.

This also fixed the namespace pollution introduced by `django`.

This PR deals with syntactic changes only. If you spot a semantic change, please tell, that's a bug.


components/prism-parser.js
```diff
@@ -1,67 +1,73 @@
...
-		lookbehind: true
-	},
-	'variable': {
-		pattern: /(^|[^^])\B\$(?:\w+|(?=[.{]))(?:(?:\.|::?)\w+)*(?:\.|::?)?/,
-		lookbehind: true,
-		inside: {
-			'punctuation': /\.|:+/
+(function (Prism) {
+
+	var parser = Prism.languages.parser = Prism.languages.extend('markup', {
```
mAAdhaTTah: I think we could update all of the initial assignments like this:

```js
var language = Prism.languages.language = { ... object ... };
```

which would be a little nicer than having it on a separate line.

RunDevelopment: The whole idea of these assignments was that we do not have to write the whole name again, but I agree that your idea seems nicer.

Like this?
```js
Prism.languages.language = { ... object ... };

var language = Prism.languages.language;

Comment (mAAdhaTTah) [reply components/prism-parser.js]: No, I just mean to suggest, when we want to also assign it to a separate variable, do it on one line, rather than two.

**Comment by RunDevelopment** [comment]
@mAAdhaTTah 

Variables are now assigned on a separate line.

**Comment by mAAdhaTTah** [comment]
@RunDevelopment I'm sorry, that's the opposite of what I intended. I think I confused you because I commented on a line that was _correct_, not one to fix. My apologies for that.

To be clear, I like assign the variable as well as to `Prism.languages.language` all on one line, the way you had it. When I saw it, I commented, intending to say "let's do it this way", fixing the ones that were on two (specifically asciidoc, and if there were others that I hadn't gotten to yet).

Sorry about that!

File: components/prism-django.js [hunk 2]
```diff
@@ -1,41 +1,45 @@
...
+				'important': /\b-?\d+(?:\.\d+)?\b/,
+				'variable': /\b\w+?\b/,
+				'punctuation': /[[\];(),.:]/
+			}
 		}
-	}
-};
+	};
+
+	Prism.languages.django = Prism.languages.extend('markup', { 'comment': /(?:<!--|{#)[\s\S]*?(?:#}|-->)/ });
```
mAAdhaTTah: Update `django` as well?

RunDevelopment:
@mAAdhaTTah 
Hahaha. Happens to the best of us.

RunDevelopment:
@mAAdhaTTah 
Ok done.

Unrelated: Maybe we should expand the tests for `django`. I just made a mistake messing up the language definition a little. It wasn't detected by the tests...

mAAdhaTTah:
> Unrelated: Maybe we should expand the tests for django. I just made a mistake messing up the language definition a little. It wasn't detected by the tests...

This would be great. You should be able to add additional tests in a separate PR without conflicting with this one.

mAAdhaTTah:
This looks good to me. A little concerned about introducing a lot of conflicts merging this. There's a PR for plsql I think that will be rough to land after this.

RunDevelopment:
That's right, we should land #1501 before this.

RunDevelopment:
Using a little copy function, I compressed AsciiDoc even further. This should also be a plus in terms of maintainability.

mAAdhaTTah:
@RunDevelopment If we resolve these conflicts, I'll give this a final review and get this in.

RunDevelopment:
Done.
Django is now left unchanged.

PR #1493, opened 2018-07-22
Replaced all occurrences of `new RegExp` with `RegExp`

RunDevelopment:
Since the `RegExp` function is equal to the `new RegExp` constructor but shorter, I replaced all occurrences of `new RegExp` with `RegExp`.


mAAdhaTTah:
This seems fine, and less code.

PR #1499, opened 2018-07-27
SCSS: Fixed exponential selector pattern

### LINKED ISSUE #1498: SCSS map var crashes browser
Prism 1.15.0 & the SCSS language component, the snippet below will crash the browser/tab.

- Chrome 67.0.3396.99
- Firefox 61.0.1

```
$state-selectors: (
    focus: '&:focus, &.is-focused',
    active: '&:active, &.is-active',
    empty: '&:empty, &.is-empty',
    target: '&:target, &.is-target',
    enabled: '&:enabled, &.is-enabled',
    disabled: '&:disabled, &.is-disabled',
    valid: '&:valid, &.is-valid',
    invalid: '&:invliad, &.is-invalid',
    checked: '&:checked, &.is-checked',
    default: '&:default, &.is-default',
    in-range: '&:in-range, &.is-in-range',
);
```

This was the minimal amount of code I came across that replicates the issue. Remove a single line from that map variable, it will take a long time but will occasionally work, anything more and it always crashes.

I can't pin down exactly the issue with the snippet, but I suspect it has something to do with the `&:` & `&.` tokens.

I can't imagine anyone ever having scss like this, but this also crashes the browser and has a similar selector character pattern as the snippet above.

```
$crash-the-browser: '&: &. &: &. &: &. &: &. &: &. &: &. &: &. &: &. &: &. &: &.',
```

An interesting thing to note, is that the SASS language handles these variables just fine.

RunDevelopment:
This resolves #1498.


mAAdhaTTah:
Just to confirm, `&.class-name` would no longer select the `&` symbol as part of it?

RunDevelopment:
No, it's still part of it.
`&` is included in `[^@;{}()]`, so nothing changed apart from performance. I don't know why it was extra in the first place...

PR #1504, opened 2018-07-29
Support for pattern generators

### LINKED ISSUE #1495: Objects as patterns
This is based on the idea of @LeaVerou ([#1479 (comment)](https://github.com/PrismJS/prism/pull/1479#issuecomment-406857890))

> We could also have factory helpers, e.g. Prism.keywords("foo", "bar", ...) that generate regexes for common things

The only difference is that `Prism.keywords` will not generate a regex but an array with a method/property `generate`. This method will be called when the pattern is used for the first time and returns a regex.
Like this, we can have lists of keywords which can be used and modified by other languages.

(To be unabbigous, this will only be allowed for the `pattern` property.)

Thoughts?

RunDevelopment:
# Pattern generators

This PR adds pattern generators.
They can be used to keep patterns extendable (for derived languages) and to create complex patterns from simpler building blocks.

A pattern generator is an object with a `generate` function.
There are two types of generators:

1. RegExp Generators which generate a regular expression, and
1. Pattern Object Generators which generate an object of the form:
    ```js
    {
        pattern: RegExp|RegExpGenerator,
        greedy?: boolean,
        lookbehind?: boolean,
        inside?: Object,
        alias?: string|string[]
    }
    ```

The `generate` function of a pattern generator will be called with `null` as the this-argument. The first parameter will be the generator object itself.
If a generator is used to generate the `pattern` property of a pattern object, the pattern object will be passed as the second argument. This will enable the RegExp generator to set values like `lookbehind`.

This resolves #1495.


mAAdhaTTah:
@LeaVerou @Golmote I marked this one as "needs consensus" as we're going to have to discuss whether and how to leverage this.

mAAdhaTTah:
My overall issue w/ the pattern builder / generator stuff you've proposed so far is that is isn't clear to me that the introduction of those constructs results in an overall reduction of code, as the code to support the features is pretty lengthy.

RunDevelopment:
This is a prerequisite for features like the keyword-lists mentioned in #1495.

But I'm closing this now anyway because new functions like the mentioned `keywordlist` are required for this to have any benefit at all.

PR #1506, opened 2018-07-30
Improved DFS

RunDevelopment:
A slight improvement for `DFS`.

- The results of `o[i]` and `_.util.type(o[i])` are now cached which yield better performance and code compression.
- `DFS` now calls itself and not `Prism.languages.DFS` (which is only most likely itself).


components/prism-core.js
```diff
@@ -139,19 +139,24 @@ var _ = _self.Prism = {
...
 
-					if (_.util.type(o[i]) === 'Object' && !visited[_.util.objId(o[i])]) {
-						visited[_.util.objId(o[i])] = true;
-						_.languages.DFS(o[i], callback, null, visited);
+					var property = o[i],
+					    propertyType = _.util.type(property);
+
+					var objId = _.util.objId;
+
+					if (propertyType === 'Object' && !visited[objId(property)]) {
```
mAAdhaTTah: Do we want to cache `objId(property)` instead of just the `objId` function?

RunDevelopment: I just wanted to cache the function itself but now that I look at it, we might as well cache the result.

RunDevelopment: I tried to cache the result and you can't really do it without producing ugly code. 
That's because you can only call `objId(property)` after you know that it's an object/array, otherwise, `Object.defineProperty` will throw an error.

Issue #1492, opened 2018-07-21
Greedy matching bug


RunDevelopment:
Consider the following code:

```js
{
	'template-string-pattern': /`(?:\\[\s\S]|\${[^}]+}|[^\\`])*`/,
	'a': 0,
	'b': 1
};
{
	'string' : /(`|')(?:\\.|(?!\1)[^\\\r\n])*\1/m,
	'c': 0
};
```

Highlighted using Test Drive:

![image](https://user-images.githubusercontent.com/20878432/43036031-c6528d0c-8cf9-11e8-9e78-087bd1c6a8fd.png)

What's happening is quite simple:

1. `template-string` matches `(?:\\[\s\S]|\${[^}]+}|[^\\` and `/,   'a': 0    .......    'string': /(` (plus the grave accents).
2. `string` matches `'template-string-pattern'` and `'c'`. The rest is blocked by `template-string`.
3. `regex` breaks the matches of `template-string` and highlights the regexes.
4. `string` gets a chance from `regex` to match and does so matching `'a'`. But that's it, because `oneshot=true`.

Btw: You can see this bug all over the place when you look at the Prism code with lots of language definitions.

RunDevelopment:
Here is a minimal example showing the bug in action:

<details>
<summary>Language definition</summary>

I used aliases for styling.

```js
Prism.languages['greedy-test'] = {
	'a': {
		pattern: /'[^'\r\n]*'/,
		alias: 'comment'
	},
	'b': {
		pattern: /"[^"\r\n]*"/,
		greedy: true,
		alias: 'string'
	},
	'c': {
		pattern: /<[^>\r\n]*>/,
		greedy: true,
		alias: 'keyword'
	}
}
```

</details>

**Code:**

```
<'> '' ''
<"> "" ""
```

(Each line is one example.)

**Result:**
![image](https://user-images.githubusercontent.com/20878432/54353724-0c0f3b00-4655-11e9-8b50-0ffecad12f60.png)

**Notes:**
1. The order of `a` and `b` is not significant. They just have to be before `c`.
2. Both example lines of code are independent of each other.

PR #1510, opened 2018-08-01
SQL: Quoted variables are now greedy

### LINKED ISSUE #1509: Single-quoted variables not working for SQL highlighting
According to the examples, the following should work, but it does not
`SET @'quoted-variable' = 3; `
All other quoted variable examples do work, i.e. the following:
```
SET @variable = 1;
SET @$_ = 2;
SET @"quoted-variable" = 3;
SET @`quoted-variable` = 3;
```

RunDevelopment:
This resolves #1509.

---

I added a small fix to bypass #1492.
It now (mostly) works because the quoted variables are now matched before strings. Therefore strings (of quoted variables) are always contained by the variable which leads to there match being discarded.

This little trick also reduces the greedy matching overhead because it doesn't have to rematch any patterns. Maybe we should include this trick into the docs as a best practice?


mAAdhaTTah:
Got conflicts here, probs a result of the other SQL PR I just merged.

mAAdhaTTah:
Are there styles for `quoted-variable`? I don't _think_ that's a currently-existing token...

RunDevelopment:
I know any styles for `quoted-variable` either. It's just a temporary name because it has to be matched before `string`.
But we could move `variable` before `string` as well and merge the two into one `variable` token without any side effects (I think).

RunDevelopment:
I did implement the merged `variable` token.

Is it ready to merge?

Issue #1518, opened 2018-08-08
JSX tag-class (HTML Element vs Component Element)


Templarian:
In JSX it's a subtle feature, but editors [Ex: VS Code] syntax highlight the difference between native HTML elements (Ex: `<a>`, `<span>`) and component classes (Ex: `<DropDown>`, `<Foo>`).

Basically, if it does not have a capital `[^a-z]` for the first character it would be helpful to have another class name appended. Like `tag-class` or something.

```
<foo  = <span class="token tag"><span class="token punctuation">&lt;</span>foo</span>

<Foo  = <span class="token tag tag-class"><span class="token punctuation">&lt;</span>Foo</span>
                               ^ Goal: Add this class name
```

I was looking below, but I can't get my head around how to achieve this for a PR (so I made this issue).

https://github.com/PrismJS/prism/blob/master/components/prism-jsx.js

Very low priority, but trying to make custom components in ReactJS documentation stand out more.

PR #1519, opened 2018-08-08
JSX/TSX: Tags of React components are styled as classes

RunDevelopment:
This PR implements #1518 but it uses `class-name` in favor of the proposed `tag-class`. Like this, we don't have to change the themes.

The PR adds a `class-name` alias to all the tags of React components. Tags which start with an uppercase letter are considered React components. 
HTML tags remain unchanged.

The new styling is similar to VS Code's.

Example: (taken from the [React tutorial](https://reactjs.org/tutorial/tutorial.html)) 
image

VS Code:
image


mAAdhaTTah:
Hey @RunDevelopment – Just wanted to say I appreciate the reviews & PRs you've made to Prism. I haven't had a lot of time to dig into all of them, and a lot of the regex changes I'm unfortunately not really qualified to review, but I just wanted to let you know we see the work you've done and thank you for it.

Templarian:
@RunDevelopment Thanks for writing this so quickly, exactly what I needed. The code/results looks good to me. 😃

atomiks:
Could this be changed to `tag-component` or `tag-class` instead of `class-name`. The reason is to desire as much granularity with the highlighting as possible, if we don't want the JavaScript class color to match the component tag color.

Templarian:
@atomiks The theme will just use a more specific CSS selector.

```css
.token.tag.class-name {
    color: red;
}
```

atomiks:
Ah ok, looks good then!

mAAdhaTTah:
Can we get tests for this?

RunDevelopment:
Unfortunately, it is not possible to test this feature with our usual tests.
Because this PR only adds an alias to a token, the alias will not show up in the JSON (see `jsx/tag_feature.test` which I did not have to change, but is styled like a class in the test page.).

mAAdhaTTah:
But you could confirm that `<Button />` styles `Button` as a `class-name`?

RunDevelopment:
> But you could confirm that <Button /> styles Button as a class-name?
Depends:
With the current implementation of this PR and the test suite, definitely no.

But, we could change the implementation of this PR so that it will replace the string `Button` with a new token with the type `class-name`. instead of adding an alias to the `tag` token. That will be testable.

mAAdhaTTah:
I dig this, much closer to what I was thinking.

PR #1520, opened 2018-08-08
Added support for Java stack traces

RunDevelopment:
Since the original PR for Java stack traces #916  wasn't continued I decided to try it. So, here is my take on it.

This PR adds support for Java stack traces.

There are 3 main elements which are highlighted:

1. The summary.
    This is the head of the stack trace and includes the name of the exception and a message.
    `Exception in thread "main" java.lang.RuntimeException: A test exception`
2. Stack frames.
    These make up the actual stack trace and include the full name of the method and the source.
    `at com.stackify.stacktrace.StackTraceExample.methodB(StackTraceExample.java:13)`
3. Omitted frames.
    This is a simple message that some frames were omitted.
    `... 27 more`

Everything besides these 3 element types will not be highlighted.

## Example
image


mAAdhaTTah:
Thanks for this. One quick note: let's leave the package-lock.json out of this PR for now, since we don't have one currently. If we're going to add it, we should do that separately.

RunDevelopment:
Oops, my bad. I forgot to delete it.

mAAdhaTTah:
Tests all look pretty good here, but I'll need @Golmote or @LeaVerou to review the Regex here. Also got a conflict that will need to be resolved.

mAAdhaTTah:
If we resolve these conflicts, we can merge this PR.

PR #1521, opened 2018-08-11
Added .scss files

TokugawaTakesi:
Because of [this issue](https://github.com/sass/node-sass/issues/2362), I offer you to add .scss versions of .css files. In the future, you can write your completely on SASS and .css files will be the output, however currently I just copied css files and changed extension.

### DIFF
--- a/plugins/autolinker/prism-autolinker.scss
+++ b/plugins/autolinker/prism-autolinker.scss
@@ -1,0 +1,3 @@
+.token a {
+	color: inherit;
+}
--- a/plugins/command-line/prism-command-line.scss
+++ b/plugins/command-line/prism-command-line.scss
@@ -1,0 +1,33 @@
+.command-line-prompt {
+	border-right: 1px solid #999;
+	display: block;
+	float: left;
+	font-size: 100%;
+	letter-spacing: -1px;
+	margin-right: 1em;
+	pointer-events: none;
+
+	-webkit-user-select: none;
+	-moz-user-select: none;
+	-ms-user-select: none;
+	user-select: none;
+}
+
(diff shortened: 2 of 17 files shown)


RunDevelopment:
It'd be good to also have a proper workflow to compile the SCSS files if we were to add this to Prism.
Otherwise, these files just sit there, not doing anything.

The repo you mentioned links a [gulp plugin for SASS](https://github.com/dlmanning/gulp-sass), so maybe we could use it. 
Thoughts?

TokugawaTakesi:
> It'd be good to also have a proper workflow to compile the SCSS files if we were to add this to Prism.
Of course. But maybe we need to reconsider the project file system for the logical separation source and destination folders.

> Otherwise, these files just sit there, not doing anything.
Libraries often included in other files. Same for **prismjs**: .scss could be included in other .scss or .sass files, so at least sass warning that I mentioned will disappear. 

> The repo you mentioned links a gulp plugin for SASS, so maybe we could use it. Thoughts?
I think it will be very good decision for the future library development. Please let me know how you want to re-organize your project file system before adding gulp.

mAAdhaTTah:
If we're going to add scss, we should take advantage of what scss provides us. As @RunDevelopment mentioned, if we're just adding a bunch of new files that don't do anything, there's not a lot of value to adding them and it's an extra maintenance burden to keep them in sync.

I'm definitely interested in bringing sass in as tool, but I wouldn't do it piecemeal like this. We haven't settled on how we'd organize a project, but code talks; if you have an idea, we'd be interested in seeing it!

I'm going to close this PR, but feel free to open a new one if you have more fleshed out ideas!

PR #1524, opened 2018-08-15
Added `addAlias` method for tokens

RunDevelopment:
This PR adds an `addAlias` method to the `Token` class.

The new method appends the given arguments to the existing aliases of a token. Nothing will be returned.
This behavior is similar to `Array.prototype.push`.

Examples:
```js
let t = new Token(type, content, "alias", ...etc);
console.log(t.alias); // "alias"
t.addAlias("newAlias");
t.addAlias("alias 1", "alias 2");
console.log(t.alias); // ["alias", "newAlias", "alias 1", "alias 2"]
```

`addAlias` will always change the data representation of `alias` to an array.

```js
let t = new Token(type, content);
console.log(t.alias); // undefined
t.addAlias();
console.log(t.alias); // []
```


mAAdhaTTah:
This might be more interesting / useful if you demonstrate where this could be used. Can you update the code this would simplify?

RunDevelopment:
I haven't found any language or plugin which currently has the need to add aliases to tokens, so there is no code to be simplified **right now**.

However, #1519 has to do exactly that and it can be compressed into a single `if`-statement using `addAlias`:
```js
// class-name alias to tags which refer to classes
if (token.type === 'tag' && token.content.length === 2 && token.content[0].type === 'punctuation'
    && typeof token.content[1] === 'string' && /^[A-Z]/.test(token.content[1]))
	token.addAlias('class-name');
```

---

I added `addAlias` because after writing #1519, I wrote another hook and had to implement it again. This PR is for eliminating this redundancy.

Plus, I think that something as basic as adding an alias to a token should be simple.

RunDevelopment:
Well, your right.

PR #1527, opened 2018-08-16
JS: Updated `catch` and `finally`

### LINKED ISSUE #1526: `catch` should not be a keyword when used in a Promise chain
Currently, when parsing javascript, `catch` will be rendered as a keyword. Since ES6, `catch` is also a method on the Promise prototype, which is probably best rendered as a `function` in Prism:

```
fetch('some-resource.json')
    .then(response => response.json())
    .catch(console.error)
```

Hopefully, there is an easy way to test for this. Thanks!

RunDevelopment:
This PR resolves #1526.

1. `catch` and `finally` will only be considered keywords if they are preceded by a `}`.
1. The exception variable will not be highlighted as `class-name`. (This was because of the c-like `class-name` pattern.)
1. `trait` is removed from the list of words preceding a class name.


components/prism-javascript.js
```diff
@@ -1,11 +1,19 @@
...
+		},
+		/\b(?:as|async|await|break|case|class|const|continue|debugger|default|delete|do|else|enum|export|extends|for|from|function|get|if|implements|import|in|instanceof|interface|let|new|null|of|package|private|protected|public|return|set|static|super|switch|this|throw|try|typeof|var|void|while|with|yield)\b/
+	],
 	'number': /\b(?:0[xX][\dA-Fa-f]+|0[bB][01]+|0[oO][0-7]+|NaN|Infinity)\b|(?:\b\d+\.?\d*|\B\.\d+)(?:[Ee][+-]?\d+)?/,
 	// Allow for all non-ASCII characters (See http://stackoverflow.com/a/2008444)
 	'function': /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*\()/i,
 	'operator': /-[-=]?|\+[+=]?|!=?=?|<<?=?|>>?>?=?|=(?:==?|>)?|&[&=]?|\|[|=]?|\*\*?=?|\/=?|~|\^=?|%=?|\?|\.{3}/
 });
 
+Prism.languages.javascript['class-name'].pattern = /(\b(?:class|interface|extends|implements|instanceof|new)\s+)[\w.\\]+/
```
mAAdhaTTah: What's the story with this line in this PR?

RunDevelopment: This line changes the `class-name` pattern to remove `catch` to no longer highlight the exception variable as a class name. (I added a new test for that)
`trait` was also removed, because this isn't a keyword in JS.

I didn't overwrite `class-name` in `extend` because it would have required more code to do so. (`lookbehind` and `inside`)
Or should I do it anyway for better readability?

mAAdhaTTah: No, I think this is fine. Maybe a comment would be helpful?

Could we also get a test for `promise().catch().finally()` to make sure they're caught as functions?

mAAdhaTTah: Wait, we have that already, my bad.

PR #1528, opened 2018-08-17
C/C++/Java operator update

RunDevelopment:
This PR extends and updates the operators of C, C++, and Java.

Java and C++ now support the `->` operator. 
(C++ had a bug.)

The tests of all languages were updated and the operator pattern for each language is now shorter.


mAAdhaTTah:
This seems good, thank you for you contributions!

PR #1529, opened 2018-08-17
More persistent test page

RunDevelopment:
This PR makes the test page remember the language and code last used.

The language is stored in the URL hash similar to the download page.
The code last used is stored in `sessionStorage` like the current theme.

This feature has been tested in Chrome, Firefox, and Opera.
Edge has [severe problems with `sessionStorage` and `localStorage`](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8816771/) when the test page is opened from the local file system. But it does work when served from a (local) server.


test.html
```diff
@@ -95,6 +95,15 @@ <h2>Test drive</h2>
     languages = components.languages,
     highlightCode = function() { Prism.highlightElement(code); };
 
+
+function updateUrl(lang) {
+	var href = window.location.href.split('#')[0] + '#';
+	if (lang)
+		href += 'language=' + lang;
```
mAAdhaTTah: Add `{}`.

test.html
```diff
@@ -95,6 +95,15 @@ <h2>Test drive</h2>
...
+	window.location.href = href;
```
mAAdhaTTah: Do we want to use `pushState` or `replaceState` here?

mAAdhaTTah:
If Edge has problems w/ local/session storage, do we need try / catch's for it?

RunDevelopment:
All done.
I choose to use `pushState`, so you can switch between languages by going back and forward.

PR #1531, opened 2018-08-18
`insertBefore` now correctly updates references

### LINKED ISSUE #1530: `insertBefore` does not update references correctly
`insertBefore` does not update references in the current language.

## Example

```js
Prism.languages.testLang = Prism.languages.extend('clike', {
	'token': /token/
});

Prism.languages.insertBefore('testLang', 'function', {
	'newToken': {
		pattern: /newToken/,
		inside: Prism.languages.testLang
	}
});

Prism.languages.testLang == Prism.languages.testLang.newToken.inside // false
```

RunDevelopment:
This PR fixes #1530.


mAAdhaTTah:
I don't suppose we can capture this in a test somehow? Does this have a real-world impact on any of the language definitions?

RunDevelopment:
> I don't suppose we can capture this in a test somehow?

We could do that with ease. The last expression of #1530 is basically that.
But, I don't think that there are any tests concerning Prism core functionality.

> Does this have a real-world impact on any of the language definitions?

No, it doesn't have any impact right now because there is an easy workaround:
Setting `token.inside.rest` after `insertBefore` was called.

This is also the workaround I used when [discovering the bug](https://github.com/PrismJS/prism/pull/1523#issuecomment-413996089).
I believe, there are also other languages using this workaround.

mAAdhaTTah:
> I believe, there are also other languages using this workaround.

If there are, can we remove those workarounds? That would also help to ensure the fix produces the result we expect.

RunDevelopment:
Ok, so I went through all languages and found that only JS and PHP used the workaround I described.

What's even more interesting is that some languages didn't even care about the bug. They used different versions of their own language and the tests did not capture it.
The ones I found are Crystal and Ruby.

Despite the different language definitions, I wasn't able to find any differences in their examples. We should probably extend the tests of both languages.

mAAdhaTTah:
This looks good to me!

Issue #1494, opened 2018-07-22
Reusable tokens


mAAdhaTTah:
Suggestion by @LeaVerou (https://github.com/PrismJS/prism/pull/1479#issuecomment-406857890):

> Btw we should probably have common tokens as re-usable variables or something (E.g. `Prism.NUMBER`). Currently every language needs to redefine numbers, and in most cases they're the same... We could also have factory helpers, e.g. `Prism.keywords("foo", "bar", ...)` that generate regexes for common things.
> What do you think?

RunDevelopment:
What are the use cases of this?

Let's take e.g. `NUMBER`:
What should it be like? Only decimal numbers? If so, are floating point numbers include (with or without exponents)? What about constants like `NaN`, `Infinity` or `∞`?
And when you think about hexadecimal and binary numbers, and optional underscores within numbers and the whole thing becomes a mess. (And I don't need something like `\d+`)

I don't think that there are many places where we can get away with `'number': Prism.NUMBER` or even `'number': [ number-style 1, 2, 3, 4, simple pattern for edge cases ]`.

Patterns like `COMMENT` or `STRING` look more promising but they also will fall short in many cases.

The cases where you can easily share patterns are between languages with very similar syntax and these are already taken care of by `Prism.languages.extend`.

So, where will they be used?

mAAdhaTTah:
> `'number': Prism.NUMBER`

I think places like these is the intention. If the same pattern isn't used in multiple places, then we don't get much and isn't worth it.

LeaVerou:
Also the helper that generates keyword regexes could be general enough to be able to combine other regexes as alternatives. Then we could have a `Prism.NUMBER` for the actual numbers, and then we could do something like `Prism.or(Prism.NUMBER, 'NaN', 'Infinity')` for JS numbers for example. I think something like that has the potential to simplify many grammars, if done well.

RunDevelopment:
> Prism.or(Prism.NUMBER, 'NaN', 'Infinity')

Doesn't this imply that `NUMBER` will be a rather broad pattern, matching all kinds of numbers? 
So in general, are we going to tolerating false positives to increases the applicability of these constants?

It might be ok because we don't valid syntax, but do we really want to be that inclusive?

RunDevelopment:
> How does it imply that?

If `NUMBER` wouldn't include stuff like e.g. hexadecimal numbers, it wouldn't work for JS.
Or did I take your example to literal? If so, I'm sorry.

[...]

Issue #1476, opened 2018-07-12
Pattern Builder


RunDevelopment:
When working with language definitions, you sometimes have parts which are used across multiple patterns (e.g. names). This redundancy makes it harder to change languages and blows up the file size.

My proposal is to add a function to build patterns.

## Behavior

The function `build(basePattern, replacements)` will take a regular expression `basePattern` and an object containing regular expressions (or strings) `replacements`. 

The source of the base patterns will contain placeholders where replacements will be inserted. The placeholder will hold the key to which replacement will be used.

This is a simple text replacement, so surprises may come. 
To minimize this risk, there are some restrictions and each of the replacements will be wrapped into a non-capturing group.

The source of the base pattern with all placeholders replaced plus the flags of the base pattern will form the pattern to be returned.

The flags of replacements will be ignored. 

### Restrictions 

To minimize the risk of text replacement in the source code, there will be a bunch of restrictions to both the base pattern and the replacements:

Placeholder positioning:
1. Cannot be inside a char set.
2. Cannot be after an unescaped backslash

Replacements:
1. No capturing groups. They could mess with backreferences in the base expression.
2. No backreferences.

### Examples

I will use the placeholder `<<\w+>>` where `\w+` will be used as the key to get a replacement.

(If you have a better placeholder, please tell me. I only choose this one because it isn't regex syntax and can be easily spotted. But I don't really like it...)

```js
build(/a<<b>>?/i, {b: /b+/i}) == /a(?:b+)?/i
build(/a<<0>>?/i, [/b+/.source]) == /a(?:b+)?/i
build(/<<0>>/m, [/^\w+/]) == /(?:^\w+)/ // warning: the meaning of ^ might have changed
build(/(a)<<0>>\1/, [/(b)/]) // error: replacement contains capturing group
build(/a<<1>>/, [/b/]) // error: replacements["1"] undefined
```

## Why regular expressions all the way?

It will be cheaper to use strings as base patterns and replacements. That's true. 
But strings do not provide two things that regular expressions do.

1. They are easier to write. 
    Inline regexes are more convenient than strings and are supported by IDEs with features like syntax highlighting.
2. They will catch error early on.
    Each of the patterns will be compiled by the browser (or node), so they are guaranteed to have correct syntax.

Of course, these things only really matter to developers but not to the end user who's computer will have to deal with additional overhead.

## Minimizing overhead

To minimize the overhead created by using and checking a bunch of regular expressions, we can do multiple things:

[...]

PR #1538, opened 2018-08-22
Pattern builder

RunDevelopment:
This PR is an implementation of #1476.

I decided to make a new namespace `Prism.patterns` (\*hint\* \*hint\* #1494) to keep things organized because `build` doesn't really belong in any of the other namespaces.

The validation method of the base pattern and replacements is located in a new file called `tests/helper/prism-core-tester.js`.
This file's purpose is to add testing functionality to existing Prism methods __only__ when running `npm test`. This has the following advantages:

- `prism-core.js` stays small but still validates on all tested components.
- The validation is not done on our client's computers.

---

Maybe we should make this a resource (#1539)?

---

Btw. This and #1537 go together really well.


RunDevelopment:
It intrigues me that the build failed despite the fact that I only changed some comments...

It seems like the `npm install -g gulp` command took longer than 10 minutes. How can that be?
Node v6 took each about 7 minutes for `npm install` and `nom install -g gulp`.

RunDevelopment:
Now it worked again.
And it took at most 20 seconds for the `npm install` commands.

mAAdhaTTah:
Travis does weird stuff sometimes. I wouldn't worry too much about it.

RunDevelopment:
This is just an idea for now but with ES6 [tagged template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) we could make a function `Prism.regex` which will behave like `Prism.patterns.build` but has a nice syntax:

```js
const foo = /\bfoo\b/;

Prism.regex`\w\s${foo}\s`; // returns /\w\s(?:\bfoo\b)/

// instead of
Prism.patterns.build(/\w\s<<0>>/, [foo]);
```

There a few drawbacks (flags, browser compatibility, no regex syntax highlighting) but maybe someone likes it.

RunDevelopment:
Closed in favor of #2190.

Issue #1475, opened 2018-07-11
Documentation highlighting


RunDevelopment:
Many languages have an (at least somewhat) standardized syntax for code documentation like [JavaDoc](http://www.oracle.com/technetwork/java/javase/tech/index-137868.html), [JSDoc ](http://usejsdoc.org/index.html), [PHPDoc ](https://docs.phpdoc.org/guides/docblocks.html), [.Net XML doc](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/xml-documentation-comments) (for C#, F# and VB.Net) and others.

Similar to most IDEs, the inline documentation comment should be displayed in a different color than normal comments. Also keywords, type and parameter references, examples and similar should be highlighted.

This means that every the themes will have to be changed to reflect the new syntax highlighting.
I am no designer, so it would be great to get some help.

Because the documentation of many languages is inspired by JavaDoc, it should not be too hard to write an abstract base of doc comment highlighting which is then extended by  different languages.

## Examples

PrismJS
![image](https://user-images.githubusercontent.com/20878432/42592247-3f2b6568-8549-11e8-92fa-0abbad424ff2.png)

GitHub
```js
/**
 * Trims a given string.
 * 
 * @param {string} [str=""] the string.
 * @returns {string} something.
 * @throws {TypeError} if the argument is not a string.
 * @example trim(" hello ")
 */
function trim(str = "") {
	if (typeof str !== "string")
		throw new TypeError("str has to be a string");
	return str.trim();
}
```

Notepad++
![image](https://user-images.githubusercontent.com/20878432/42591957-8195fae0-8548-11e8-9c03-c4ef1139d8cb.png)

VS Code
![image](https://user-images.githubusercontent.com/20878432/42591930-6e97ca86-8548-11e8-9963-81b3bc4b791b.png)

mAAdhaTTah:
I could see this being provided as an additional language; maybe a "language" that enhances existing languages?

RunDevelopment:
I originally imagined this as an extension of existing languages, but making them into their languages might be better idea. 
Like this we could also support wildly different styles, which might be necessary if we ever wanted to support a language where there is no one dominating standard.

Using this approach we could do it like this:

```js
// Prism.languages.java
'doc-comment': {
    pattern: /\/\*\*[\s\S]*?\*\//,
    greedy: true
},
...
```

Something like `Prism.languages.javadoc` would then add itself as `inside`.

[...]

PR #1541, opened 2018-08-23
Doc comment highlighting

RunDevelopment:
This PR implements #1475.
It adds support for JavaDoc (Java), PHPDoc (PHP), and JSDoc (JS, JSX, N4JS, TS, TSX).
This will also affect languages which include any of the above languages.

__Notes:__
- The current implementation of PHPDoc does not support annotations.
- TypeScript actually has its own standard called TSDoc but this is only a subset of JSDoc, so we should be fine.

### TypeScript

image

### JavaScript

image

### Java

image

### PHP

image


components/prism-javadoclike.js
```diff
@@ -0,0 +1,78 @@
...
+				if (token[i] instanceof RegExp)
+					token[i] = { pattern: token[i] };
+				callback(token[i]);
+			}
+		} else {
+			callback(token);
+		}
+	}
+
+	Object.defineProperty(javaDocLike, 'addSupport', {
```
mAAdhaTTah: j/c but why `defineProperty`?

RunDevelopment: Because otherwise, Prism will assume that `addSupport` is a normal token and throw an error.
But a property defined using `Object.defineProperty` [cannot be enumerated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Description) and is effectively hidden.

components/prism-javadoc.js
```diff
@@ -0,0 +1,56 @@
+(function (Prism) {
+
+	var javaDocLike = Prism.languages.javadoclike;
```
mAAdhaTTah: I remember we're doing this for minification reasons, but if we're only using it once, do we actually get those benefits here?

RunDevelopment:
It's been a while since I last looked at this. 
I'd actually like to make a few changes, so please wait a little bit longer.

components/prism-phpdoc.js
```diff
@@ -0,0 +1,28 @@
+(function (Prism) {
+
+	var typeExpression = /(?:[a-zA-Z]\w*|[|\\[\]])+/.source;
+
+	Prism.languages.phpdoc = Prism.languages.extend('javadoclike', {
```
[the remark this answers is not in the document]
RunDevelopment: Yes, but not here. Further down below there is an `insertBefore`, which changes the object associated with `Prism.languages.phpdoc`.

But we could do:
```js
var phpdoc = Prism.languages.insertBefore('phpdoc', 'keyword', { ... });
```

components/prism-javadoc.js
RunDevelopment: You're right! I will remove it.

troy351:
I could not find this PR works, any ideas?
image

https://codepen.io/troy351/pen/rNeZErz

PR #1542, opened 2018-08-23
JS: Support for BigInt literals

RunDevelopment:
This PR adds support for [BigInt](https://github.com/tc39/proposal-bigint) literals to JS and languages extending or including JS.


RunDevelopment:
And I really want to thank you for that! 😄

Issue #1513, opened 2018-08-01
Add support for the Racket lisp


Eugleo:
Racket is basically like Scheme when it comes to syntax highlighting, it only has some minor differences. I have made a complete Racket grammar ([link](https://github.com/Eugleo/magic-racket/blob/master/syntaxes/racket.tmLanguage.json)), but I think just copying the Scheme one and adding support for `racket` language will be cleaner and function just as well.

If you want me I can open a PR, just need to decide if we want the complex grammar or rather the simple one.

EDIT: Updated the link to point to the renamed repo.

mAAdhaTTah:
I think the simplest implementation is just another alias for the current `lisp` language.

Otherwise, simpler is generally going to better if it's "good enough". We're not a parser, so perfection isn't the goal. Would def be interested in a PR if you want to give it a shot.

jcubic:
Isn't racket just scheme with brackets in some places? it should work the same as for scheme syntax. If there is something missing in scheme I would add that to that language and use scheme for racket. Or if aliases are possible use racket as aliast for scheme.

Eugleo:
Well, mostly it is. There are some minor differences, for example the aforementioned braces and brackets, or syntax identifiers, such as `#'` and `#\``. 

Now that you've reminded me of this old issue, I think duplicating the scheme file and making some changes to it (i.e. creating a whole new language, not just aliasing to the old one) would be the best course of action. Or is there some downside to it that I'm missing? I'd rather not break (=ammend) the scheme syntax highlighting.

RunDevelopment:
Hmmm if it's really just brackets and very minor differences, couldn't we just write a little something with go through all scheme patterns and replaces all `(` and `)` characters with their racket equivalents? 

<details>
<summary>Like this:</summary>

The following requires that if a character class in scheme contains both "(" and ")", they have to be at the beginning of the character class. E.g. `[^()\s]` but not `[\s()]`

```js
Prism.languages.racket = Prism.languages.extend('scheme', {});
Prism.languages.DFS(Prism.languages.racket, function (key, value) {
    if (Prism.util.type(value) === "RegExp") {
        var source = value.source;
        // replace lone "(" and ")" characters with a character class "[({[]" and "[)}\]]"
        source = source.replace(/((?:^|[^\\])(?:\\{2})*)\\\(/g, "$1[({\\[]");
        source = source.replace(/((?:^|[^\\])(?:\\{2})*)\\\)/g, "$1[)}\\]]");
[...]

PR #1547, opened 2018-09-04
Add basic syntax support for Racket

Eugleo:
As per #1513. It is based on the file for Scheme; I decided not to use the advanced regex file, and opted for the basic one instead.


RunDevelopment:
Thanks for the implementation!

Now, there a few things we will have to do before we can accept this:

1. Add Racket to `components.json`. Otherwise, Prism doesn't know your language definition exists.
1. Run gulp via the `npx gulp` command. This will rebuild Prism and (among other things) add a minified version of your language definition.

It would also be great if you were to add a few [tests](https://prismjs.com/test-suite.html).

Eugleo:
@RunDevelopment Sure, I wondered whether some "registration" of the new file will be needed. Done! 
I'm quite bussy at the moment, but rather than simply copying over the tests from Scheme, I'll reserve some time in the future to do some better ones. So, no tests now, if it's ok.

components.json
```diff
@@ -614,6 +614,11 @@
 			"title": "R",
 			"owner": "Golmote"
 		},
+		"racket": {
+			"title": "Racket",
+			"owner": "Eugleo"
+		},
+
```
RunDevelopment: Why is this extra line here?

RunDevelopment:
@Eugleo Sure, take your time for the tests 😄

Btw. Don't worry about the failed checks. It's not your fault.

components.json
Eugleo: No idea. Will fix, sorry.

Eugleo:
@RunDevelopment Should be ok now. Thanks for the heads up about the checks — I'm fairly new to GitHub and don't really understand these things 😃 Will you merge now or after the tests? I'm not sure when I'll be able to add them...

RunDevelopment:
@Eugleo 
> Will you merge now or after the tests?

I'm sorry but I can't merge anything. A maintainer will have to do that. But before that, we will thoroughly review your language definition.

I will most likely review it in the next few days or so.

RunDevelopment:
@Eugleo 
While using Scheme as the base for Racket is not a bad idea, please use the `Prism.languages.extend` method to do so.

Also, Racket supports some features which Scheme does not and they should be added as well (maybe not all of them).
I can really recommend [Beau­ti­ful Racket](https://beautifulracket.com/#explainers) which has short explainations on all Racket features many of which differ from Scheme. E.g. [name keyword arguments](https://beautifulracket.com/explainer/functions.html#input), [condinitionals with `[]`](https://beautifulracket.com/explainer/booleans-and-conditionals.html#conditionals), and more.

Eugleo:
@RunDevelopment I know Racket well, but from the conversation in the issue I linked I got the feeling that it will be better to have a minimal implementation. But I agree with you; I'll work on a more complete grammar/regex definition for Racket and rework this PR.

RunDevelopment:
@Eugleo 
It's totally fine to keep it minimalistic for now. We can always add more complex language features after we have base support.

It's just that some Racket features (like [conditionals with `[]`](https://beautifulracket.com/explainer/booleans-and-conditionals.html#conditionals)) will break Scheme highlighting.
I don't know Racket very well, but I think that we should cover commonly used features.

mAAdhaTTah:
@Eugleo @RunDevelopment Is there still work to be done on this PR?

mAAdhaTTah:
@Eugleo No worries! Just wanted to make sure we didn't leave you hanging.

PR #1549, opened 2018-09-06
Add Java 10 support

winterbe:
Add new keyword `var` to Java language. The keyword was introduced in Java Version 10.

See: https://developer.oracle.com/java/jdk-10-local-variable-type-inference


mAAdhaTTah:
LGTM!

PR #1552, opened 2018-09-09
C-like: Simplified function

RunDevelopment:
This PR simplifies the `function` pattern of the C-like language.

`/[a-z0-9_]/i` == `/\w/`


mAAdhaTTah:
I suppose this is a nit, but wouldn't `my-function()` also get caught by this?

RunDevelopment:
No, [`\w` is equal to `[a-zA-Z0-9_]`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#character-classes). `-` is not included.

PR #1554, opened 2018-09-09
Adding option for toggling line numbers on the test output box.

kostasx:
I thought it might be useful to add the option to toggle line numbers on the output box.

Here's a preview: https://kostasx.github.io/prism/test.html

### DIFF
--- a/test.html
+++ b/test.html
@@ -52,6 +52,50 @@ textarea {
 	}
 
 </style>
+<!-- Line Numbers -->
+<style>
+	pre[class*="language-"].line-numbers {
+		position: relative;
+		padding-left: 3.8em;
+		counter-reset: linenumber;
+	}
+
+	pre[class*="language-"].line-numbers > code {
+		position: relative;
+		white-space: inherit;
+	}
+
+	.line-numbers .line-numbers-rows {
+		position: absolute;
+		pointer-events: none;
+		top: 0;
+		font-size: 100%;
+		left: -3.8em;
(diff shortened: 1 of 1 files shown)


mAAdhaTTah:
I like the concept, but it looks like you inlined the CSS / JS from the plugin. Can we reuse the code we have so we don't have it duplicated in both places?

kostasx:
Made some changes. Should be OK now.

test.html
```diff
@@ -195,8 +200,18 @@ <h2>Test drive</h2>
 	highlightCode();
 }).call(textarea);
 
+// Line Numbers Switcher
+var $lineNumbersBtn = document.getElementById("line-numbers");
+$lineNumbersBtn && $lineNumbersBtn.addEventListener("click", function(){
```
mAAdhaTTah: We shouldn't need to check its existence, since it'll always be there.

test.html
```diff
@@ -195,8 +200,18 @@ <h2>Test drive</h2>
...
+
+	document.querySelector("pre").classList.toggle("line-numbers");
```
mAAdhaTTah: Let's query for this above and reuse the result. Also, let's use `this.checked` to determine whether or not to add / remove the class rather than relying on `toggle` always matching the value.

mAAdhaTTah:
Sorry for the delay. On board with the suggestion, but needs some minor tweaks to get merged.

PR #1557, opened 2018-09-10
Markdown: Fixed headers

RunDevelopment:
This PR fixed markdown header of the form

```markdown
Überschrift
-----------
```
(Example from the [German Markdown Wikipedia page](https://de.wikipedia.org/wiki/Markdown).)

This however also embraces the ambiguity of horizontal lines and headers. But Since a single empty line resolves the ambiguity (that is built into markdown! ), I think it's ok.


tests/languages/markdown/title_feature.test
```diff
@@ -4,6 +4,9 @@ title 1
 foobar
 ======
 
+Überschrift
+===========
```
mAAdhaTTah: Can we add a test for the example in your PR comment?

```
Überschrift
-----------
```

RunDevelopment: That, I cannot do because of the way `.test` files are parsed, so I added a similar test.

mAAdhaTTah: Fair enough. Looks like we had a similar test already but no harm in another.

PR #1559, opened 2018-09-11
Removed extraneous travis configs, test against latest node

amilajack:
Remove travis support for legacy node

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ branches:
 before_script:
 - npm install -g gulp
 - gulp
+test:
+  - npm test
 deploy:
   provider: npm
   email: lea@verou.me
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,7 +12,6 @@ branches:
 before_script:
 - npm install -g gulp
 - gulp
-script: npm test
```
RunDevelopment: Why did you remove the testing part itself ?? 😕

amilajack: Travis defaults to this. No need to include. But if you want me to revert this I can

.travis.yml
```diff
@@ -12,7 +12,8 @@ branches:
...
+test:
+  - npm test
```
mAAdhaTTah: This should remain `script: npm test`. `test` is not a valid configuration for Travis.

RunDevelopment:
Closed because #1651 was merged,

PR #1562, opened 2018-09-11
Markdown: Added support for code blocks

RunDevelopment:
This PR adds support for markdown [backtick code blocks](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code).

~The new code is intentionally written so that we can easily write a plugin which highlights the code block with the given language.~
~I actually intended that the language does that itself, but that turned out to be really difficult and it lacks certain features, so a plugin it is then.~

It now also supports syntax highlighting!

The highlighting is done in two steps:

1. Add an alias `language-****` containing the given language to the `code-block` token.
    This happens in the `after-tokenize` hook.
2. Highlight the code with the `wrap` hook.

This is to get around the encoding (`util.encode`) of tokens in `Prism.highlight`.

By using this procedure we get the correct execution of the `before-tokenize`, `after-tokenize`, and `wrap` hook for all included languages.

Btw: This implementation would have benefitted from #1524.

### Example

image


ripperhe:
How I use this for Objective-C？ What I should write after "```"?

RunDevelopment:
@ripperhe 
Use `objectivec` after the first tripple backtick.
The language names are the same as you would use after `language-` in `code` or `pre` tags.

ripperhe:
But it not works for me. When I use it for ruby or bash is OK,  only use if for Objective-C is not available. 

I have add this to index.html

```html
<script src="//unpkg.com/prismjs/components/prism-objectivec.min.js"></script>
```

But it not works for me . This is the final result.

[https://ripperhe.com/ZYDebugo/#/Guide/quick-login]( https://ripperhe.com/ZYDebugo/#/Guide/quick-login)

This is GitHub link

[https://github.com/ripperhe/ZYDebugo/blob/master/docs/Guide/quick-login.md](https://github.com/ripperhe/ZYDebugo/blob/master/docs/Guide/quick-login.md)

I had add  `objectivec`.

RunDevelopment:
@ripperhe That's because you are using https://unpkg.com/prismjs@1.15.0/ which does not include this PR.

ripperhe:
@RunDevelopment So, how can I get the latest prism code for Objective-C? What should I write in index.html? Please tell the solution, thank you!  🙏

RunDevelopment:
@ripperhe 
This is an **open** pull request as of right now, which means that until properly reviewed by one of the maintainers, it is not even clear if this PR is going to be included into Prism's code base.
It is also very likely that there will be changes made to this PR during the review process.

I highly suggest you wait until this PR is reviewed and (hopefully) included in the latest release (which might take a few weeks). 

If you **really** need this feature for your website right now, you can do the following:
1. Download [this branch of my fork](https://github.com/RunDevelopment/prism/tree/markdown-code-blocks).
2. Include the downloaded files into the JS folder of your website.
3. Change all references to `prism-markdown.min.js` from `https://unpkg.com/prismjs/components/prism-markdown.min.js` to `path_to_js_folder/prism/components/prism-markdown.min.js`.

ripperhe:
@RunDevelopment OK, I see. Thank you very much. 🙏

mAAdhaTTah:
Looks like there are some conflicts here to resolve.

components/prism-markdown.js
```diff
@@ -137,3 +155,68 @@ Prism.languages.markdown['italic'].inside['bold'] = Prism.languages.markdown['bo
 Prism.languages.markdown['italic'].inside['strike'] = Prism.languages.markdown['strike'];
 Prism.languages.markdown['strike'].inside['bold'] = Prism.languages.markdown['bold'];
 Prism.languages.markdown['strike'].inside['italic'] = Prism.languages.markdown['italic'];
+
+Prism.hooks.add('after-tokenize', function (env) {
+	if (env.language !== 'markdown')
+		return;
```
mAAdhaTTah: I'm not sure what our style policy is on this, but I generally prefer `{}` around `if` blocks like this.

mAAdhaTTah: Yeah, looks like core uses `{}` with `if` blocks. Otherwise, code looks good. Love this improvement.

mAAdhaTTah:
Thought for the future: can we integrate w/ the autoloader so you can write Markdown & it'll autoload the languages you need?

RunDevelopment:
> write Markdown & it'll autoload the languages you need

I also thought about but for that would we need the Autoloader itself to expose the API and find a clever way to highlight the code asynchronously after the markdown pass.

PR #1588, opened 2018-10-21
fix: invoke callback after `after-highlight hook`

Jocs:
Hi, first thanks for you for provide so cool library.

when I use `prism.highlightElement` with `keep markup` plugin. I have encountered some problems.

```javascript
 const code = '<span class="highlight">const a = "nice"</span>'
 const lang = 'javascript'
 const wrapper = document.createElement('div')
 wrapper.classList.add(`language-${lang}`)
 wrapper.innerHTML = code
 prism.highlightElement(wrapper, false, function () {
    const highlightedCode = this.innerHTML
})
```

expect highlightedCode contains with `span.highlight` markup, but it does not. because the `callback` called before `after-highlight` hook?


RunDevelopment:
`callback` according to the [documentation](https://prismjs.com/extending.html#highlight-all):

> An optional callback to be invoked after the highlighting is done. [...]

Now, when exactly do we consider that _"highlighting is done"_? Right now that is after the highlighted code is inserted into its element.

Wouldn't it be better if the callback were invoked after the `complete` hook was run so that all plugins (including [line numbers](https://prismjs.com/plugins/line-numbers/) and [command line](https://prismjs.com/plugins/command-line/)) take effect?

---

Also, `callback` is invoked in [line 252](https://github.com/PrismJS/prism/blob/master/components/prism-core.js#L252) and [line 270](https://github.com/PrismJS/prism/blob/master/components/prism-core.js#L270) so we have to change both.

Jocs:
>Wouldn't it be better if the callback were invoked after the complete hook?

Yeh, I think it's better, I'll update this PR.

Jocs:
@RunDevelopment Updated this PR, can you help review it?

RunDevelopment:
@Jocs Well, I can do that but please run `gulp` before.

Jocs:
@RunDevelopment I have run `gulp` and build the `min` file. 😸

RunDevelopment:
I looked at the code and everything seems to be allright. The Prism website works as expected.

@mAAdhaTTah What do you think about this change?

RunDevelopment:
I looked at the code and everything seems to be allright. The Prism website works as expected.

@mAAdhaTTah What do you think about this change?

RunDevelopment:
I looked at the code and everything seems to be allright. The Prism website works as expected.

@mAAdhaTTah What do you think about this change?

RunDevelopment:
I looked through the code and everything seems to be all right. The Prism website works as expected.

@mAAdhaTTah What do you think about this change?

RunDevelopment:
I looked through the code and everything seems to be all right. The Prism website works as expected.

@mAAdhaTTah What do you think about this change?

RunDevelopment:
I looked through the code and everything seems to be all right. The Prism website works as expected.

@mAAdhaTTah What do you think about this change?

mAAdhaTTah:
Intuitively, this feels "right", in that highlighting isn't "done" until after everything has run. I guess I'm a little concerned this could be a breaking change for anyone relying on `callback` to be run before those hooks.

Jocs:
@mAAdhaTTah I think we can give a notification in the `changelog`. After all, according to the documentation, the callback should be placed after after-highlight, and I really need such a callback.

mAAdhaTTah:
Merged as a bugfix.

Jocs:
@mAAdhaTTah BTW, when will release this bugfix version?

Jocs:
@mAAdhaTTah thank you!

PR #1591, opened 2018-10-22
JS: Regex example

RunDevelopment:
This PR fixes a regex example which Prism does not highlight correctly because it is preceded by a comment.

### Before
image

### After
image

### DIFF
--- a/examples/prism-javascript.html
+++ b/examples/prism-javascript.html
@@ -49,7 +49,7 @@ var foo = /a/, bar = 3/4;</code></pre>
 
 <h2>ES6 features</h2>
 <pre><code>// Regex "y" and "u" flags
-/[a-zA-Z]+/gimyu
+var a = /[a-zA-Z]+/gimyu;
 
 // for..of loops
 for(let x of y) { }


mAAdhaTTah:
I will merge this fix, but should we file this as a bug for the JS language?

RunDevelopment:
You're right that I just treated the symptoms of the underlying problem but that's because the underlying problem is not really fixable.

The regex in the _before_ state did not match because a regex literal has to be preceded by a character like `=`, `,`, `+`, `/`, etc. (__not__ including alphanumeric characters!) followed by any number of spaces (`\s`). But because of the single-line comment, the character preceding our regex is an `s`, so no match.

To correctly match the regex, we have to detect the preceding single-line comment.
_Just include a `//.*` into the lookbehind and done, right?_
Well, no. Consider the following example:

```js
var foo = 1, g = 2;
var a = /* // this will make our lives hard */ 5
/foo/g;
```
(GitHub gets this wrong. It thinks that `/foo/g` is a regex.)

`//` can also be part of strings and even regexes to make matters worse.

So to correctly match the regex we need to correctly match the preceding single-comment for which need to consider comments, strings, and regexes.

Put simply, it's impossible (IMO) to fix this bug with the methods Prism currently offers.
(And I don't see the point in opening an issue which we cannot fix.)

mAAdhaTTah:
Fair enough. This also seems like something that won't come up a lot in practice. Thanks for explaining!

PR #1593, opened 2018-10-23
New language: CIL

sbrl:
I discovered that Prism.js doesn't have CIL (_Common Intermediate Language_, from .NET) support when I [wrote a blog post](https://starbeamrainbowlabs.com/blog/article.php?article=posts%2F322-Disassembly-with-Mono.html) that contained some - so I decided to create my own grammar, since the process looked fairly easy (it was!).


mAAdhaTTah:
Couple of quick notes:

1. Filename should be `prism-cil.js`, to match the rest of the languages.
2. Run `gulp` to generate the minfied version.
3. Update the `components.js` file with the language information.

components/prism-cil.js
```diff
@@ -0,0 +1,26 @@
+Prism.languages.cil = {
+	'displayTitle': "CIL",
```
mAAdhaTTah: `displayTitle` doesn't go here, you already have it in the components.json.

mAAdhaTTah:
Thanks for the contribution. One minor comment, but I'll need someone with stronger regex skills to give this a full review. Also, if this is a .NET extension, should this extend from the `csharp` language?

RunDevelopment:
CIL is a language on its own and C# is to CIL basically what C is to Assembly; Namely, C# compiles to CIL.
So, I don't think that extending C# is going to help very much.

components/prism-cil.js
```diff
@@ -0,0 +1,26 @@
...
+	},
+	
+	// Actually an assembly reference
+	'variable': /\[\w+\]/,
+	
+	'boolean': /\b(?:true|false)\b/,
+	'number': /\b(0x)?[0-9a-fA-F]+\b/i,
+	'punctuation': /[{}[\];(),:]|IL_[0-9A-Za-z]+/,
+	
+	'keyword': /\b(iant|currency|syschar|void|bool|int8|int16|int32|int64|float32|float64|error|unsigned int8|unsigned int16|unsigned int32|unsigned int64|nativeType|decimal|date|bstr|lpstr|lpwstr|lptstr|objectref|iunknown|idispatch|struct|interface|int|unsigned int|nested struct|byvalstr|ansi bstr|tbstr|variant bool|lpstruct|static|public|private|family|final|specialname|virtual|abstract|assembly|default|cil managed|famandassem|famorassem|privatescope|hidebysig|newslot|rtspecialname|unmanagedexp|reqsecobj|pinvokeimpl|value|enum|interface|sealed|abstract|auto|sequential|explicit|ansi|unicode|autochar|import|serializable|nested public|nested private|nested family|nested assembly|nested famandassem|nested famorassem|beforefieldinit|specialname|rtspecialname|class|instance|string|extern|extends)\b/,
```
RunDevelopment: A few things about this keyword list:

1. `initonly` is missing.
2. Please split things like `unsigned int32` into `unsigned` and `int32`. This will reduce the number of keywords.
3. Please sort the keywords alphabetically. This makes the list easier to maintain and you will notice the keywords which are included twice.

components/prism-cil.js
```diff
@@ -0,0 +1,26 @@
...
+		lookbehind: true,
+		pattern: /([^a-zA-Z0-9])\.[a-z]+\b/
+	},
+	
+	// Actually an assembly reference
+	'variable': /\[\w+\]/,
+	
+	'boolean': /\b(?:true|false)\b/,
+	'number': /\b(0x)?[0-9a-fA-F]+\b/i,
+	'punctuation': /[{}[\];(),:]|IL_[0-9A-Za-z]+/,
```
RunDevelopment: `.` and `=` should also be punctuation.

components/prism-cil.js
```diff
@@ -0,0 +1,26 @@
...
+	'class-name': {
+		lookbehind: true,
+		pattern: /([^a-zA-Z0-9])\.[a-z]+\b/
+	},
+	
+	// Actually an assembly reference
+	'variable': /\[\w+\]/,
+	
+	'boolean': /\b(?:true|false)\b/,
+	'number': /\b(0x)?[0-9a-fA-F]+\b/i,
```
RunDevelopment: What about floating point numbers?

components/prism-cil.js
```diff
@@ -0,0 +1,26 @@
...
...\.u|conv\.u1|conv\.u2|conv\.u4|conv\.u8|cpblk|cpobj|div|div\.un|dup|endfault|endfilter|endfinally|initblk|initobj|isinst|jmp|ldarg|ldarg\.0|ldarg\.1|ldarg\.2|ldarg\.3|ldarg\.s|ldarga|ldarga\.s|ldc\.i4|ldc\.i4\.[0-9]+|ldc\.i4\.m1|ldc\.i4\.M1|ldc\.i4\.s|ldc\.i8|ldc\.r4|ldc\.r8|ldelem|ldelem\.i|ldelem\.i1|ldelem\.i2|ldelem\.i4|ldelem\.i8|ldelem\.r4|ldelem\.r8|ldelem\.ref|ldelem\.u1|ldelem\.u2|ldelem\.u4|ldelem\.u8|ldelema|ldfld|ldflda|ldftn|ldind\.i|ldind\.i1|ldind\.i2|ldind\.i4|ldind\.i8|ldind\.r4|ldind\.r8|ldind\.ref|ldind\.u1|ldind\.u2|ldind\.u4|ldind\.u8|ldlen|ldloc|ldloc\.[0-9]+|ldloc\.s|ldloca|ldloca\.s|ldnull|ldobj|ldsfld|ldsflda|ldstr|ldtoken|ldvirtftn|leave|leave\.s|localloc|mkrefany|mul|mul\.ovf|mul\.ovf\.un|neg|newarr|newobj|nop|not|or|pop|readonly\.|refanytype|refanyval|rem|rem\.un|ret|rethrow|shl|shr|shr\.un|sizeof|starg|starg\.s|stelem|stelem\.i|stelem\.i1|stelem\.i2|stelem\.i4|stelem\.i8|stelem\.r4|stelem\.r8|stelem\.ref|stfld|stind\.i|stind\.i1|stind\.i2|stind\.i4|stind\.i8|stind\.r4|stind\.r8|stind\.ref|stloc|stloc\.0|stloc\.1|stloc\.2|stloc\.3|stloc\.s|stobj|stsfld|sub|sub\.ovf|sub\.ovf\.un|switch|tail\.|throw|unaligned\.|alignment|unbox|unbox\.any|volatile\.|xor)\b/
```
RunDevelopment: 1. Why do some end with `.`?
2. Please sort this list by descending length. 
    Because of the way alternations work and the fact that `.` is not included in `\w` (which is why `\b` won't help), some function names are not going to be matched correctly.
    E.g.: `add.ovf` will be matched as `[add].ovf`. 
    This is because alternations will use the first one that matches (in this case `add`).
3. Please compress the list a little.
    There is a lot of redundancy which can be used to make the list smaller.
    E.g.: All functions which start with `conv` can be matched with the following pattern: `conv\.(?:[iu][1248]?|ovf\.[iu][1248]?(?:\.un)?|r\.un|r4|r8)`.

components.json
```diff
@@ -57,6 +57,10 @@
 			},
 			"option": "default"
 		},
+		"cil": {
```
RunDevelopment: You have to place this further down below (under C++, I think).
These here are special languages for Prism.

components/prism-cil.js
```diff
@@ -0,0 +1,26 @@
+Prism.languages.cil = {
+	'displayTitle': "CIL",
+	
+	'comment': /\/\/[^\n]*(?=\r?\n)/,
```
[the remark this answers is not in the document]
RunDevelopment: Also, does CIL support multi-line comments?

components/prism-cil.js
```diff
@@ -0,0 +1,26 @@
...
+	
+	'string': {
+		pattern: /(["'])(?:\\(?:\r\n|[\s\S])|(?!\1)[^\\\r\n])*\1/,
+		greedy: true
+	},
+	
+	'class-name': {
```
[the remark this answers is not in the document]
sbrl: I tried to pick the most appropriate token out of my prism css file I had on-hand.

How do I define an alias?

components/prism-cil.js
sbrl: Done. I originally pulled out of the Wikipedia article with a bit of JS and a regex or two to process it.

sbrl:
All done! Is that better?

components/prism-cil.js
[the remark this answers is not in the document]
RunDevelopment: Via an `alias` property, so in this case `alias: 'class-name'`.

RunDevelopment:
Looks good!

There are still a few things I noticed:

1. Please use non-capturing groups, if the group is not used as a lookbehind.
2. Again, please remove the `package-lock.json` and revert the changes made to `prism.js`. They simply do not belong to this PR.
3. Could you please consider adding some tests? It'd be great to have a test to cover all functions and keywords to verify that everything is matched correctly.

sbrl:
I've made the changes requested (even though [it's recommended](https://stackoverflow.com/a/44210813/1460422) that you commit `package-lock.json`).

How do I add tests? Is there any documentation on this?

RunDevelopment:
[Here's the documentation](https://prismjs.com/test-suite.html) on how to add tests.

sbrl:
Ah, thanks!

Looking at it, I think test cases are going to be challenging as I don't know the language well enough to create a valid cut-down program that would do the job as a test case. In addition, there's a lot of boilerplate code that's appears to be common across multiple IL files, which further complicates the process of writing tests.

Thoughts?

RunDevelopment:
Prism test cases usually test specific features (e.g. numbers, comments, small aspects of a language in general). So don't have to come up with a minimal program which includes all features. Just write small tests which test one feature each.

For examples on how to do tests right, I suggest looking at the JavaScript and C-like tests.

sbrl:
Tests added! Let's see if Travis likes them.

tests/languages/cil/keywords.test
```diff
@@ -0,0 +1,80 @@
+abstract ansi assembly auto autochar beforefieldinit bool bstr byvalstr cil class currency date decimal default enum error explicit extends extern famandassem family famorassem final float32 float64 hidebysig iant idispatch import initonly instance int int16 int32 int64 int8 interface iunknown lpstr lpstruct lptstr lpwstr managed nativeType nested newslot objectref pinvokeimpl private privatescope public reqsecobj rtspecialname sealed sequential serializable specialname static string struct syschar tbstr unicode unmanagedexp unsigned value variant virtual void
```
RunDevelopment: How about a few new lines?

RunDevelopment:
@sbrl Could you please fix the instructions test. After the test passes, we should be able to merge this.

sbrl:
@RunDevelopment Hrm. It looks to be failing, but I don't understand what [Travis is trying to tell me](https://travis-ci.org/PrismJS/prism/jobs/447741928#L2002).

RunDevelopment:
@sbrl 
Travis gives you the output of `npm test`, so you can run all tests locally on your computer with the `npm test` command (or `npm test -- --lanuage=cil` for CIL related test). No need to ask Travis.

The error message you get is because the token stream (the one that Prism currently creates; L1996) is not equal to the expected token stream (the JSON part of your test file; L1999).
The following red and green lines are a git-like difference view.

I suggest to use the local `test.html` to visualize your test case, so you can make out the errors.
Adjust the tested instruction list and the `function` pattern until everything is highlighted nicely. 

Then you can use my favorite dirty trick:
Insert your updated instruction list into the instructions test file, run `npm test` and let it fail. Then you can just copy the token stream Prism outputted into the error message and insert it as the expected token stream into your test file. (Maybe do some formatting.)
Still, be sure to check that the token stream is correct.

I hope that helps.

sbrl:
Thanks for the helpful advice, @RunDevelopment! That test tool is very helpful (for my own reference, make sure you re-select the language every time you refresh the page :P).

I've improved the handling of command prefixes and made the tests pass.

RunDevelopment:
@sbrl I'm glad that it helped you out! (I also know how annoying it is to re-select the language, hence #1529)

The tests all pass, but I have one last nit before we can merge this: 
The `function` pattern still contains a capturing group `(i[1248]?|r[48]|ref)`. Please make it non-capturing.

components/prism-cil.js
```diff
@@ -0,0 +1,27 @@
+Prism.languages.cil = {
+	'comment': /\/\/.*/,
+	
+	'string': {
+		pattern: /(?:["'])(?:\\(?:\r\n|[\s\S])|(?!\1)[^\\\r\n])*\1/,
```
RunDevelopment: The first group has to be capturing otherwise, strings won't be highlighted.
Maybe a string test would be good?

components/prism-cil.js
```diff
@@ -0,0 +1,27 @@
...
+		greedy: true
+	},
+	
+	'directive': {
+		pattern: /([^a-zA-Z0-9])\.[a-z]+\b/,
```
RunDevelopment: 1. `.method` will not be matched. Please change the first group to `(^|\W)` (`[^a-zA-Z0-9]` == `\W`).
2. `.ctor()` will be matched. Isn't there always at least one space after a directive, so why don't we use `(?=\s)` instead of `\b`?

components/prism-cil.js
```diff
@@ -0,0 +1,27 @@
...
+		lookbehind: true,
+		alias: 'class-name'
+	},
+	
+	// Actually an assembly reference
+	'variable': /\[\w+\]/,
```
RunDevelopment: What about `[System.Core]`?

RunDevelopment:
While playing around I also found a few minor things:

1. `object`, `method` (for overwrites), `native` (IntPtr), `literal` (consts and enums), `valuetype` (enums), `runtime` (delegates), and `implements` (interfaces) should be keywords.
2. try-catch/try-finally blocks don't have `catch`/`finally` highlighted.

components/prism-cil.js
sbrl: Good thinking! I didn't know JS supported positive lookaheads.

sbrl:
Thanks for the comments! PR updated. This is a _much_ longer process than I thought it would be :P

RunDevelopment:
@sbrl Looks good!

I found another missing keyword :) `char`. Also, all of the unsigned integers are missing, so `uint` and all `uint<N>`.
(And please update the keyword test to include the newly added ones.)

Don't worry, this is going to be the last nit.

sbrl:
Done, @RunDevelopment

RunDevelopment:
Looks good! Tests pass. @mAAdhaTTah I think we are ready to merge this.

sbrl:
What's the status on this please, @RunDevelopment? I thought I'd done everything required, but it's been a while and this hasn't been merged yet? Did I miss something?

RunDevelopment:
@sbrl You've done everything we ask of and you definitely have my OK. 
It's just that @mAAdhaTTah seems to be quite busy at the moment and hasn't had the time to look at changes himself.

So don't worry! This has not been forgotten.

mAAdhaTTah:
@RunDevelopment @sbrl I haven't forgotten! Holidays are busy :). I want to take a pass through Prism's PRs sometime this week(end) so I'll be taking a closer look at this soon.

mAAdhaTTah:
@sbrl If you can resolve the conflict (just merge up from `master` & rerun `gulp`), I can merge this.

sbrl:
Done, @mAAdhaTTah :D

mAAdhaTTah:
@sbrl Thank you for the contribution!

Issue #1590, opened 2018-10-21
Support for comments in JSON


mariusschulz:
What's your take on supporting single-line and block comments in JSON? Several popular projects (e.g. TypeScript) allow comments in JSON configuration files:

```json5
{
  "compilerOptions": {
    "target": "es5",      /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
    "module": "commonjs", /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
    // "lib": [],         /* Specify library files to be included in the compilation. */
    // "allowJs": true,   /* Allow javascript files to be compiled. */
    // "checkJs": true,   /* Report errors in .js files. */
    // "jsx": "preserve", /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */

    // ...
}
```

Although the JSON spec doesn't include comments, I think it would be helpful to have support for them in Prism; comments can be particularly helpful in technical blog posts.

What do you think? I'll be happy to send a PR! :)

mAAdhaTTah:
I don't know as we need to modify JSON; if you want comments, couldn't you use the JS language instead? I also find most of the editors I use don't highlight comments in JSON files correctly.

That said, I'm open to being convinced otherwise.

mariusschulz:
That's exactly what I did to unblock myself for now. I highlighted my code example using the JS language, but the results aren't great since quoted properties and string literals use the same color:

![A JSON object that was highlighted using the JS language](https://i.imgur.com/5XrmF7h.png)

If you feel uncomfortable extending JSON directly, we could introduce a new [JSON5](https://json5.org/) language that extends the existing JSON language. However, I'd prefer to add comments to JSON directly so that we can use comments to indicate that some parts have been omitted:

```json5
{
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    // ...
}
```

mariusschulz:
For further consideration, here's what the above screenshot would look like if the JSON language supported comments. I think it looks much better! 

![A JSON object that was highlighted using the JSON language with support for comments](https://i.imgur.com/yYH89Ed.png)

We could reuse the comment patterns from the `clike` language:

```diff
Prism.languages.json = {
+  comment: [
+    {
+      pattern: /(^|[^\\])\/\*[\s\S]*?(?:\*\/|$)/,
[...]

PR #1595, opened 2018-10-24
Adds support for comments in JSON

mariusschulz:
Adds support for comments in JSON, as discussed in https://github.com/PrismJS/prism/issues/1590.

A JSON object that was highlighted using the JSON language with support for comments

@RunDevelopment You [mentioned](https://github.com/PrismJS/prism/issues/1590#issuecomment-432559854) that `property` needs to be greedy now that it's no longer the first pattern. Do you have an example JSON file that would be highlighted incorrectly if `greedy` isn't `true`?


RunDevelopment:
Any JSON which has properties that contain comment-like text will be highlighted incorrectly.
E.g.

```json
{
    "//": 5,
    "/*": "I have a problem",
    "*/": null
}
```

RunDevelopment:
The code looks good and there is even a test to check the new feature.

So, @mAAdhaTTah. What do you think?

mAAdhaTTah:
We're on board with this change.

PR #1598, opened 2018-10-27
HTTP: Improved content highlighting

RunDevelopment:
This PR extends the content highlighting capabilities of HTTP.

1. Addes support for `application/javascript` and `text/css` content types.
2. Changed the languages associated with certain types:
    1. `application/json` now uses `json` with `javascript` as a backup instead of just `javascript`.
    2. All XML types now use `Prism.languages.xml` instead of `markup`.
    3. The HTML type now uses `html` instead of `markup`.
3. Added support for highlighting based on [suffixes](https://en.wikipedia.org/wiki/Media_type#Suffix) for XML and JSON.
    E.g. `image/svg+xml`, `application/xhtml+xml`, `application/atom+xml` will use `xml` and `application/calendar+json` will use `json`.
4. Stylistic changes:
    1. Fixed indentation.
    2. Used single quotes for token names.


mAAdhaTTah:
Based on the tests, I like this change. I'll have to dig through the regexes to give this a full review.

PR #1599, opened 2018-10-27
Latte templating language support

jahudka:
This PR adds support for the Latte templating language used by the Nette PHP framework. It's similar to Smarty; the most obvious difference is the support for `n:attribute` macros.


components/prism-latte.js
```diff
@@ -0,0 +1,178 @@
+/* TODO
+	Add support for the n:syntax attribute macro
+	Add support for string interpolation using {$advancedSyntax} within macros
+*/
+
+(function(Prism) {
+
+	var string_interpolation = {
```
RunDevelopment: This variable is only used once in `inside.string[1].inside.interpolation`, so why not inline it?

components/prism-latte.js
```diff
@@ -0,0 +1,178 @@
...
+		],
+		'word': {
+			pattern: /:?[a-zA-Z0-9_.-]+(?::[a-zA-Z0-9_.-]+)*[:!]?(?=[\s,\])]|$)/,
+			alias: 'string'
+		},
+		'operator': /[+\-*\/%]|==?=?|[!<>]=?|&&|\|\|?/,
+		'punctuation': /[\[\]().,:`]+|->|=>/
+	};
+
+	function mkmacro(begin, content, closing) {
```
RunDevelopment: Please give this function a more readable name and add some documentation.

components/prism-latte.js
```diff
@@ -0,0 +1,178 @@
...
+						inside: inside
+					}
+				}
+			}
+		}, Prism.languages.markup.tag);
+	}
+
+	// Tokenize all inline Latte expressions
+	Prism.hooks.add('before-tokenize', function(env) {
+		if (!/(?:\{[a-zA-Z$_!])/ig.test(env.code)) {
```
RunDevelopment: Why the group? Do the `i` and `g` flags have any effect?

components/prism-latte.js
```diff
@@ -0,0 +1,178 @@
...
+
+			var valid, m;
+
+			switch (syntaxMode) {
+				case 0: return false;
+				case 1: valid = match.charAt(1) !== '{'; break;
+				case 2: valid = match.charAt(1) === '{'; break;
+			}
+
+			if (m = valid ? syntaxPattern.exec(match) : null) {
```
RunDevelopment: Please extract this assignment.

examples/prism-latte.html
```diff
@@ -0,0 +1,63 @@
...
+
+</code></pre>
+
+<h2>Syntax options</h2>
+<pre><code>{syntax double}
+	{{$foo}}
+	&lt;script>
+		(function() { /* This is JS, not Smarty */ } ());
+	&lt;/script>
+{/syntax}
```
RunDevelopment: This is not highlighted.

components/prism-latte.js
```diff
@@ -0,0 +1,178 @@
...
+			{
+				pattern: /\b(?:null|as)\b/,
+			},
+			{
+				pattern: /\b(?:or|and)\b/,
+				alias: 'operator'
+			}
+		],
+		'word': {
+			pattern: /:?[a-zA-Z0-9_.-]+(?::[a-zA-Z0-9_.-]+)*[:!]?(?=[\s,\])]|$)/,
```
RunDevelopment: Please use `\w` instead of `a-zA-Z0-9_`.
There are also a few other places where the same should be done.

components/prism-latte.js
```diff
@@ -0,0 +1,178 @@
...
+		],
+		'filter': {
+			pattern: /((?:^|[^|])\|\s*)(?!\d)\w+/,
+			lookbehind: true,
+			alias: 'function'
+		},
+		'function': /\b(?!\d)\w+(?=\s*\()/,
+		'keyword': [
+			{
+				pattern: /\b(?:true|false)\b/,
```
RunDevelopment: Why is this a keyword and a boolean? Almost all other languages (including PHP) have a `boolean` pattern and that's it, no `keyword`.

RunDevelopment:
See #2140.

PR #1600, opened 2018-10-27
NEON configuration language support

jahudka:
NEON is a configuration language similar to YAML, with the notable addition of so-called "NEON entities". NEON is used mainly by the Nette PHP framework.


RunDevelopment:
See #2140.

PR #1607, opened 2018-11-12
Show invisibles: changed color

RunDevelopment:
This PR changes the color of the placeholders of invisible characters.
They will now use the default text color of the theme as the base and blend in instead of a fixed color for all themes.

(I also used tabs for indentation.)

### Before

Default:
image

Tomorrow night:
image

### After

Default:
image

Tomorrow night:
image


mAAdhaTTah:
Love this. I always thought the dark theme dots were too bright.

PR #1610, opened 2018-11-12
Show invisibles inside tokens

### LINKED ISSUE #1608: Show invisibles does not show invisibles inside of tokens
Invisible characters within tokens (such as comments, strings, CSS selectors, etc.) are not displayed.

image

[More examples](https://prismjs.com/plugins/show-invisibles/)

RunDevelopment:
This PR fixes #1608.

The patterns for invisible characters (ICs) are now also recursively added to the `inside` of every pattern.

I also had to change the color of ICs because in #1607 I changed the color to `inherit` which causes the ICs to have different colors depending on which token they are inside of.
So, to make everything consistent, I changed the color to grey (`rgb(128, 128, 128)`) and made it half transparent so it would blend in. Looks really similar to #1607.

---

Most of the code is just traversing the language definition, handling the different kinds of tokens (simple regex, array, pattern object), and trying not to fall because of the added tokens for ICs and the `rest` property.


plugins/show-invisibles/prism-show-invisibles.js
```diff
@@ -1,21 +1,80 @@
...
+	 * @param {Object} grammar
+	 */
+	function addInvisibles(grammar) {
+		if (!grammar || grammar['tab'])
+			return;
+
+		// assign invisibles here to "mark" the grammar in case of self references
+		for (var name in invisibles) {
+			if (invisibles.hasOwnProperty(name))
+				grammar[name] = invisibles[name];
```
mAAdhaTTah: `{}`

plugins/show-invisibles/prism-show-invisibles.js
```diff
@@ -1,21 +1,80 @@
...
+		}
+
+		for (var name in grammar) {
+			if (grammar.hasOwnProperty(name) && !invisibles[name]) {
+				if (name === 'rest')
+					addInvisibles(grammar['rest']);
+				else
+					handleToken(grammar, name);
```
mAAdhaTTah: And last one I think.

Issue #1609, opened 2018-11-12
Show Invisibles: Potential problems


RunDevelopment:
Because Show Invisibles modifies the language definition itself, there might be problems with:

1. The `rest` property is "unfolded" **after** we add the patterns for ICs which results in the tokens of `rest` being highlighted **after** the IC tokens.
2. Other plugins (e.g. Autolinker) also modify language definitions at run-time (meaning via hooks). Show Invisibles has to be run after these plugins to be compatible.

RunDevelopment:
I finally happened. Show Invisibles modify grammars broke something. Details here: #3273.

PR #1612, opened 2018-11-12
Show Invisibles: Corrected load order

RunDevelopment:
This PR corrects the load order of Show Invisibles, Autolinker, and Data-URI Highlight so that Show Invisibles is loaded after the other ones.

Details as to why in #1609.


mAAdhaTTah:
Bah, conflicts here too. Tough merging so many PRs at once :/

PR #1620, opened 2018-11-18
Add N1QL language support

### LINKED ISSUE #1619: Add support for N1QL
[N1QL](https://www.couchbase.com/products/n1ql), is a query language for the Couchbase NoSQL database. 

While similar to SQL, the classifications of keywords, operators and booleans differ. There are also some structural differences. This means using the SQL syntax highlighter doesn't quite hit the spot. 

I already have a pull request ready to go, but realised I had no issue to link it to. I will raise the request soon!

TMWilds:
[N1QL](https://www.couchbase.com/products/n1ql) is a query language for the Couchbase NoSQL database. Fixes #1619.

N1QL, while intentionally similar to SQL, is not an extension or variant of SQL. As such, there are a number of differences in the classification of general keywords, operators and booleans, along with some other differences and structural changes.

This contribution adds syntax highlighting for N1QL, including the full set of builtin [functions](https://docs.couchbase.com/server/6.0/n1ql/n1ql-language-reference/functions.html) and [operators](https://docs.couchbase.com/server/6.0/n1ql/n1ql-language-reference/operators.html). 

Some examples:
n1qlselectexample
n1qlselectadvancedexample


components/prism-n1ql.js
```diff
@@ -0,0 +1,22 @@
...
+	'boolean': /\b(?:TRUE|FALSE)\b/i,
+	'number': /\b[+\-]?(?:0|[1-9]\d*)(?:\.\d*)?(?:[eE][+\-]?\d+)?\b|\b\d+\.?\d*|\B\.\d+\b/i,
+	'operator': /[-+*\/=%]|!=|==?|\|\||<(?:[>=])?|>=?|\b(?:AND|ANY|ARRAY|BETWEEN|CASE|ELSE|END|EVERY|EXISTS|FIRST|IN|LIKE|NOT|OR|THEN|VALUED|WHEN|WITHIN)\b/i,
```
RunDevelopment: `(?:[>=])?` == `[>=]?`

components/prism-n1ql.js
```diff
@@ -0,0 +1,22 @@
...
+	'parameter': [
+		/\$[\w.]+/
+	],
+	'string': {
+		pattern: /(^|[^@\\])("|')(?:\\[\s\S]|(?!\2)[^\\]|\2\2)*\2/,
+		greedy: true,
+		lookbehind: true
+	},
+	'identifier': {
+		pattern: /(^|[^@\\])(`)(?:\\[\s\S]|(?!\2)[^\\]|\2\2)*\2/,
```
RunDevelopment: Also, this pattern can be simplified to
```js
/(^|[^@\\])`(?:\\[\s\S]|[^\\`]|``)*`/
```

components/prism-n1ql.js
```diff
@@ -0,0 +1,22 @@
+Prism.languages.n1ql = {
+	'comment': /(?:\/\*[\s\S]*?\*\/)/,
+	'parameter': [
+		/\$[\w.]+/
+	],
+	'string': {
+		pattern: /(^|[^@\\])("|')(?:\\[\s\S]|(?!\2)[^\\]|\2\2)*\2/,
```
RunDevelopment: `string` and `identifier` both cannot be preceded by an `@`. I'm guessing that you copied this from SQL where strings and variables interfere with each other but since this is not the case in N1QL, you can remove it.

The backslash is also bugging me. I don't understand why it is there neither here nor in SQL. Am I missing something?

RunDevelopment:
Here is my full review:

Your current implementation does not support [object construction](https://docs.couchbase.com/server/6.0/n1ql/n1ql-language-reference/constructionops.html#object-construction) and [array slicing](https://docs.couchbase.com/server/5.5/n1ql/n1ql-language-reference/nestedops.html#array-slicing).

These two and my comments are all I could find. Really good work!

TMWilds:
@RunDevelopment Thank you for the feedback! Seems to mostly come down to me not being the best with regexes, and a few oversights also... 
I'll aim to make the changes and additions you suggested soon. Thanks again!

components/prism-n1ql.js
TMWilds: You're completely right with the `@`. I did borrow heavily from the SQL regexes here, which is why this slipped in. 
I believe the backslash is so you cannot have a single backslash at the end of the string. 
For example, this prevents us matching `"foo\"`, which is an invalid string due to the last quotation mark.

RunDevelopment: If I were to guess it's probably a relic of the time when Prism did not support greedy matching.
We generally don't worry about highlighting syntactically incorrect code, so that should not concern us.
It's probably safe to remove the whole lookbehind.

TMWilds:
Took me much longer than I expected to get round to doing this, but the changes @RunDevelopment have been added!

RunDevelopment:
The only real conflict is in `prism-show-language.js`. 
To replace

```
<<<<<<< master
var Languages = /*languages_placeholder[*/{ ... }/*]*/;
======= 
var Languages = /*languages_placeholder[*/{ ... }/*]*/;
>>>>>>> master
```

with

```
var Languages = /*languages_placeholder[*/{ ... }/*]*/;
```

What's between the brackets doesn't really matter, cause `gulp` will fill it in.

`components.js` and `prism-show-language.min.js` are both generated so running `gulp` will resolve these.

TMWilds:
Thanks @RunDevelopment, that seems to have gone well!

RunDevelopment:
Thank you for contributing!

PR #1622, opened 2018-11-19
Resolves #1618, preserve the markup if it's the last element

### LINKED ISSUE #1618: Keep Markup: the markup is removed if it's the last element
I don't if this is intended or not but if the markup (in the exemple below `<div class="keep"></div>"`) is the last element, the Keep Markup plugin will remove it:

```html
<!DOCTYPE html>
<html>
  <head>
    <link href="prism.css" rel="stylesheet" />
    <style>
    .keep {
      height: 1rem;
      width: 1rem;
      background-color: red;
    }
    </style>
  </head>
  <body>
    <pre><code class="language-css"><div class="keep"></div>p { color: red }<div class="keep"></div></code></pre>
  </body>
</html>
```

And here's the result:

error

As you can see the first `<div class="keep"></div>` is preserved but the last is removed.
If I add some content after the `div` then it's working as expected:

good

Mogztter:


RunDevelopment:
@Mogztter No, plugins are not tested (at least right now).

RunDevelopment:
It seems like the current implementation of Keep Markup assumes that every tag contains at least one character, causing #1618.

This PR does handle the case of the last tag correctly but it handles some other case worse than the current implementation.

1. Order swapping:
    The order of all zero-length tags which have the same start position will the reversed. E.g.: 
    `<x></x><y></y>` -> `<y></y><x></x>`
1. Tag inclusion:
    Zero-length tags after a closing tag will be included into the preceding tag (in reversed order). E.g.:
    `<x>foo</x><y></y><z></z>` -> `<x>foo<z></z><y></y></x>`

Tag inclusion is probably just a symtom of #1640.

Right now, this implementation does more harm than good, so we will have to address the above points before we can merge this.

Mogztter:
OH! we really need tests :smile: 
Can I submit a proposal to add unit tests on this plugin ? Otherwise it will be a nightmare to fix :neutral_face:

RunDevelopment:
Of course, you can submit the proposal but it will require to implement a very different test suite than the one we have right now. So, no small task.

> Otherwise it will be a nightmare to fix

I don't see why that is. A little hacky but I just made a little test page based on the [index page of the plugin](https://github.com/PrismJS/prism/blob/master/plugins/keep-markup/index.html). Insert the markup to be tested and verify yourself. Not pretty but it works.

mAAdhaTTah:
If we wanted to go all-out, something w/ JSDOM + snapshot tests could work. We have a (admittedly complex) mocha/chai setup for the current tests. Might be nice to get some tests going for the plugins.

Mogztter:
> If we wanted to go all-out, something w/ JSDOM + snapshot tests could work. 

That's exactly what I had in mind :+1: 
I've added a failing tests but I don't know how to solve this issue. To be honest I'm a bit lost with all the start/end position...

plugins/keep-markup/tests/test.js
```diff
@@ -0,0 +1,88 @@
+const expect = require('chai').expect;
+const jsdom = require('jsdom')
+const { JSDOM } = jsdom
+
+require('../../../prism')
+// fake DOM
+global.self = {}
+global.self.Prism = Prism
+global.document = {}
+document.createRange = function () {
```
Mogztter: `createRange` is not implemented in JSDOM: https://github.com/jsdom/jsdom/issues/317

plugins/keep-markup/tests/test.js
```diff
@@ -0,0 +1,88 @@
...
+}
+global.self.document = document
+
+require('../prism-keep-markup')
+
+describe('Prism Keep Markup Plugin', function () {
+
+	function execute (code) {
```
Mogztter: I'm using a fake `createRange` function to register the "actions"

plugins/keep-markup/tests/test.js
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	it('should keep <span> markup', function () {
+		const result = execute(`<code class="language-javascript">x<span>a</span>y</code>`)
+		expect(result.start.length).to.equal(1)
+		expect(result.end.length).to.equal(1)
+		expect(result.nodes.length).to.equal(1)
+		expect(result.nodes[0].nodeName).to.equal('SPAN')
+	})
+	it('should preserve markup order', function () {
```
Mogztter: This test currently fails, `<a></a><b></b>` are inverted in the result.

**Input**
```html
<code class="language-javascript">x<a></a><b></b>y</code>
```
**Output**
```html
<code class="language-javascript">x<b></b><a></a>y</code>
```

mAAdhaTTah:
@Mogztter I really appreciate the enthusiasm and the contribution. I think if we're going to do this (introduce testing for plugins), we should break up this PR into one that introduces the framework and maybe some basic, passing tests, then a separate PR that adds tests for & fixes this bug. That'll allow us to evaluate each of those things separately.

Mogztter:
> I think if we're going to do this (introduce testing for plugins), we should break up this PR into one that introduces the framework and maybe some basic, passing tests, then a separate PR that adds tests for & fixes this bug. 

@mAAdhaTTah Sure no worries.
My [initial bugfix is a one-liner](https://github.com/PrismJS/prism/pull/1622/files#diff-53db008d4081ae1e03b14b6b780ea7bcR58). I can revert this change and only add a test that check that the extension has been registered ? ie. `Prism.hooks.all['before-highlight']` and `Prism.hooks.all['after-highlight']` are not empty ? Does it sound good to you ?

mAAdhaTTah:
@Mogztter I'd prefer to evaluate the bugfix in a separate PR. It's not clear to me [this comment](https://github.com/PrismJS/prism/pull/1622#discussion_r237923134) isn't a result of introducing that change. What I'd suggest is ending up with 2 PRs:

1. Introduce the testing infrastructure, including some basic tests for this plugin that pass, without changing any plugins. We can then start adding tests for the plugins as we go.
2. A PR that fixes this bug and introduces tests to cover the bugfix.

How you want to break that up (modifying this PR to be just tests, or closing and creating them separately) is up to you, but that's the place I'd like to end up if we want to add tests at this time.

The smaller the PR, the easier it will be to get each of them in, and I'd like to evaluate & think through the testing infrastructure as well as whatever tests you're introducing without also needing to evaluate the change(s) you're making to the plugin.

Does that make sense? My apologies for making this complex; I threw out the suggesting not expecting you to run with it so fast. And I appreciate it, but it does ramp up the overall complexity of what you're doing, and I just want to simplify it as much as possible.

Mogztter:
@mAAdhaTTah Done, see https://github.com/PrismJS/prism/pull/1646

mAAdhaTTah:
Let's get the tests in first, then we'll get this fix.

Mogztter:
@mAAdhaTTah Rebased :+1: 
@RunDevelopment You're right, I can reproduce the "order swapping" regression. With this fix the test "should preserve markup order" is failing :wink:

mAAdhaTTah:
Hey, sorry for the delay on this, but coming back to this now. I see one code change, which I _guess_ is fine? Does this change the commented out test? We should either delete it and replace it with a test that represents the intended behavior or update it.

That said, if this change removes that `<a>` tag in the test, that seems like a breaking change to me, and if the underlying issue is #1640, then I think this fix is invalid, unfortunately. We should probably tackle that problem holisitically.

I'm going to go ahead & close, but happy to reopen if anything above I've said it wrong.

PR #1631, opened 2018-11-23
Removed unused `insertBefore` overload

### LINKED ISSUE #1505: Prism.languages.insertBefore overload
`insertBefore` has a strange [overload](https://github.com/PrismJS/prism/blob/master/components/prism-core.js#L109) with two arguments where it just copies everything from `before` (which is used like `insert`) into a language definition.

This overload isn't documented ~anywhere~ well and I couldn't find any references to it in our code base.

Should we just remove it?

RunDevelopment:
This PR removes the unused `Prism.languages.insertBefore` overload mentioned in #1505.

This PR resolves #1505.


mAAdhaTTah:
Looks good; needs conflicts resolved :/

PR #1639, opened 2018-11-29
Python: Decorator support

### LINKED ISSUE #1635: Python - add decorator detection
It would be nice to recognize decorator invocation, eg:

```python
@some.decorator
def foo(bar):
    pass
```
Decorator invocation can only appear as first token on a line.

This definition should work:
```js
    'decorator': {
        pattern: /(^\s*)@\w+(\.\w+)?/i,
        lookbehind: true
    },
```

RunDevelopment:
This PR adds support for Python decorators implementing #1635 using the solution proposed by @spectras.

I also added 2 aliases: `annotation` to be like Java and `punctuation` to change the actual color (like Java).

This resolves #1635.


sardmhmd:
You need to leave a comment indicating the requested changes.

tests/languages/python/decorator_feature.test
```diff
@@ -0,0 +1,28 @@
+@some.decorator(foobar)
```
mAAdhaTTah: Do we want to highlight the whole line as a decorator or just the `@` symbol? How do other highlighters do it?

RunDevelopment: I err towards highlighting `@some.decorator` as Notepad++ and VS both do.
(VS Code doesn't highlight anything)

spectras:
Wow that was quick.
Just a short comment to thank all for the amazing work. :+1:

PR #1642, opened 2018-12-01
Python: F-strings and string prefixes

### LINKED ISSUE #1636: Python: F-string support
The python language definition does not support [strings interpolation](https://www.python.org/dev/peps/pep-0498/) aka. f-strings.

RunDevelopment:
This PR adds support for [string interpolation](https://www.python.org/dev/peps/pep-0498/) (aka. f-strings) and makes the [string prefixes](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals) part of the string.

### Screenshot
image

### Known issues

I assumes that strings inside the interpolation expression are 'nice'. So strings with unfortunate numbers of curley braces will cause incorrect highlighting: E.g.: `f"{'}'}"`.

---

This PR resolves #1636.


mAAdhaTTah:
Can we include the "known failure" in our list of list of known failures for Python?

RunDevelopment:
I also removed the old known issue which works now.

image

mAAdhaTTah:
Does that mean #1464 is no longer necessary?

RunDevelopment:
It is still necessary because it has 3 feature ([empty triple-quoted strings](https://github.com/PrismJS/prism/pull/1464/files#diff-e99f677a9aaad5549db82df4b410cb1fR11), the example, and a fix for [comments](https://github.com/PrismJS/prism/pull/1464/files#diff-e99f677a9aaad5549db82df4b410cb1fR7)) of which this only covers the example part.

PR #1645, opened 2018-12-01
Fix typos

Gelma:


RunDevelopment:
Thank you for contributing to Prism!

RunDevelopment:
Thank you for contributing to Prism!

But you only need to modify `prism-core.js` and `plugins/file-higlight/prism-file-highlight.js`. 
Then you can use the `npx `[`gulp`](https://github.com/gulpjs/gulp) command to rebuild Prism and copy your changes to to `prism.js`.

RunDevelopment:
Captured by #1755.

PR #1648, opened 2018-12-03
add the pipe operators in reasonml

Schniz:
`|>` and `->` will now be considered as single tokens instead of a composition of 2 tokens, making ligatures work


RunDevelopment:
Thank you for your contribution!

Please also run [gulp](https://github.com/gulpjs/gulp) to generate the minified files.

Schniz:
oh, sure! no problem. thanks! 😸

PR #1650, opened 2018-12-03
CSS extras: Highlighting for pseudo class arguments

RunDevelopment:
This PR adds highlighting for the arguments of CSS pseudo-classes e.g `:not(.foo)`. It's a first in that it changes the otherwise uniformly colored CSS selectors.
Expressions of the form [an+b](https://drafts.csswg.org/css-syntax/#anb) are also highlighted.

This also fixes the bug of pseudo-classes with arguments matching waaay too much because of `\(.*\)` in e.g. `foo:not(.bar) > :not(:hover)`.

### Example

image

---

The pseudo-class pattern implementation of this PR is very different from the current one in the way that it will not attempt to match the arguments at all. This greatly simplifies the overall complexity when trying to handle nested pseudo-classes/selectors and is similar to the way C-like handles functions.


components/prism-css-extras.js
```diff
@@ -35,7 +35,16 @@ Prism.languages.css.selector = {
 				],
 				'operator': /[|~*^$]?=/
 			}
-		}
+		},
+		'n-th': {
```
mAAdhaTTah: `n-th` is a class in our themes already?

RunDevelopment: No, but it's a necessary wrapper for the `number` and `operator` tokens.

PR #1652, opened 2018-12-04
PHP: Fixed closing tag issue

### LINKED ISSUE #1582: Highlighting issue with PHP when code contains xml snippet
Hello,

With the following code:

```php
<?php

  echo '<?xml version="1.0" encoding="UTF-8"?><tracker />';
  echo PHP_EOL;

```
PrismJS does not syntax highlight the second statement (and the following ones, if any).

Tested on PrismJS 1.15.0

Best regards,
Nicolas

RunDevelopment:
This PR fixes #1582.

image

---

I will quickly explain the changes because otherwise, this will be a hell to review. Let's start with the easy ones:

I changed the `delimiter` and inserted it before `comment` because otherwise, `<?php // ?>` and `<? # ?>` (among others) will be highlighted incorrectly. Also, there has to be [a single whitespace after `<?php`](http://php.net/manual/en/language.basic-syntax.phptags.php#118827).

In `before-tokenize`, I changed the pattern to check if templating is even necessary (first line). I did this because whenever `<?php` matches, so will `<?`. Therefore `<?php` can be removed.

Now for the juicy part of this PR: `phpPattern`.
`(?:<\?php|<\?)` was adjusted similarly to the reasoning above and `[\s\S]*?` was adjusted to take comments and strings into account. Note that this does not include nowdoc and heredoc strings. The pattern is already complex enough.
The pattern consists of:
- `("|')(?:\\[\s\S]|(?!\1)[^\\])*\1` Strings (single and double quoted).
- `\/\*[\s\S]*?(?:\*\/|$)` C-style multiline comments
- `(?:\/\/|#)(?:[^?\n\r]|\?[^>\n\r])*` C-style and shell singleline comments which are NOT allowed to [contain a closing tag](http://php.net/manual/en/language.basic-syntax.phptags.php#122618). These comments are only included because otherwise you could use them to trip strings and multiline comments e.g.:
    ```php
    <?php
    # '
    echo '?>';
    ?>
    ```
- `[^"'/#]|\/[^*/]` Everything besides the above.

I hope this makes everything clear.


RunDevelopment:
Tests failed because I deleted the branch faster than Travis could fetch. (I will keep that in mind from now on.)

Master branch tests pass, so no problem.

PR #1658, opened 2018-12-10
Add Vala language

TemplarVolk:


components/prism-vala.js
```diff
@@ -0,0 +1,68 @@
...
+			// class Foo : Bar
+			pattern: /(\b(?:class|interface)\s+[A-Z]\w*(?:\.\w+)*\s*:\s*)[A-Z]\w*(?:\.\w+)*\b/,
+			lookbehind: true,
+			inside: {
+				punctuation: /\./
+			}
+		},
+		{
+			// class Foo
+			pattern: /((?:\b(?:class|interface|new)\s+)|(?:catch\s+\())[A-Z]\w*(?:\.\w+)*\b/,
```
RunDevelopment: This does not include [structs and enums](https://wiki.gnome.org/Projects/Vala/Tutorial#Value_Types).

components/prism-vala.js
```diff
@@ -0,0 +1,68 @@
...
+Prism.languages.insertBefore('vala','string', {
+	'raw-string': {
+		pattern: /(""")[\s\S]*?\1/,
+		greedy: true,
+		alias: 'string'
+	},
+	'template-string': {
+		pattern: /@"[\s\S]*?"/,
+		greedy: true,
+		alias: 'string',
```
RunDevelopment: It's better to have a `'string': /[\s\S]+/` after `interpolation` instead of this alias, so that the code inside `interpolation` is not highlighted like a string.

components/prism-vala.js
```diff
@@ -0,0 +1,68 @@
...
+		inside: {
+			'interpolation': {
+				pattern: /\$(?:\([^)]*\)|[a-zA-Z]\w*)/,
+				inside: {
+					'delimiter': {
+						pattern: /^\$\(?|\)$/,
+							alias: 'punctuation'
+				},
```
RunDevelopment: Please correct the indentation here.

components/prism-vala.js
```diff
@@ -0,0 +1,68 @@
...
+		}
+	],
+	'constant': /\b[A-Z0-9_]+\b/,
+	'keyword': /\b(?:bool|char|double|float|null|size_t|ssize_t|string|unichar|void|int|int8|int16|int32|int64|long|short|uchar|uint|uint8|uint16|uint32|uint64|ulong|ushort|class|delegate|enum|errordomain|interface|namespace|struct|break|continue|do|for|foreach|return|while|else|if|switch|assert|case|default|abstract|const|dynamic|ensures|extern|inline|internal|override|private|protected|public|requires|signal|static|virtual|volatile|weak|async|owned|unowned|try|catch|finally|throw|as|base|construct|delete|get|in|is|lock|new|out|params|ref|sizeof|set|this|throws|typeof|using|value|var|yield)\b/i,
+	'operator': /\+\+|--|&&|\|\||<<=?|>>=?|~|[+\-*\/%&^|=!<>]=?|\?\??/
+});
+
+Prism.languages.insertBefore('vala','string', {
+	'raw-string': {
+		pattern: /(""")[\s\S]*?\1/,
```
RunDevelopment: Why not `/"""[\s\S]*?"""/`?

components/prism-vala.js
```diff
@@ -0,0 +1,68 @@
...
+			pattern: /((?:\b(?:class|interface|new)\s+)|(?:catch\s+\())[A-Z]\w*(?:\.\w+)*\b/,
+			lookbehind: true,
+			inside: {
+				punctuation: /\./
+			}
+		}
+	],
+	'constant': /\b[A-Z0-9_]+\b/,
+	'keyword': /\b(?:bool|char|double|float|null|size_t|ssize_t|string|unichar|void|int|int8|int16|int32|int64|long|short|uchar|uint|uint8|uint16|uint32|uint64|ulong|ushort|class|delegate|enum|errordomain|interface|namespace|struct|break|continue|do|for|foreach|return|while|else|if|switch|assert|case|default|abstract|const|dynamic|ensures|extern|inline|internal|override|private|protected|public|requires|signal|static|virtual|volatile|weak|async|owned|unowned|try|catch|finally|throw|as|base|construct|delete|get|in|is|lock|new|out|params|ref|sizeof|set|this|throws|typeof|using|value|var|yield)\b/i,
+	'operator': /\+\+|--|&&|\|\||<<=?|>>=?|~|[+\-*\/%&^|=!<>]=?|\?\??/
```
RunDevelopment: 1. `=>` for lambdas is missing.
2. `...` should also be an operator.

components/prism-vala.js
```diff
@@ -0,0 +1,68 @@
+Prism.languages.vala = Prism.languages.extend('clike', {
+	// Classes copied from prism-csharp
+	'class-name': [
+		{
+			// (Foo bar, Bar baz)
+			pattern: /\b[A-Z]\w*(?:\.\w+)*\b(?=\s+\w+)/,
```
RunDevelopment: It's quite common for types to be nullable. E.g.: `Color? c = getColor();`
Also pointers? (`Foo* f_ptr = f;`)

components/prism-vala.js
```diff
@@ -0,0 +1,68 @@
...
+				},
+					rest: Prism.languages.vala
+				}
+			}
+		}
+	}
+});
+
+Prism.languages.insertBefore('vala', 'keyword', {
+	'regex': /\/(\[(?:[^\]\\\r\n]|\\.)*]|\\.|[^/\\\[\r\n])+\/[gimyu]{0,5}(?=\s*($|[\r\n,.;})\]]))/
```
RunDevelopment: Wrong [flags](https://wiki.gnome.org/Projects/Vala/Tutorial#Regular_Expression_Literals).

components/prism-vala.js
TemplarVolk: I copied the classes from the C# component, so I'm guessing pointers should be added there as well.

RunDevelopment: That's not necessary. Pointers have a few constraints in C#, so that shouldn't be necessary (as far as I know).

components/prism-vala.js
```diff
@@ -0,0 +1,71 @@
...
+			inside: {
+				punctuation: /\./
+			}
+		}
+	],
+	'constant': /\b[A-Z0-9_]+\b/,
+	'function': /\w+(?=\s*\()/,
+	'keyword': /\b(?:bool|char|double|float|null|size_t|ssize_t|string|unichar|void|int|int8|int16|int32|int64|long|short|uchar|uint|uint8|uint16|uint32|uint64|ulong|ushort|class|delegate|enum|errordomain|interface|namespace|struct|break|continue|do|for|foreach|return|while|else|if|switch|assert|case|default|abstract|const|dynamic|ensures|extern|inline|internal|override|private|protected|public|requires|signal|static|virtual|volatile|weak|async|owned|unowned|try|catch|finally|throw|as|base|construct|delete|get|in|is|lock|new|out|params|ref|sizeof|set|this|throws|typeof|using|value|var|yield)\b/i,
+	'number': /(?:\b0x[\da-f]+\b|(?:\b\d+\.?\d*|\B\.\d+)(?:e[+-]?\d+)?)(?:f|u?l?)?/i,
+	'operator': /\+\+|--|&&|\|\||<<?=?|(?<!-)>>?=?|=>|~|-(?!>)=?|[+\*\/%&^|=!]=?|\?\??|\.\.\./,
```
RunDevelopment: You cannot use `(?<!-)` which is a lookbehind, which is not part of the ECMAScript standard. (It does work in few browsers and Node, but we generally can't use it for the whole web. Maybe in future.)
(This is why the tests fail for older Node versions.)

TemplarVolk: Oh right, sorry. How should i go about doing it then.

The `->` is being parsed as `-` and `>` operators, but if i use punctuations over operators the `...` will be parsed as 3 `.` punctuations

RunDevelopment: Then let's make this really simple for ourselves and just make `->` an operator. 
(C# also does this. My bad...)

components/prism-vala.js
```diff
@@ -0,0 +1,71 @@
...
+					rest: Prism.languages.vala
+				}
+			},
+			'string': /[\s\S]+/
+		}
+	}
+});
+
+Prism.languages.insertBefore('vala', 'keyword', {
+	'regex': /\/(\[(?:[^\]\\\r\n]|\\.)*]|\\.|[^/\\\[\r\n])+\/[imsx]{0,4}(?=\s*($|[\r\n,.;})\]]))/
```
RunDevelopment: Shouldn't `regex` be greedy?
E.g. `/"foo"/i`

RunDevelopment:
Please re-run `gulp` and then I will gladly merge this.

PR #1660, opened 2018-12-11
Markup: Added support for CSS and JS inside of CDATAs

### LINKED ISSUE #1656: Fix case when <script> or <style> contains a <!⁠[CDATA[…]]> section
When `<script>` or `<style>` contains a `<!⁠[CDATA[…]]>` section then code inside the section isn’t properly highlighted:
```
<script type="application/javascript"><![CDATA[
let stuff;
alert("</script>");
]]></script>

<style type="text/css"><![CDATA[
.stuff {
	…
}
]]></style>
```

## See for a live example:
- https://developer.mozilla.org/docs/User:ExE-Boss/Prism_issues#CDATA_in_script_or_style

RunDevelopment:
This PR adds resolves #1656.

image


RunDevelopment:
I should have asked this a lot sooner but is that even valid (X)HTML?

It looks valid but I wasn't able to get any browser to parse a `CDATA` in a `script` element (in HTML) correctly when containing a `"</script>"` string. I think HTML ignores `CDATA`s because `script` and `style` tags are parsed like CDATA content anyway (without the closing `]]>` of course).
I wasn't even able to get XHTML to recognize CDATA at all.
(All findings for Chrome, FF, IE)

So, what is this change for?
Because if it's for XML, then you won't like #1603.

ExE-Boss:
It’s for XHTML, and from my testing (in&nbsp;Firefox) XHTML supports CDATA.

Also, are you sure you weren’t accidentally loading the XHTML document as HTML?

RunDevelopment:
> Are you sure you weren’t accidentally loading the XHTML document as HTML?

Pretty sure. I got an error message that the `<!CDATA[[` inside `script` cannot be parsed which I don't get in HTML.
But maybe I'm doing something wrong. Could you please provide an example of a valid XHTML page with `CDATA`s?

RunDevelopment:
Any number of `CDATA` can now be included in `<style>` and `<script>` tags.

---

The whole job is done by `Prism.languages.markup.tag.addInlined` (rename this?) which adds support for languages which are contained by special tags such as `<style>`.

(I actually wanted to add this function to `markup` instead of `markup.tag` but that means that `addInlined` will be lost after `insertBefore`.)

ExE-Boss:
This needs rebasing.

(also, I&nbsp;thought this was merged already)

ExE-Boss:
review?(@mAAdhaTTah)

RunDevelopment:
Sorry @ExE-Boss!
I completely forgot about this. I'll make it ready tomorrow.

Issue #1539, opened 2018-08-23
Resources


RunDevelopment:
I would like to propose the concept for resources to Prism.

## Concept

Resources are small JS files which contain one feature and can be required by plugins, languages, and other resources. They contain functions or constants and do not add new languages or plugin functionalities themselves.

Resources are supposed to encourage code reuse between different components and the addition of new features (even to Prism-core). Like this, even features which are only used by a small number of components and therefore disqualified from being added directly to Prism-core can be shared between said components.

## Examples

One example of a resource we have right now is the `markup-templating` "language". It does not add any grammar to Prism and therefore should disqualify as a language.

The generated objects from gulp used by the AutoLoader and ShowLanguage plugin could be outsourced as resources in the case that other plugins need them as well.

Future resources could be shared patterns (#1494) and other functionality like #1476 or #1524.

## Implementation

Resource should be located under `./resources` and similar to other components have names of the form `prism-{name}.js`. Minified files will be generated by gulp.

They will be registered inside the new `resources` object in `components.json`.

This also means that we will have to change the current dependency system to allow for dependencies between different resource kinds (`languages`, `plugins`, `resources`).

RunDevelopment:
One note on the required dependency system:

I'm currently working on fixing #1490 and [mAAdhaTTah/babel-plugin-prismjs#3](https://github.com/mAAdhaTTah/babel-plugin-prismjs/issues/3). To do that, I wrote a new implementation of widely used dependency-related methods, which are used by the website, `loadLanguages` and our testing suite.

This means that it will be very easy to implement such a feature.

PR #1664, opened 2018-12-18
Generate nested patterns

RunDevelopment:
This PR adds the `Prism.patterns.nested` function which creates patterns to match recursively nested structures.

This PR is intended to go together with #1538.

#1444 will profit from this.

---

I really want to make this a resource (#1539) so that it doesn't add weight to core.


RunDevelopment:
I'm going to close this now because there are better ways to implement methods for generating nested patterns.

Issue #1667, opened 2018-12-21
Unproper alignment of line numbers


eakarpov:
Hi! Why when I use a plugin for "line-numbers", in my code block numbers are not aligned as left side relatively with my code in it.

![image](https://user-images.githubusercontent.com/16017789/50331565-229f6d80-0510-11e9-878f-93961611db76.png)

RunDevelopment:
Could you please add a description on how to reproduce the bug?
A link to the website or JSFiddle showcasing the bug would be even better.

eakarpov:
> Could you please add a description on how to reproduce the bug?
> A link to the website or JSFiddle showcasing the bug would be even better.

I can't properly tune Prism via CDN, so...: https://jsfiddle.net/65ernfb7/
I use the same code in my problem place:
```html
 <pre class="command-line line-numbers" ref="pre">
    <code class="language-cpp" ref="code">
       {{code}}
    </code>
</pre>
```

RunDevelopment:
@mAAdhaTTah The problem of spaces between `<pre>` and `<code>` has also caused some problems in the past, I believe. Why not do a kind of preprocessing step to correct that automatically?

eakarpov:
> @eakarpov The issue is the indentation. You would need to write it as such:
> 
> ```
>  <pre class="command-line line-numbers" ref="pre"><code class="language-cpp" ref="code">{{code}}</code></pre>
> ```

Hmm.... This fixed the issue... But why is it important? It is rather strange to have necessity of single-lined code-block.

RunDevelopment:
It's important because of the way the line number plugin is currently implemented:
The line numbers list you see starts at the `<code>` element and goes down from there as a block. This means that any initial offset (e.g. the indentation of `<code>`) will move the whole line numbers list to the right.

mAAdhaTTah:
@RunDevelopment That's what [normalize whitespace](https://prismjs.com/plugins/normalize-whitespace/) does.

@eakarpov Sounds like your issue is resolved, so I'm going to close this, but feel free to ask any follow up questions if you need help.

Issue #1604, opened 2018-11-05
line numbers overlap with code


amikrop:
When using the line-numbers plugin, the line numbers (apart from the first line) overlap with the code, as shown in the screenshot. I also include some of the HTML source.
[code.txt](https://github.com/PrismJS/prism/files/2550421/code.txt)
![screenshot from 2018-11-05 21-49-40](https://user-images.githubusercontent.com/24046/48023866-a6132400-e147-11e8-991d-9b4ec69c8cc9.png)

RunDevelopment:
This bug seems to be cause by the spaces between the opening `pre` and `code` tags.

```html
<pre class="line-numbers">
    <code class="language-python">from __future__ import unicode_literals
```

Prism does not expect that there are any characters between the `pre` and `code` tags.
Simply removing these spaces will solve the problem.

---

One small correction: The first line also overlaps with the list of line numbers; it's just that the first line is also indented with 4 spaces, so it looks right but really isn't.

amikrop:
Thanks!

PR #1668, opened 2018-12-21
Trim spaces inside highlighted <pre> tags

RunDevelopment:
The problem of spaces between `<pre>` and `<code>` has caused some problems (#1667, #1604).

This PR removes any leading and trailing spaces in the parent `<pre>` tag in the sense of `pre.innerHTML = pre.innerHTML.trim()`.

This change does not affect the code inside the `<code>` tag but it will change the behavior of this:

```html
<pre>
    <code>var a = 0;</code>
</pre>
```

This spaces before and after the `code` element will be removed meaning that the code will not be strangely indented (the problem) and that there will be no additional newline at the end (good?).


RunDevelopment:
> the solution is in formatting the HTML properly, not magically transforming the HTML for them

That's true.
Then I'll just leave the magic to Normalize Whitespace as you [mentioned](https://github.com/PrismJS/prism/issues/1667#issuecomment-449400029). 😉

Issue #1623, opened 2018-11-21
Increase token granularity for improved syntax highlighting


atomiks:
It would be cool to add more tokens so different parts of the code can be highlighted differently. Prism seems more granular than CodeMirror, but it's still missing some nice tokens compared to Atom and VS Code.

For JavaScript, it would be nice to have these:

- `token parameter`

```js
function add(a, b) { // <-- `a` and `b` receive `parameter`
  return a + b;
}
```

- `token property definition` and `token property access`

```js
const obj = {
  example: true // <-- `example` receives `property` and `definition`
};
obj.example; // <-- `example` receives `property` and `access`
```

- `token function method`

```js
this.func(); // `func` receives `method` if being called on an object
```

- `token keyword module`

All keywords here should receive the `module` token.

```js
import { thing as thingy } from 'thing';
export default thingy;
```

- `token keyword special`

"special"(?) keywords like `this`

- `token variable dom`

```js
window
document
navigator
performance
localStorage
```

- `token console`

```js
console.log('hello'); // <-- `console` receives this token
```

- `token operator spread`

```js
const copy = [...arr]; // <- `...` receives this token
```

RunDevelopment:
If you want more fine-grained control about the way keywords are highlighted, I recommend the [Highlight keywords](https://prismjs.com/plugins/highlight-keywords/) plugin.

Also, `token parameter` is kind of covered by #1446.

atomiks:
Thanks, I didn't see that plugin.

And the PR looks good. I'll create my own language extension and copy some of that code while the PR waits to be merged.

Do you have ideas for the regex necessary for the other tokens listed in the OP? I'm really not great with regex and can't figure it out on my own. 😭

RunDevelopment:
Insert this code into `prism-javascript.js` in line 37 and you will have most of the features.

But let me warn you: 
1. I did this quick and dirty, so it's not perfect and no guarantees. 
2. It really doesn't look nice IMO, you will probably have to adjust the default theme you're using.

Keep in mind that if you want to use the minified file, you will have to rebuild Prism (refer to `README.md`).

---

```js
Prism.languages.insertBefore('javascript', 'punctuation', {
	'definition': {
		pattern: /([{,]\s*)[a-z]\w*(?=\s*:)/i,
		lookbehind: true,
		alias: 'property'
	},
	'access': {
		pattern: /(\.\s*)[a-z]\w*/i,
		lookbehind: true,
		alias: 'property'
	},
[...]

PR #1671, opened 2018-12-22
CSS: Highlight attribute selector

RunDevelopment:
This PR adds additional highlighting for attribute selectors.

Spec: [CSS 3](https://drafts.csswg.org/selectors-3/#attribute-selectors), [CSS 4](https://drafts.csswg.org/selectors-4/#attribute-selectors)

### Example

image


RunDevelopment:
Are you ok with this @LeaVerou ? I'm asking because of your comment [here](https://github.com/PrismJS/prism/pull/375#issuecomment-57237888).
(Same for #1650)

LeaVerou:
If the others are ok with it!

RunDevelopment:
Since [people seem to be interested](https://github.com/PrismJS/prism/issues/1623#issuecomment-463428551) in this feature, I'm going to merge this if @mAAdhaTTah gives his OK.

mAAdhaTTah:
LGTM!

atomiks:
How do I use CSS extras? Is it documented somewhere?

RunDevelopment:
You load both languages and CSS Extras will add new features to CSS, so you just use the normal `language-css` class.

atomiks:
I'm using a plugin for this which indirectly loads them for me, is the name `language-css-extras`?

RunDevelopment:
Are you using the [Autoloader](https://prismjs.com/plugins/autoloader/)?

atomiks:
It's a gatsby plugin, here's the source: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-prismjs/src/load-prism-language.js

RunDevelopment:
That seems to be a problem with the library. Seems like they don't support [loading language extensions](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-prismjs/src/highlight-code.js).
It would probably be best to open an issue over there.

PR #1682, opened 2018-12-30
Regex language

RunDevelopment:
This PR adds a Regex language based on the JavaScript Regex syntax.
Edit: It now supports most of the regex syntax of C# and Java.

This is based on #1468 with a few improvements.

Also, I'm thinking about editing the themes to show the results.


shreyasminocha:
Thank you so much for this. I believe Prism is the only mainstream syntax highlighter with regex highlighting.

> Also, I'm thinking about editing the themes to show the results.

I'm assuming you didn't get around to doing this yet. Meanwhile, I'm using [a stylesheet I wrote](https://gist.github.com/shreyasminocha/a2df081273ada0fec8b3d8faa6941e7b) with colours lifted from [RegExr](https://regexr.com) for anyone looking for that.

RunDevelopment:
> I'm assuming you didn't get around to doing this yet.

Thanks for reminding me. I completely forgot about this.
However, I won't change the themes for this as there are [quite a few](https://github.com/PrismJS/prism-themes) to change. I'll add a few aliases to given them reasonable highlighting which hopefully will look reasonably good in all themes.

PR #1683, opened 2018-12-30
Show language improvements

RunDevelopment:
This PR fixes the bug that aliases without a title (e.g. `js` for JavaScript) were not included in the language map.

It also improves the guessing of a language's title by capitalizing 'script' to correctly guess ActionScript, AppleScript, CoffeeScript, JavaScript, LiveScript, and TypeScript.


gulpfile.js
```diff
@@ -76,19 +76,49 @@ gulp.task('languages-plugins', function (cb) {
...
-				var ucfirst = p.substring(0, 1).toUpperCase() + p.substring(1);
-				if (title !== ucfirst) {
-					languagesMap[p] = title;
-				}
+
+				addLanguageTitle(p, title);
 
 				for (var name in data.languages[p].aliasTitles) {
-					languagesMap[name] = data.languages[p].aliasTitles[name];
+					addLanguageTitle(name, data.languages[p].aliasTitles[name]);
```
mAAdhaTTah: I don't _think_ we just want to run this through `addLanguageTitle`, as we do want to add the name even if the guessed title doesn't match the title.

mAAdhaTTah: Oh wait, this is the gulpfile, not show-language,

RunDevelopment:
This also includes a different version of `java.min.js`. I don't know why uglify chooses different variables names, but those are the ones you get with a fresh installation, so it should be alright, I think.

Issue #1070, opened 2016-12-05
Ditch gh-pages?


LeaVerou:
Now that Github supports running Github Pages from master, is there still any reason to have two branches?

Golmote:
Probably not. This situation has been confusing for some users in the past, and surely still is.

LeaVerou:
Is there any info in master that we should retain or is it sufficient to delete the master branch and rename gh-pages to master?

Golmote:
I just compared the two branches. There does not seem to be any info in master that is not already in gh-pages (we just did some weird reverting once after an accidental merge of a PR directly on master).

So I think your idea is the way to go ;)

LeaVerou:
Awesome thanks, will do it in a few days (currently rushing to meet a deadline)!

mAAdhaTTah:
@Golmote Can we go ahead with this?

Golmote:
@mAAdhaTTah Having never used Github pages, I must admit I'm not confident at all doing this. I was hoping @LeaVerou would handle it.

Also, I think some people used to reference the `gh-pages` branch directly. See Bower-related issues https://github.com/PrismJS/prism/issues/137#issuecomment-21528734, https://github.com/PrismJS/prism/issues/165#issuecomment-47699782, https://github.com/PrismJS/prism/issues/180#issuecomment-28613947, https://github.com/PrismJS/prism/issues/231#issuecomment-48849176, https://github.com/PrismJS/prism/issues/257#issue-34122668, https://github.com/PrismJS/prism/issues/497#issuecomment-83185647, and the more recent one using npm https://github.com/PrismJS/prism/issues/966#issuecomment-379167653. See also this Github search: https://github.com/search?q=%22prism.git%23gh-pages%22&type=Code.
So... will this be considered a breaking change? I'm a bit worried after v1.14's mess.

Finally, we do need to remember we are currently referencing the `gh-pages` branch in our code too, at least in [`download.js`](https://github.com/PrismJS/prism/blob/gh-pages/download.js#L13) and in [`examples.js`](https://github.com/PrismJS/prism/blob/gh-pages/examples.js#L9). We'll need to update all references when making the change.

mAAdhaTTah:
Fair enough. I don't really feel strongly that we need to change it. We're working with `gh-pages` as a "development" branch, with `master` matching what's on `npm`, so we can just leave it and go with that approach.

Golmote:
[...]

PR #1684, opened 2018-12-30
GitHub API calls now use the master branch

### LINKED ISSUE #1624: Download page: No filesize for GML
The list item of the GameMakerLangauge does not display its file size when loading the page. 
The file size will only the displayed after GML was selected.

image

RunDevelopment:
This resolves #1624.

It fixes the missing examples for newly added language in the [example page](https://prismjs.com/examples.html).

### DIFF
--- a/download.js
+++ b/download.js
@@ -10,7 +10,7 @@ var minified = true;
 
 var dependencies = {};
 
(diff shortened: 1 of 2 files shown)


RunDevelopment:
This might be interesting for #1070.

mAAdhaTTah:
The open question in #1070 is just what to do with the open PRs, so this doesn't resolve it but definitely fixes a lingering issue caused by it.

PR #1478, opened 2018-07-13
Add file extensions support for major languages

ybiquitous:
This adds aliases for file extensions of some major languages.

- `.coffee` -> `.coffeescript`
- `.md` -> `.markdown`
- `.py` -> `.python`
- `.rb` -> `.ruby`
- `.yml` -> `.yaml`

Such as:

https://github.com/PrismJS/prism/blob/7d084d5f17fa40d3451d46cac605308382c9c499/components/prism-javascript.js#L56

PR #1685, opened 2019-01-01
Add extension shortcut for Asciidoc

oncletom:
In the continuity of #1478.


RunDevelopment:
Thank you for contributing!

Please also add this alias to the `asciidoc` entry in `components.json` and run the `npx `[`gulp`](https://github.com/gulpjs/gulp) command to generate the minified files.

PR #1688, opened 2019-01-02
Improve recognition of constants in PHP

### LINKED ISSUE #1687: Recognition of constants in PHP
According to <https://secure.php.net/manual/en/language.constants.php>, a valid constant name in PHP can be expressed by the following regular expression:

```php
[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*
```

Looks like Prism violates the rule. It doesn’t recognize constants with names containing lower­‑case Latin letters and/⁠or consisting of only one underscore or Latin letter.

#### Input

```php
_;
X;
```

#### Expected output

```html
<span class="token constant">_</span><span class="token punctuation">;</span>
<span class="token constant">X</span><span class="token punctuation">;</span>
```

#### Actual output

```html
_<span class="token punctuation">;</span>
X<span class="token punctuation">;</span>
```

#### Input

```php
MyConstant;
```

#### Expected output

```html
<span class="token constant">MyConstant</span><span class="token punctuation">;</span>
```

#### Actual output

```html
MyConstant<span class="token punctuation">;</span>
```

volado:
Resolves #1687.


RunDevelopment:
I don't think it's a good idea to say that any word which isn't highlighted as anything else should be a constant.

Let's say there is a feature of PHP which Prism doesn't support (yet). All of the keywords/names related to this feature will be highlighted as constants.
Today, these features are [type declarations](http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration), [group use declarations](http://php.net/manual/en/language.namespaces.importing.php#language.namespaces.importing.group), [return types](http://php.net/manual/en/functions.returning-values.php#functions.returning-values.type-declaration), and [type casting](http://php.net/manual/de/language.types.type-juggling.php#language.types.typecasting) just to name a few. These are only partially supported or not at all.

I encourage you to play around with the examples on the PHP.net page and see for yourself what would be highlighted as constants.

But apart from that, isn't it a naming convention to not use lowercase characters for constants?
Even [the page you linked](https://secure.php.net/manual/en/language.constants.php) to follows it.

volado:
> All of the keywords/names related to this feature will be highlighted as constants.

That’s the way PHP behaves in some circumstances, such as `echo X`. However, I shall agree that there are other contexts, such as `function f($x): type {}`.

Prism definitely shouldn’t mark unsupported elements as constants.

OK, let’s stick with [PSR‑1](https://www.php-fig.org/psr/psr-1/#41-constants), not language itself, i. e., underscore as a word separator only, no lower­‑case letters. Should we have single­‑letter constants (`X`), then? I think we should. It doesn’t affect type declarations, group use declarations, etc.

RunDevelopment:
> Should we have single­‑letter constants (`X`), then?

Sure!

RunDevelopment:
Thank you for contributing!

volado:
Thank you!

But it’s sad that the PR was merged **before** Travis CI has finished building :(

RunDevelopment:
Don't worry about that.
When you look at the detail, you'll see that _all tests passed_ on 2/4 Node version and the other 2 couldn't find this PR (probably because it was merged already).

volado:
OK, thank you!

PR #1690, opened 2019-01-03
Update the list of PHP keywords

### LINKED ISSUE #1689: Update the list of PHP keywords
Here are some suggestions:

- add `__halt_compiler`, `callable`, `clone`, `empty`, `eval`, `exit`, `insteadof`, `isset`, `list`, `unset`. See <https://secure.php.net/manual/en/reserved.keywords.php>;
- remove `cfunction` **or** add `old_function`. Both are defined in PHP 4 only. See <https://web.archive.org/web/20090122062341/http://php.net/manual/en/reserved.keywords.php>.

Presently, `empty()`, `eval()`, `exit()`, `isset()`, `list()`, `unset()` are recognized as functions because of braces. But `__halt_compiler()`, `array()`, `die()`, `empty()`, `eval()`, `exit()`, `isset()`, `list()`, `unset()` are not functions: they are keywords.

Depending on your decision, I can make a PR with the appropriate RegExp. An example:

```javascript
/\b(?:__halt_compiler|abstract|and|array|as|break|callable|case|catch|cfunction|class|clone|const|continue|declare|default|die|do|echo|else|elseif|empty|enddeclare|endfor|endforeach|endif|endswitch|endwhile|eval|exit|extends|final|finally|for|foreach|function|global|goto|if|implements|include|include_once|instanceof|insteadof|interface|isset|list|namespace|new|null|old_function|or|parent|print|private|protected|public|require|require_once|return|static|switch|throw|trait|try|unset|use|var|while|xor|yield)\b/i
```

Is it alright to arrange keywords in alphabetical order? Is there a meaning in the current order? Is it connected with optimization?

I also have suggestions regarding `null` and `parent`, but it’s better to start a separate topic for that.

volado:
Resolves #1689.


volado:
You’re welcome! Thank you for your work!

Issue #1691, opened 2019-01-03
Operators should not have a background colour


lippirk:
This relates somewhat to #1347, but in that issue the raiser had manually changed the background. I have simply referenced the official prism CSS with the default theme, and am getting incorrect background colours.

Html: 
```
<span class="token operator">==</span>
```

Expected:
![image](https://user-images.githubusercontent.com/24416311/50661789-1c8f9f00-0f9c-11e9-91b9-eaba124895b7.png)

Actual:
![image](https://user-images.githubusercontent.com/24416311/50661824-45b02f80-0f9c-11e9-925a-9f8cde2d0e4e.png)

Removing the background css property from `.token.operator` fixes this

RunDevelopment:
Sorry, but I'll give [the same answer](https://github.com/PrismJS/prism/issues/1347#issuecomment-372159127) as @Golmote. We won't change the themes.

That aside:
Add a `<style>` tag anywhere **after** the `<link>` you load `prism.css` with and insert the following rule:

```css
.token.operator {
    background: none;
}
```

That should solve the issue.

Alternatively, the rule can also be included in another `.css` file, it just has to be loaded **after** `prism.css`.

Another alternative is use the rule:

```css
.token.operator {
    background: none !important;
}
```

I don't like `!important` but like this, load order doesn't matter.

lippirk:
Okay no problem - thanks for having a look!

PR #1692, opened 2019-01-03
Remove background property from .token.operator

lippirk:
Suggested fix for issue #1691


lippirk:
Maintainers do not want to change the theme

PR #1699, opened 2019-01-11
Added T4 Text Templates

RunDevelopment:
This PR adds the [T4](https://en.wikipedia.org/wiki/Text_Template_Transformation_Toolkit) [Text Templates](https://docs.microsoft.com/en-us/visualstudio/modeling/design-time-code-generation-by-using-t4-text-templates?view=vs-2017) language to Prism.

### Examples

image
[Source](https://github.com/mono/t4/blob/master/dotnet-t4/test.tt)

image


components.json
```diff
@@ -764,6 +764,16 @@
...
+		"t4": {
+			"title": "T4 Text Templates",
+			"require": ["csharp", "visual-basic"],
+			"alias": ["t4-cs", "t4-vb"],
+			"aliasTitles": {
+				"t4-cs": "T4 C#",
+				"t4-vb": "T4 Visual Basic"
+			},
+			"owner": "RunDevelopment"
+		},
```
mAAdhaTTah: The way the code is set up, I'm not sure these aliases are correct; only `t4-cs` is an alias. The reason this might be an issue is at least the babel plugin is going to have to implement [aliases](https://github.com/mAAdhaTTah/babel-plugin-prismjs/issues/4), which is going to have to search the aliases and replace them with the source, so in this case, searching for t4-cs would load t4, which is not what is really wanted.

Given the relationship to the languages they're extending from, I'm not really sure what the best solution is here.

RunDevelopment: One solution would be to split it into 3 different languages each with their own files: `t4-cs`, `t4-vb`, and `t4-template` (for creating the T4 variants).

Or we could fix the Babel plugin. In this case, it would be good to also adjust our tests to be able to deal with aliases. I'd know if I can do that though.

mAAdhaTTah: I haven't implemented the alias usage in the babel plugin, hence the linked issue, but given the meaning of "alias", I'm not sure it's fixable because the meaning "alias" in this case is a bit different from the others. We could hard-code an exception, but I'd rather not do that, tbh.

If there's no way to derive a single T4 language, I _kinda_ think 3 separate T4 languages might be best? But I'm open to other solutions.

@Golmote @LeaVerou Thoughts?

RunDevelopment: _Note to self:_ Depending on how this goes out, I will have to change #1603.

RunDevelopment: I split T4 into 3 different languages as suggested.

components/prism-t4-vb.min.js
```diff
@@ -0,0 +1 @@
+Prism.languages["t4-vb"]=Prism.languages["t4-templating"].createT4("visual-basic");
```
RunDevelopment: Ladies and Gentleman,
I present to you: __The smallest languages definition!__

PR #1704, opened 2019-01-13
Fixed#1696 - Swapped ABAP and Erland commits

dellagustin:
At changlog entry 1.8.4 the commits for ABAP and Erlang were
swapped.


RunDevelopment:
Thank you for contributing!

Issue #1706, opened 2019-01-14
bug(plugin): command-line data-filter-out not working


muuvmuuv:
I'm trying to get this to work:

```html
<pre class="command-line language-bash" data-filter-output="(out)" data-user="user" data-host="remotehost">
  <code>
  pwd
  (out)/usr/home/user/bin
  ls -la
  (out)total 2
  (out)drwxr-xr-x   2 user  group     11 Jan 10 16:48 .
  (out)drwxr--r-x  45 user  group     92 Feb 14 11:10 ..
  (out)-rwxr-xr-x   1 user  group    444 Aug 25  2013 backup
  (out)-rwxr-xr-x   1 user  group    642 Jan 17 14:42 deploy
  </code>
</pre>
```

I'm loading the core and plugins like this:

```js
import 'prismjs/prism.js';
import 'prismjs/plugins/unescaped-markup/prism-unescaped-markup.min.js';
import 'prismjs/plugins/normalize-whitespace/prism-normalize-whitespace.min.js';
import 'prismjs/plugins/autolinker/prism-autolinker.min.js';
import 'prismjs/plugins/command-line/prism-command-line.min.js'; // BUG: output not rendered correctly
import 'prismjs/plugins/line-numbers/prism-line-numbers.min.js';
```

and prism is rendering it to this:

![bildschirmfoto 2019-01-14 um 10 56 42](https://user-images.githubusercontent.com/11534760/51106134-2fe80900-17eb-11e9-800d-11b1192b3552.png)
![bildschirmfoto 2019-01-14 um 10 56 16](https://user-images.githubusercontent.com/11534760/51106135-2fe80900-17eb-11e9-8f32-ebde5e633eda.png)

I have tried disabling all plugins but still not the result I expected.

RunDevelopment:
The leading spaces are the problem (I think). Because of them, you don't have any lines with the prefix `(out)` but a few with the prefix `[ ][ ](out)` (`[ ]` for one space).

Also, please don't indent your code. For `pre` elements, every (leading) space counts. They will not be removed as in other tags.

muuvmuuv:
I have created a fiddle: https://jsfiddle.net/muuvmuuv/tckqsa1j/

muuvmuuv:
I have found the issue, I had not imported the required languages... 😆 sorry, so obvious

RunDevelopment:
It should work even without the grammar being imported, so I still think that this needs fixing.

[...]

PR #1709, opened 2019-01-15
The `insert-before` hook is now called properly

RunDevelopment:
This PR fixes the bug described in #1706 (also includes a demo).

The problem was that the `insert-before` hook was not called if `grammar` is undefined.
This also fixes that `callback` was not called in that case and that `env` of the `complete` hook did not have a `highlightedCode` property.

All other hooks are unchanged. The `insert-before` hook is only used by the Command line plugin, so there should be no other side effects.


components/prism-core.js
```diff
@@ -212,33 +212,37 @@ var _ = _self.Prism = {
...
-			}
+		if (!env.code) {
 			_.hooks.run('complete', env);
 			return;
 		}
 
 		_.hooks.run('before-highlight', env);
 
+		if (!env.grammar) {
+			insertHighlightedCode(_.util.encode(env.code));
```
mAAdhaTTah: Why is `util.encode` used here?

mAAdhaTTah: Ah, and this was handled previous here:

```js
env.element.textContent = env.code;
```

which uses `textContent` rather than `innerHTML`.

PR #1710, opened 2019-01-15
Added HDL (Hardware Description Language) support

stephanmax:
I added support for the declarative Hardware Description Language used in the Coursera Nand2Tetris (https://phoenix.goucher.edu/~kelliher/f2015/cs220/hdlSurvivalGuide.pdf).


components/prism-hdl.js
```diff
@@ -0,0 +1,18 @@
+Prism.languages.hdl = {
+	'comment': [
+		{
+			pattern: /(^|[^\\])\/\*[\s\S]*?(?:\*\/|$)/,
```
RunDevelopment: You probaböy borrowed this from C-like, but this language isn't as complex as C, C++, etc., so we can simplify and combine the two comment patterns to this:

```js
/\/\/.*|\/\*[\s\S]*?(?:\*\/|$)/
```

No `lookbehind` or `greedy` required.

components/prism-hdl.js
```diff
@@ -0,0 +1,18 @@
...
+		},
+		{
+			pattern: /(^|[^\\:])\/\/.*/,
+			lookbehind: true,
+			greedy: true
+		}
+	],
+	'keyword': /\b(?:CHIP|IN|OUT|PARTS|BUILTIN|CLOCKED)\b/,
+	'boolean': /\b(?:true|false)\b/,
+	'number': /\b\d+\b/i,
```
RunDevelopment: You don't need the `i` flag.

tests/languages/hdl/boolean_feature.test
```diff
@@ -0,0 +1,15 @@
+true
+false
+trueVal
```
RunDevelopment: I don't think you have to test for negatives. (In keywords and numbers as well.)

stephanmax:
Thanks for your feedback, @RunDevelopment. I applied all changes as requested. PS. Say Hi to my hometown Leipzig 😃

components/prism-nand2tetris-hdl.js
```diff
@@ -0,0 +1,8 @@
+Prism.languages['nand2tetris-hdl'] = {
+	'comment': /\/\/.*|\/\*[\s\S]*?(?:\*\/|$)/,
+	'keyword': /\b(?:CHIP|IN|OUT|PARTS|BUILTIN|CLOCKED)\b/,
+	'boolean': /\b(?:true|false)\b/,
+	'number': /\b\d+\b/,
+	'operator': /=/,
```
RunDevelopment: There also seems to be a [range operator `..`](https://github.com/pmenon/nand-2-tetris/blob/master/projects/05/Memory.hdl#L35) (I guess).

RunDevelopment:
Nice work.

Wouldn't it be possible to use the C-like function here? I don't know if stuff like `Not`, `And`, and so on are called functions here but they do look like it.

Also, is there a spec for this language?

components/prism-nand2tetris-hdl.js
stephanmax: True, I accidentally put the single dot as punctuation, but it is rather just a double dot range operator here. `'operator': /=|\.\./` should do the trick, right?

RunDevelopment: Yes, that should work.

stephanmax:
The semantics behind `And` and `Not` is actually a bit different than the notion of functions. They have to be HDL scripts in the same folder with the same name. Actually, they could be called `Bananarama` as long as there is a `Bananarama.hdl` file next to it.

There is only an informal spec in the appendix of the course‘s accompanying book.

RunDevelopment:
> There is only an informal spec in the appendix of the course‘s accompanying book.

Well, can't do anything about that.

> The semantics behind `And` and `Not` is actually a bit different than the notion of functions. They have to be HDL scripts in the same folder with the same name. Actually, they could be called `Bananarama` as long as there is a `Bananarama.hdl` file next to it.

So the function name has to relate to a file name (excluding builtins like `And` I suppose).
I think this pretty much falls within the concept of a function so can we land this feature?

examples/prism-nand2tetris-hdl.html
```diff
@@ -0,0 +1,23 @@
...
+false</code></pre>
+
+<h2>Full example</h2>
+<pre><code>/*
+ * Checks if two input bits are equal
+ */
+ 
+CHIP Eq {
+  IN a, b;
+  OUT out; // True iff a=b
```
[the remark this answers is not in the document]
stephanmax: You mean "iff"? No, that‘s just short for "if and only if".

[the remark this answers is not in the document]
RunDevelopment: TIL: English has some nice abbreviations. 😄

stephanmax:
> So the function name has to relate to a file name (excluding builtins like `And` I suppose).
> I think this pretty much falls within the concept of a function so can we land this feature?

You got me convinced! Yes, it is actually like a function 👍

components/prism-nand2tetris-hdl.js
```diff
@@ -2,6 +2,7 @@ Prism.languages['nand2tetris-hdl'] = {
 	'comment': /\/\/.*|\/\*[\s\S]*?(?:\*\/|$)/,
 	'keyword': /\b(?:CHIP|IN|OUT|PARTS|BUILTIN|CLOCKED)\b/,
 	'boolean': /\b(?:true|false)\b/,
+	'function': /[A-Za-z][A-Za-z0-9]*(?=\()/,
```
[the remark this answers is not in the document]
RunDevelopment: Alright. It's good as is then.

components.json
```diff
@@ -342,6 +342,10 @@
 			"title": "HCL",
 			"owner": "outsideris"
 		},
+		"nand2tetris-hdl": {
```
RunDevelopment: One last thing: Please insert this at its alphanumerical correct position, after N4JS.

stephanmax: Great catch. Done.

RunDevelopment:
Thank you for contributing!

volado:
It seems to me that `components/prism-hdl.min.js` shouldn’t have been committed.

RunDevelopment:
Indeed! I'll fix it.

PR #1711, opened 2019-01-17
Modify attr-name regex to include fields with arguments

kuldar:
I'm very green at Prism themes and Regex, so I'd love if someone would sanity-check this. But it seems like the current GraphQL component doesn't apply `attr-name` class to schema fields with attributes (e.g. `user(id: ID!): User`).

I did my best Googling and tried to modify the regex syntax to also include cases where colon appears after a set of parenthesis.

screenshot 2019-01-17 at 16 51 34


components/prism-graphql.js
```diff
@@ -11,7 +11,7 @@ Prism.languages.graphql = {
 		pattern: /@[a-z_]\w*/i,
 		alias: 'function'
 	},
-	'attr-name': /[a-z_]\w*(?=\s*:)/i,
+	'attr-name': /[a-z_]\w*((?=\s*:)|(?=\s*\([^\)]*\):))/i,
```
RunDevelopment: 2 nits:

1. Please use `[^()]` inside the parentheses.
2. I think the lookahead would be easier to understand like this: `(?=\s*(?:\([^()]*\))?:)`.
    This also removes the capturing group (we want to avoid these if possible).

kuldar: Thank you for the suggestion, Michael! The new version reads better for sure. 👍 I also ran Gulp to include the minified version.

RunDevelopment:
Thank you for contributing!

PR #1712, opened 2019-01-17
Deleted prism-hdl.min.js

RunDevelopment:
This deletes `prism-hdl.min` as pointed out [here](https://github.com/PrismJS/prism/pull/1710#issuecomment-455257120).


RunDevelopment:
I'll just go ahead and merge this.

PR #1714, opened 2019-01-23
Add support for styled-components syntax highlighting

mxstbr:
[styled-components](https://github.com/styled-components/styled-components) is a widely used CSS-in-JS library. It lets users write their CSS in JavaScript as CSS strings. For example:

```
const Button = styled.button`
  color: blue;
  background: red;
`
```

This PR adds support for highlighting the styled-components template strings as CSS.

We have been running this exact code [on the styled-components website for almost two years](https://github.com/styled-components/styled-components-website/blob/bbfcb0d21aaec9e8202609a67e7bba5278042056/utils/prismTemplateString.js) and it has worked perfectly.

I have added tests for the feature and am willing to maintain this going forward.

/cc @kitten @probablyup the other core maintainers of the styled-components package


components/prism-javascript.js
```diff
@@ -76,6 +76,31 @@ Prism.languages.insertBefore('javascript', 'string', {
 	}
 });
 
+Prism.languages.insertBefore("javascript", "template-string", {
+  "styled-components-template-string": {
+    pattern: /(styled(\.\w+|\([^\)]*\))(\.\w+(\([^\)]*\))*)*|css|createGlobalStyle|keyframes)`(?:\$\{[^}]+\}|\\\\|\\?[^\\])*?`/,
```
RunDevelopment: 1. Please don't use capturing groups if not necessary.
2. You don't need to escape `)` in a char set.

A corrected minimal pattern could look like this: `styled(?:\([^)]*\))?(?:\.\w+(?:\([^)]*\))*)+` (plus the rest)

mxstbr: That specific RegExp does not correctly match

```JS
styled(Button)`
  background: red;
`
```

Results in this test failure:

```
  1) Testing language 'javascript' – should pass test case 'styled-components_feature':

      + expected - actual

           "punctuation"
           ")"
         ]
         [
      -    "template-string"
      +    "styled-components-template-string"
           [
             [
               "string"
      -        "`\n  background: red;\n`"
      +        "`\n  background: red"
             ]
      +      ";"
      +      [
      +        "string"
      +        "\n`"
      +      ]
           ]
         ]
```

RunDevelopment: You're right, my bad. The last `+` has to be a `*`, so `styled(?:\([^)]*\))?(?:\.\w+(?:\([^)]*\))*)*`.

components/prism-javascript.js
```diff
@@ -76,6 +76,26 @@ Prism.languages.insertBefore('javascript', 'string', {
...
+		greedy: true,
+		inside: {
+			interpolation: {
+				pattern: /\$\{[^}]+\}/,
+				inside: {
+					"interpolation-punctuation": {
+						pattern: /^\$\{|\}$/,
+						alias: "punctuation"
+					},
+					rest: Prism.languages.javascript
```
RunDevelopment: You don't want string highlighting?

mxstbr:
Actually, this does not correctly highlight CSS, right? 🤔

RunDevelopment:
> Actually, this does not correctly highlight CSS, right? 🤔

What does?

mxstbr:
I mean, the CSS part inside the tagged template strings is not highlighted as CSS:

```JS
styled.button`
  color: blue;
`
```

mAAdhaTTah:
Sorry for the dlay.

Besides the conflicts we now have, this feels like it might be better as a `js-extras` language, rather than part of JS code, since most users won't need this functionality.

Thoughts?

mxstbr:
@mAAdhaTTah if anything, it should be part of JSX since it is React-specific. What would you think about that? 

Happy to resolve the conflicts tomorrow if we can agree on moving this to JSX!

mAAdhaTTah:
@mxstbr `jsx-extras`? Since this is targeted at a subset of people who use JS(X), I'd prefer people to opt into it.

RunDevelopment:
Moving this to JS-Extras will mean that the `createTemplate` method which creates the `template-string` token has to remain in JS. The only line moved to JS-Extras will be:

```js
if (Prism.languages.css) {
    Prism.languages.javascript['template-string'].unshift(createTemplate('css', /\b(?:styled(?:\([^)]*\))?(?:\.\w+(?:\([^)]*\))*)*|css|createGlobalStyle|keyframes)/.source));
}
```

Technically, I also think that this feature should be moved of JS-Extras. All three lines of it...

RunDevelopment:
Also, JS-Extras has its dependencies set up in a way so that it doesn't affect TypeScript.
This feature should also be available for TypeScript, so we would have to either 

1. Change the dependencies and make JS-Extras compatible with TypeScript.
2. Make a new language/extension similar to JS-Extras exclusively for JS/TS and JSX/TSX for just this feature (all three lines of it).
3. Just leave it as it is now...

It might seem like overkill, but I think that option 2 is the best one.

RunDevelopment:
Conflicts are resolved.

Regarding offloading the CSS templates:
How about we leave it as it is for now (so as a part of JS) and make a new 'language' (something like JS-Templates) in the future which only handles different languages embedded in template strings?
When adding this new 'language' we can add support for [Polymer's `html`](https://polymer-library.polymer-project.org/3.0/docs/devguide/dom-template) as well.

karlhorky:
Ah, just opened a new issue which is a superset of this (#1930 - for handling all languages tagged template literals) and didn't see this while searching for similar issues.

I will mention this in the other issue. This could also be a good starting point for the other languages (and also for other CSS-in-JS libs like `styled-jsx`).

karlhorky:
> How about we leave it as it is for now (so as a part of JS)

I would vote to add it as part of the main default JS language, since tagged template strings with different languages are common in places other than React / JSX / other frameworks.

RunDevelopment:
I'll close this now as all of the changes in this were also part of #1931.

PR #1719, opened 2019-02-07
Line Numbers plugin instructions clarifications

bruceontheloose:
* Fix a typo.
* Clarify some of the instructions, including in "How to use" and by adding a note to "Soft wrap support" that is in the same spirit as the `data-start` note that appears under the "HTML" heading.


RunDevelopment:
Thank you for contributing!

PR #1722, opened 2019-02-11
Improve JavaScript parameter recognization

RexSkz:
Some improvements:

- Fix: not recognizing function name with uppercase letter
- Add: test cases for destructuring
- Add: support default value with bracket (function call)


components/prism-javascript.js
```diff
@@ -29,26 +29,26 @@ Prism.languages.insertBefore('javascript', 'keyword', {
...
 	// This must be declared before keyword because we use "function" inside the look-forward
 	'function-variable': {
-		pattern: /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*[=:]\s*(?:async\s*)?(?:\bfunction\b|(?:\([^()]*\)|[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)\s*=>))/i,
+		pattern: /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*[=:]\s*(?:async\s*)?(?:\bfunction\b|(?:\(.*\)|[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)\s*=>))/i,
 		alias: 'function'
 	},
 	'parameter': [
 		{
-			pattern: /(function(?:\s+[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)?\s*\(\s*)[^\s()][^()]*?(?=\s*\))/,
+			pattern: /(function(?:\s+[_$A-Za-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)?\s*\(\s*)[^\s].*?(?=\s*\)\s*\{)/,
```
RunDevelopment: 1. Why don't you use the `i` flag here as well? It uses fewer characters than `A-Z`.
2. `[^\s]` == `\S`
2. Please use a pattern for recursive expressions.
3. I get why you changed the lookahead but this won't be necessary after we use a pattern for recursive expression.

components/prism-javascript.js
```diff
@@ -29,26 +29,26 @@ Prism.languages.insertBefore('javascript', 'keyword', {
...
 			inside: Prism.languages.javascript
 		},
 		{
-			pattern: /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*=>)/,
+			pattern: /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*=>)/i,
 			inside: Prism.languages.javascript
 		},
 		{
-			pattern: /(\(\s*)[^\s()][^()]*?(?=\s*\)\s*=>)/,
+			pattern: /(\(\s*)[^\s].*?(?=\s*\)\s*=>)/i,
```
RunDevelopment: 1. The `i` flag doesn't do anything here.
2. `[^\s]` == `\S`.
3. Please use a pattern for recursive expressions.

components/prism-javascript.js
```diff
@@ -29,26 +29,26 @@ Prism.languages.insertBefore('javascript', 'keyword', {
...
 			lookbehind: true,
 			inside: Prism.languages.javascript
 		},
 		{
-			pattern: /((?:\b|\s|^)(?!(?:as|async|await|break|case|catch|class|const|continue|debugger|default|delete|do|else|enum|export|extends|finally|for|from|function|get|if|implements|import|in|instanceof|interface|let|new|null|of|package|private|protected|public|return|set|static|super|switch|this|throw|try|typeof|var|void|while|with|yield)(?![$\w\xA0-\uFFFF]))(?:[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*\s*)\(\s*)[^\s()][^()]*?(?=\s*\)\s*\{)/,
+			pattern: /((?:\b|\s|^)(?!(?:as|async|await|break|case|catch|class|const|continue|debugger|default|delete|do|else|enum|export|extends|finally|for|from|function|get|if|implements|import|in|instanceof|interface|let|new|null|of|package|private|protected|public|return|set|static|super|switch|this|throw|try|typeof|var|void|while|with|yield)(?![$\w\xA0-\uFFFF]))(?:[_$A-Za-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*\s*)\(\s*)[^\s].*?(?=\s*\)\s*\{)/,
```
RunDevelopment: 1. `[^\s]` == `\S`.
3. Please use a pattern for recursive expressions.

components/prism-javascript.js
```diff
@@ -29,26 +29,26 @@ Prism.languages.insertBefore('javascript', 'keyword', {
 	},
 	// This must be declared before keyword because we use "function" inside the look-forward
 	'function-variable': {
-		pattern: /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*[=:]\s*(?:async\s*)?(?:\bfunction\b|(?:\([^()]*\)|[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)\s*=>))/i,
+		pattern: /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*[=:]\s*(?:async\s*)?(?:\bfunction\b|(?:\(.*\)|[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)\s*=>))/i,
```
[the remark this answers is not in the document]
RexSkz: You're right... I want to match recursive brackets, but it's not the thing that pure regex can do, so I thought matching the start `(` and the end `)` is enough.

The finite recursive depth trick is fine in most cases, I'll take your advice.

We can have neither strings nor comments in parameters because they'll split the codes before `function-variable` and `parameter` get to work, this is another example that comments break the highlight in Prism (what a pity):

```jsx
<Button
    prop1={value1}
    // some comments
/>
```

components/prism-javascript.js
RexSkz: The keyword `function` is case-sensitive, so `i` flag can't be used here.

components/prism-javascript.js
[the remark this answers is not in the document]
RunDevelopment: Yeah, comments and strings are a huge pain...

components/prism-javascript.js
RunDevelopment: Oh, you're right! I missed that part. My bad!

components/prism-javascript.js
```diff
@@ -29,26 +29,26 @@ Prism.languages.insertBefore('javascript', 'keyword', {
...
 	// This must be declared before keyword because we use "function" inside the look-forward
 	'function-variable': {
-		pattern: /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*[=:]\s*(?:async\s*)?(?:\bfunction\b|(?:\([^()]*\)|[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)\s*=>))/i,
+		pattern: /[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*[=:]\s*(?:async\s*)?(?:\bfunction\b|(?:\((?:[^()]|\([^()]*\))*\)|[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)\s*=>))/i,
 		alias: 'function'
 	},
 	'parameter': [
 		{
-			pattern: /(function(?:\s+[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)?\s*\(\s*)[^\s()][^()]*?(?=\s*\))/,
+			pattern: /(function(?:\s+[_$A-Za-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*)?\s*\(\s*)\S(?:[^()]|\([^()]*\))*(?=\s*\))/,
```
RunDevelopment: I think that `\S` isn't an optimal solution because it could match `(` or `)`.
How about this:

```js
/(... \(\s*)(?!\s)(?:[^()]|\([^()]*\))+?(?=\s*\))/
```

RexSkz: Yeah, negative lookahead is better here.

RunDevelopment:
The top-level group of the expression before the lookahead has to be reluctant because we will match too many spaces otherwise. So `+?`.

RexSkz:
@RunDevelopment Finished~

RunDevelopment:
Thank you for contributing!

PR #1723, opened 2019-02-11
Autolinker: Fixed that all tokens were url-decoded

### LINKED ISSUE #1721: Autolinker plugin strips out percent signs and some digits in Java (and maybe others)
With the Autolinker plugin enabled, Prism mistakenly replaces some percent signs and digits. I've only seen this in Java but other languages might be affected as well.

```
("%010d")
```

would become:

```
("0d")
```

Here's a minimal reproduction: https://codesandbox.io/s/r5o7ny0x74

Something important to note here though, is that this doesn't happen if **only** the Java language is included in the Prism object. It's when you add all the others that this is reproducible, so I'm guessing it's some sort of conflict with the grammar or hooks for another language.

Reference: https://github.com/refined-bitbucket/refined-bitbucket/issues/264

RunDevelopment:
This PR fixes #1721.

The problem was that the url-decoding introduced in #1173 was applied to all tokens.


RunDevelopment:
🎉

PR #1725, opened 2019-02-12
Support JSX/TSX class-name with dot

RexSkz:
Support class names like `<Tree.TreeNode.Item>`.


RunDevelopment:
I really like this change but I also think that we should include a `punctuation` pattern for the dots.

RexSkz:
It's easy to add `punctuation` inside, but both VSCode and GitHub treats dot as a normal character in JSX tag, and there's no other special characters in it, so I think it's not necessary to recognize dot here.

RunDevelopment:
Well, it is as you said, it's easy to add later on.

RunDevelopment:
Thank you for contributing!

PR #1726, opened 2019-02-12
Fix param missing for toolbar plugin sample code

RexSkz:
Missing `env` in sample code.


RunDevelopment:
Thank you for contributing!

PR #1727, opened 2019-02-13
fix(themes): Highlight `.token.null`

ExE-Boss:
The&nbsp;**JSON** syntax highlighting uses `.token.null` to style `null`: https://github.com/PrismJS/prism/blob/4362e42cd4c0271abd976145a85098092a51db80/components/prism-json.js#L15

This updates the&nbsp;Prism stylesheet to&nbsp;style it&nbsp;using the&nbsp;style for&nbsp;`.token.boolean` and&nbsp;`.token.number`.

### DIFF
--- a/themes/prism-coy.css
+++ b/themes/prism-coy.css
@@ -118,6 +118,7 @@ pre[class*="language-"]:after {
 .token.property,
 .token.tag,
 .token.boolean,
+.token.null,
 .token.number,
 .token.function-name,
 .token.constant,
--- a/themes/prism-dark.css
+++ b/themes/prism-dark.css
@@ -76,6 +76,7 @@ pre[class*="language-"] {
 .token.property,
 .token.tag,
 .token.boolean,
+.token.null,
 .token.number,
 .token.constant,
 .token.symbol {
--- a/themes/prism-funky.css
+++ b/themes/prism-funky.css
@@ -65,6 +65,7 @@ code[class*="language-"] {
 .token.property,
 .token.tag,
 .token.boolean,
+.token.null,
 .token.number,
 .token.constant,
 .token.symbol {
--- a/themes/prism-okaidia.css
+++ b/themes/prism-okaidia.css
@@ -71,6 +71,7 @@ pre[class*="language-"] {
 }
 
 .token.boolean,
+.token.null,
 .token.number {
 	color: #ae81ff;
 }
--- a/themes/prism-solarizedlight.css
+++ b/themes/prism-solarizedlight.css
@@ -97,6 +97,7 @@ pre[class*="language-"] {
 .token.property,
 .token.tag,
 .token.boolean,
+.token.null,
 .token.number,
 .token.constant,
 .token.symbol,
--- a/themes/prism-tomorrow.css
+++ b/themes/prism-tomorrow.css
@@ -70,6 +70,7 @@ pre[class*="language-"] {
 }
 
 .token.boolean,
+.token.null,
 .token.number,
 .token.function {
 	color: #f08d49;
--- a/themes/prism-twilight.css
+++ b/themes/prism-twilight.css
@@ -90,6 +90,7 @@ code[class*="language-"]::selection, code[class*="language-"] ::selection {
 
 .token.tag,
 .token.boolean,
+.token.null,
 .token.number,
 .token.deleted {
 	color: hsl(14, 58%, 55%); /* #CF6A4C */
--- a/themes/prism.css
+++ b/themes/prism.css
@@ -83,6 +83,7 @@ pre[class*="language-"] {
 .token.property,
 .token.tag,
 .token.boolean,
+.token.null,
 .token.number,
 .token.constant,
 .token.symbol,


mAAdhaTTah:
It seems curious that JSON is the only language that has `.token.null` for `null`. I think we highlight it as a `keyword` in other languages. Might make more sense to change how we highlight in JSON.

@RunDevelopment Thoughts?

ExE-Boss:
Well, PHP used `.token.null` at&nbsp;one point too: https://github.com/PrismJS/prism/blob/c5160a1908d0f2a752cae394b1289a5b7ef956f1/components/prism-php.js#L53-L58

And there might be&nbsp;other languages that do&nbsp;the&nbsp;same.

RunDevelopment:
At this point in time, the only languages which have a `null` token without aliases are JSON and SCSS.

> I think we highlight it as a `keyword` in other languages.

Depends on the language. Most use `keyword`, some use `boolean`, few use others.

> Might make more sense to change how we highlight in JSON.

I also think it's better for JSON to have null as a `keyword` instead of changing the themes to include `null` for a few reasons:

1. You only have to change one file. `keyword` is supported by all themes.
2. JS uses `keyword` for null and JSON is supposed to be a subset of JS (it's not, but almost).
3. We can always use the [Highlight Keywords](https://prismjs.com/plugins/highlight-keywords/) plugin to change the style of `null`.

ExE-Boss:
> 2. JS uses `keyword` for null and JSON is supposed to be a subset of JS (it's not, but almost).

[It is a subset of ES2019+](https://github.com/tc39/proposal-json-superset#subsume-json-aka-json--ecmascript), as ES2019 now allows unescaped `\u2028` and `\u2029` in strings:
```js
let str = "a b"; // This string contains an unescaped Line Separator
```

---

That said, I’d still like to fix highlighting of `.token.null`.

RunDevelopment:
> That said, I’d still like to fix highlighting of `.token.null`.

`.token.null` is pretty much a non-standard token name, so I think it's ok to ignore it.
Plus, changing JSON and SCSS is simply easier.

mAAdhaTTah:
Agreed. I'd prefer to use a token we reuse across a lot of languages than handling a token used once or twice in a theme. I find it a little curious some of them are using `boolean` for `null`; `keyword` seems to make the most sense, so maybe we just decide now to consistently use `keyword` for it?

Appreciate you raising this to our attention!

RunDevelopment:
> `keyword` seems to make the most sense, so maybe we just decide now to consistently use `keyword` for it?

Well, it's not that easy... 
Most languages in which null (so similar) is a language construct highlight it as `keyword`. But we can't just make null a `keyword` in all languages.
I.e. in C/C++ NULL is a macro, in PHP it's a constant.

But I think that null (or conceptually similar) should not be `boolean` as in Python (None), SQL, and Twig.

mAAdhaTTah:
Fair enough; choosing something that makes sense for each language is fine. I'm admittedly less familiar with those languages, but I think a null token for JSON & SCSSS only doesn't really make sense. So we should update the JSON + SCSS languages to give `null` different tokens. Agree?

ExE-Boss:
I dunno, the `null` token makes a lot of sense for JSON, maybe we could add `keyword` as an alias?

RunDevelopment:
> So we should update the JSON + SCSS languages to give `null` different tokens. Agree?

Maybe the alias idea from @ExE-Boss is even better. I mean, we used these token names for quite some time now, so there might be some folks out there how incorporated `null` into their themes. With an alias, it's unlikely (but still possible) that we break them.

mAAdhaTTah:
I'm on board with that.

RunDevelopment:
Just for clarification: We are not going to change the themes now, are we? @ExE-Boss @mAAdhaTTah

ExE-Boss:
I’d like to&nbsp;fix the&nbsp;themes.

RunDevelopment:
@ExE-Boss With or without the aliases? Because after the aliases, there's nothing to fix.

ExE-Boss:
Well, even with the aliases, `.token.null` should still take precedence over `.token.keyword`.

RunDevelopment:
So you want to change the look of `null` to be highlighted the different from `keyword`, I suppose.
I think is better if JSON's nulls are highlighted like JS's nulls. Consistency.

RunDevelopment:
I feel like this question should have been asked a lot sooner, but why do you want `null` to look like `boolean`?
Null and true/false are not conceptually close (IMO) and most editors highlight both like keywords, so why `boolean`?

ExE-Boss:
After a&nbsp;day of&nbsp;thinking it&nbsp;over, I&nbsp;think it might be better to&nbsp;just highlight it&nbsp;using `keyword` as&nbsp;an&nbsp;alias.

mAAdhaTTah:
I'm personally opposed to changing the theme here, as I think handling a token in all the themes that only appears in two languages is unnecessary and the correct fix is the keyword change, as suggested.

RunDevelopment:
> I'm personally opposed to changing the theme here

Agreed. After #1735 there's nothing left to fix.

I'm closing this now. Feel free to reopen this if further discussion is necessary.

PR #1731, opened 2019-02-13
Add "yarn add" to bash functions

CupOfTea696:
Add "yarn add" to bash functions, so that it highlights any `yarn add` commands in the same way it highlights `npm install`.


RunDevelopment:
LGTM.

Thank you for contributing!

PR #1733, opened 2019-02-13
fix(json): Add `keyword` alias to `null`

ExE-Boss:
Opened&nbsp;based on&nbsp;the&nbsp;discussion in&nbsp;https://github.com/PrismJS/prism/pull/1727.


components/prism-json.js
```diff
@@ -11,8 +11,14 @@ Prism.languages.json = {
 	'number': /-?\d+\.?\d*(e[+-]?\d+)?/i,
 	'punctuation': /[{}[\],]/,
 	'operator': /:/,
-	'boolean': /\b(?:true|false)\b/,
-	'null': /\bnull\b/
+	'boolean': {
+		pattern: /\b(?:true|false)\b/,
+		alias: 'keyword'
```
RunDevelopment: Should `boolean` stay as is?

ExE-Boss: Probably.

PR #1734, opened 2019-02-13
feat(bash): Add `pnpm`

ExE-Boss:
Since #1731 added `yarn` to `prism-bash.js`, this adds highlighting for [`pnpm`](https://pnpm.js.org), which is the third largest Node package manager.

`pnpm`’s command&#x2011;line syntax is otherwise the same as for `npm`, and as it’s the third largest, I&nbsp;believe highlighting support for&nbsp;it&nbsp;should be&nbsp;added.

## See also:
- https://www.kochan.io/nodejs/why-should-we-use-pnpm.html
- https://medium.com/pnpm/flat-node-modules-is-not-the-only-way-d2e40f7296a3

/cc @zkochan


RunDevelopment:
Thank you for contributing!

PR #1739, opened 2019-02-14
JS: Fixed `function-variable`

RunDevelopment:
The `function-variable` pattern should not be case insensitive because of the keywords in the pattern.


RunDevelopment:
Conflict resolved

PR #1743, opened 2019-02-15
JS Extras

RunDevelopment:
This PR adds extras for Javascript. 
It's based on @atomiks's [comment](https://github.com/PrismJS/prism/issues/1623#issuecomment-463422719) and heavily modified by me.

All languages extending Javascript (expect for JSX) are loaded before JS Extras and therefore unaffected.

By default, this only changes the highlighting of known variables and previously undetected known class names.

### Keyword groupings

Some keywords get new aliases which might be useful for highlighting:

- `as`, `default`, `export`, `from`, and `import` get the `module` alias.
- `null` and `undefined` get the `nil` alias.

It's implemented in a way that is not (currently) testable (unfortunately) to be compatible with the [Highlight Keywords plugin](https://prismjs.com/plugins/highlight-keywords/) which only checks the [token name](https://github.com/PrismJS/prism/blob/master/plugins/highlight-keywords/prism-highlight-keywords.js#L11).

### Known class names

All [JS standard built types](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects) and all error types will be recognized.

### Known variables

Some variables in JS are fixed. Here are some of them:

- `document`, `location`, `navigator`, `performance`, `localStorage`, `sessionStorage`, and `window` get the `dom` and `variable` classes.
- `console` gets the `console` and `class-name` classes.

### Special operators

Some operators will get their own token name:

- `=>` gets `arrow`.
- `...` gets `spread`.

### Property accesses

All accesses of an object property using literals like `foo.bar`, `String.raw`, `foo.bar.baz` but not including accesses via computed keys like `foo['bar']`.

### Methods

Methods are functions with are being accessed from an object. This also includes the static methods of a class.
E.g.: `console.log()`, `foo.bar()`, `String.fromCodePoint(72334)`

This also includes variables like `foo.bar = function() { this.baz = true; }`.

All methods are also _property accesses_.

### Maybe class names

These are names which we are pretty sure are class names but not quite 100% sure, thus maybe class names. All names with start with an upper case letter are assumed to may be class names.

All `property-access` and `function` tokens are checked for maybe class names.

## Not included

There are some features for @atomiks work I didn't include.

- Some keyword groupings I did not include because they didn't seem useful to me.
- (Property) definitions (e.g. `{ a: 1, b: 2 }`, here `a` and `b`) I couldn't include because I couldn't find a way to fix false positives caused by the conditional operator (e.g. `cond ? foo: bar`).
    My basic approach was to say that a property definition has to be preceded by either `{` or `,`. That works great but not with comments.
    ```js
    {
        foo: null, // explain foo
        bar: false // not preceded by ','
    }
    ```
    If we ever get something which allows us to ignore comments, this will work. Until then, it's not included.


components/prism-js-extras.js
```diff
@@ -0,0 +1,100 @@
...
+		]
+	});
+
+	Prism.languages.javascript['keyword'].unshift(
+		{
+			pattern: /\b(?:as|default|export|from|import)\b/,
+			alias: 'module'
+		},
+		{
+			pattern: /\b(?:null|undefined)\b/,
```
ExE-Boss: Shouldn’t `null` get the `null` alias to match JSON (see: #1727, #1733, #1735)?
```suggestion
			pattern: /\bnull\b/,
			alias: 'null'
		},
		{
			pattern: /\bundefined\b/,
```

RunDevelopment: That's true.

RunDevelopment:
The failed tests, failed because of `git clone`... 
2/4 node versions and my local system all passed the tests.

PR #1745, opened 2019-02-16
JSONP support fix

RunDevelopment:
This PR fixes that JSON did not support [JSONP](https://en.wikipedia.org/wiki/JSONP) callback functions.

I also added a JSON example page.


components/prism-json.js
```diff
@@ -9,7 +9,9 @@ Prism.languages.json = {
 		greedy: true
 	},
 	'number': /-?\d+\.?\d*(e[+-]?\d+)?/i,
-	'punctuation': /[{}[\],]/,
+	// function is for JSONP
+	'function': /[_$a-zA-Z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*\()/,
```
ExE-Boss: This should be&nbsp;moved into the&nbsp;`jsonp` language which would be&nbsp;extending&nbsp;`json`.

RunDevelopment: Alright, there you go!
I also think that it's a cleaner solution like that.

Issue #1748, opened 2019-02-16
Test page: Allow variable sources


RunDevelopment:
Add an option to the [Test page](https://prismjs.com/test.html) where the user can specify the source of the language definitions tested.

This feature can be used to compare the highlighting differences between various versions of Prism from the PrismJS repo, forks of Prism, and external sources.

PR #1758, opened 2019-02-23
Test page: Added variable source

RunDevelopment:
This implements #1748 by using the URL hash to specify a `source` parameter.
E.g.: `https://prismjs.com/test.html#source=<URL>`

If `source` is not specified, the test page will behave as is does now.
If `source` is specified, the test page will reload `components.js` from `source` and fetch all other Prism files via the specified URL (e.g. `<URL>components/prism-c.js`).

## Themes

Themes are not affected by `source`.

Please note that the current `components.js` of the PrismJS website will be loaded regardlessly. So we load _our_ `components.js`, handle _our_ themes, and then load the `components.js` from `source` (if specified) for language dependencies and to create the list of languages.

## Why is this useful?

My main motivation for this is reviewing.
Using [combinatronics.com](https://combinatronics.com/) we can directly try the changes proposed in PRs (e.g. `https://prismjs.com/test.html#source=https://combinatronics.com/RunDevelopment/prism/html-attr-fix/`).

My intention is to write a GitHub bot which automatically comments this link so we and other users can play with the changes.
This will only affect new PRs which change the languages of course.

## Notes

Because `#` is an invalid character for file names, it is not possible to specify `source` if the file is opened directly from the local file system (AFAIK). A local server will work of course.

I use a polyfill for `URLSearchParams` (used to parse the URL hash).

This works for the latest versions of Chrome, Firefox, Edge, and IE11 on my machine.


ExE-Boss:
Actually, files loaded using the `file:` protocol can have URL fragments (the&nbsp;part after&nbsp;`#`)

RunDevelopment:
I gave this a quick overhaul. 

Instead of the hash, this is now implemented using the search part of the URL. So changing the source means reloading the page (this makes the implementation a little easier).

I also added a small panel for displaying the current source and for users to set the source. It has 3 buttons that will spawn browser prompts.

<details>

image

Prompts:

__From URL__
image

__From GitHub__
image

</details>

The _From GitHub_ button is especially useful because you can just past the user name + branch from a pull request.
image

RunDevelopment:
I forget that I can't test the bot because the Danger code executed is from prism/master.

RunDevelopment:
Update:

I scaled down the PR. It now only implements support for the `source` URL search parameter. All UI changes have been removed. I don't think it's necessary for now. If need be, I'll just re-add it.

Our lovely bot will now also add a link to the test page for the current PR. This will make it easy for anyone to try out language changes/additions.

Here's an example of how a bot message may look like:

---

## JS File Size Changes (gzipped)

A total of 1 files have changed, with a combined diff of +26 B (+1.4%).

<details>

| file | master | pull | size diff | % diff |
| components/prism-php.min.js | 1.9 KB | 1.92 KB | +26 B | +1.4% |

</details>

## Try it out

I detect that you changed languages. Everyone can [try out the changes here](http://prismjs.com/test.html?source=https://combinatronics.com/RunDevelopment/prism/issue2817#language=php).
<sub>(The link will stop working once the PR branch was deleted.)</sub>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 5a4cbcf93bf3de5e4a231d0ce4174b7d1121a569
</p>

RunDevelopment:
@mAAdhaTTah What do you think?

This is something that will make contributing and reviewing easier, so I would like to merge this ASAP. 
(E.g. I can just send someone a link with a PR as the source and ask them if their issue is completely fixed.)

RunDevelopment:
While a nice quality-of-life feature, I don't think it is vital for this project and adds quite a bit of complexity to the test page.

For those reasons, I'll now close this.

PR #1759, opened 2019-02-25
Fixed dependencies of Pug and Pure

RunDevelopment:
There were a few problems with the dependencies of Pug and Pure.

1. Pug did not require [markup](https://github.com/PrismJS/prism/blob/master/components/prism-pug.js#L50).
2. Pug had `ejs`, `hogan`, `mustache`, `plates`, and `swig` as peer dependencies, none of which are languages (or aliases) Prism supports.
3. Pure had `ats` and `dsp` as peer dependencies, none of which are languages (or aliases) Prism supports.

I removed the unsupported languages for now.


RunDevelopment:
Might have to adjust Pug after #1769.

PR #1760, opened 2019-02-25
Some encouragement for visitors of PrismJS.com to request new languages

RunDevelopment:


mAAdhaTTah:
If we're going to do this, we should add an issue template for it so we can ask for more info.

RunDevelopment:
I was thinking the same but I also think that we should also include templates for other common issue types while we're at it.

Issue #1764, opened 2019-02-26
<pre> line breaks get deleted when calling prism.highlightElement(code)


SetTrend:
I set up my HTML using the following snippet:

```html
<div>
 <pre><code class=" language-js" contenteditable="true"></code></pre>
</div>
```

When I call `prism.highlightElement(code);` from the following TypeScript snippet:

```ts
import * as prism from "prismjs";

const code: HTMLElement = document.getElementsByTagName("code")[0];

prism.highlightElement(code);
```

... all line breaks in the `<code>` element get deleted when calling `prism.highlightElement(code)`.
<br/>

### before

![multiple lines](https://user-images.githubusercontent.com/9283914/53380173-c40dca00-396c-11e9-9170-303de94268ea.png)

### after

![one liner](https://user-images.githubusercontent.com/9283914/53380178-c839e780-396c-11e9-88c6-a11e9cf93732.png)
<br/>

#### I believe this is not supposed to happen?
<br/>

RunDevelopment:
I assume that you wrote the highlighted code yourself (because of `contenteditable="true"`).

The problem here is that Prism uses a literal string line break (so `\n` or `\r\n`).
But that is not what the browser will insert when you press <kbd>Enter</kbd>. It will insert a new `<br>` which Prism will ignore (just like all other HTML elements).

The best solution in that is to use keyboard event to handle <kbd>Enter</kbd> yourself.
A working example of this [dabblet](http://dabblet.com/). The script responsible for this behavior can be found [here](https://github.com/LeaVerou/dabblet/blob/master/code/editor.js).

> I believe this is not supposed to happen?

If my assumption is correct, this is supposed to happen.
Otherwise, please provide the full example producing the bug.

SetTrend:
> It will insert a new `<br>` which Prism will ignore (just like all other HTML elements).

Nope. The `<code>` element lies within a `<pre>` element, so `\r\n` and `\n` are both interpreted as newlines by the browser. I debugged my session and I can confirm that `\n` characters are being inserted.

I will upload a sample in short time.

SetTrend:
Is there a public CDN available for Prism I can link to in my sample page?

RunDevelopment:
https://cdnjs.com/libraries/prism will work I guess.

RunDevelopment:
Prism will also add the `language-js` CSS class to the `<pre>` element.
[...]

PR #1766, opened 2019-02-26
Fixed issue with <br> elements

SetTrend:
# Synopsis
Prism ignored `<br>` elements and removed them from the code. This behaviour was based on the fact that `Prism.highlightElement()` used `element.textContent`, which expunges `<br>` elements (because they are no text nodes).

I replaced the use of `element.textContent` with `element.innerText` (which correctly produces a node's text content whitespace) and removed a call to `replace()` that was removing newlines from the tokens.

So, this commit retains the new-lines correctly.

See issue #1764 .
<hr/>

### Remarks
I only tested my personal use case.


components/prism-core.js
```diff
@@ -28,7 +28,7 @@ var _ = _self.Prism = {
 			} else if (_.util.type(tokens) === 'Array') {
 				return tokens.map(_.util.encode);
 			} else {
-				return tokens.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/\u00a0/g, ' ');
```
RunDevelopment: Why did you remove this?

SetTrend: `<pre>` is supposed to retain newline characters, so it doesn't make sense to remove them (in the original code, see merge request).

SetTrend: This line caused text tokens to be altered from `\nTest\n` to ` Test `.

So what was ...

```javascript
var i;

Test

var j;
```

... in the original `<code>` element became ...

```javascript
var i; Test var j;
```

SetTrend: See my sample HTML file in issue #1764.

Check the result when you alter the code content to:

```javascript
var i = 1;

Test

function()
{}
```

RunDevelopment: This line only escapes text to be valid HTML, so that e.g. the code `<p>` is not interpreted as a tag but as text.

I'm asking why you removed the removal of No-break spaces. I don't see how this relates to new lines.
(TBH, I don't know why we remove NBSPs in the first place)

SetTrend: D'oh! You are right. I misread the `\u00a0` as `\u000a`.

The Firefox debugger I'm using is not very stable. While debugging it turned out that the debugger stripped the new-lines from my test when I stepped over that function. I then examined that function and misread the literal value.

Yet, it's like you wrote: There is no need to escape the non-breaking space. I can remove the change, though, if you prefer.

RunDevelopment: Yes, please remove this change.
It's better to keep things simple.

ExE-Boss:
I think a better solution would be a hook that would auto&#x2011;convert `<br>` elements into `\n`.

SetTrend:
As @RunDevelopment correctly mentioned in [his reply to #1764](/PrismJS/prism/issues/1764#issuecomment-467501221), `innerText` also adds `\n` for non-nested block elements, mimicking what the user sees. I guess the complex calculation for what's supposed to add a `\n` character is what's making `innerText` slower than `textContent`.

SetTrend:
As I wrote yesterday in #1764, I'm now going to close this pull-request.

I'd be interested, though, to learn how many users preferred the current behaviour against the suggested behaviour. I believe the current behaviour is odd, but I'm open to opinions, though.

RunDevelopment:
I'm going to make a new issue for that.

PR #1768, opened 2019-02-26
Ruby/ERB: Fixed block comments

### LINKED ISSUE #1767: support block comments in erb
ERB has block comments like

```erb
<%# this is a block comment %>
<%
=begin %>
                block comment
    (both lines of both the begin and end tags must be at the start of their lines)
<%
=end %>

<%# this is not %>
                <%
                =begin %>
                not a comment
                <%
                =end %>
```

Would love to see Prism support that!

RunDevelopment:
This PR fixes #1767.

- Added support for block comments containing closing tags.
- Ruby block comments can now have text after `=begin`. E.g.
    ```rb
    =begin foo
    =end
    ```

## Example

image


olets:
Thanks @RunDevelopment 👏

Issue #1215, opened 2017-11-01
Third-party language definitions - prism-ejs-language


dutchenkoOleg:
I wrote tiny extend for EJS highlighting  
maybe someone will handy

https://github.com/dutchenkoOleg/prism-ejs-language

Also, it would be great to hear criticism if I did something wrong ))

Golmote:
You could make a PR if you want EJS to be available directly from this repo, and on Prism website.
You'd have to add [tests](https://github.com/PrismJS/prism/tree/gh-pages/tests/languages) ([Test suite](http://prismjs.com/test-suite.html)), though.

PR #1769, opened 2019-02-26
EJS language

RunDevelopment:
This PR adds support of [EJS](https://github.com/mde/ejs).

## Limitation

This does not include support for custom delimiters (e.g., `<? ?>`)

## Example

image


ExE-Boss:
How does this compare to&nbsp;https://github.com/PrismJS/prism/issues/1215?

---

Also, I’d like to add this to MDN.

/cc @Elchi3 @escattone @jwhitlock

RunDevelopment:
> How does this compare to #1215?

I didn't know #1215 exists...
The main difference is that I use markup-templating which is usually better suited for this kind of work than extending and modifying `markup`.

PR #1774, opened 2019-02-27
Docs: Added missing `Prism.highlight` parameter

RunDevelopment:


RunDevelopment:
We can simply ignore this when #1777 get merged.

PR #1778, opened 2019-02-28
fix(core): Don’t redefine `clone` variable

ExE-Boss:
This shows up when setting
```json
"javascript.implicitProjectConfig.checkJs": true
```
in&nbsp;**VS&nbsp;Code**’s `settings.json`.

This also makes the&nbsp;minified files slightly smaller, as&nbsp;the&nbsp;`clone` variable is&nbsp;defined only in&nbsp;one&nbsp;place, instead&nbsp;of&nbsp;two:

**Before:**
```js
clone:function i(e,t){var r=a.util.type(e);switch(t=t||{},r){
case"Object":var n=a.util.objId(e);if(t[n])return t[n];var l={};t[n]=l;for(var o in e)e.hasOwnProperty(o)&&(l[o]=i(e[o],t));return l;
case"Array":var n=a.util.objId(e);if(t[n])return t[n];var l=[];return t[n]=l,e.forEach(function(e,a){l[a]=i(e,t)}),l}
/*default:*/return e}}
```
**After:**
```js
clone:function i(e,t){var n,r,l=a.util.type(e);switch(t=t||{},l){
case"Object":if(r=a.util.objId(e),t[r])return t[r];n={},t[r]=n;for(var o in e)e.hasOwnProperty(o)&&(n[o]=i(e[o],t));return n;
case"Array":return r=a.util.objId(e),t[r]?t[r]:(n=[],t[r]=n,e.forEach(function(e,a){n[a]=i(e,t)}),n)}
/*default:*/return e}}
```
**After https://github.com/PrismJS/prism/commit/a17303c40541eff04da06f35af7f077035aaa5f8:**
```js
clone:function i(e,t){var n,r,l=a.util.type(e);switch(t=t||{},l){
case"Object":if(r=a.util.objId(e),t[r])return t[r];n={},t[r]=n;for(var o in e)e.hasOwnProperty(o)&&(n[o]=i(e[o],t));return n;
case"Array":return r=a.util.objId(e),t[r]?t[r]:(n=[],t[r]=n,e.forEach(function(e,a){n[a]=i(e,t)}),n);
default:return e}}}
```


RunDevelopment:
Shouldn't you get the same warning for `id` as well? Because I do.

ExE-Boss:
Right, `id` is `string`, so the default TypeScript configuration doesn’t complain.

RunDevelopment:
Ahh, so it complained because of the different types. SonarLint complains because the variable `id` is redeclared in the same (function) scope.

It would be nice to fix that as well.

components/prism-core.js
```diff
@@ -80,6 +80,7 @@ var _ = _self.Prism = {
 					return clone;
 			}
 
+			// default:
```
RunDevelopment: There seems to be some problem here.

ExE-Boss: That shouldn’t be an issue, I just put that there so that I would remind myself that it’s only run when the above `switch` doesn’t match anything.

RunDevelopment: Wouldn't it be a good idea to make a `default` case then?

ExE-Boss:
Results:
With https://github.com/PrismJS/prism/commit/a17303c40541eff04da06f35af7f077035aaa5f8: 8&nbsp;bytes&nbsp;saved.
Without https://github.com/PrismJS/prism/commit/a17303c40541eff04da06f35af7f077035aaa5f8: 16&nbsp;bytes&nbsp;saved.

RunDevelopment:
I think the gained readability (it just looks cleaner IMO) is more important.

And then again: It's only 8 bytes.

RunDevelopment:
Thank you for contributing!

PR #1779, opened 2019-03-01
core(package): Update gulp to version 4.0.0

ExE-Boss:
This updates the repository to use gulp&nbsp;4.

PR #1780, opened 2019-03-01
gulp: Refactoring

RunDevelopment:
Since we just switched to gulp v4 (#1779) and dropped support for Node 4 some time ago, I took the chance to refactor our `gulpfile.js`.

I used ES6 syntax and removed [`gulp.task`](https://gulpjs.com/docs/en/api/task) which is not recommended anymore.

(Let's see if Node v6 can handle the changes. _Edit:_ It worked.)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -69,9 +69,12 @@ function minifyPlugins(cb) {
 	pump([src(paths.plugins), ...minifyJS(), rename({ suffix: '.min' }), dest('plugins')], cb);
 }
 function build(cb) {
-	pump([src(paths.main), header('\n/* **********************************************\n' +
-		'     Begin <%= file.relative %>\n' +
-		'********************************************** */\n\n'), concat('prism.js'), dest('./')], cb);
+	pump([src(paths.main), header(`
+/* **********************************************
+     Begin <%= file.relative %>
+********************************************** */
+
+`), concat('prism.js'), dest('./')], cb);
 }
 
 function componentsJsonToJs(cb) {
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,110 +1,98 @@
-var gulp   = require('gulp'),
-	rename = require('gulp-rename'),
-	uglify = require('gulp-uglify'),
-	header = require('gulp-header'),
-	concat = require('gulp-concat'),
-	replace = require('gulp-replace'),
-	pump = require('pump'),
-	fs = require('fs'),
+// @ts-check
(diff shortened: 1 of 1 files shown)


RunDevelopment:
@mAAdhaTTah

gulpfile.js
```diff
@@ -1,110 +1,96 @@
...
+function minifyComponents(cb) {
+	pump([src(paths.components), ...minifyJS(), rename({ suffix: '.min' }), dest('components')], cb);
+}
+function minifyPlugins(cb) {
+	pump([src(paths.plugins), ...minifyJS(), rename({ suffix: '.min' }), dest('plugins')], cb);
+}
+function build(cb) {
+	pump([src(paths.main), header('\n/* **********************************************\n' +
+		'     Begin <%= file.relative %>\n' +
+		'********************************************** */\n\n'), concat('prism.js'), dest('./')], cb);
```
mAAdhaTTah: If we're ES6-ifying everything, we can make this a template string.

gulpfile.js
```diff
@@ -1,110 +1,96 @@
...
+
+
+function minifyComponents(cb) {
+	pump([src(paths.components), ...minifyJS(), rename({ suffix: '.min' }), dest('components')], cb);
+}
+function minifyPlugins(cb) {
+	pump([src(paths.plugins), ...minifyJS(), rename({ suffix: '.min' }), dest('plugins')], cb);
+}
+function build(cb) {
+	pump([src(paths.main), header('\n/* **********************************************\n' +
```
ExE-Boss:
```suggestion
	pump([src(paths.main), header(`
\n/* **********************************************
```

gulpfile.js
```diff
@@ -1,110 +1,96 @@
...
+		'     Begin <%= file.relative %>\n' +
```
ExE-Boss:
```suggestion
     Begin <%= file.relative %>
```

gulpfile.js
ExE-Boss:
```suggestion
********************************************** */

`), concat('prism.js'), dest('./')], cb);
```

RunDevelopment: I didn't do this because then the line ends of `prism.js` depend on the line ends of `gulpfile.js`.

But if you think that that's not a problem I'll change it.

RunDevelopment: nvm. Template strings ftw! I already use them anyway for `components.js`.

PR #1781, opened 2019-03-01
Website: Added script directory

RunDevelopment:
To decrease the number of files in the root directory of the Prism repo, I added a `script` directory which holds all `.js` files which are used exclusively by the prismjs.com website.


ExE-Boss:
This will make things easier on&nbsp;#1796, once I&nbsp;resolve the&nbsp;conflicts.

PR #1782, opened 2019-03-01
Added JSDoc

### LINKED ISSUE #1555: Docs: Extending Prism page does not mention `Prism.languages.extend`
We should add a few lines explaining the method.

### LINKED ISSUE #1777: JSDoc: Generated documentation
Why don't we use [JSDoc](https://github.com/jsdoc3/jsdoc) to add generated code documentation to the [prismjs.com](https://prismjs.com/index.html) website? I'm thinking about something like `prismjs.com/docs/`.

The generated documentation will include all _documented variables/functions_ of `components/prism-core.js`. Functions that aren't documented won't show up in the doc.
I already tested this a little and it works quite well.

This is intended to replace the API section of [extending Prism](https://prismjs.com/extending.html#api) to keep Core and its documentation more in sync to avoid issues like #1774.

(We can also use this include API docs for plugins and components like markup templating in the future.)

Thoughts?

RunDevelopment:
This PR adds generated [JSDoc](https://github.com/jsdoc3/jsdoc) documentation to Prism (#1777) using the [Minami theme](https://github.com/Nijikokun/minami). The generated `docs` directory will be accessible from `prismjs.com/docs`.

This contains the doc of all highlight functions, `insertBefore`, `extend`, all hook functions, and `Prism.Token`. I used the docs from our [extending page](https://prismjs.com/extending.html#api) as a base and added a few sentences.
This includes complete type information for all documented function (except for hooks where the type of the `env` parameter depends on the hook).

Because JSDoc requires a README for the `index.html`, I used Prism's README. We can also make one specifically for the doc. (I just didn't know what to write.)

[Interactive demo](https://rundevelopment.github.io/docs/index.html). (Will only be available during the time this PR is open.)

Right now, only `prism-core` is used to generate the doc. Other documented plugins and languages will need some adjustment before JSDoc can handle them. And there's also the question of whether these components should even be part of the doc.

## Implementation

I used [`gulp-jsdoc3`](https://github.com/mlucool/gulp-jsdoc3), so `npx gulp` will also re-generate the doc (this takes less time than minifying the language definitions, so the build process won't take too much longer).

Because of some ... features of JSDoc, I had to make some debatable adjustments to the doc generation process:

1. JSDoc only _adds_ the newly generated doc. It doesn't remove outdated files, i.e. if the file documented was deleted or renamed. So to fix this, gulp will first delete the whole `doc` directory.
2. You cannot disable the generation time in the Minami theme, so I use a regex on the HTML files to remove the date.
3. JSDoc also includes a few font files which our theme doesn't use, so those get removed.

Another rather unfortunate thing is that JSDoc doesn't use Prism to highlight source code.
We could make it with a few regex replacements but that would only add another point on my list of debatable decisions.

---

This resolves #1777.
This resolves #1555.
This resolves #1774.


ExE-Boss:
It works decently with [TypeScript’s `checkJs`](https://github.com/Microsoft/TypeScript/wiki/JsDoc-support-in-JavaScript), although that doesn’t yet support `@memberof`.

RunDevelopment:
I looked a little into this and there don't seem to be any workarounds for this, are there?

ExE-Boss:
The only solution would be if TypeScript would add support for `@memberof` or&nbsp;if&nbsp;we&nbsp;would migrate portions&nbsp;of&nbsp;this project to&nbsp;TypeScript.

RunDevelopment:
Well, there is an [issue](https://github.com/Microsoft/TypeScript/issues/7237) so the hope is there...

I'm actually quite a fan of TS but I think that:

1. Porting the project for the sake of JS-Check isn't worth it.
2. TS isn't necessary for Prism.

Also, TS is not as popular as JS, so some people might get held back from contributing.

components/prism-core.doc.js
```diff
@@ -0,0 +1,37 @@
...
+ * @typedef Grammar
+ * @type {Object.<string, RegExp | TokenObject | Array.<RegExp | TokenObject>>}
+ * @property {Grammar} [rest] An optional grammar object that will appended to this grammar.
+ */
+
+/**
+ * A function which will invoked after an element was successfully highlighted.
+ *
+ * @callback HighlightCallback
+ * @param {HTMLElement} element The element successfully highlighted.
```
ExE-Boss: This doesn’t need to be an `HTMLElement`:
```suggestion
 * @param {Element} element The element successfully highlighted.
```

components/prism-core.js
```diff
@@ -168,10 +204,35 @@ var _ = {
...
 	plugins: {},
 
+	/**
+	 * This is the most high-level function in Prism’s API.
+	 * It fetches all the elements that have a `.language-xxxx` class and then calls {@link Prism.highlightElement} on
+	 * each one of them.
+	 *
+	 * This is equivalent to `Prism.highlightAllUnder(document, async, callback)`.
+	 *
+	 * @param {HTMLElement} container The root element, whose descendants that have a `.language-xxxx` class will be highlighted.
```
ExE-Boss: There’s no `container` parameter for this function.
```suggestion
```

components/prism-core.js
```diff
@@ -168,10 +204,35 @@ var _ = {
...
 
+	/**
+	 * Fetches all the descendants of `container` that have a `.language-xxxx` class and then calls
+	 * {@link Prism.highlightElement} on each one of them.
+	 *
+	 * The following hooks will be run:
+	 * 1. `before-highlightall`
+	 * 2. All hooks of {@link Prism.highlightElement} for each element.
+	 *
+	 * @param {HTMLElement} container The root element, whose descendants that have a `.language-xxxx` class will be highlighted.
```
ExE-Boss: `Document` isn’t an instance of `ParentNode`, which is why this causes issues with `tsc`.
```suggestion
	 * @param {ParentNode} container The root element, whose descendants that have a `.language-xxxx` class will be highlighted.
```

components/prism-core.js
```diff
@@ -187,6 +248,30 @@ var _ = {
...
+	 *
+	 * The following hooks will be run:
+	 * 1. `before-sanity-check`
+	 * 2. `before-highlight`
+	 * 3. All hooks of {@link Prism.highlightElement}. These hooks will only be run by the current worker if `async` is `true`.
+	 * 4. `before-insert`
+	 * 5. `after-highlight`
+	 * 6. `complete`
+	 *
+	 * @param {HTMLElement} element The element containing the code.
```
ExE-Boss: This doesn’t need to be an HTMLElement:
```suggestion
	 * @param {Element} element The element containing the code.
```

components/prism-core.js
```diff
@@ -447,15 +589,50 @@ var _ = {
...
+	 * @member {number}
+	 */
 	this.length = (matchedStr || "").length|0;
+	/**
+	 * Whether the pattern that created this token is greedy or not.
+	 * @member {boolean}
+	 * @see TokenObject
+	 */
 	this.greedy = !!greedy;
 }
```
ExE-Boss: This adds JSDoc for&nbsp;the&nbsp;`Token.stringify(…)` static method.
```suggestion

/**
 * Converts the token to an HTML representation.
 *
 * @param {string|Token|Token[]} o The token to convert to an HTML representation.
 * @param {string} language The current programming language.
 * @param {string|Token|Token[]} [parent] The parent token, if any.
 * @return {string} The HTML representation of the token.
 */
```

RunDevelopment: I added docs for it but made them private because I don't think that this method, which only ever used by prism-core itself, should appear in the _public_ API doc.

ExE-Boss: It’s still good to have that for type checking.

RunDevelopment: Well, it still isn't working properly because of the `@memberof`s...

examples.js
```diff
@@ -200,7 +200,7 @@ function loadLanguage (lang)
  * Returns all dependencies (as identifiers) of a specific language
  *
  * @param {string} lang
- * @returns {Array.<string>} the list of dependencies. Empty if the language has none.
+ * @returns {Array<string>} the list of dependencies. Empty if the language has none.
```
ExE-Boss:
```suggestion
 * @returns {string[]} the list of dependencies. Empty if the language has none.
```

tests/helper/test-case.js
```diff
@@ -135,7 +135,7 @@ module.exports = {
 	 *
 	 * @private
 	 * @param {string} filePath
-	 * @returns {{testSource: string, expectedTokenStream: Array.<Array.<string>>, comment:string?}|null}
+	 * @returns {{testSource: string, expectedTokenStream: Array<Array<string>>, comment:string?}|null}
```
ExE-Boss:
```suggestion
	 * @returns {{testSource: string, expectedTokenStream: string[][], comment:string?}|null}
```

tests/helper/test-discovery.js
```diff
@@ -51,7 +51,7 @@ module.exports = {
 	 * in the given src directory
 	 *
 	 * @param {string} src
-	 * @returns {Array.<string>}
+	 * @returns {Array<string>}
```
ExE-Boss:
```suggestion
	 * @returns {string[]}
```

tests/helper/test-discovery.js
```diff
@@ -67,7 +67,7 @@ module.exports = {
 	 *
 	 * @param {string} src
 	 * @param {string|string[]} languages
-	 * @returns {Array.<string>}
+	 * @returns {Array<string>}
```
ExE-Boss:
```suggestion
	 * @returns {string[]}
```

tests/helper/test-discovery.js
```diff
@@ -99,7 +99,7 @@ module.exports = {
 	 *
 	 * @private
 	 * @param {string} src
-	 * @returns {Array.<string>}
+	 * @returns {Array<string>}
```
ExE-Boss:
```suggestion
	 * @returns {string[]}
```

RunDevelopment:
That's a good idea. I'll change all similar occurrences.

components/prism-core.js
```diff
@@ -447,18 +588,65 @@ var _ = {
...
 	Token: Token
 };
-
 _self.Prism = _;
 
+/**
+ * Creates a new token.
+ *
+ * @param {string} type See {@link Prism.Token#type type}
+ * @param {string | Prism.Token | Array<string|Prism.Token>} content See {@link Prism.Token#content content}
```
ExE-Boss:
```suggestion
 * @param {Prism.TokenStream} content See {@link Prism.Token#content content}
```

components/prism-core.doc.js
```diff
@@ -0,0 +1,37 @@
...
+ * This can be used for recursive language definitions.
+ *
+ * Note that this can cause infinite recursion.
+*/
+
+/**
+ * @typedef Grammar
+ * @type {Object<string, RegExp | TokenObject | Array<RegExp | TokenObject>>}
+ * @property {Grammar} [rest] An optional grammar object that will appended to this grammar.
+ */
```
ExE-Boss:
```suggestion
 * @memberof Prism
 */

/**
 * @typedef TokenStream A stream of tokens.
 * @type {string | Prism.Token | Array<string | Prism.Token>}
 * @memberof Prism
 */
```

components/prism-core.js
```diff
@@ -447,18 +588,65 @@ var _ = {
...
+	 *
+	 * This is usually the key of a pattern in a {@link Grammar}.
+	 * @member {string}
+	 */
 	this.type = type;
+	/**
+	 * The strings or tokens contained by this token.
+	 *
+	 * This will be a token stream if the pattern matched also defined an `inside` grammar.
+	 * @member {string | Prism.Token | Array<string|Prism.Token>}
```
ExE-Boss:
```suggestion
	 * @member {Prism.TokenStream}
```

components/prism-core.js
```diff
@@ -447,18 +588,65 @@ var _ = {
...
 	this.greedy = !!greedy;
 }
 
+/**
+ * Converts the given token or token stream to an HTML representation.
+ *
+ * The following hooks will be run:
+ * 1. `wrap`: On each {@link Prism.Token}.
+ *
+ * @param {string | Prism.Token | Array<string|Prism.Token>} o The token or token stream to be converted.
```
ExE-Boss:
```suggestion
 * @param {Prism.TokenStream} o The token or token stream to be converted.
```

RunDevelopment:
I actually also had a `TokenStream` type at one point but discarded it because of two reasons:

1. I want to keep it simple and the `TokenStream` type has essentially zero value documentation wise.
    This is about documentation, not types (though, both are related of course).
2. The generated doc-website doesn't have a nice hover-for-type-definition-summary feature. This means that the `TokenStream` type in a e.g. function definition will hide the actual type information until one clicks the link to go to the definition.
    So instead of adding `TokenStream`, which makes the code a little cleaner, I choose to write its definition again and again to make the doc easier to read.

components/prism-core.doc.js
ExE-Boss:
```suggestion
 * @memberof Prism
 */
```

RunDevelopment: I didn't want this to be a member of Prism because of TSC.

LeaVerou:
What happened here? It seems this was just abandoned, which is a shame.

RunDevelopment:
Approval pending happened.

And also a change of heart in myself. I tried [TSDoc](https://github.com/microsoft/tsdoc) and ohh boi it's not as good as you think it is. Both TSDoc and JSDoc are a pain to setup and use.

<details>
<summary>JSDoc vs TSDoc for Prism</summary>

While JSDoc is kinda nice when you want to get a nice documentation page out, it can't do anything else. JSDoc type annotations are compatible with TS but that's it. Everything else like namespaces and membership is not supported by TS, so the namespaces of documentation don't work in code. Luckily, everything in Prism is in the `Prism` namespace, so nothing (except trivial stuff) works. 
You need these annotations because for JSDoc everything is declared in global scope unless declared otherwise and if everything is global, you don't get the nice structure you want which makes JSDoc useless. So if we really use JSDoc, we break with TS which I do NOT want. I love TS.

So what about TSDoc? Well, if you don't write everything in TS, it's almost useless. It can parse JS files like JSDoc but it doesn't understand the annotations which make JSDoc useful, so everything is global. You now have two choices: 1) convert to TS or 2) use declaration files. 1) is way to much work just for documentation, so 2) it is. The problem with declaration files is twofold. 
Firstly, your types and documentation are now far away from the actual code, so they can easily become outdated. And secondly, you know have to fight the TypeScript ecosystem. 
I love TS and its strict compiler but when you venture outside Typescript's usual system of static imports (of whatever kind) and dare to use features like [ref comments](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) it's not a pleasant experience. If we just don't check the types (why have them then?), the types itself are a problem too. 
Prism has a few things which cannot be expressed in TS's type system. A good example are Prism grammars. You can't express them in TS because of `rest` properties. It's not possible to declare an object in which all keys have a certain type except for one key that has a different one. You cannot do it (related: [this](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0d04fc967106fe48895d1dfdcbde070072a1b153/types/prismjs/index.d.ts#L172) simply doesn't work). Tbh, most of these problems are suboptimal design on our side (e.g. why can't you have a token named `rest`? Similar for a language named `extend`).

You can make TS work with Prism but we have to make 2 breaking changes:

1. Remove `rest`. I plan to do that someday anyway. The spread syntax is a thing after all and `rest` also has other problems.
2. Move the `Prism.languages` functions. I have no idea why `DFS` is not in `util` but `extend` and `insertBefore` also have to be moved.

We can talk about these changes some other day. I just wanted to show that full types, while impossible right now, are not far off.

---

</details>

> It seems this was just abandoned

That's because I kinda did abandon JSDoc. 

I want Prism to move closer TS and a major step toward that goal was TSDoc. The problem: I failed miserably one 3 separate occasions. I just couldn't find a way to make it work with Prism without basically rewriting 50% of the library.

I'm now at the point of giving up on TSDoc for now. So yeah... 

JSDoc:
It's still a pain to use, I mean, I even had to hack in something as basic as determinism and not even the code display works out-of-the-box. Also, almost all changes to any file with documentation will cause the generated website to change because the line numbers of function/classes/types change.

However, my main problem is this: JSDoc can generate some nice-looking docs but it messes with TS in all the wrong ways. It annoys me to no end that I have to decide whether I want type annotations that actually work and help me code (TS) or whether I want type annotations that generate good doc. __You can't have both__! (note: Maybe you can right now? This was the state of things one year ago but I doubt that it changed and put simply: I'm not willing to put any more days (yes, days) of my life into marrying JSDoc and TS. If there is a reliably way, I'll do it but I won't put any more hours into finding that way.)

So let's get this over with: __Before I touch this again, I want a bit of clarity. What do we want?__

I say it as it is: I want TS and JSDoc blocks that path as it is now. 
I want to slowly move Prism, so that one day the whole codebase is still JS but most strongly typed via comments so that the TS can check everything. There already have been a few occasions where TS would have pointed out a bug but that didn't happen, so some user reported it instead.

LeaVerou:
Thank you! I was considering JSDoc for another project, and this further reassures me I made the right choice deciding against it.

RunDevelopment:
Ok, so no JSDoc. That's good.

But also bad because some way of publishing our documentation would be great. When I made my research a year, there were some competitor but all of them were basically abandoned at this point in time. Are there any other static site documentation generators or something else we can use?

LeaVerou:
Do we want generation from source or a separate app?

I can see the DRY argument for docs from source, but it can also be quite restrictive in a number of ways.
Since the argument for generating from source is obvious, I'm going to play devil's advocate here and argue keeping docs separate from comments, but don't assume that's my position. I'm fine with whatever you folks decide :)

- Often the kind of docs one wants in source are not the docs one wants on a website. It can be extremely tedious to scroll through several paragraphs of documentation comments to spot the code, so there's a tendency to keep them terser and with fewer examples than if they were separate.
- Comments are not meant for long-form text of that sort. E.g. line breaking needs to be done manually and is quite a pain to align multiline documentation for params (especially when documenting object literal params)
- Generators tend to follow a specific, more rigid syntax and a more formulaic output, whereas with a separate app, the sky is the limit, with interactive filters, custom structured data, live demos etc.

RunDevelopment:
The counterexample here would be [Rust](https://doc.rust-lang.org/stable/rust-by-example/meta/doc.html). They even do simple tests in doc comments. It's actually pretty awesome.

> or a separate app

What can app X do that our website can't do?

Maybe the CS student with limited experience shouldn't talking like he knows stuff but:

From what I see we need 2 types of documentation. The first is the one we have right now. It's the short documentation that tells how you use the library without going into the nitty-gritty details that you only need to know when you want to dive deep. This also includes docs for our development setup and everything else that can't be a comment in our code.
The second kind is the one we lack. It's the detailed documentation you need when you can't go the easy way (e.g. Prism's automatic highlighting) and have to do things yourself. This should include a detailed description of every public and internal function/class/constant.

For the first kind, a website like we already have is perfect. Elaborate articles and demo. You can do it all.
For the second kind, _I think_ static site generation from doc comments is ideal. Sure, it's limited but you don't need interactive demos in technical documentation (at least, I never did). The problem with those generators is outlined in the comment above.

> Comments are not meant for long-form text of that sort.

I disagree with that. Sure, your next novel shouldn't be published as a doc comment but up to 200 lines are fine IMO. Everything beyond that should probably be an article for the first kind of documentation anyway.

Also, what you then describe seems to be a general lack of comment formatters in JavaScript (after a quick google search, I found none). Java has excellent comment formatters with both the Eclipse IDE and IntelliJ. In C/C++ you can count on clang-format for at least reflow your comments which is usually all you need.

(Looking at from a different perspective: It's the state of JS tooling when it comes to documentation pretty bad? Shouldn't documentation be a major part of writing code? Especially since you don't have types that by there name alone do have the documenting in other languages.)

---

How do others do this? 
I've never seen a tools like JSDoc being used in the wild. In languages like Java and C/C++ you see these static sites generated from comment for many projects.

Also, I think I'll look into JSDoc once more. There seems to be some recent activity on their GitHub, so hopefully, things improved a little.

RunDevelopment:
Ok. It's done.

After [this comment](https://github.com/PrismJS/prism/issues/1777#issuecomment-649838950), I gave it another try. This time with the goal of just getting something that works even if it barely works.

The main problem with marrying TS and JSDoc was that TS couldn't do namespaces and wasn't smart enough to figure out some things about the structure of Prism instances. That was a year ago. TS has since become a bit smarter and we had a PR that made it easier for TS to figure out Prism. So TS could now understand Prism. One problem solved.
Now namespaces. Screw namespaces! I just didn't use them. Without using namespaces, the generated doc is somewhat odd because all classes and type definitions are part of the global namespace. This isn't really scalable documentation-wise but TS can understand it, so sure why not. We only have 5 types in global space right now, so it isn't a big deal.

After some tinkering and hacking, I finally had something that worked with both TS and JSDoc.
I also used a different theme that is a lot more configurable and just plain better than the previous one.

It finally works and it even looks kinda nice. [Demo](https://rundevelopment.github.io/docs/index.html).

__The hacking__

I still have a small CSS file to change/fix the theme but it's better now.

The real hacking comes with JSDoc's parser. It can't deal with some TS syntax (e.g. intersection types and type imports) and will throw an error. 
To fix that, I added a handler for JSDoc parser events that will just remove all type imports via a small regex and hopes that it works then. This is nothing more than a hack but it works and I don't plan to change anything about it because the JSDoc team doesn't intend to support this syntax, so this is the best I can do.

__TS__

One of the main intentions was to get good type checking, so did that turnout?

Amazing. It works really well. You have to kinda work around some issue but it's doable. My goal of moving Prism toward strict typing via doc comments is almost fulfilled. There are only 2 type problems left in Core. For that alone, I want to see this merged.

__Other nice things__

Documentation is opt-in. You can add doc comments however you like and for whatever you like. Only if you mark a doc comment as `@public` will it be part of the generated documentation site.

Thanks to the new theme, we now have a search bar and we can add your own links to the nav.

__Not so nice things__

The doc site includes a "jump to source" function. This means that every change to Core will result in changes in the doc site. To mitigate this a little, the `gulp docs` task is not part of the default build task. If you make changes to the doc, you have to rebuild the site manually. Not nice, but not too bad either.

__Closing thoughts__

It's not perfect but it works. You can write some doc and it will generate a nice site for you. 

I criticize JSDoc and my solutions to all the problems I had to deal with a lot but I'm actually really happy with this result.

Let me know what you think. 
Awaiting review.

components/prism-core.js
```diff
@@ -114,8 +180,8 @@ var _ = {
 			if (typeof document === 'undefined') {
 				return null;
 			}
-			if ('currentScript' in document) {
-				return document.currentScript;
+			if ('currentScript' in document && 1 < 2) {
```
mAAdhaTTah, line 183: Why is this extra conditional here?

components/prism-core.js
```diff
@@ -181,7 +247,41 @@ var _ = {
...
+		 * @param {string} id The id of the language to extend. This has to be a key in `Prism.languages`.
+		 * @param {Grammar} redef The new tokens to append.
+		 * @returns {Grammar} The new language created.
+		 * @public
+		 * @example
+		 * Prism.languages['css-with-colors'] = Prism.languages.extend('css', {
+		 *     // this will overwrite the 'comment' token of CSS at its original position
+		 *     'comment': { ... },
+		 *     // this will be appended
+		 *     'color': /\b(?:red|green|blue)\b/
```
mAAdhaTTah, line 282: I think these examples need some explanation in the prose. Why is `color` appended & `comment` overridden?

components/prism-core.js
```diff
@@ -395,6 +675,15 @@ var _ = {
 			hooks[name].push(callback);
 		},
 
+		/**
+		 * Runs a hook invoking all registered callbacks with the given environment variables.
+		 *
+		 * Callbacks will be invoked synchronously and in the order in which they were registered.
+		 *
+		 * @param {string} name The name of the hook.
+		 * @param {Object<string, any>} env The environment variables of the hook passed to all callbacks registered.
```
mAAdhaTTah, line 684: Can we define the `env` structure more explicitly?

RunDevelopment: No. You can do custom hooks (see Line numbers), so we technically can't even assume that `env` is an object.

components/prism-core.js
RunDevelopment: Ahhh. My bad. TS thinks that `document` always has a `currentScript` but this means that the below IE code is dead code, so TS displays an error. To remove this "error", I tricked TS' flow analysis with `1 < 2`. Obviously, `1 < 2` will always be true, but TS isn't smart enough to figure that out, so TS thinks that the IE code might be executed. Tada! No error.

(Uglify will even optimize this away.)

components/prism-core.js
RunDevelopment: I'll add something.

On that note: ["override" or "overwrite"](https://stackoverflow.com/a/8651622/7595472)?

package.json
```diff
@@ -6,6 +6,7 @@
   "style": "themes/prism.css",
   "scripts": {
     "build": "gulp",
+    "generate-docs": "gulp docs",
```
mAAdhaTTah, line 9: Maybe?

```suggestion
    "docs:generate": "gulp docs",
```

mAAdhaTTah: Actually!
```suggestion
    "build:docs": "gulp docs",
```

RunDevelopment: I went with `generate-docs` because that's [what our theme suggested](https://github.com/clenemt/docdash#usage-npm).

mAAdhaTTah:
In another project, I actually wrote everything in JS and managed to generate `.d.ts` files from them. If we're going to put all this work into writing & maintaining the documentation, is it possible to do that here? Not a requirement for this PR, by any means, just throwing it out there.

components/prism-core.js
mAAdhaTTah: "overwrite", I think. We're fully replacing the object, yeah?

RunDevelopment: We do. Then I'll stick with "overwrite".

package.json
RunDevelopment: Actually, `build:docs` was my first thought as well. I decided against it because people might think that they also need to run the `build:docs` task if they were told to "rebuild Prism".

RunDevelopment:
> is it possible to do that here?

Should be possible. Let's make an issue of this after this PR is through.

components/prism-core.js
RunDevelopment: I'll just add a small comment to explain this.

package.json
mAAdhaTTah: I don't think that confusion is super likely & isn't terrible if it does happen, and `build:docs` is more consistent with how we define other tasks.

RunDevelopment: Should it be part of the default build task then? (To be consistent with `test` and `test:*`.)

components/prism-core.js
```diff
@@ -410,18 +700,86 @@ var _ = {
...
+	 * @see GrammarToken
+	 * @public
+	 */
 	this.type = type;
+	/**
+	 * The strings or tokens contained by this token.
+	 *
+	 * This will be a token stream if the pattern matched also defined an `inside` grammar.
+	 *
+	 * @member {string | Token | TokenStream}
```
RunDevelopment, line 739: Should we narrow this down to `string | TokenStream`? (#2071)

package.json
mAAdhaTTah: Doesn't have to be, imo.

mAAdhaTTah:
I'm generally ok with punting off defining TokenStream to a future PR and not blocking this on it.

package.json
RunDevelopment: It might be better to have it be part of the default build task. I really want to avoid the case whether somebody makes some changes to `prism-core.js` (or any of the file from which the doc is generated from) and then the next person wants to make a PR, runs the `build:docs` task and then has to ask whether they should also commit that as well.

Let's make part of the default build task. If it's too slow or causes problems otherwise, we can kick it out later.

RunDevelopment:
Well, if we really narrow down the definition of a Token after this, then that's a breaking change.

Actually, let's just take the current definition of Token and TokenStream as they are in here right now. This will resolve #2071.

mAAdhaTTah:
I'm definitely on board with these API docs in general. I would love to eventually get to a world where we generate both types of docs into a single cohesive website, but this works for now.

RunDevelopment:
I should probably point out that we now have API doc, but I haven't actually linked to it yet... Let just me do that real quick.

RunDevelopment:
Minor change: I narrowed down the type definition of a `Token` ([reason](https://github.com/PrismJS/prism/pull/2070#issuecomment-650648282)).
I'm gonna merge now.

PR #1776, opened 2019-02-27
Core: VSCode friendly

RunDevelopment:
This PR makes the `prism-core.js` more compatible with editors like VSCode.
VSCode will now be able to provide useful suggestions:

image

I also made `_self` a parameter for better code compression.

PR #1783, opened 2019-03-01
refactor(gulp): Use `async` functions

ExE-Boss:
Depends&nbsp;on https://github.com/PrismJS/prism/pull/1780 and [dropping Node&nbsp;<7.6.0](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/async_function#Browser_compatibility).


RunDevelopment:
I was also thinking about using async functions when writing #1780 but didn't because [Node v6 isn't dead yet](https://github.com/nodejs/Release#release-schedule) even though it will be very soon.

So I really don't know if we want to drop Node v6 just to have a nicer build process.
I don't think it's worth it.

RunDevelopment:
Let's just keep this open for a month or two until Node v6 is dead.

@ExE-Boss You might want to remove Node v6 from [`.travis.yml`](https://github.com/PrismJS/prism/blob/master/.travis.yml#L3) to make the build succeed.

ExE-Boss:
In the meantime, you should merge #1780 and&nbsp;#1776.

ExE-Boss:
**Travis** was acting&nbsp;up, closing and reopening solves&nbsp;it in&nbsp;most cases.

gulpfile.js
```diff
@@ -6,6 +6,7 @@ const header = require('gulp-header');
 const concat = require('gulp-concat');
 const replace = require('gulp-replace');
 const pump = require('pump');
+const util = require('util');
```
RunDevelopment: Is this some kind of default package or did you forget to commit your `package.json`?

ExE-Boss: This is&nbsp;a&nbsp;built&#x2011;in&nbsp;API: https://nodejs.org/api/util.html

gulpfile.js
```diff
@@ -77,100 +78,86 @@ function build(cb) {
...
-
-			stream.on('error', done);
-			stream.on('end', done);
-		}
-	});
+	await Promise.all(tasks.map(task => new Promise(resolve => {
+		const stream = src(task.plugin)
+			.pipe(replace(/\/\*languages_placeholder\[\*\/[\s\S]*?\/\*\]\*\//, '/*languages_placeholder[*/' + task.map + '/*]*/'))
+			.pipe(dest(task.plugin.substring(0, task.plugin.lastIndexOf('/'))));
+		stream.on('error', resolve);
```
mAAdhaTTah: Should we `reject` on `error`?

ExE-Boss: This is the current behaviour.

RunDevelopment: The current behavior is probably a bug, so why not fix it.

RunDevelopment:
@ExE-Boss 
Node v6 is (finally) dead, so we can merge this now after the conflicts are resolved and the last comment is addressed.

RunDevelopment:
@ExE-Boss Is this still active?
If you're too busy right now, I'll gladly take over. Just let me know.

Same for #1796.

PR #1788, opened 2019-03-02
GraphQL improvements

### LINKED ISSUE #1786: Wrong GraphQL tokenization
First of all, the GraphQL Support (#971) focus only on the client queries side, not the schema definition in SDL.

So there are a lot of missing keywords:

> schema, type, input, scalar, union, enum, interface, implements 

https://graphql.org/learn/schema/

Also, it tokenize the fields names as `attr-name` and not tokenize the field names with args:

image

image

It would be great if you could add support for it @Golmote 💪

RunDevelopment:
This PR resolves #1786 

Changes to the GraphQL language definition:

1. Added missing keywords.
2. Fixed field names with arguments. (Small correction of #1711)
3. Added missing `|` operator for unions.
4. Added highlighting of fragments.
5. Added highlighting for class-names after type definition keywords.
6. Added constants. These match all identifiers which do not contain lower case letters (`ID` is the only exception here). This is intended to highlight enum values.

## Examples

<details>

Better `attr-name`:
image

New: Highlighting of fragments
image

New: Highlighting of constants/enum values (detected using naming conventions)
image

</details>


19majkel94:
@RunDevelopment 
Is there any ETA or some blockers for this feature?
I would love to release it ASAP and update Prism in Docusaurus for my docs page 😃

PR #1793, opened 2019-03-06
JSONP highlight: Fixed minified adapter names

### LINKED ISSUE #1673: JSONP highlight and UglifyJS
Right now, we use UglifyJS to compress our `.js` files. 

UglifyJS also replaces the names of most function with a new shorter name to decrease file size.
But this is a problem because it changes that function's `.name` property upon which JSONP Highlight's [`getAdapter`](https://github.com/PrismJS/prism/blob/master/plugins/jsonp-highlight/prism-jsonp-highlight.js#L15) relies.

As a consequence, `Prism.plugins.jsonphighlight.removeAdapter("github")` will remove the `github` adapter when using the development version, but it doesn't do anything when using the minified version.

### Possible solutions

1. We could set `keep_fnames: true` in the mangle options of UglifyJS but this will also blow up the file size of (almost) every other file with named functions.
2. Change the definition of `registerAdapter` to take an optional name, like so:
    ```js
    function registerAdapter(adapter, name) {
    	name = name || adapter.name;
    	...
    }
    ```
   We would have to change the underlying adapter list as well, but that's not a problem because we don't expose it.
3. Change the definition of `registerAdapter` to require a name similar to `Prism.hooks.add`. 
    This is a _breaking change_.

RunDevelopment:
This PR fixes #1673.


plugins/jsonp-highlight/prism-jsonp-highlight.js
```diff
@@ -1,54 +1,72 @@
...
-	 * @param {string|Function} adapter The adapter itself or the function name of an adapter.
-	 * @returns {(response: any, pre?: HTMLPreElement) => string} A registered adapter or `null`.
+	 * @param {string|Function} adapter The adapter itself or the name of an adapter.
+	 * @returns {Adapter} A registered adapter or `null`.
 	 */
 	function getAdapter(adapter) {
 		if (typeof adapter === "function") {
-			return adapters.filter(function (fn) { return fn.valueOf() === adapter.valueOf(); })[0];
+			for (var i = 0, item; item = adapters[i++];) {
+				if (item.adapter.valueOf() === adapter.valueOf()) {
```
mAAdhaTTah: What's the backstory on `valueOf` here?

RunDevelopment: No idea, I just kept it as is.

Issue #1578, opened 2018-10-02
When can we drop IE11 support?


LeaVerou:
Once we can drop IE11, we will be able to depend on CSS variables.
This means we can finally get rid of the whole problem "how do all themes work with all languages?". Themes would just define variables, language definitions would just use these variables.

What do you think?

mAAdhaTTah:
I'm personally OK w/ dropping IE11 support, but I know projects I work on generally still support it as MSFT is still supporting that browser. Additionally, IE11 still works w/ Windows 10, so MSFT may continue to support it for a while.

This is also definitely a breaking change, so we should release as a 2.0. If we drop IE11 support, we could probably go all-in on ES6 as well, if that helps the code at all.

Just some thoughts. Not really sure where I stand, tbh.

ExE-Boss:
We could maybe support CSS variables and use [`postcss-custom-properties`](https://github.com/postcss/postcss-custom-properties) for rudimentary IE11&nbsp;support.

LeaVerou:
I'm wondering if we could start using CSS variables without completely dropping IE11 support. We could use PostCSS or something to add fallbacks, and IE11 would just use those fallbacks. So people on IE11 would still get highlighting, just not customized highlighting. Would that make sense?

RunDevelopment:
[Microsoft is retiring IE11 in June **2022**](https://blogs.windows.com/windowsexperience/2021/05/19/the-future-of-internet-explorer-on-windows-10-is-in-microsoft-edge/).

joshgoebel:
Are you planning to support it until then?

RunDevelopment:
Given the current progress on v2, I can't blame you for thinking that but no :)

mAAdhaTTah:
Technically, dropping IE11 support is a breaking change, so we really _should_ support it until then, unfortunately.

mAAdhaTTah:
Unless I misunderstood the comment/question. To rephrase, v1 should support IE11. v2 was never intended to. So really, we'll support IE11 until whenever v2 gets out. Or at least, that's what I think we should do.

RunDevelopment:
> To rephrase, v1 should support IE11. v2 was never intended to.

Yup, that's pretty much it. We'll support IE11 for the whole of v1.x and then drop it in v2.0.0.

ljharb:
"users stop using it" is a date that is not tightly correlated to "a company stops supporting it".

LeaVerou:
[...]

PR #1795, opened 2019-03-07
Tests: Refactoring

RunDevelopment:
I refactored all test related files to use ES6 features.


mAAdhaTTah:
My only concern about this is it's a little weird to only allow ES6 in the tests but not in the core of the library. Maybe we can enforce this with ESLint?

RunDevelopment:
Why is this weird? Everything made exclusively for a node environment is ES6 and the code which is also available for browser is ES5.

Well, we could use Babel to solve this and make everything ES6 or we could wait for #1578.

> Maybe we can enforce this with ESLint?

Enforce what?

mAAdhaTTah:
I'm pretty sure we can enforce es5 env in the core / languages / plugins area and enable es6 env in the tests with ESLint.

mAAdhaTTah:
It's weird mostly cuz, as a developer, I have to remember "I'm in core, I can only use ES5" to "now I'm in tests, I can use ES6+".

RunDevelopment:
We enforce ES5 for core / languages / plugins already because Uglify doesn't support ES6.

PR #1796, opened 2019-03-07
chore(package): Add ESLint

ExE-Boss:
This adds a&nbsp;basic ESLint configuration to&nbsp;the&nbsp;project which doesn’t require changing almost every file like&nbsp;#1483 (the&nbsp;goal is&nbsp;to&nbsp;slowly enable&nbsp;rules across&nbsp;multiple&nbsp;PRs).


.eslintrc
```diff
@@ -0,0 +1,91 @@
...
+		"no-empty": "off",
+		"no-inner-declarations": "off",
+		"no-mixed-spaces-and-tabs": "error",
+		"no-redeclare": "off",
+		"no-sparse-arrays": "off",
+		"no-unused-vars": [
+			"warn",
+			{
+				"args": "none",
+				"argsIgnorePattern": "^_",
```
RunDevelopment: We don't check arguments anyway, so we don't need this pattern.

.eslintrc
```diff
@@ -0,0 +1,91 @@
...
+				"varsIgnorePattern": "^_",
```
RunDevelopment: There are no unused variables which match this pattern.

gulpfile.js
```diff
@@ -61,6 +62,13 @@ function minifyJS() {
 	];
 }
 
+function lint(cb) {
+	return pump([
+		src(['**/*.js', '!node_modules/**'], { base: './' }),
+		eslint(),
+		eslint.format(),
+	], cb);
```
RunDevelopment: Do you really want to not [fail on lint errors](https://github.com/PrismJS/prism/blob/662bbfe5828a0160d2035940469f8a9339fc3f22/gulpfile.js#L53)?

ExE-Boss: I&nbsp;accidentally forgot to&nbsp;add&nbsp;it.

.eslintrc
ExE-Boss: There&nbsp;is&nbsp;now:

https://github.com/PrismJS/prism/blob/3fb78077e23ddc2cdd102506d1a43735764e329e/code.js#L21

RunDevelopment: Hmmm, unused variables are usually a sign that something went wrong during refactoring or writing, so why would we want to ignore these?

(`_`)`after` is a good example: Why is it there? Does it do anything?
`nextSibling` should be side-effect free, so why _not_ remove it. I don't see the benefit in ignoring a potential mistake because of its name.
(I mean, even after the ESLint warning disappears, I have like 3 other linters which all scream at me to remove it...)

Also, even after you remove the ignore pattern, `after` stays the only unused variable.

.eslintrc
```diff
@@ -0,0 +1,100 @@
...
+				"Promise": "readonly",
+				"components": "readonly",
+				"setTheme": "readonly",
+				"PrefixFree": "readonly",
+				"StyleFix": "readonly",
+			},
+			"rules": {
+				"no-cond-assign": "off",
+				"no-mixed-spaces-and-tabs": "off",
+				"semi": "off",
```
RunDevelopment: Why did you turn this off?

(There are 2 missing semicolons in these files.)

ExE-Boss: Ok, I’ll&nbsp;fix&nbsp;it.

I&nbsp;originally tried to&nbsp;avoid changing `code.js` and&nbsp;`downloads.js` too&nbsp;much.

.eslintrc
```diff
@@ -0,0 +1,100 @@
...
+			"env": {
+				"es6": true,
+			},
+			"parserOptions": {
+				"ecmaVersion": 2015,
+			},
+			"rules": {
+				"comma-dangle": [
+					"error",
+					"always-multiline",
```
RunDevelopment: Why did you change the behavior here?

.eslintrc
```diff
@@ -0,0 +1,100 @@
...
+				],
+			},
+		},
+		{
+			"files": [
+				"code.js",
+				"download.js",
+				"examples.js",
+				"utopia.js",
```
RunDevelopment: I'm not really sure if we should lint this file. It feels more like a library to me.

@LeaVerou Isn't this the predecessor to [Bliss](https://github.com/LeaVerou/bliss)?

ExE-Boss: Yes, this is the predecessor to Bliss, but it’s not a 100% drop&#x2011;in replacement.

.eslintrc
ExE-Boss: Because `gulpfile.js` already has traling commas in&nbsp;most places.

RunDevelopment: But do we really want to force people to use commas because of that?
Also, isn't ESLint support to make the code style consistent?

.eslintrc
RunDevelopment: Wait. Don't we ignore utopia anyway? (`.eslintignore`)

.eslintrc
RunDevelopment: I think we can just remove the unused variable.

.eslintrc
ExE-Boss: Right, we&nbsp;do…

.eslintrc
ExE-Boss: I eventually want to enforce <code>[comma‑dangle](https://eslint.org/docs/rules/comma-dangle): "always‑multiline"</code> everywhere.

RunDevelopment: I can see why you want this but it literally requires us to change (almost) every language definition.

I'm personally quite a fan of `"only-multiline"`.
We should get some more opinions on this.

.eslintrc
RunDevelopment: Can't we remove the ignore pattern now or is there a reason to keep it?

code.js
```diff
@@ -198,7 +198,7 @@ function listPlugins(ul) {
 		$u.element.create('li', {
 			contents: {
 				tag: 'a',
-				prop: {
+				properties: {
```
RunDevelopment: Nice catch!

components/prism-core.js
```diff
@@ -1,7 +1,7 @@
+/* global WorkerGlobalScope */
```
RunDevelopment: I get that it's convenient to just declare it here but wouldn't it be better to make an overwrite in `.eslintrc` for `components/prism-core.js` and `prism.js` to add this global variable?

examples.js
```diff
@@ -166,8 +166,8 @@ function update(id) {
  * Loads a language, including all dependencies
  *
  * @param {string} lang the language to load
- * @type {Promise} the promise which resolves as soon as everything is loaded
- */
+ * @returns {Promise<Event|void>} the promise which resolves as soon as everything is loaded
+s */
```
RunDevelopment: s?

examples.js
```diff
@@ -166,8 +166,8 @@ function update(id) {
  * Loads a language, including all dependencies
  *
  * @param {string} lang the language to load
- * @type {Promise} the promise which resolves as soon as everything is loaded
- */
+ * @returns {Promise<Event|void>} the promise which resolves as soon as everything is loaded
```
RunDevelopment: Missing `.` after `Promise`?

package.json
```diff
@@ -23,8 +23,10 @@
   },
   "devDependencies": {
     "chai": "^4.2.0",
+    "eslint": "^5.15.1",
```
RunDevelopment: Isn't just `gulp-eslint` enough? Why do we also need `eslint` itself?

ExE-Boss: Because `gulp‑eslint` declares `eslint` as&nbsp;a&nbsp;`peerDependency` (or&nbsp;at&nbsp;least it&nbsp;did when I&nbsp;last checked)

examples.js
ExE-Boss: Typo

examples.js
ExE-Boss: It’s unnecessary, also I use the [TypeScript style syntax for JSDoc types](https://github.com/Microsoft/TypeScript/wiki/JsDoc-support-in-JavaScript).

package.json
RunDevelopment: [ES-lint is a normal dependency](https://github.com/adametry/gulp-eslint/blob/master/package.json#L43) now.

examples.js
RunDevelopment: Seems like JSDoc is also able to handle this syntax.

gulpfile.js
```diff
@@ -184,14 +193,14 @@ function changelog(cb) {
...
 }
 
 const components = minifyComponents;
 const plugins = series(languagePlugins, minifyPlugins);
 
-
+exports.lint = lint;
 exports.watch = watchComponentsAndPlugins;
-exports.default = parallel(components, plugins, componentsJsonToJs, build);
+exports.default = series(parallel(components, plugins, componentsJsonToJs, build), lint);
```
RunDevelopment: Why can't we do it in parallel?

ExE-Boss: I’ve made the&nbsp;`lint` task depend on&nbsp;everything in&nbsp;`default`, since if&nbsp;the&nbsp;`prism.js` file hasn’t finished building yet and the&nbsp;`lint` task tries to&nbsp;lint&nbsp;it while it’s&nbsp;being&nbsp;built, it will cause the&nbsp;build to&nbsp;fail.

code.js
ExE-Boss: The `prop` alias is&nbsp;only available as&nbsp;a&nbsp;fallback when passed to&nbsp;the&nbsp;`$u.element.set(…)`&nbsp;method (it’s&nbsp;ignored&nbsp;in&nbsp;`$u.element.create(…)`, because `create(…)` always creates a&nbsp;`properties`&nbsp;key).

LeaVerou:
As I have expressed in the other issue, I disagree with any linting being classed as "error". It should be warnings, and errors should be reserved for actual JS errors. A wrong indent is not the same importance as a JS parse error that breaks the whole script.

.eslintrc
LeaVerou: I'm not a fan of dangling commas. I see the benefit in multiline cases, but they are also less natural and I think optimizing our code for tooling is a slippery slope. But my objection is not so strong that is should hold this up if the rest of you agree. If you don't however, do count it as an objection.

mAAdhaTTah:
@LeaVerou Is that to suggest we can merge in PRs with warnings? We could switch warnings -> errors in CI if not, so they're just warnings during development.

LeaVerou:
Oh I wasn't even thinking about CI. I primarily find ESLint useful as an authoring environment plugin that highlights errors and warnings inline. I would find it frustrating if it highlighted syntax errors and stylistic issues the same way.

gulpfile.js
mAAdhaTTah: We probably shouldn't lint `prism.js`, as that's a generated file. That said we should probably not build if `lint` errors, right?

mAAdhaTTah:
Sounds good, so we should bump warnings -> errors in CI and downgrade style issues to warnings during development.

ExE-Boss:
And how do you propose to do that?

mAAdhaTTah:
[This might work](https://github.com/eslint/eslint/issues/2309#issuecomment-219828044). Another suggestion is to use a `.js` file for config and use env vars. There's a package called `is-ci` that could help here.

RunDevelopment:
We could also check the output of [eslint](https://www.npmjs.com/package/gulp-eslint#eslintresultsaction) and to if we got any warnings.

components/prism-core.js
```diff
@@ -549,5 +548,5 @@ if (typeof module !== 'undefined' && module.exports) {
 
 // hack for components to work correctly in node.js
 if (typeof global !== 'undefined') {
-	global.Prism = Prism;
+	global['Prism'] = Prism;
```
mAAdhaTTah: Not a fan of whatever's causing this. I saw this below as well.

RunDevelopment: That's a dirty hack to remove an error message from [TSCheck](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html).

Not a fan either btw.

.eslintrc
mAAdhaTTah: I kinda like dangling commas, but in this case, I think the churn on the language definitions makes it not worth it.

I'd also prefer not to have a different rule for the `gulpfile.js` but instead make the changes in the `gulpfile.js` to bring it in line with the rest of the codebase.

.eslintrc
```diff
@@ -0,0 +1,113 @@
...
+					"always-multiline",
+				],
+			},
+		},
+		{
+			"files": [
+				"code.js",
+				"download.js",
+				"examples.js",
+			],
```
mAAdhaTTah: I believe these files will need to change as a result of #1781.

ExE-Boss: Yeah, they will…

.eslintrc
mAAdhaTTah: Agreed, remove the ignore pattern.

.eslintrc
```diff
@@ -0,0 +1,113 @@
...
+			"only-multiline",
+		],
+//		"indent": [
+//			"error",
+//			"tab",
+//			{
+//				"SwitchCase": 1,
+//				"outerIIFEBody": 0,
+//			}
+//		],
```
mAAdhaTTah: There are a couple commented out rules that should be cleaned up.

.eslintignore
```diff
@@ -0,0 +1,4 @@
+*.min.js
+/components.js
+/utopia.js
+/vendor/**
```
mAAdhaTTah: This file will also need to change because of #1781.

tests/helper/test-case.js
```diff
@@ -30,9 +30,6 @@ var TokenStreamTransformer = require("./token-stream-transformer");
  * If the file contains more than three parts, the remaining parts are just ignored.
  * If the file however does not contain at least two parts (so no expected token stream),
  * the test case will later be marked as failed.
- *
- *
- * @type {{runTestCase: Function, transformCompiledTokenStream: Function, parseTestCaseFile: Function}}
```
mAAdhaTTah: I don't think this should be removed?

download.js
```diff
@@ -558,8 +560,8 @@ function generateCode(){
 }
 
 function buildCode(promises) {
-	var i = 0,
-	    l = promises.length;
+	var i = 0;
+	var l = promises.length;
```
mAAdhaTTah: These files all have changes which add `var` statements, despite the prism-core files all allowing multiline vars. What's the story here?

ExE-Boss: This is because of [`no-mixed-spaces-and-tabs`](https://eslint.org/docs/rules/no-mixed-spaces-and-tabs) (see commit https://github.com/PrismJS/prism/pull/1796/commits/53a54b40a433ca390918a7d3d0514718fc438ac8).

I’ll probably use the `smart-tabs` option which I didn’t know about until now.

tests/helper/test-case.js
RunDevelopment: It should because the typing is incomplete and only gets in the way of TS's type inference and the doc/types provided by the JSDoc of the functions.

download.js
RunDevelopment: [Here](https://github.com/PrismJS/prism/pull/1796#pullrequestreview-212046340)'s the story.

The problem is that tabs followed by spaces are considered mixed indentation and some auto formatter just replace the 4 spaces with a tab. Most of Prism core has already fallen victim to these formatters. But instead of saving the wounded I think it's easier to just finish them off. (Enough metaphors for today)

The idea is that if you just write the additional `var`s you don't have these problems (mixed indentation, war against formatters).

.eslintrc
ExE-Boss: The thing is that some code is written as if [`comma‑dangle`](https://eslint.org/docs/rules/comma-dangle) was `"always‑multiline"` and other as if it was `"only‑multiline"`.

LeaVerou:
I saw some back and forth on dangling commas. My opinion is that if there is not near 100% consensus on a linting rule, I'd suggest not enforcing one either way. We don't need to control every single aspect of code style via linting, especially if they haven't been enforced so far at all. We don't need to provide a value for every single rule ESLint supports. Linting should help our work, not get in the way, and overly strict linting can easily get in the way.
Just my 2 cents, take it or leave it!

gulpfile.js
```diff
@@ -59,10 +60,18 @@ function inlineRegexSource() {
...
+		uglify(),
 	];
 }
 
+function lint(cb) {
+	return pump([
+		src(['**/*.js', '!node_modules/**'], { base: './' }),
+		eslint(),
+		eslint.format(),
+		eslint.failAfterError(),
```
RunDevelopment: Implementing a `failAfterWarningOrError` should be trivial looking at the [implementation of `failAfterError`](https://github.com/adametry/gulp-eslint/blob/master/index.js#L161).

RunDevelopment:
@ExE-Boss Sorry, I did expect to be able to push to your branch. I reverted everything and make a PR.

RunDevelopment:
The only things left to discuss now are:

1. How to handle [joint variable declarations](https://github.com/PrismJS/prism/pull/1796#pullrequestreview-212046340).
    Do we want to leave the tabs as is or should we add a few more `var`s.
    Using spaces is not possible because of `"no-mixed-spaces-and-tabs": "error"`. [Smart tabs](https://eslint.org/docs/rules/no-mixed-spaces-and-tabs#smart-tabs) are possible but a real pain because some formatters (e.g. the one I'm using) don't support them.
2. The [TSC hack](https://github.com/PrismJS/prism/pull/1796#discussion_r268459515).

.eslintrc
```diff
@@ -0,0 +1,109 @@
...
+		],
+		"linebreak-style": [
+			"error",
+			"unix",
+		],
+		"no-console": "off",
+		"no-empty": "off",
+		"no-inner-declarations": "off",
+		"no-mixed-spaces-and-tabs": "error",
+		"no-redeclare": "off",
```
RunDevelopment: I think that this should be set to `"error"`.

When we switch to ES6 and replace all `var`s with either `const` or `let` in the future, we will have to deal with these errors anyway, so we might as well do it now.

.eslintrc
```diff
@@ -0,0 +1,109 @@
...
+		"comma-dangle": [
+			"warn",
+			"only-multiline",
+		],
+		"linebreak-style": [
+			"error",
+			"unix",
+		],
+		"no-console": "off",
+		"no-empty": "off",
```
RunDevelopment: An empty block is usually a sign that you forgot to do something, so I think that this should be a warning.

One notable exception is [empty `catch` blocks](https://eslint.org/docs/rules/no-empty#allowemptycatch).

```suggestion
		"no-empty": [
			"warn",
			{
				"allowEmptyCatch": true
			}
		],
```

plugins/jsonp-highlight/prism-jsonp-highlight.js
```diff
@@ -39,15 +39,16 @@
 	 * @returns {Adapter} A registered adapter or `null`.
 	 */
 	function getAdapter(adapter) {
+		var i = 0, item;
```
RunDevelopment: You don't need to initialize `i`. The for loops do that already.

```suggestion
		var i, item;
```

plugins/command-line/prism-command-line.js
```diff
@@ -34,9 +34,11 @@ Prism.hooks.add('before-highlight', function (env) {
 
 	var outputSections = pre.getAttribute('data-output');
 	var outputFilter = pre.getAttribute('data-filter-output');
+
+	var i = 0;
```
RunDevelopment:
```suggestion
	var i;
```

mAAdhaTTah:
Ooof, letting this lag will make it very difficult to merge. Are we all aligned on the rules presented here? I might go so far as to suggest slotting this into a potential 2.0 and introduce it in one sweeping change early on there. Thoughts?

LeaVerou:
>  Using spaces is not possible because of `"no-mixed-spaces-and-tabs": "error"`. [Smart tabs](https://eslint.org/docs/rules/no-mixed-spaces-and-tabs#smart-tabs) are possible but a real pain because some formatters (e.g. the one I'm using) don't support them.

We should never use tabs for alignment. The codebase uses smart tabs. If we don't want to enforce them with ESLint we should avoid any rules related to indentation for now, not inserting rules that will force us to worsen how we do indentation and alignment! Again, we don't need to lint everything. We can start with a small set of rules and expand as we go.

(Also, if a tool doesn't support smart tabs, I'd consider that a bug and open an issue with them)

RunDevelopment:
About alignment: Why don't we use Prettier and end the formatting debate?

LeaVerou:
Does Prettier support smart tabs?

RunDevelopment:
Just checked. It doesn't and I'd even argue that it doesn't matter. I have yet to come across a formatter that handles smart tabs correctly.
Examples: ESLint doesn't (see above). Clang format (C/C++ formatter that can also do JS) can kinda do it but it assumes that tabs are 8 spaces wide, so the alignment is all messed up with my usual 4 spaces = 1 tab. VSCode's builtin formatter can't do it.

Point is: Which formatter has good support for smart tabs? If there aren't any good formatters that support it, then I don't think this should be a criterion. Manual formatting just isn't feasible and no formatter is a lot worse than a formatter that uses, god forbid, tabs for alignment, IMO.

Don't get me wrong: I think you're correct. Tabs for indentation and spaces for alignment is the indented use for tabs and spaces but if all our automated tools work against us on this matter, it might be easiest to just not care about it.

RunDevelopment:
Closed in favor of #2831.

PR #1800, opened 2019-03-09
Django/Jinja2 improvements

### LINKED ISSUE #1643: Expand tests for django
As per [this comment](https://github.com/PrismJS/prism/pull/1489#issuecomment-406820767).

### LINKED ISSUE #1797: prism cant work
i just include prism
```html
<script type="text/javascript" src ="/prism/prism.js"></script>
```
but when i opened   the page ，in chrome console it show error 
`Uncaught TypeError: Cannot set property 'inside' of undefined`

the error part is

```javascript
Prism.languages.django.script.inside.string.inside = _django_template
```
i dont know why ， pls help me

RunDevelopment:
I completely remade the Django/Jinja2 language definition using markup templating.

Because [Django templates](https://docs.djangoproject.com/en/2.1/topics/templates/) are (almost) a subset of [Jinja2 templates](http://jinja.pocoo.org/docs/2.10/) both remain one language definition.

The new definition only half the previous size because it uses a positional approach for tags, keywords and filters instead of keyword lists.
<details>
<summary>For future reviewers:</summary>

This is the code I used to test the language:

```Django
{% extends layout_template if layout_template is defined else 'master.html' %}

{% from 'forms.html' import input as input_field, textarea %}

{% include "sidebar.html" ignore missing %}
{% include "sidebar.html" ignore missing with context %}
{% include "sidebar.html" ignore missing without context %}

{% get_language_info_list for available_languages as langs %}
{% for lang in langs %} ... {% endfor %}

{% for athlete in athlete_list %}
    <li>{{ athlete.name }}</li>
{% empty %}
    <li>Sorry, no athletes in this list.</li>
{% endfor %}

{% trans "starting point" as start %}
{% trans "end point" as end %}
{% trans "La Grande Boucle" as race %}
{% blocktrans with amount=article.price %}
That will cost $ {{ amount }}.
{% endblocktrans %}

{% blocktrans with myvar=value|filter %}
This will have {{ myvar }} inside.
{% endblocktrans %}

{% load l10n %}

{% localize on %}
    {{ value }}
{% endlocalize %}

{% localize off %}
    {{ value }}
{% endlocalize %}

{% timezone None %}
    Server time: {{ value }}
{% endtimezone %}

{% load tz %}
{% get_current_timezone as TIME_ZONE %}
<form action="{% url 'set_timezone' %}" method="POST">
    {% csrf_token %}
    <label for="timezone">Time zone:</label>
    <select name="timezone">
        {% for tz in timezones %}
        <option value="{{ tz }}"{% if tz == TIME_ZONE %} selected{% endif %}>{{ tz }}</option>
        {% endfor %}
    </select>
    <input type="submit" value="Set">
</form>

{{ my_date|date:"Y-m-d" }}
{{ value|truncatechars:9 }}
{{ name|striptags|title }}
{{ listx|join(', ') }}
{{ "%s - %s"|format("Hello?", "Foo!") }}
{{ users|join(', ', attribute='username') }}
{{ mytext|urlize(40, true) }}
{{ [1, 2, -3]|max }}
{{ foo['bar'] }}

<ul{{ {'class': 'my_list', 'missing': none,
        'id': 'list-%d'|format(variable)}|xmlattr }}>

<img src="{{ static('path/to/company-logo.png') }}" alt="Company Logo">
<a href="{{ url('admin:index') }}">Administration</a>

{%+ if something %}yay{% endif %}
{%- endfor %}

{% if messages|length >= 100 %}
   You have lots of messages today!
{% endif %}

{% if name is defined %}
{% if loop.index is divisibleby 3  %}
{% if loop.index is divisibleby(3) %}
{% if foo.attribute is sameas false %}
{% if somevar is True %}
{% if somevar is not True %}

{% for value in values %}
    {% if loop.previtem is defined and value > loop.previtem %}
        The value just increased!
    {% endif %}
    {{ value }}
    {% if loop.nextitem is defined and loop.nextitem > value %}
        The value will increase even more!
    {% endif %}
{% endfor %}
{%- for item in sitemap recursive %}
    <li><a href="{{ item.href|e }}">{{ item.title }}</a>
    {%- if item.children -%}
        <ul class="submenu">{{ loop(item.children) }}</ul>
    {%- endif %}</li>
{%- endfor %}
{% for user in users if not user.hidden %}
    <li>{{ user.username|e }}</li>
{% endfor %}

{% regroup cities|dictsort:"country" by country as country_list %}

{% load i18n %}
<html>
  <head>
    <title>{% trans 'Homepage - Hall of Fame' %}</title>
  </head>
  <body>
    {# Translated in the view: #}
    <h1>{{ message }}</h1>
    <p>
      {% blocktrans count member_count=bands.count %}
      Here is the only band in the hall of fame:
      {% plural %}
      Here are all the {{ member_count }} bands in the hall of fame:
      {% endblocktrans %}
    </p>
    <ul>
    {% for band in bands %}
      <li>
        <h2><a href="{{ band.get_absolute_url }}">{{ band.name }}</a></h2>
        {% if band.can_rock %}<p>{% trans 'This band can rock!' %}</p>{% endif %}
      </li>
    {% endfor %}
    </ul>
  </body>
</html>

<style>
a {
	color: {{ bar }};
}
</style>
```

</details>

## Before

image

## After

image

---

This resolves #1643.
This fixes #1797.


components.json
```diff
@@ -231,11 +231,7 @@
 		},
 		"django": {
 			"title": "Django/Jinja2",
-			"require": "markup",
-			"peerDependencies": [
-				"css",
-				"javascript"
-			],
```
mAAdhaTTah: I think we don't want to remove the `peerDependencies` here.

components/prism-django.js
```diff
@@ -1,41 +1,63 @@
...
-	Prism.languages.insertBefore('inside', 'atrule', {'tag': _django_template.property}, Prism.languages.django.style);
-	Prism.languages.django.style.inside.string.inside = _django_template;
-}
-
-// Add an Jinja2 alias
-Prism.languages.jinja2 = Prism.languages.django;
+(function (Prism) {
+
+	var string = /("|')(?:\\.|(?!\1)[^\\\r\n])*\1/;
+	var number = /\b\d+(?:\.\d+)?\b/;
```
mAAdhaTTah: These 2 aren't repeated. Why are they extracted here?

tests/languages/django!+css/inclusion.test
```diff
@@ -0,0 +1,43 @@
+<style>
+a {
+	color: {{ bar }};
```
mAAdhaTTah: Should we test with `{% %}` as the delimiter? I thought this was only for filters.

components.json
RunDevelopment: Yes we do because they aren't really peer dependencies in the sense that Django modifies JS or CSS. It just has to be loaded after them.

The new implementation doesn't care about the markup language because markup-templating takes care of that, so we can safely remove them.

components/prism-django.js
RunDevelopment: Seems like I forgot to inline them.

RunDevelopment:
The difference between `{{ }}` and `{% %}` is that the first can only contain simple _expressions_ (e.g. `{{ foo.bar|baz:"arg" }}`) while the latter can hold more complex _statements_ (e.g. `{% if foo.bar|baz:"arg" is defined or x > 5 %}`).
Therefore it's usually sufficient to just test the more complex case (also most of the features implemented are for statements). The cases where I do use `{{}}` are usually code I copied from the official specs.

tests/languages/django!+css/inclusion.test
RunDevelopment: If one works, the other should also work. But sure, better be safe than sorry.

PR #1801, opened 2019-03-09
Tests: Pretty-printed token stream

RunDevelopment:
This adds support for pretty-printed simplified token streams via the `--pretty` modifier.
E.g. `npm test -- --pretty` or `npm test -- --language=xxxx --pretty`

Only the current token stream of failed tests will be pretty-printed. This can then be copypasted into `.test`-files.
Because most consoles convert tabs into 8 spaces, I used 4 spaces for indentation.

This should make writing tests **a lot** more comfortable.

## Example

Now:

```json
[["tag",[["tag",[["punctuation","<"],"div"]],["attr-name",["class"]],["attr-value",[["punctuation","="],["punctuation","\""],"test",["punctuation","\""]]],["attr-name",["foo"]],["attr-name",["bar"]],["attr-value",[["punctuation","="],"baz"]],["punctuation",">"]]]]
```

Pretty:

```json
[
    ["tag", [
        ["tag", [
            ["punctuation", "<"],
            "div"
        ]],
        ["attr-name", [
            "class"
        ]],
        ["attr-value", [
            ["punctuation", "="],
            ["punctuation", "\""],
            "test",
            ["punctuation", "\""]
        ]],
        ["attr-name", [
            "foo"
        ]],
        ["attr-name", [
            "bar"
        ]],
        ["attr-value", [
            ["punctuation", "="],
            "baz"
        ]],
        ["punctuation", ">"]
    ]]
]
```

### DIFF
--- a/tests/helper/token-stream-transformer.js
+++ b/tests/helper/token-stream-transformer.js
@@ -19,8 +19,7 @@ module.exports = {
 				.map(this.simplify.bind(this))
 				.filter(function (value) {
 					return !(Array.isArray(value) && !value.length) && !(typeof value === "string" && !value.trim().length);
-				}
-				);
+				});
 		}
 		else if (typeof tokenStream === "object") {
 			return [tokenStream.type, this.simplify(tokenStream.content)];
--- a/tests/helper/token-stream-transformer.js
+++ b/tests/helper/token-stream-transformer.js
@@ -11,7 +11,7 @@ module.exports = {
 	 *
 	 *
 	 * @param {Array} tokenStream
-	 * @returns {Array.<string[]|Array>}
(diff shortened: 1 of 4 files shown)


tests/helper/token-stream-transformer.js
```diff
@@ -20,13 +20,53 @@ module.exports = {
 				.filter(function (value) {
 					return !(Array.isArray(value) && !value.length) && !(typeof value === "string" && !value.trim().length);
 				}
-			);
+				);
```
mAAdhaTTah: Nit: this should be on the same line as the closing `}`.

RunDevelopment: Done.

PR #1803, opened 2019-03-12
Tests: Additional checks for Prism functions

### LINKED ISSUE #1717: `insertBefore` should check that `before` exists
[Details](https://github.com/PrismJS/prism/pull/1677#discussion_r249607614)

RunDevelopment:
This adds some additional checks for the inputs of some Prism functions.
They work by replacing a function to test with another function calling both the test code and the original function.

This allows us to add complex checks without adding weight to Prism core.

Right now, I added checks for `Prism.languages.extend` and `Prism.languages.insertBefore`.

---

This resolves #1717.


tests/checks/insert-before.js
```diff
@@ -0,0 +1,39 @@
...
+
+
+function testFunction(name, object, tester) {
+	const func = object[name];
+
+	object[name] = function () {
+		tester.apply(this, arguments);
+		return func.apply(this, arguments);
+	};
+}
```
mAAdhaTTah: Since these are only used in tests, we can put this function in a shared module.

RunDevelopment:
This feels __really__ hacky but it even gets us IDE support.
I added a _pseudo_ require.

tests/helper/prism-loader.js
```diff
@@ -90,7 +90,16 @@ module.exports = {
...
-			context.Prism = this.runFileWithContext(testSource, { Prism: context.Prism }).Prism;
+			context.Prism = this.runFileWithContext(testSource, {
+				Prism: context.Prism,
+				require(path) {
+					if (path.startsWith('./')) {
+						return require('./../checks/' + path.substr(2));
+					} else {
+						return require(path);
+					}
+				}
```
mAAdhaTTah: What's the story here and could we add a comment explaining it?

RunDevelopment: Sure!

PR #1809, opened 2019-03-14
npm: Fixed `test` script

RunDevelopment:
Because of #1646 the command `mocha tests/run.js` was no longer the last command in the chain which means that arguments like `npm test -- --language=css` were not passed to it.
This fixes the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "prism.js",
   "style": "themes/prism.css",
   "scripts": {
-    "test": "mocha tests/testrunner-tests.js && mocha tests/run.js && mocha tests/plugins/**/*.js"
+    "test": "mocha tests/testrunner-tests.js && mocha tests/plugins/**/*.js && mocha tests/run.js"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "test:runner": "mocha tests/testrunner-tests.js",
     "test:languages": "mocha tests/run.js",
     "test:plugins": "mocha tests/plugins/**/*.js",
-    "test": "npm run test:tests && npm run test:languages && npm run test:plugins"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "prism.js",
   "style": "themes/prism.css",
   "scripts": {
-    "test": "mocha tests/testrunner-tests.js && mocha tests/run.js && mocha tests/plugins/**/*.js"
+    "test": "mocha tests/testrunner-tests.js && mocha tests/plugins/**/*.js && mocha tests/run.js"
```
mAAdhaTTah: We should split this into 3 commands, and have `test` run them, so if someone wants to test the languages, they could do `npm run test:languages -- --language=css` without issue. It also allows you to e.g. just test plugins without needing to run the other tests around it.

RunDevelopment:
Like this @mAAdhaTTah ?
This, of course, has the disadvantage that none of the arguments of `npm test` are passed to either of the called commands.

mAAdhaTTah:
@RunDevelopment Yeah, that's what I was thinking. If you're just testing a single language, you wouldn't run the main command, you'd just run `npm run test:languages` with the appropriate flag. You wouldn't want to run all the other tests at the same time.

package.json
```diff
@@ -5,7 +5,10 @@
   "main": "prism.js",
   "style": "themes/prism.css",
   "scripts": {
-    "test": "mocha tests/testrunner-tests.js && mocha tests/run.js && mocha tests/plugins/**/*.js"
+    "test:tests": "mocha tests/testrunner-tests.js",
```
mAAdhaTTah:
```suggestion
    "test:runner": "mocha tests/testrunner-tests.js",
```

RunDevelopment:
Alright, then I'll adjust the docs and then this should be gtg.

PR #1814, opened 2019-03-16
Scheme: Minor improvements

RunDevelopment:
This makes some minor improvements to the Scheme language definition.

This resolves the multiline string issue from #1451.


components/prism-scheme.js
```diff
@@ -17,17 +17,18 @@ Prism.languages.scheme = {
 		lookbehind: true
 	},
 	'number': {
-		pattern: /(\s|[()])[-+]?\d*\.?\d+(?:\s*[-+]\s*\d*\.?\d+i)?\b/,
+		newPattern: /(?:#d(?:#[ei])?|#[ei](?:#d)?)?/i,
```
mAAdhaTTah: Is `newPattern` a thing? And should there be associated tests for this?

RunDevelopment: Ahhh, big mistake on my side. I sometimes test out new patterns like this.
No, `newPattern` is not a thing.

PR #1818, opened 2019-03-16
Toolbar: Minor improvements

RunDevelopment:
Some minor improvements for the Toolbar plugin:

1. More JSDoc and type information for `registerButton`.
2. A new check for already registered keys.
    It now logs a warning and returns without registering the button. This slightly changes the current behavior but I don't think that registering for the same __key__ twice should be expected to work anyway.


plugins/toolbar/prism-toolbar.js
```diff
@@ -43,6 +50,11 @@
 			};
 		}
 
+		if (key in map) {
+			console.warn('The is a button with the key "' + key + '" registered already.');
```
mAAdhaTTah: Typo in the below (`the` -> `there`). Also added a prefix to make it clearer where the error is coming from.

```suggestion
			console.warn('[Prism.plugins.toolbar#registerButton]: There is a button with the key "' + key + '" registered already.');
```

RunDevelopment: Typo: Point taken.

`console.warn` will come with a neat stack trace in modern browsers, so I don't quite see the point of the prefix.

mAAdhaTTah: Yeah, and we don't prefix [other warnings](https://github.com/PrismJS/prism/blob/fedf03f61645dc5d9372453e421e3697cb922fe7/plugins/copy-to-clipboard/prism-copy-to-clipboard.js#L7).

PR #1819, opened 2019-03-17
Travis: Fail when changed files are detected

### LINKED ISSUE #1785: Travis: Fail if files change after gulp
A lot of times, contributors (myself included) forgot to build or commit the minified files.

Since Travis is running `gulp` anyway, why not check for changes?
The current gulp build process is deterministic and only depends on the input files, so successive build won't change any files. So if files do change, we know that someone forgot to rebuild Prism.

We should be able to use `git status` to detect the changed files.

RunDevelopment:
This resolves #1785.

---

I'll make some testing on this PR, so there might be some failed tests here and there. I'll comment again when I'm ready.


RunDevelopment:
Alright, it failed when some generated files were changed and it succeeds when Prism is built properly. 

Seems to work.

mAAdhaTTah:
Love the idea. Two nits:

1. In the `gulpfile.js`, we tend to define the function separately from the exports, so we should do the same here. I'm specifically thinking this will be useful once we land #1796, we can add eslint + this check to like a "premerge" task. Which leads to...
2. ...maybe we should change the task name to `premerge` or something of that nature? Bikeshedding, but I don't love that name. I might suggest `needsRebuild` as the function name, and `premerge` as the task name, so we can expand `premerge` later, but I'm open to other ideas.

RunDevelopment:
Alright, funny story:

The build fails because some minified files changed. I had to make a fresh npm installation until these files also changed on my system.
I don't know the concrete behavior of `npm install` but it doesn't seem to update dependencies when the new release is only a patch. A patch which is uglify's case changed the minification behavior.

So yeah. With the current state of this PR, the success of our builds will be at the mercy of Uglify's minification behavior.

Maybe it's time for a `package-lock.json`?

mAAdhaTTah:
Yeah, and we'll need to update the install step to use `npm ci`. That can go in this PR, since both are dealing with Travis.

RunDevelopment:
Minor problem: The npm version of Node v6 doesn't support `npm ci`. Any ideas?

mAAdhaTTah:
Upgrade `npm` in travis?

RunDevelopment:
Updating to the latest npm version works but it feels kind of hacky...

Well, anyway. @mAAdhaTTah, you might want to have another look.

mAAdhaTTah:
This works for me. Since we're only updating to use `npm ci`, which relies solely on the `package-lock.json`, I don't think we're likely to run into issues if we stick with the latest version.

PR #1820, opened 2019-03-17
Adding initial support for CMake language

mjrogozinski:
I know it's not perfect but it is good enough for me. Sorry for not being an expert on the subject, it's my first time doing it.


components/prism-cmake.js
```diff
@@ -0,0 +1,8 @@
...
...USER|CTEST_EXTRA_COVERAGE_GLOB|CTEST_GIT_COMMAND|CTEST_GIT_INIT_SUBMODULES|CTEST_GIT_UPDATE_CUSTOM|CTEST_GIT_UPDATE_OPTIONS|CTEST_HG_COMMAND|CTEST_HG_UPDATE_OPTIONS|CTEST_LABELS_FOR_SUBPROJECTS|CTEST_MEMORYCHECK_COMMAND|CTEST_MEMORYCHECK_COMMAND_OPTIONS|CTEST_MEMORYCHECK_SANITIZER_OPTIONS|CTEST_MEMORYCHECK_SUPPRESSIONS_FILE|CTEST_MEMORYCHECK_TYPE|CTEST_NIGHTLY_START_TIME|CTEST_P4_CLIENT|CTEST_P4_COMMAND|CTEST_P4_OPTIONS|CTEST_P4_UPDATE_OPTIONS|CTEST_RUN_CURRENT_SCRIPT|CTEST_SCP_COMMAND|CTEST_SITE|CTEST_SOURCE_DIRECTORY|CTEST_SUBMIT_URL|CTEST_SVN_COMMAND|CTEST_SVN_OPTIONS|CTEST_SVN_UPDATE_OPTIONS|CTEST_TEST_LOAD|CTEST_TEST_TIMEOUT|CTEST_TRIGGER_SITE|CTEST_UPDATE_COMMAND|CTEST_UPDATE_OPTIONS|CTEST_UPDATE_VERSION_ONLY|CTEST_USE_LAUNCHERS|CYGWIN|ENV|EXECUTABLE_OUTPUT_PATH|GHS-MULTI|IOS|LIBRARY_OUTPUT_PATH|MINGW|MSVC|MSVC10|MSVC11|MSVC12|MSVC14|MSVC60|MSVC70|MSVC71|MSVC80|MSVC90|MSVC_IDE|MSVC_TOOLSET_VERSION|MSVC_VERSION|MSYS|PROJECT_BINARY_DIR|PROJECT_DESCRIPTION|PROJECT_HOMEPAGE_URL|PROJECT_NAME|PROJECT_SOURCE_DIR|PROJECT_VERSION|PROJECT_VERSION_MAJOR|PROJECT_VERSION_MINOR|PROJECT_VERSION_PATCH|PROJECT_VERSION_TWEAK|UNIX|WIN32|WINCE|WINDOWS_PHONE|WINDOWS_STORE|XCODE|XCODE_VERSION)\b/,
```
RunDevelopment: We use single quotes for all our token names.

components/prism-cmake.js
```diff
@@ -0,0 +1,8 @@
+Prism.languages.cmake = Prism.languages.extend("python", {
```
RunDevelopment: Why did you choose to extend Python?

I'm not very familiar with the CMake syntax but they don't seem to be that much alike.

(Also, single quotes please.)

components/prism-cmake.js
```diff
@@ -0,0 +1,8 @@
...
+    "boolean": /\b(ON|OFF)\b/,
+    "namespace": /\b(PROPERTIES|SHARED|PRIVATE|STATIC|PUBLIC|INTERFACE|TARGET_OBJECTS)\b/,
+    operator: /[-+]{1,2}|!=?|<{1,2}=?|>{1,2}=?|\->|:{1,2}|={1,2}|\^|~|%|&{1,2}|\|?\||\?|\*|\/|\b(and|and_eq|bitand|bitor|not|not_eq|or|or_eq|xor|xor_eq)\b/,
```
RunDevelopment: `[+-]{1,2}`: Does that mean that `+-` is a valid operator?

RunDevelopment: Instead of defining a bunch of single characters via alternation, it's usually better to make them a character set.
So `\^|~|%|&{1,2}|\|?\||\?|\*|\/` -> `&&?|\|\|?|[~^%?*/]`

components/prism-cmake.js
```diff
@@ -0,0 +1,8 @@
...
...VS_SHADER_ENABLE_DEBUG|VS_SHADER_ENTRYPOINT|VS_SHADER_FLAGS|VS_SHADER_MODEL|VS_SHADER_OBJECT_FILE_NAME|VS_SHADER_OUTPUT_HEADER_FILE|VS_SHADER_TYPE|VS_SHADER_VARIABLE_NAME|VS_STARTUP_PROJECT|VS_TOOL_OVERRIDE|VS_USER_PROPS|VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION|VS_WINRT_COMPONENT|VS_WINRT_EXTENSIONS|VS_WINRT_REFERENCES|VS_XAML_TYPE|WILL_FAIL|WIN32_EXECUTABLE|WINDOWS_EXPORT_ALL_SYMBOLS|WORKING_DIRECTORY|WRAP_EXCLUDE|XCODE_ATTRIBUTE_.*|XCODE_EMIT_EFFECTIVE_PLATFORM_NAME|XCODE_EXPLICIT_FILE_TYPE|XCODE_EXPLICIT_FILE_TYPE|XCODE_FILE_ATTRIBUTES|XCODE_LAST_KNOWN_FILE_TYPE|XCODE_PRODUCT_TYPE|XCODE_SCHEME_ADDRESS_SANITIZER|XCODE_SCHEME_ADDRESS_SANITIZER_USE_AFTER_RETURN|XCODE_SCHEME_ARGUMENTS|XCODE_SCHEME_DISABLE_MAIN_THREAD_CHECKER|XCODE_SCHEME_DYNAMIC_LIBRARY_LOADS|XCODE_SCHEME_DYNAMIC_LINKER_API_USAGE|XCODE_SCHEME_ENVIRONMENT|XCODE_SCHEME_EXECUTABLE|XCODE_SCHEME_GUARD_MALLOC|XCODE_SCHEME_MAIN_THREAD_CHECKER_STOP|XCODE_SCHEME_MALLOC_GUARD_EDGES|XCODE_SCHEME_MALLOC_SCRIBBLE|XCODE_SCHEME_MALLOC_STACK|XCODE_SCHEME_THREAD_SANITIZER|XCODE_SCHEME_THREAD_SANITIZER_STOP|XCODE_SCHEME_UNDEFINED_BEHAVIOUR_SANITIZER|XCODE_SCHEME_UNDEFINED_BEHAVIOUR_SANITIZER_STOP|XCODE_SCHEME_ZOMBIE_OBJECTS|XCTEST)\b/,
```
RunDevelopment: Here, I'm guessing, but you probably don't want `.*` for two reasons:

1. I guess that `_OUTPUT_NAME` isn't a valid function, right? So please use the `+` operator. If it is, please ignore this point.
2. The dot `.` will match any character except line breaks. This includes spaces, brackets, punctuation, etc meaning that the `\b` you use to guard the properties doesn't mean much. In this case, you should probably use `\w` (== `[A-Za-z0-9_]`) instead of `.`.

Also, could you please group all words starting with `.*` (or `\w+`) together to get `.*(?:_OUTPUT_NAME|_POSTFIX|...)`?

All of these points also apply to `keyword`.

components/prism-cmake.js
```diff
@@ -0,0 +1,8 @@
...
...|enable_testing|endforeach|endfunction|endif|endmacro|endwhile|exec_program|execute_process|export|export_library_dependencies|file|find_file|find_library|find_package|find_path|find_program|fltk_wrap_ui|foreach|function|get_cmake_property|get_directory_property|get_filename_component|get_property|get_source_file_property|get_target_property|get_test_property|if|include|include_directories|include_external_msproject|include_guard|include_regular_expression|install|install_files|install_programs|install_targets|link_directories|link_libraries|list|load_cache|load_command|macro|make_directory|mark_as_advanced|math|message|option|output_required_files|project|qt_wrap_cpp|qt_wrap_ui|remove|remove_definitions|return|separate_arguments|set|set_directory_properties|set_property|set_source_files_properties|set_target_properties|set_tests_properties|site_name|source_group|string|subdir_depends|subdirs|target_compile_definitions|target_compile_features|target_compile_options|target_include_directories|target_link_directories|target_link_libraries|target_link_options|target_sources|try_compile|try_run|unset|use_mangled_mesa|utility_source|variable_requires|variable_watch|while|write_file)\b/,
```
RunDevelopment: Guessing from the test, could it be that a function is always followed by `(`?

If so, we can simplify this expression to `/\b[a-z_]\w*(?=\s*\()/i`.

components.json
```diff
@@ -925,6 +925,11 @@
 		"yaml": {
 			"title": "YAML",
 			"owner": "hason"
+		},
+		"cmake": {
```
RunDevelopment: The language list is largely sorted alphabetically, so please insert this before CoffeeScript.

tests/languages/cmake/namespace_feature.test
```diff
@@ -0,0 +1,11 @@
...
+
+----------------------------------------------------
+
+[
+	["namespace", "STATIC"]
+]
+
+----------------------------------------------------
+
+Checks for all keywords.
```
RunDevelopment: "Checks for all namespaces."

tests/languages/cmake/property_feature.test
```diff
@@ -0,0 +1,11 @@
...
+
+----------------------------------------------------
+
+[
+	["property", "ANDROID_GUI"]
+]
+
+----------------------------------------------------
+
+Checks for all keywords.
```
RunDevelopment: "Checks for all properties."

components/prism-cmake.js
mjrogozinski: Sure thing.

components/prism-cmake.js
mjrogozinski: Two reasons:
1. # comments
2. parenthesis
I haven't found any generic script-like syntax so I chose the simplest possible syntax :)

components/prism-cmake.js
mjrogozinski: CMake has different ones. This entire line will be replaced.

mjrogozinski: This entire line will be replaced.

components/prism-cmake.js
mjrogozinski: Unfortunately, it's not that simple. if() is also followed by (. This is the list of built-in functions generated by CMake's help. You can define your own functions, but it's quite rare and there is no way to know it with a regex.

components/prism-cmake.js
mjrogozinski: Since the order is the same as generated from CMake's help `cmake --help-command-list` `cmake --help-property-list` `cmake --help-variable-list`, I propose to use the `\w+` approach and not group them.

components/prism-cmake.js
RunDevelopment: But `if` is already matched as a `keyword`, so it can't be matched again as a `function`.
Other programming languages (every C-style one) have the same issue with `if`, `while`, etc. and they use this exact trick.

So after matching the keywords, are there any function-like constructs left?

components/prism-cmake.js
RunDevelopment: Hmmm, the CMake doc for `--help-command-list` says:

> [The list contains all commands for which help may be obtained by using the --help-command option followed by a command name.](https://cmake.org/cmake/help/v3.1/manual/ccmake.1.html#options)

Similar for the other ones. So are these lists complete?

components/prism-cmake.js
RunDevelopment: Wait. `if` is also in the functions list.

components/prism-cmake.js
RunDevelopment: So, the only things we need from Python are comments, strings, and punctuation. Is that right?

In that case we could just copy them over from Python.

RunDevelopment:
This is a more general question:

You assigned the outputs of these commands the following token names, right?

- `cmake --help-command-list` -> `function`
- `cmake --help-property-list` -> `property`
- `cmake --help-variable-list` ->`keyword`

The command list contains stuff like `if`, `while`, etc., so could we make this `keyword` to make it more consistent with other languages?
We could give the variables list the `variable` highlighting as well.

examples/prism-cmake.html
```diff
@@ -0,0 +1,86 @@
...
+<h2>Properties</h2>
+<pre><code>CUDA_STANDARD
+CUDA_STANDARD_REQUIRED
+CXX_EXTENSIONS
+CXX_STANDARD</code></pre>
+
+<h2>Full example</h2>
+<pre><code>cmake_minimum_required(VERSION 3.13)
+
+        project(crypto)
```
RunDevelopment: Do we really want this indentation here?

components/prism-cmake.js
mjrogozinski: Welcome to CMake world ;)

examples/prism-cmake.html
mjrogozinski: Of course not :)

mjrogozinski: I somehow couldn't get it to work. How to see if this example works? When I open examples.html, it says that CMake has no examples.

components/prism-cmake.js
mjrogozinski: In general constructs like ifs and whiles are used very sparingly during CMake "development" so treating them like keywords or functions does not really matter. I think it's enough they are "recognized" by the highlighter and coloured in some way.

components/prism-cmake.js
mjrogozinski: It covers much more than is used in everyday development. I will give it a spin on my blog and evaluate whether we need some additional commands/functions/keywords.

components/prism-cmake.js
mjrogozinski: Ok

mjrogozinski: Could you give me a hand with punctation?
We need ( ) $< >. $< is a generator expression, so we have to match those two characters, not only <

examples/prism-cmake.html
RunDevelopment: No, unfortunately, we cant.
The examples page gets the info of what examples are available from Github using the master branch of this project. So it won't be accessible until it gets merged.

I know, not the best setup but you can always the [test page](https://github.com/PrismJS/prism/blob/master/test.html) of your local project to see the results.

components/prism-cmake.js
RunDevelopment: Ok, so until proven otherwise, we'll just assume that they are complete. Works for me.

components/prism-cmake.js
RunDevelopment: You probably want `${...}` as well, right?
If so: `/[()>}]|\$[<{]/`

components/prism-cmake.js
```diff
@@ -1,8 +1,18 @@
-Prism.languages.cmake = Prism.languages.extend('python', {
+Prism.languages.cmake = {
+    'comment': {
+		pattern: /(^|[^\\])#.*/,
```
RunDevelopment: Now that we control the comment pattern, we can simplify it!
This is simpler and should also work:
```js
'comment': /#.*/
```

components/prism-cmake.js
```diff
@@ -1,8 +1,18 @@
...
+		lookbehind: true
+    },
+    'string': {
+		pattern: /(?:[rub]|rb|br)?("|')(?:\\.|(?!\1)[^\\\r\n])*\1/i,
```
RunDevelopment: Now that we control strings, we might as well remove the `(?:[rub]|rb|br)?` prefix (and the `i` flag).

I also have a few questions about strings:

1. Do backslash escapes work? (e.g. `"\n\t"`)
2. Are multiline strings allowed?

components/prism-cmake.js
```diff
@@ -1,8 +1,18 @@
...
+		greedy: true
+    },
+    'punctuation': /[()]/,
+    'number': /(?:\b(?=\d)|\B(?=\.))(?:0[bo])?(?:(?:\d|0x[\da-f])[\da-f]*\.?\d*|\.\d+)(?:e[+-]?\d+)?j?\b/i,
```
RunDevelopment: Just a little question: Why apart from `cmake_minimum_required (VERSION 2.8.11)` are numbers used? Aren't [math expressions done via strings](https://cmake.org/cmake/help/v3.13/command/math.html)?

components/prism-cmake.js
mjrogozinski: Yes, that's right. Thank you!

components/prism-cmake.js
mjrogozinski: In CMake everything is a string. The only numbers look like the one you pasted. Numbers are used for versioning mainly and are separated with a dot.

components/prism-cmake.js
mjrogozinski: Multiline strings - yes.
Bashslash escapes - no.

mjrogozinski: Simplified.

RunDevelopment: Alright, then this should work:

```js
'string': {
    pattern: /"(?:[^\\"]|\\.)*"/,
    greedy: true
}
```

components/prism-cmake.js
RunDevelopment: Then should the more simple pattern: `\d+(?:\.\d+)*` also work?
The Python number pattern also handles binary, hexadecimal, and complex numbers and scientific notation. We don't need that, do we?

components/prism-cmake.js
```diff
@@ -0,0 +1,18 @@
+Prism.languages.cmake = {
+    'comment': {
+		pattern: /#.*/,
+		lookbehind: true
```
RunDevelopment: You don't need the lookbehind anymore.
It can just be 
```js
'comment': /#.*/,
... other tokens
```

components/prism-cmake.js
RunDevelopment: I just think that these, well, _feel_ like keywords, so I would much rather have them highlighted as such. `function` could then match user-defined functions and macros using the pattern: `/\b[a-z_]\w*(?=\s*\()/i`.

What do you think?

mjrogozinski: I agree that user-defined functions should be highlighted differently. However e.g. WIN32 should be highlighted differently than add_library. All of the IDEs supporting CMake work this way and it makes reading code easier.

I propose to use the `keyword` highlighting for `add_library, find_package, if, while` - everything that's generally known to CMake and written with lowercase with parens after.

I agree with the function proposal of yours. Let's use it.

Let's leave properties as they are.

Now, what can we do with the current `keyword` marking? It's different than the others. Can we use some other class?

components/prism-cmake.js
```diff
@@ -0,0 +1,17 @@
+Prism.languages.cmake = {
+    'comment': {
+	  pattern: /#.*/,
```
RunDevelopment: You seem to have mixed indentation here.

tests/languages/cmake/property_feature.test
RunDevelopment: This still needs fixing.

components/prism-cmake.js
RunDevelopment: Also, you can inline the regex to get

```js
'comment': /#.*/,
...
```

tests/languages/cmake/property_feature.test
mjrogozinski: Fixed

components/prism-cmake.js
RunDevelopment: Looks good!
As suggested [here](https://github.com/PrismJS/prism/pull/1820#issuecomment-474110085) we could  the current `keyword`

components/prism-cmake.js
mjrogozinski: Fixed

examples/prism-cmake.html
```diff
@@ -0,0 +1,86 @@
...
+        crypto::helpers
+    )
+
+set_target_properties(analyzers-obj
+    PROPERTIES
+        POSITION_INDEPENDENT_CODE ON
+    )
+
+add_library(analyzers SHARED
+    $<TARGET_OBJECTS:analyzers-obj>
```
RunDevelopment: The `<` should be escaped using `&lt;`.

components/prism-cmake.js
mjrogozinski: Yes, we might use that. However, we need a bit more sophisticated pattern. It matches the 4 from e.g. CTEST_P4_CLIENT and somehow \r\n.
I will revisit this issue after examining other comments.

examples/prism-cmake.html
mjrogozinski: I guess the other one with `&gt;`? I still haven't figured out how to run this example, hence the trivial mistakes.

components/prism-cmake.js
mjrogozinski: Sorry about that, I think I have fixed them properly. However, I might still need to convert it to tabs. Isn't there some automatic an formatter like clang-format in C++?

components/prism-cmake.js
mjrogozinski: As you pointed out: "We could give the variables list the `variable` highlighting as well."
The variable concept matches what's underneath the current `keyword` as e.g. `CMAKE_ABSOLUTE_DESTINATION_FILES` is a variable holding a list of files and can be used using variable semantics.

components/prism-cmake.js
RunDevelopment: My bad, I forgot the boundaries. The following should work.

```js
/\b\d+(?:\.\d+)*\b/
```

components/prism-cmake.js
RunDevelopment: I'm using VSCode. It can auto convert back and forth between spaces and tabs.

If you don't find any better solution, a replace regex usually works wonders.

examples/prism-cmake.html
RunDevelopment: > I still haven't figured out how to run this example

[As I said](https://github.com/PrismJS/prism/pull/1820#discussion_r266665156), you can't until we merge this PR.

components/prism-cmake.js
mjrogozinski: I've pushed my proposal.
It lacks function tag as I tried the proposed one and it did not match what I expected i.e. `bar` in expression `bar()`.

components/prism-cmake.js
mjrogozinski: ok

components/prism-cmake.js
mjrogozinski: Strange, according to Regexr, your regex works just fine. I will debug this a bit more.

examples/prism-cmake.html
mjrogozinski: Ok, sorry about missing this answer earlier.

components/prism-cmake.js
RunDevelopment: That because you already match the `(` in `punctuation`. Is usually a good idea to match operators and punctuation last.

components/prism-cmake.js
mjrogozinski: Fixed.

components/prism-cmake.js
mjrogozinski: Yes, you are right.
I've pushed the correction.

components/prism-cmake.js
```diff
@@ -0,0 +1,18 @@
...
+	'boolean': /\b(ON|OFF|TRUE|FALSE)\b/,
+	'namespace': /\b(PROPERTIES|SHARED|PRIVATE|STATIC|PUBLIC|INTERFACE|TARGET_OBJECTS)\b/,
+	'operator': /\b(NOT|AND|OR|MATCHES|LESS|GREATER|EQUAL|STRLESS|STRGREATER|STREQUAL|VERSION_LESS|VERSION_EQUAL|VERSION_GREATER|DEFINED)\b/,
+	'comment': {
```
RunDevelopment: Please move comment and string before all other tokens. This will solve the current issue with:

```
# ON OFF add_compile_options
"ON"
```

Also, please make it `'comment': /#.*/`. You don't need to create a new object.

mjrogozinski: done

components/prism-cmake.js
```diff
@@ -0,0 +1,18 @@
...
+	'boolean': /\b(ON|OFF|TRUE|FALSE)\b/,
```
RunDevelopment: Please only use non-capturing groups, so `\b(?:ON|...)\b` instead of  `\b(ON|...)\b`.

This also applied for all other tokens.

mjrogozinski: Of course, but not to comments, right?

components/prism-cmake.js
RunDevelopment: You still need to reorder.

components/prism-cmake.js
mjrogozinski: pushed

RunDevelopment: Comments?

mjrogozinski: `	'comment': /#.*/,` Should this be a non-capturing group?

mjrogozinski: Well, it's not a group, so probably not :)

components/prism-cmake.js
mjrogozinski: Ok, sure I misunderstood that.

components/prism-cmake.js
RunDevelopment: No, we don't need to introduce an unnecessary group, so you can leave it as is.

components/prism-cmake.js
mjrogozinski: Now it should be fine

components/prism-cmake.js
```diff
@@ -0,0 +1,16 @@
...
+	'boolean': /\b(?:ON|OFF|TRUE|FALSE)\b/,
+	'namespace': /\b(?:PROPERTIES|SHARED|PRIVATE|STATIC|PUBLIC|INTERFACE|TARGET_OBJECTS)\b/,
+	'operator': /\b(?:NOT|AND|OR|MATCHES|LESS|GREATER|EQUAL|STRLESS|STRGREATER|STREQUAL|VERSION_LESS|VERSION_EQUAL|VERSION_GREATER|DEFINED)\b/,
+	'number': /\b\d+(?:\.\d+)*\b/,
+	'function': /\b[a-z_]\w*(?=\s*\()\b/,
```
RunDevelopment: This needs an `i` flag, otherwise, we don't match `foobar()`.

mjrogozinski: How is it possible?
foobar is matched without this flag in the function_feature.test.

components/prism-cmake.js
```diff
@@ -0,0 +1,16 @@
...
...|enable_testing|endforeach|endfunction|endif|endmacro|endwhile|exec_program|execute_process|export|export_library_dependencies|file|find_file|find_library|find_package|find_path|find_program|fltk_wrap_ui|foreach|function|get_cmake_property|get_directory_property|get_filename_component|get_property|get_source_file_property|get_target_property|get_test_property|if|include|include_directories|include_external_msproject|include_guard|include_regular_expression|install|install_files|install_programs|install_targets|link_directories|link_libraries|list|load_cache|load_command|macro|make_directory|mark_as_advanced|math|message|option|output_required_files|project|qt_wrap_cpp|qt_wrap_ui|remove|remove_definitions|return|separate_arguments|set|set_directory_properties|set_property|set_source_files_properties|set_target_properties|set_tests_properties|site_name|source_group|string|subdir_depends|subdirs|target_compile_definitions|target_compile_features|target_compile_options|target_include_directories|target_link_directories|target_link_libraries|target_link_options|target_sources|try_compile|try_run|unset|use_mangled_mesa|utility_source|variable_requires|variable_watch|while|write_file)\b/,
```
RunDevelopment: While testing I found this:

```cmake
include_directories(~/exploringBB/extras/cmake/studentlib_shared/include)
```

The last `include` is highlighted as a keyword.
How about we replace a `(?=\s*\()` to the pattern to eliminate these false positives? This assumes that all keywords are followed by a `(`.

components/prism-cmake.js
RunDevelopment: Sorry, I mistyped. Fixed it.

components/prism-cmake.js
mjrogozinski: Good idea. Keyword has to be followed by '('.

mjrogozinski: pushed

components/prism-cmake.js
RunDevelopment: @mjrogozinski The `i` flag is necessary, i.e. `Foobar()`.

mjrogozinski: `i` means case insensitive. Now I get it.
I need to learn more about those regex things.

RunDevelopment:
@mjrogozinski With that this PR is _almost_ ready to merge. Only 2 more things:

1. Please use optimized keyword lists. Right now, the language definition is 24Kb large with most of the size coming from `variable` and `property`. So by using the Regex syntax to group the words, we can drastically reduce the file size. 
    I also used another trick to reduce the size of `variable`: It contains `CMAKE_\w+`, meaning that all words stating with `CMAKE_` are variables. As a result we can remove all other words which start with `CMAKE_` from the list.
    The resulting new patterns are the following:
    ```js
    'variable': /\b(?:CMAKE_\w+|\w+_(?:VERSION(?:_MAJOR|_MINOR|_PATCH|_TWEAK)?|(?:BINARY|SOURCE)_DIR|DESCRIPTION|HOMEPAGE_URL|ROOT)|(?:CTEST_CUSTOM_(?:MAXIMUM_(?:(?:FAIL|PASS)ED_TEST_OUTPUT_SIZE|NUMBER_OF_(?:ERROR|WARNING)S)|ERROR_(?:P(?:OST|RE)_CONTEXT|EXCEPTION|MATCH)|P(?:OST|RE)_MEMCHECK|WARNING_(?:EXCEPTION|MATCH)|(?:MEMCHECK|TESTS)_IGNORE|P(?:OST|RE)_TEST|COVERAGE_EXCLUDE)|ANDROID|APPLE|BORLAND|BUILD_SHARED_LIBS|CACHE|CPACK_(?:ABSOLUTE_DESTINATION_FILES|COMPONENT_INCLUDE_TOPLEVEL_DIRECTORY|ERROR_ON_ABSOLUTE_INSTALL_DESTINATION|INCLUDE_TOPLEVEL_DIRECTORY|INSTALL_DEFAULT_DIRECTORY_PERMISSIONS|INSTALL_SCRIPT|PACKAGING_INSTALL_PREFIX|SET_DESTDIR|WARN_ON_ABSOLUTE_INSTALL_DESTINATION)|CTEST_(?:BINARY_DIRECTORY|BUILD_COMMAND|BUILD_NAME|BZR_COMMAND|BZR_UPDATE_OPTIONS|CHANGE_ID|CHECKOUT_COMMAND|CONFIGURATION_TYPE|CONFIGURE_COMMAND|COVERAGE_COMMAND|COVERAGE_EXTRA_FLAGS|CURL_OPTIONS|CUSTOM_(?:COVERAGE_EXCLUDE|ERROR_EXCEPTION|ERROR_MATCH|ERROR_POST_CONTEXT|ERROR_PRE_CONTEXT|MAXIMUM_FAILED_TEST_OUTPUT_SIZE|MAXIMUM_NUMBER_OF_(?:ERRORS|WARNINGS)|MAXIMUM_PASSED_TEST_OUTPUT_SIZE|MEMCHECK_IGNORE|POST_MEMCHECK|POST_TEST|PRE_MEMCHECK|PRE_TEST|TESTS_IGNORE|WARNING_EXCEPTION|WARNING_MATCH)|CVS_CHECKOUT|CVS_COMMAND|CVS_UPDATE_OPTIONS|DROP_LOCATION|DROP_METHOD|DROP_SITE|DROP_SITE_CDASH|DROP_SITE_PASSWORD|DROP_SITE_USER|EXTRA_COVERAGE_GLOB|GIT_COMMAND|GIT_INIT_SUBMODULES|GIT_UPDATE_CUSTOM|GIT_UPDATE_OPTIONS|HG_COMMAND|HG_UPDATE_OPTIONS|LABELS_FOR_SUBPROJECTS|MEMORYCHECK_(?:COMMAND|COMMAND_OPTIONS|SANITIZER_OPTIONS|SUPPRESSIONS_FILE|TYPE)|NIGHTLY_START_TIME|P4_CLIENT|P4_COMMAND|P4_OPTIONS|P4_UPDATE_OPTIONS|RUN_CURRENT_SCRIPT|SCP_COMMAND|SITE|SOURCE_DIRECTORY|SUBMIT_URL|SVN_COMMAND|SVN_OPTIONS|SVN_UPDATE_OPTIONS|TEST_LOAD|TEST_TIMEOUT|TRIGGER_SITE|UPDATE_COMMAND|UPDATE_OPTIONS|UPDATE_VERSION_ONLY|USE_LAUNCHERS)|CYGWIN|ENV|EXECUTABLE_OUTPUT_PATH|GHS-MULTI|IOS|LIBRARY_OUTPUT_PATH|MINGW|MSVC(?:10|11|12|14|60|70|71|80|90|_IDE|_TOOLSET_VERSION|_VERSION)?|MSYS|PROJECT_(?:BINARY_DIR|DESCRIPTION|HOMEPAGE_URL|NAME|SOURCE_DIR|VERSION|VERSION_(?:MAJOR|MINOR|PATCH|TWEAK))|UNIX|WIN32|WINCE|WINDOWS_PHONE|WINDOWS_STORE|XCODE|XCODE_VERSION))\b/,
    'property': /\b(?:(?:ARCHIVE_OUTPUT_(?:DIRECTORY|NAME)|COMPILE_DEFINITIONS|COMPILE_PDB_NAME|COMPILE_PDB_OUTPUT_DIRECTORY|EXCLUDE_FROM_DEFAULT_BUILD|IMPORTED_(?:IMPLIB|LIBNAME|LINK_DEPENDENT_LIBRARIES|LINK_INTERFACE_LANGUAGES|LINK_INTERFACE_LIBRARIES|LINK_INTERFACE_MULTIPLICITY|LOCATION|NO_SONAME|OBJECTS|SONAME)|INTERPROCEDURAL_OPTIMIZATION|LIBRARY_OUTPUT_DIRECTORY|LIBRARY_OUTPUT_NAME|LINK_FLAGS|LINK_INTERFACE_LIBRARIES|LINK_INTERFACE_MULTIPLICITY|LOCATION|MAP_IMPORTED_CONFIG|OSX_ARCHITECTURES|OUTPUT_NAME|PDB_NAME|PDB_OUTPUT_DIRECTORY|RUNTIME_OUTPUT_DIRECTORY|RUNTIME_OUTPUT_NAME|STATIC_LIBRARY_FLAGS|VS_CSHARP|VS_DOTNET_REFERENCEPROP|VS_DOTNET_REFERENCE|VS_GLOBAL_SECTION_POST|VS_GLOBAL_SECTION_PRE|VS_GLOBAL|XCODE_ATTRIBUTE)_\w+|\w+_(?:CLANG_TIDY|COMPILER_LAUNCHER|CPPCHECK|CPPLINT|INCLUDE_WHAT_YOU_USE|OUTPUT_NAME|POSTFIX|VISIBILITY_PRESET)|ABSTRACT|ADDITIONAL_MAKE_CLEAN_FILES|ADVANCED|ALIASED_TARGET|ALLOW_DUPLICATE_CUSTOM_TARGETS|ANDROID_(?:ANT_ADDITIONAL_OPTIONS|API|API_MIN|ARCH|ASSETS_DIRECTORIES|GUI|JAR_DEPENDENCIES|NATIVE_LIB_DEPENDENCIES|NATIVE_LIB_DIRECTORIES|PROCESS_MAX|PROGUARD|PROGUARD_CONFIG_PATH|SECURE_PROPS_PATH|SKIP_ANT_STEP|STL_TYPE)|ARCHIVE_OUTPUT_DIRECTORY|ARCHIVE_OUTPUT_NAME|ATTACHED_FILES|ATTACHED_FILES_ON_FAIL|AUTOGEN_(?:BUILD_DIR|ORIGIN_DEPENDS|PARALLEL|SOURCE_GROUP|TARGETS_FOLDER|TARGET_DEPENDS)|AUTOMOC|AUTOMOC_(?:COMPILER_PREDEFINES|DEPEND_FILTERS|EXECUTABLE|MACRO_NAMES|MOC_OPTIONS|SOURCE_GROUP|TARGETS_FOLDER)|AUTORCC|AUTORCC_EXECUTABLE|AUTORCC_OPTIONS|AUTORCC_SOURCE_GROUP|AUTOUIC|AUTOUIC_EXECUTABLE|AUTOUIC_OPTIONS|AUTOUIC_SEARCH_PATHS|BINARY_DIR|BUILDSYSTEM_TARGETS|BUILD_RPATH|BUILD_RPATH_USE_ORIGIN|BUILD_WITH_INSTALL_NAME_DIR|BUILD_WITH_INSTALL_RPATH|BUNDLE|BUNDLE_EXTENSION|CACHE_VARIABLES|CLEAN_NO_CUSTOM|COMMON_LANGUAGE_RUNTIME|COMPATIBLE_INTERFACE_(?:BOOL|NUMBER_MAX|NUMBER_MIN|STRING)|COMPILE_(?:DEFINITIONS|FEATURES|FLAGS|OPTIONS|PDB_NAME|PDB_OUTPUT_DIRECTORY)|COST|CPACK_DESKTOP_SHORTCUTS|CPACK_NEVER_OVERWRITE|CPACK_PERMANENT|CPACK_STARTUP_SHORTCUTS|CPACK_START_MENU_SHORTCUTS|CPACK_WIX_ACL|CROSSCOMPILING_EMULATOR|CUDA_EXTENSIONS|CUDA_PTX_COMPILATION|CUDA_RESOLVE_DEVICE_SYMBOLS|CUDA_SEPARABLE_COMPILATION|CUDA_STANDARD|CUDA_STANDARD_REQUIRED|CXX_EXTENSIONS|CXX_STANDARD|CXX_STANDARD_REQUIRED|C_EXTENSIONS|C_STANDARD|C_STANDARD_REQUIRED|DEBUG_CONFIGURATIONS|DEBUG_POSTFIX|DEFINE_SYMBOL|DEFINITIONS|DEPENDS|DEPLOYMENT_ADDITIONAL_FILES|DEPLOYMENT_REMOTE_DIRECTORY|DISABLED|DISABLED_FEATURES|ECLIPSE_EXTRA_CPROJECT_CONTENTS|ECLIPSE_EXTRA_NATURES|ENABLED_FEATURES|ENABLED_LANGUAGES|ENABLE_EXPORTS|ENVIRONMENT|EXCLUDE_FROM_ALL|EXCLUDE_FROM_DEFAULT_BUILD|EXPORT_NAME|EXPORT_PROPERTIES|EXTERNAL_OBJECT|EchoString|FAIL_REGULAR_EXPRESSION|FIND_LIBRARY_USE_LIB32_PATHS|FIND_LIBRARY_USE_LIB64_PATHS|FIND_LIBRARY_USE_LIBX32_PATHS|FIND_LIBRARY_USE_OPENBSD_VERSIONING|FIXTURES_CLEANUP|FIXTURES_REQUIRED|FIXTURES_SETUP|FOLDER|FRAMEWORK|Fortran_FORMAT|Fortran_MODULE_DIRECTORY|GENERATED|GENERATOR_FILE_NAME|GENERATOR_IS_MULTI_CONFIG|GHS_INTEGRITY_APP|GHS_NO_SOURCE_GROUP_FILE|GLOBAL_DEPENDS_DEBUG_MODE|GLOBAL_DEPENDS_NO_CYCLES|GNUtoMS|HAS_CXX|HEADER_FILE_ONLY|HELPSTRING|IMPLICIT_DEPENDS_INCLUDE_TRANSFORM|IMPORTED|IMPORTED_(?:COMMON_LANGUAGE_RUNTIME|CONFIGURATIONS|GLOBAL|IMPLIB|LIBNAME|LINK_DEPENDENT_LIBRARIES|LINK_INTERFACE_(?:LANGUAGES|LIBRARIES|MULTIPLICITY)|LOCATION|NO_SONAME|OBJECTS|SONAME)|IMPORT_PREFIX|IMPORT_SUFFIX|INCLUDE_DIRECTORIES|INCLUDE_REGULAR_EXPRESSION|INSTALL_NAME_DIR|INSTALL_RPATH|INSTALL_RPATH_USE_LINK_PATH|INTERFACE_(?:AUTOUIC_OPTIONS|COMPILE_DEFINITIONS|COMPILE_FEATURES|COMPILE_OPTIONS|INCLUDE_DIRECTORIES|LINK_DEPENDS|LINK_DIRECTORIES|LINK_LIBRARIES|LINK_OPTIONS|POSITION_INDEPENDENT_CODE|SOURCES|SYSTEM_INCLUDE_DIRECTORIES)|INTERPROCEDURAL_OPTIMIZATION|IN_TRY_COMPILE|IOS_INSTALL_COMBINED|JOB_POOLS|JOB_POOL_COMPILE|JOB_POOL_LINK|KEEP_EXTENSION|LABELS|LANGUAGE|LIBRARY_OUTPUT_DIRECTORY|LIBRARY_OUTPUT_NAME|LINKER_LANGUAGE|LINK_(?:DEPENDS|DEPENDS_NO_SHARED|DIRECTORIES|FLAGS|INTERFACE_LIBRARIES|INTERFACE_MULTIPLICITY|LIBRARIES|OPTIONS|SEARCH_END_STATIC|SEARCH_START_STATIC|WHAT_YOU_USE)|LISTFILE_STACK|LOCATION|MACOSX_BUNDLE|MACOSX_BUNDLE_INFO_PLIST|MACOSX_FRAMEWORK_INFO_PLIST|MACOSX_PACKAGE_LOCATION|MACOSX_RPATH|MACROS|MANUALLY_ADDED_DEPENDENCIES|MEASUREMENT|MODIFIED|NAME|NO_SONAME|NO_SYSTEM_FROM_IMPORTED|OBJECT_DEPENDS|OBJECT_OUTPUTS|OSX_ARCHITECTURES|OUTPUT_NAME|PACKAGES_FOUND|PACKAGES_NOT_FOUND|PARENT_DIRECTORY|PASS_REGULAR_EXPRESSION|PDB_NAME|PDB_OUTPUT_DIRECTORY|POSITION_INDEPENDENT_CODE|POST_INSTALL_SCRIPT|PREDEFINED_TARGETS_FOLDER|PREFIX|PRE_INSTALL_SCRIPT|PRIVATE_HEADER|PROCESSORS|PROCESSOR_AFFINITY|PROJECT_LABEL|PUBLIC_HEADER|REPORT_UNDEFINED_PROPERTIES|REQUIRED_FILES|RESOURCE|RESOURCE_LOCK|RULE_LAUNCH_COMPILE|RULE_LAUNCH_CUSTOM|RULE_LAUNCH_LINK|RULE_MESSAGES|RUNTIME_OUTPUT_DIRECTORY|RUNTIME_OUTPUT_NAME|RUN_SERIAL|SKIP_AUTOGEN|SKIP_AUTOMOC|SKIP_AUTORCC|SKIP_AUTOUIC|SKIP_BUILD_RPATH|SKIP_RETURN_CODE|SOURCES|SOURCE_DIR|SOVERSION|STATIC_LIBRARY_FLAGS|STATIC_LIBRARY_OPTIONS|STRINGS|SUBDIRECTORIES|SUFFIX|SYMBOLIC|TARGET_ARCHIVES_MAY_BE_SHARED_LIBS|TARGET_MESSAGES|TARGET_SUPPORTS_SHARED_LIBS|TESTS|TEST_INCLUDE_FILE|TEST_INCLUDE_FILES|TIMEOUT|TIMEOUT_AFTER_MATCH|TYPE|USE_FOLDERS|VALUE|VARIABLES|VERSION|VISIBILITY_INLINES_HIDDEN|VS_(?:CONFIGURATION_TYPE|COPY_TO_OUT_DIR|DEBUGGER_(?:COMMAND|COMMAND_ARGUMENTS|ENVIRONMENT|WORKING_DIRECTORY)|DEPLOYMENT_CONTENT|DEPLOYMENT_LOCATION|DOTNET_REFERENCES|DOTNET_REFERENCES_COPY_LOCAL|GLOBAL_KEYWORD|GLOBAL_PROJECT_TYPES|GLOBAL_ROOTNAMESPACE|INCLUDE_IN_VSIX|IOT_STARTUP_TASK|KEYWORD|RESOURCE_GENERATOR|SCC_AUXPATH|SCC_LOCALPATH|SCC_PROJECTNAME|SCC_PROVIDER|SDK_REFERENCES|SHADER_(?:DISABLE_OPTIMIZATIONS|ENABLE_DEBUG|ENTRYPOINT|FLAGS|MODEL|OBJECT_FILE_NAME|OUTPUT_HEADER_FILE|TYPE|VARIABLE_NAME)|STARTUP_PROJECT|TOOL_OVERRIDE|USER_PROPS|WINRT_COMPONENT|WINRT_EXTENSIONS|WINRT_REFERENCES|XAML_TYPE)|WILL_FAIL|WIN32_EXECUTABLE|WINDOWS_EXPORT_ALL_SYMBOLS|WORKING_DIRECTORY|WRAP_EXCLUDE|XCODE_(?:EMIT_EFFECTIVE_PLATFORM_NAME|EXPLICIT_FILE_TYPE|FILE_ATTRIBUTES|LAST_KNOWN_FILE_TYPE|PRODUCT_TYPE|SCHEME_(?:ADDRESS_SANITIZER|ADDRESS_SANITIZER_USE_AFTER_RETURN|ARGUMENTS|DISABLE_MAIN_THREAD_CHECKER|DYNAMIC_LIBRARY_LOADS|DYNAMIC_LINKER_API_USAGE|ENVIRONMENT|EXECUTABLE|GUARD_MALLOC|MAIN_THREAD_CHECKER_STOP|MALLOC_GUARD_EDGES|MALLOC_SCRIBBLE|MALLOC_STACK|THREAD_SANITIZER(?:_STOP)?|UNDEFINED_BEHAVIOUR_SANITIZER(?:_STOP)?|ZOMBIE_OBJECTS))|XCTEST)\b/
    ```
    With these new patterns in place the language definition is only ~11Kb large.
2. Please rerun gulp.

mjrogozinski:
> @mjrogozinski With that this PR is _almost_ ready to merge. Only 2 more things:
> 
> 1. Please use optimized keyword lists. Right now, the language definition is 24Kb large with most of the size coming from `variable` and `property`. So by using the Regex syntax to group the words, we can drastically reduce the file size.
>    I also used another trick to reduce the size of `variable`: It contains `CMAKE_\w+`, meaning that all words stating with `CMAKE_` are variables. As a result we can remove all other words which start with `CMAKE_` from the list.
>    The resulting new patterns are the following:
>    ```js
>    'variable': /\b(?:CMAKE_\w+|\w+_(?:VERSION(?:_MAJOR|_MINOR|_PATCH|_TWEAK)?|(?:BINARY|SOURCE)_DIR|DESCRIPTION|HOMEPAGE_URL|ROOT)|(?:CTEST_CUSTOM_(?:MAXIMUM_(?:(?:FAIL|PASS)ED_TEST_OUTPUT_SIZE|NUMBER_OF_(?:ERROR|WARNING)S)|ERROR_(?:P(?:OST|RE)_CONTEXT|EXCEPTION|MATCH)|P(?:OST|RE)_MEMCHECK|WARNING_(?:EXCEPTION|MATCH)|(?:MEMCHECK|TESTS)_IGNORE|P(?:OST|RE)_TEST|COVERAGE_EXCLUDE)|ANDROID|APPLE|BORLAND|BUILD_SHARED_LIBS|CACHE|CPACK_(?:ABSOLUTE_DESTINATION_FILES|COMPONENT_INCLUDE_TOPLEVEL_DIRECTORY|ERROR_ON_ABSOLUTE_INSTALL_DESTINATION|INCLUDE_TOPLEVEL_DIRECTORY|INSTALL_DEFAULT_DIRECTORY_PERMISSIONS|INSTALL_SCRIPT|PACKAGING_INSTALL_PREFIX|SET_DESTDIR|WARN_ON_ABSOLUTE_INSTALL_DESTINATION)|CTEST_(?:BINARY_DIRECTORY|BUILD_COMMAND|BUILD_NAME|BZR_COMMAND|BZR_UPDATE_OPTIONS|CHANGE_ID|CHECKOUT_COMMAND|CONFIGURATION_TYPE|CONFIGURE_COMMAND|COVERAGE_COMMAND|COVERAGE_EXTRA_FLAGS|CURL_OPTIONS|CUSTOM_(?:COVERAGE_EXCLUDE|ERROR_EXCEPTION|ERROR_MATCH|ERROR_POST_CONTEXT|ERROR_PRE_CONTEXT|MAXIMUM_FAILED_TEST_OUTPUT_SIZE|MAXIMUM_NUMBER_OF_(?:ERRORS|WARNINGS)|MAXIMUM_PASSED_TEST_OUTPUT_SIZE|MEMCHECK_IGNORE|POST_MEMCHECK|POST_TEST|PRE_MEMCHECK|PRE_TEST|TESTS_IGNORE|WARNING_EXCEPTION|WARNING_MATCH)|CVS_CHECKOUT|CVS_COMMAND|CVS_UPDATE_OPTIONS|DROP_LOCATION|DROP_METHOD|DROP_SITE|DROP_SITE_CDASH|DROP_SITE_PASSWORD|DROP_SITE_USER|EXTRA_COVERAGE_GLOB|GIT_COMMAND|GIT_INIT_SUBMODULES|GIT_UPDATE_CUSTOM|GIT_UPDATE_OPTIONS|HG_COMMAND|HG_UPDATE_OPTIONS|LABELS_FOR_SUBPROJECTS|MEMORYCHECK_(?:COMMAND|COMMAND_OPTIONS|SANITIZER_OPTIONS|SUPPRESSIONS_FILE|TYPE)|NIGHTLY_START_TIME|P4_CLIENT|P4_COMMAND|P4_OPTIONS|P4_UPDATE_OPTIONS|RUN_CURRENT_SCRIPT|SCP_COMMAND|SITE|SOURCE_DIRECTORY|SUBMIT_URL|SVN_COMMAND|SVN_OPTIONS|SVN_UPDATE_OPTIONS|TEST_LOAD|TEST_TIMEOUT|TRIGGER_SITE|UPDATE_COMMAND|UPDATE_OPTIONS|UPDATE_VERSION_ONLY|USE_LAUNCHERS)|CYGWIN|ENV|EXECUTABLE_OUTPUT_PATH|GHS-MULTI|IOS|LIBRARY_OUTPUT_PATH|MINGW|MSVC(?:10|11|12|14|60|70|71|80|90|_IDE|_TOOLSET_VERSION|_VERSION)?|MSYS|PROJECT_(?:BINARY_DIR|DESCRIPTION|HOMEPAGE_URL|NAME|SOURCE_DIR|VERSION|VERSION_(?:MAJOR|MINOR|PATCH|TWEAK))|UNIX|WIN32|WINCE|WINDOWS_PHONE|WINDOWS_STORE|XCODE|XCODE_VERSION))\b/,
>    'property': /\b(?:(?:ARCHIVE_OUTPUT_(?:DIRECTORY|NAME)|COMPILE_DEFINITIONS|COMPILE_PDB_NAME|COMPILE_PDB_OUTPUT_DIRECTORY|EXCLUDE_FROM_DEFAULT_BUILD|IMPORTED_(?:IMPLIB|LIBNAME|LINK_DEPENDENT_LIBRARIES|LINK_INTERFACE_LANGUAGES|LINK_INTERFACE_LIBRARIES|LINK_INTERFACE_MULTIPLICITY|LOCATION|NO_SONAME|OBJECTS|SONAME)|INTERPROCEDURAL_OPTIMIZATION|LIBRARY_OUTPUT_DIRECTORY|LIBRARY_OUTPUT_NAME|LINK_FLAGS|LINK_INTERFACE_LIBRARIES|LINK_INTERFACE_MULTIPLICITY|LOCATION|MAP_IMPORTED_CONFIG|OSX_ARCHITECTURES|OUTPUT_NAME|PDB_NAME|PDB_OUTPUT_DIRECTORY|RUNTIME_OUTPUT_DIRECTORY|RUNTIME_OUTPUT_NAME|STATIC_LIBRARY_FLAGS|VS_CSHARP|VS_DOTNET_REFERENCEPROP|VS_DOTNET_REFERENCE|VS_GLOBAL_SECTION_POST|VS_GLOBAL_SECTION_PRE|VS_GLOBAL|XCODE_ATTRIBUTE)_\w+|\w+_(?:CLANG_TIDY|COMPILER_LAUNCHER|CPPCHECK|CPPLINT|INCLUDE_WHAT_YOU_USE|OUTPUT_NAME|POSTFIX|VISIBILITY_PRESET)|ABSTRACT|ADDITIONAL_MAKE_CLEAN_FILES|ADVANCED|ALIASED_TARGET|ALLOW_DUPLICATE_CUSTOM_TARGETS|ANDROID_(?:ANT_ADDITIONAL_OPTIONS|API|API_MIN|ARCH|ASSETS_DIRECTORIES|GUI|JAR_DEPENDENCIES|NATIVE_LIB_DEPENDENCIES|NATIVE_LIB_DIRECTORIES|PROCESS_MAX|PROGUARD|PROGUARD_CONFIG_PATH|SECURE_PROPS_PATH|SKIP_ANT_STEP|STL_TYPE)|ARCHIVE_OUTPUT_DIRECTORY|ARCHIVE_OUTPUT_NAME|ATTACHED_FILES|ATTACHED_FILES_ON_FAIL|AUTOGEN_(?:BUILD_DIR|ORIGIN_DEPENDS|PARALLEL|SOURCE_GROUP|TARGETS_FOLDER|TARGET_DEPENDS)|AUTOMOC|AUTOMOC_(?:COMPILER_PREDEFINES|DEPEND_FILTERS|EXECUTABLE|MACRO_NAMES|MOC_OPTIONS|SOURCE_GROUP|TARGETS_FOLDER)|AUTORCC|AUTORCC_EXECUTABLE|AUTORCC_OPTIONS|AUTORCC_SOURCE_GROUP|AUTOUIC|AUTOUIC_EXECUTABLE|AUTOUIC_OPTIONS|AUTOUIC_SEARCH_PATHS|BINARY_DIR|BUILDSYSTEM_TARGETS|BUILD_RPATH|BUILD_RPATH_USE_ORIGIN|BUILD_WITH_INSTALL_NAME_DIR|BUILD_WITH_INSTALL_RPATH|BUNDLE|BUNDLE_EXTENSION|CACHE_VARIABLES|CLEAN_NO_CUSTOM|COMMON_LANGUAGE_RUNTIME|COMPATIBLE_INTERFACE_(?:BOOL|NUMBER_MAX|NUMBER_MIN|STRING)|COMPILE_(?:DEFINITIONS|FEATURES|FLAGS|OPTIONS|PDB_NAME|PDB_OUTPUT_DIRECTORY)|COST|CPACK_DESKTOP_SHORTCUTS|CPACK_NEVER_OVERWRITE|CPACK_PERMANENT|CPACK_STARTUP_SHORTCUTS|CPACK_START_MENU_SHORTCUTS|CPACK_WIX_ACL|CROSSCOMPILING_EMULATOR|CUDA_EXTENSIONS|CUDA_PTX_COMPILATION|CUDA_RESOLVE_DEVICE_SYMBOLS|CUDA_SEPARABLE_COMPILATION|CUDA_STANDARD|CUDA_STANDARD_REQUIRED|CXX_EXTENSIONS|CXX_STANDARD|CXX_STANDARD_REQUIRED|C_EXTENSIONS|C_STANDARD|C_STANDARD_REQUIRED|DEBUG_CONFIGURATIONS|DEBUG_POSTFIX|DEFINE_SYMBOL|DEFINITIONS|DEPENDS|DEPLOYMENT_ADDITIONAL_FILES|DEPLOYMENT_REMOTE_DIRECTORY|DISABLED|DISABLED_FEATURES|ECLIPSE_EXTRA_CPROJECT_CONTENTS|ECLIPSE_EXTRA_NATURES|ENABLED_FEATURES|ENABLED_LANGUAGES|ENABLE_EXPORTS|ENVIRONMENT|EXCLUDE_FROM_ALL|EXCLUDE_FROM_DEFAULT_BUILD|EXPORT_NAME|EXPORT_PROPERTIES|EXTERNAL_OBJECT|EchoString|FAIL_REGULAR_EXPRESSION|FIND_LIBRARY_USE_LIB32_PATHS|FIND_LIBRARY_USE_LIB64_PATHS|FIND_LIBRARY_USE_LIBX32_PATHS|FIND_LIBRARY_USE_OPENBSD_VERSIONING|FIXTURES_CLEANUP|FIXTURES_REQUIRED|FIXTURES_SETUP|FOLDER|FRAMEWORK|Fortran_FORMAT|Fortran_MODULE_DIRECTORY|GENERATED|GENERATOR_FILE_NAME|GENERATOR_IS_MULTI_CONFIG|GHS_INTEGRITY_APP|GHS_NO_SOURCE_GROUP_FILE|GLOBAL_DEPENDS_DEBUG_MODE|GLOBAL_DEPENDS_NO_CYCLES|GNUtoMS|HAS_CXX|HEADER_FILE_ONLY|HELPSTRING|IMPLICIT_DEPENDS_INCLUDE_TRANSFORM|IMPORTED|IMPORTED_(?:COMMON_LANGUAGE_RUNTIME|CONFIGURATIONS|GLOBAL|IMPLIB|LIBNAME|LINK_DEPENDENT_LIBRARIES|LINK_INTERFACE_(?:LANGUAGES|LIBRARIES|MULTIPLICITY)|LOCATION|NO_SONAME|OBJECTS|SONAME)|IMPORT_PREFIX|IMPORT_SUFFIX|INCLUDE_DIRECTORIES|INCLUDE_REGULAR_EXPRESSION|INSTALL_NAME_DIR|INSTALL_RPATH|INSTALL_RPATH_USE_LINK_PATH|INTERFACE_(?:AUTOUIC_OPTIONS|COMPILE_DEFINITIONS|COMPILE_FEATURES|COMPILE_OPTIONS|INCLUDE_DIRECTORIES|LINK_DEPENDS|LINK_DIRECTORIES|LINK_LIBRARIES|LINK_OPTIONS|POSITION_INDEPENDENT_CODE|SOURCES|SYSTEM_INCLUDE_DIRECTORIES)|INTERPROCEDURAL_OPTIMIZATION|IN_TRY_COMPILE|IOS_INSTALL_COMBINED|JOB_POOLS|JOB_POOL_COMPILE|JOB_POOL_LINK|KEEP_EXTENSION|LABELS|LANGUAGE|LIBRARY_OUTPUT_DIRECTORY|LIBRARY_OUTPUT_NAME|LINKER_LANGUAGE|LINK_(?:DEPENDS|DEPENDS_NO_SHARED|DIRECTORIES|FLAGS|INTERFACE_LIBRARIES|INTERFACE_MULTIPLICITY|LIBRARIES|OPTIONS|SEARCH_END_STATIC|SEARCH_START_STATIC|WHAT_YOU_USE)|LISTFILE_STACK|LOCATION|MACOSX_BUNDLE|MACOSX_BUNDLE_INFO_PLIST|MACOSX_FRAMEWORK_INFO_PLIST|MACOSX_PACKAGE_LOCATION|MACOSX_RPATH|MACROS|MANUALLY_ADDED_DEPENDENCIES|MEASUREMENT|MODIFIED|NAME|NO_SONAME|NO_SYSTEM_FROM_IMPORTED|OBJECT_DEPENDS|OBJECT_OUTPUTS|OSX_ARCHITECTURES|OUTPUT_NAME|PACKAGES_FOUND|PACKAGES_NOT_FOUND|PARENT_DIRECTORY|PASS_REGULAR_EXPRESSION|PDB_NAME|PDB_OUTPUT_DIRECTORY|POSITION_INDEPENDENT_CODE|POST_INSTALL_SCRIPT|PREDEFINED_TARGETS_FOLDER|PREFIX|PRE_INSTALL_SCRIPT|PRIVATE_HEADER|PROCESSORS|PROCESSOR_AFFINITY|PROJECT_LABEL|PUBLIC_HEADER|REPORT_UNDEFINED_PROPERTIES|REQUIRED_FILES|RESOURCE|RESOURCE_LOCK|RULE_LAUNCH_COMPILE|RULE_LAUNCH_CUSTOM|RULE_LAUNCH_LINK|RULE_MESSAGES|RUNTIME_OUTPUT_DIRECTORY|RUNTIME_OUTPUT_NAME|RUN_SERIAL|SKIP_AUTOGEN|SKIP_AUTOMOC|SKIP_AUTORCC|SKIP_AUTOUIC|SKIP_BUILD_RPATH|SKIP_RETURN_CODE|SOURCES|SOURCE_DIR|SOVERSION|STATIC_LIBRARY_FLAGS|STATIC_LIBRARY_OPTIONS|STRINGS|SUBDIRECTORIES|SUFFIX|SYMBOLIC|TARGET_ARCHIVES_MAY_BE_SHARED_LIBS|TARGET_MESSAGES|TARGET_SUPPORTS_SHARED_LIBS|TESTS|TEST_INCLUDE_FILE|TEST_INCLUDE_FILES|TIMEOUT|TIMEOUT_AFTER_MATCH|TYPE|USE_FOLDERS|VALUE|VARIABLES|VERSION|VISIBILITY_INLINES_HIDDEN|VS_(?:CONFIGURATION_TYPE|COPY_TO_OUT_DIR|DEBUGGER_(?:COMMAND|COMMAND_ARGUMENTS|ENVIRONMENT|WORKING_DIRECTORY)|DEPLOYMENT_CONTENT|DEPLOYMENT_LOCATION|DOTNET_REFERENCES|DOTNET_REFERENCES_COPY_LOCAL|GLOBAL_KEYWORD|GLOBAL_PROJECT_TYPES|GLOBAL_ROOTNAMESPACE|INCLUDE_IN_VSIX|IOT_STARTUP_TASK|KEYWORD|RESOURCE_GENERATOR|SCC_AUXPATH|SCC_LOCALPATH|SCC_PROJECTNAME|SCC_PROVIDER|SDK_REFERENCES|SHADER_(?:DISABLE_OPTIMIZATIONS|ENABLE_DEBUG|ENTRYPOINT|FLAGS|MODEL|OBJECT_FILE_NAME|OUTPUT_HEADER_FILE|TYPE|VARIABLE_NAME)|STARTUP_PROJECT|TOOL_OVERRIDE|USER_PROPS|WINRT_COMPONENT|WINRT_EXTENSIONS|WINRT_REFERENCES|XAML_TYPE)|WILL_FAIL|WIN32_EXECUTABLE|WINDOWS_EXPORT_ALL_SYMBOLS|WORKING_DIRECTORY|WRAP_EXCLUDE|XCODE_(?:EMIT_EFFECTIVE_PLATFORM_NAME|EXPLICIT_FILE_TYPE|FILE_ATTRIBUTES|LAST_KNOWN_FILE_TYPE|PRODUCT_TYPE|SCHEME_(?:ADDRESS_SANITIZER|ADDRESS_SANITIZER_USE_AFTER_RETURN|ARGUMENTS|DISABLE_MAIN_THREAD_CHECKER|DYNAMIC_LIBRARY_LOADS|DYNAMIC_LINKER_API_USAGE|ENVIRONMENT|EXECUTABLE|GUARD_MALLOC|MAIN_THREAD_CHECKER_STOP|MALLOC_GUARD_EDGES|MALLOC_SCRIBBLE|MALLOC_STACK|THREAD_SANITIZER(?:_STOP)?|UNDEFINED_BEHAVIOUR_SANITIZER(?:_STOP)?|ZOMBIE_OBJECTS))|XCTEST)\b/
>    ```
>    
>    
>    With these new patterns in place the language definition is only ~11Kb large.
> 2. Please rerun gulp.

I think we can do a bit more optimizations. I will post them soon. e.g. keyword containing `LIBRARIES`, `DIRECTORIES`, `DEPENDENCIES`, `INTERFACE`, `CXX`. It wouldn't hurt to treat everything containing those and perhaps some more words as variables or properties.
I will try to come up with an optimized regex.

There is one more thing. CMake's compilation features like `cxx_std_17`, `cxx_constexpr`, basically `cxx_\w+`. I will also take a look at that and try to push some commits regarding those as well.

RunDevelopment:
> I think we can do a bit more optimizations. I will post them soon.

I appreciate it but you really don't have to. The patterns from above have the optimizations which saved the most. You can further optimize but the gain won't be as high, so I don't think it's worth putting that much effort into it.

> It wouldn't hurt to treat everything containing those and perhaps some more words as variables or properties.

This might cause a lot of false positives, so I don't think we should lose correctness of small file sizes.

> There is one more thing. CMake's compilation features like cxx_std_17, cxx_constexpr, basically cxx_\w+. I will also take a look at that and try to push some commits regarding those as well.

Nice!

Regarding new features want to add variable interpolation as well?
E.g. `math(EXPR RES "${A} + ${B}")`

Change `string` to the following:

```js
'string': {
	pattern: /"(?:[^\\"]|\\.)*"/,
	greedy: true,
	inside: {
		'interpolation': {
			pattern: /\${(?:[^{}$]|\${[^{}$]*})*}/,
			inside: {
				'punctuation': /\${|}/,
				'variable': /\w+/
			}
		}
	}
},
```

<details>
<summary>Results</summary>

image

</details>

mjrogozinski:
> > I think we can do a bit more optimizations. I will post them soon.
> 
> I appreciate it but you really don't have to. The patterns from above have the optimizations which saved the most. You can further optimize but the gain won't be as high, so I don't think it's worth putting that much effort into it.
> 
> > It wouldn't hurt to treat everything containing those and perhaps some more words as variables or properties.
> 
> This might cause a lot of false positives, so I don't think we should lose correctness of small file sizes.
> 
> > There is one more thing. CMake's compilation features like cxx_std_17, cxx_constexpr, basically cxx_\w+. I will also take a look at that and try to push some commits regarding those as well.
> 
> Nice!
> 
> Regarding new features: Want to add variable interpolation for strings as well?
> E.g. `math(EXPR RES "${A} + ${B}")`
> 
> Change `string` to the following:
> 
> ```js
> 'string': {
> 	pattern: /"(?:[^\\"]|\\.)*"/,
> 	greedy: true,
> 	inside: {
> 		'interpolation': {
> 			pattern: /\${(?:[^{}$]|\${[^{}$]*})*}/,
> 			inside: {
> 				'punctuation': /\${|}/,
> 				'variable': /\w+/
> 			}
> 		}
> 	}
> },
> ```
> 
> Results

This is awesome. It took me some time to figure out where to put commas in unit tests but I managed :)
I've added an :: operator that is used to define targets. If we had a new tag, it could be `\w+\:\:\w+`, meaning that `whatever::whenever` should be highlighted as it is unmistakably a target in CMake's definition.

mjrogozinski:
What would you say about: `'inserted': /\b(\w+\:\:\w+)\b/,`?

mjrogozinski:
> What would you say about: `'inserted': /\b(\w+\:\:\w+)\b/,`?

Acutally `'inserted': /\b(\w+\:\:\w+)\b/i,`, since `Foo::Bar` is also a target.

RunDevelopment:
> What would you say about: `'inserted': /\b(\w+\:\:\w+)\b/`?

Make it the following and I'll take it.
```js
'inserted': {
	pattern: /\b\w+::\w+\b/,
	inside: {
		'punctuation': /::/
	}
}
```

Btw. What is this feature even? I've seen it in CMake examples before but I didn't get its purpose.

mjrogozinski:
> > What would you say about: `'inserted': /\b(\w+\:\:\w+)\b/`?
> 
> Make it the following and I'll take it.
> 
> ```js
> 'inserted': {
> 	pattern: /\b\w+::\w+\b/,
> 	inside: {
> 		'punctuation': /::/
> 	}
> }
> ```
> 
> Btw. What is this feature even? I've seen it in CMake examples before but I didn't get its purpose.
> 
> PS: Are you sure that neither of the words left and right of `::` can be matched by `variable`, `property`, etc.?

Acutally `::` is not punctuation, but part of the target name. 

What this feature does is enable CMake to know for sure that what you are linking to, it a target. If a target containing `::` is not defined, CMake can fail on generation step instead of during linking of a nonexistent library.

It is partially becuase of legacy reasons, because function `target_link_libraries` is polymorphic in a sense and it does not know whether you want to link to an existing target (often carrying stuff like include path, compiler flags etc) or just a library (an .so, .a, .dll) without additional info.

RunDevelopment:
Hmmm, in that case, it might be nice to give it some color. The `class-name` alias fits nicely. Also, you don't need the escapes for `:`, the capturing group, and the `i` flag.

Result:
```js
'inserted': {
	pattern: /\b\w+::\w+\b/,
	alias: 'class-name'
}
```

mjrogozinski:
> Hmmm, in that case, it might be nice to give it some color. The `class-name` alias fits nicely. Also, you don't need the escapes for `:`, the capturing group, and the `i` flag.
> 
> Result:
> 
> ```js
> 'inserted': {
> 	pattern: /\b\w+::\w+\b/,
> 	alias: 'class-name'
> }
> ```

Pushed

RunDevelopment:
Thank you very much for contributing @mjrogozinski!

mjrogozinski:
> Thank you very much for contributing @mjrogozinski!

Thank you very much for helping me :) It was a great pleasure to work on this with you Michael.

PR #1824, opened 2019-03-22
Change case of name in composer.json

charlieartist:
I'm looking to add Prism to a Drupal 8 project via Composer. In setting up a Packagist repository off master, I was hit with errors, since the name of the project in composer.json must be lowercase. I changed this line from: `  "name": "PrismJS/prism",` to `  "name": "prismjs/prism",`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-  "name": "PrismJS/prism",
+  "name": "prismjs/prism",
   "description": "Lightweight, robust, elegant syntax highlighting. A spin-off project from Dabblet.",
   "type": "component",
   "license": "MIT",


RunDevelopment:
> [...], since the name of the project in composer.json must be lowercase.

According to [the docs](https://getcomposer.org/doc/04-schema.md#name), the package name "_can contain any character, including white spaces, and it's case insensitive_". Though, it is recommended to use lowercase names.

Point is, our current configuration should be perfectly valid, so I don't get why it doesn't work on your side.

Anyway, because the names are supposed to be case insensitive, it _shouldn't_ be a problem to change it.
I'm not a PHP dev, so I definitely want a second opinion for this.

/cc @mAAdhaTTah

charlieartist:
@RunDevelopment This isn't per Composer, but per [Packagist](https://packagist.org/about).

Listed here:

> Note that package names are case-insensitive, but it's encouraged to use a dash (-) as separator instead of CamelCased names.

When I attempted to add the repository to Packagist, I was blocked because of the naming.

RunDevelopment:
So package names are case-insensitive __and__ can only contain lowercase characters.
Then what does _case-insensitive_ even mean here? After all, it's meaningless to say that you don't care about the case when only one case is allowed...

Can we really change the package name without breaking the dependencies of people who are using the current name?

charlieartist:
@rundevelopment I would agree. It is mystifying why it is listed as such. Ideally, Packagist would get it together. Now, I did check other Composer packages, and those all had lowercase names.

PR #1829, opened 2019-03-23
Autoloader: Added support for aliases

### LINKED ISSUE #1827: Can't load language plugins via aliases even when they are officially provided
**Information:**
- Prism version: 1.15
- Environment: Browser

**Description**
Prism fails to load language plugins via aliases because of use of aliases for plugin path resolving.

**Example**
TypeScript language plugin has `ts` alias by default but isn't loadable.

https://falsandtru.github.io/securemark/#%60%60%60ts%0Aconst%20a%20%3D%200%3B%0A%60%60%60

```
cdnjs.cloudflare.com/ajax/libs/prism/1.15.0/components/prism-ts.min.js:1 Failed to load resource: the server responded with a status of 404 ()
```

RunDevelopment:
This PR adds alias support to the [Autoloader plugin](https://prismjs.com/plugins/autoloader/).

The bug that the dependencies of forced-loaded languages in `data-dependencies` were not loaded is also fixed.
I.e. `<code class="language-pug" data-dependencies="less!">` the dependencies of `less` were not loaded.

---

This resolves #1827.


falsandtru:
Can you expose lang_aliases to enable custom aliases?

RunDevelopment:
Before thinking about custom aliases, are 'normal' aliases not enough?
For what purpose do you need them?

falsandtru:
I want to define the `hs` alias for haskell. Looks like I have to register that alias to lang_aliases.

RunDevelopment:
`hs` is the file extension of Haskell files, right? Then making `hs` an alias for `haskell` is an easy solution to your problem.
Also, there are others who might also want this alias, so making this part of Prism is probably a better idea than introducing custom aliases.

If there are any other aliases you want, feel free to request them, I'll add them.

If you want to add them yourself:
1. Add a line `Prism.language.myNewAlias = Prism.languages.mylanguage;` to the `prism-mylanguage.js` file in the `components/` directory.
2. Open `components.json` and add an `alias` property to the entry `languages.mylanguage` and set its value to the added alias. E.g. `"alias": "myNewAlias "` or `"alias": ["firstAlias", "myNewAlias "]`.
3. Run the `npx gulp` command after installing all npm dependencies.
4. Commit and make a PR (without `package-lock.json`)!

falsandtru:
I see, could you add `hs` alias with #1830? Since `hs` alias is adopted with GitHub it can be considered a common alias.

RunDevelopment:
I'll make a new PR (#1831).

falsandtru:
Thanks!

mAAdhaTTah:
We'll also want to update the gulp annotations PR, which may be worth revisiting now.

RunDevelopment:
I closed #1535 because I want to get rid of partly generate files in the end.
My goal is to implement #1539 (or similar) so that we generate files that these plugins then require. This will also get rid of the annoying merge conflicts I get almost every time in `prism-autoloader.js` and `prism-show-language.js`.

PR #1837, opened 2019-03-25
Core: Removed `parent` from `wrap`'s environment

### LINKED ISSUE #1836: What is the `parent` property in the environment of the `wrap` hook for?
I don't get what the [`parent`](https://github.com/PrismJS/prism/blob/master/components/prism-core.js#L480) property is for.

We don't have a plugin that is using it doesn't even is what the name suggests: `parent` actually holds the nearest token array; i.e. in the case that the `content` of a token is a single token, then `parent` will not be the parent token but the token stream of the parent.

Apart from that, I also don't see why it would be useful:
Assuming we wrap a token which is in a token stream: What can `parent` be used for?
We can alter token positioned before the current token but that doesn't change the output of `stringify` because these token as stringified already. Changing the current token doesn't do anything either.
The only thing `parent` can be used for is to alter the items positioned _after_ the current token. Btw. we can't add or remove items because [`stringify` uses `map`](https://github.com/PrismJS/prism/blob/master/components/prism-core.js#L468).

So, what is it for?

Also, I don't see why the `wrap` hook should be able to modify the token stream anyway.

---

[Added 11 May 2013](https://github.com/PrismJS/prism/commit/801aab58cf3c1b684dd4e2755b83fb4250a6866a)

RunDevelopment:
This resolves #1836.


RunDevelopment:
Thank you @Golmote!

PR #1840, opened 2019-03-27
Made the download button its own plugin

### LINKED ISSUE #1816: Make the 'Download this file' button of File highlight its own plugin
Right now, a 'Download this button' is part of the [File highlight plugin](https://prismjs.com/plugins/file-highlight/).

I think that this little button should be its own plugin as this is not feature directly related to File highlight.

### LINKED ISSUE #1817: File highlight: Toolbar button is registered multiple times
**Information:**
- Prism version: latest
- Environment: Browser

**Description**
The [File highlight plugin](https://prismjs.com/plugins/file-highlight/) registers the 'Download this file' every time [`fileHighlight`](https://github.com/PrismJS/prism/blob/master/plugins/file-highlight/prism-file-highlight.js#L84) is executed. 
But [Toolbar](https://prismjs.com/plugins/toolbar) [can't deal with registering multiple callbacks on the same key](https://github.com/PrismJS/prism/blob/master/plugins/toolbar/prism-toolbar.js#L46).

**Possible solutions**

1. Adjust File highlight to register the button once and only once.
2. #1816 

Regardlessly, we should add a check to Toolbar so that you can't register on the same key twice. (#1818)

RunDevelopment:
This PR makes the download button from Autoloader its own plugin.

---

This resolves #1816.
This fixes #1817.


prism.js
```diff
@@ -934,22 +934,6 @@ Prism.languages.js = Prism.languages.javascript;
...
-					return;
-				}
-				var src = pre.getAttribute('data-src');
-				var a = document.createElement('a');
-				a.textContent = pre.getAttribute('data-download-link-label') || 'Download';
-				a.setAttribute('download', '');
-				a.href = src;
-				return a;
-			});
-		}
```
mAAdhaTTah: I think we need to add this to the build so the website has it again.

RunDevelopment: Since 1) the toolbar plugin isn't loaded and 2) none of our code blocks has the `data-download-link` attribute, the plugin isn't active, so why include it again?

mAAdhaTTah: I assumed the toolbar plugin was loaded, my b.

PR #1406, opened 2018-04-25
Adds solidity highlighting

glachaud:

PR #1843, opened 2019-03-27
Solidity support revisited

### LINKED ISSUE #1404: Support for Solidity?
It would be great to have support for solidity... https://solidity.readthedocs.io/en/v0.4.23/

wmitsuda:
Fixes #1404  . Revised PR from #1406


components/prism-solidity.js
```diff
@@ -0,0 +1,37 @@
...
+					inside: {
+						punctuation: /[(){};:]/
+					}
+				}
+			}
+		},
+		{
+			pattern: /\bmapping\s*\(.*\s+=>\s+.*\)(?:\s+(?:private|public|internal|external|inherited))?\s+[A-Za-z_]\w*\b/,
+			inside: {
+				keyword: {
```
RunDevelopment: Missing single quotes.

components/prism-solidity.js
```diff
@@ -0,0 +1,37 @@
...
+					pattern: /\b(address|string|bytes\d*|int\d*|uint\d*|bool|u?fixed\d+x\d+)\b(?:\s+(?:indexed\s+)?[A-Za-z_]\w*\s*[,)])?/,
+					lookbehind: true,
+					inside: {
+						punctuation: /[(){};:]/
+					}
+				}
+			}
+		},
+		{
+			pattern: /\bmapping\s*\(.*\s+=>\s+.*\)(?:\s+(?:private|public|internal|external|inherited))?\s+[A-Za-z_]\w*\b/,
```
RunDevelopment: I have a problem with `.*` as the dot also includes most of `\s+=>\s+` which invites catastrophic backtracking.
What exactly is `.*` supposed to match?

components/prism-solidity.js
```diff
@@ -0,0 +1,37 @@
...
+				}
+			}
+		},
+	],
+	'constant': [
+		{
+			pattern: /\b(?:address|string|bytes\d*|int\d*|uint\d*|bool|u?fixed\d+x\d+)\b(?:\s+(?:indexed\s+)?[A-Za-z_]\w*\s*[,)])?/,
+			inside: {
+				'attr-name' : {
+					pattern: /\b(address|string|bytes\d*|int\d*|uint\d*|bool|u?fixed\d+x\d+)\b(?:\s+(?:indexed\s+)?[A-Za-z_]\w*\s*[,)])?/,
```
RunDevelopment: I'm not really sure but isn't `(?:indexed\s+)?` supposed to be part of the lookbehind?

Also, shouldn't `\s*[,)]` be a lookahead rather than being part of the name?

Why is the whole non-lookbehind expression optional? This might cause empty tokens.

components/prism-solidity.js
```diff
@@ -0,0 +1,37 @@
+Prism.languages.solidity = Prism.languages.extend('javascript', {
+	'keyword': [
+		/\b(?:assert|break|constant|continue|do|else|emit|external|for|function|if|import|inherited|internal|is|memory|new|payable|private|public|pure|require|returns?|revert|selfdestruct|storage|suicide|this|throw|var|view|while|_)\b/,
+		{
+			pattern: /\b(?:contract|interface|library|using|struct|function|modifier)\s+[A-Za-z_]\w*(?:\s+is\s+(?:[A-Za-z_][,\s]*)*)?\b/,
+			inside: {
+				'variable': {
+					pattern: /\b(contract|interface|library|using|struct|function|modifier)\s+[A-Za-z_]\w*(?:\s+is\s+(?:[A-Za-z_][,\s]*)*)?\b/,
```
RunDevelopment: Please move the `\s+` inside the lookbehind. Tokens should contain leading or trailing spaces.

RunDevelopment:
@mAAdhaTTah 

This language would undoubtedly benefit from #1679. Just in case you [weren't](https://github.com/PrismJS/prism/pull/1679#issuecomment-471322306) convinced already 😉

components/prism-solidity.js
wmitsuda: The first .* is supposed to be a builtin type, the second one may be any type, even another mapping. Don't know if there is a better way to express this.

wmitsuda: For more information, this is the language specification: https://solidity.readthedocs.io/en/v0.5.7/types.html#mapping-types

components/prism-solidity.js
wmitsuda: Actually, this part is from the original patch from the original author, have to admit I didn't pay much attention to it. I think the original author made a mistake. Not sure exactly what besides a simple attribute declaration he intended to match. I'll remove what I think is unnecessary to make it work and update the patch.

components/prism-solidity.js
RunDevelopment: I don't what the most complex type name could be but wouldn't `\w+` instead of `.*` also work? So `\(\s*\w+\s*=>\s*\w+\s*\)`.

And if we were to allow _any_ content, then `\([^()]*\)` might be what we want.

wmitsuda: It is possible to have a mapping of mapping, for e.g.: `mapping(address => mapping(uint => customStruct[])) someName`, and so on... how can I model this recursion on prism? Is there some good example I can follow?

RunDevelopment: Matching such expression correctly requires a context-free grammar, so we can't match them with regexes. It's fundamentally impossible.
But we can approximate by matching a limited number of recursion levels.

E.g. `\((?:[^()]|\((?:[^()]|\([^()]*\))*\))*\)` will match recusive bracket expression for up to a maximum depth of 3.

This is how we usually deal with nested expression.

davodesign84:
Any plan to merge this PR? Is the last blocker to migrate from highlight.js to Prism for us.

RunDevelopment:
Since the author never got to make the requested changes, I can make the required changes within the next few days.

PR #1851, opened 2019-03-31
Added Autoloader SRI

### LINKED ISSUE #1680: Secure autoloading by integrity
Prism doesn't check integrity of extension scripts with autoloading. Can you make Prism secure by checking integrity? All the integrity is listed in https://cdnjs.com/libraries/prism.

FYI: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

RunDevelopment:
This PR adds a plugin which extends Autoloader to add [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity).

---

This resolves #1680.


RunDevelopment:
As pointed out in #1870, we can't merge this until we have resolved the versioning issue.

We should also only merge this right before the next release. Because otherwise, until then, we'll have a plugin available on the download page which is practically useless because of the above-mentioned issue.

Maybe a milestone, so we don't forget about it?

plugins/autoloader/prism-autoloader.js
```diff
@@ -25,18 +25,37 @@
 	}
 	var config = Prism.plugins.autoloader = {
 		languages_path: languages_path,
-		use_minified: true
+		use_minified: true,
+		sri: {
```
mAAdhaTTah: Instead of an `sri` key specifically, can we modify the autoloader so it's hookable so we can implement the sri? That would be more generic and make the autoloader less likely to change in the future. May even be less code, if some of the current logic is moved into the sri plugin.

plugins/autoloader/prism-autoloader.js
```diff
@@ -25,18 +25,37 @@
...
+				return;
+			}
+		}
+
 		var s = document.createElement('script');
 		s.src = src;
+		if (integrity) {
+			s.integrity = integrity;
+			s.crossOrigin = 'anonymous';
+		}
```
mAAdhaTTah: Hook a callback in here so the sri plugin can modify the script tag?

JaKXz:
Is this still relevant (it looks good to me at glance but just curious)? Should it be added to 2.0?

RunDevelopment:
> Should it be added to 2.0?

Probably not, but I'm not too sure. The main blocker for this is that our download page gets everything straight from master. This is a problem because the SRI hashes are (Prism) version specific. 

Some users might download a small bundle from our download and then wire it up with a CDN to autoload the rest. This won't work because the SRI hashes (from autoloader included in the bundle) won't match up with the hashes of  the CDN files.

PR #1856, opened 2019-04-03
'semicolon'

andrusha19:
Add semicolon. After wro4j processing browsers display error at this row. Adding semicolon fixes this issue


RunDevelopment:
Thank you for contributing!

RunDevelopment:
#1796 will add the remaining missing semicolons.

PR #1862, opened 2019-04-18
Fix for uncaught errors when 'commandLine' is an empty object.

libantema:


plugins/command-line/prism-command-line.js
```diff
@@ -79,7 +79,7 @@ Prism.hooks.add('before-insert', function (env) {
 
 	// Reinsert the output lines into the highlighted code. -- cwells
 	var codeLines = env.highlightedCode.split('\n');
-	for (var i = 0; i < commandLine.outputLines.length; i++) {
+	for (var i = 0; i < (commandLine.outputLines.length || 0); i++) {
```
RunDevelopment: If `commandLine` empty, `commandLine.outputLines` will be `undefined` meaning that `commandLine.outputLines.length` will throw a reference error.

```suggestion
	for (var i = 0, l = (commandLine.outputLines || []).length; i < l; i++) {
```

plugins/command-line/prism-command-line.js
```diff
@@ -123,7 +123,7 @@ Prism.hooks.add('complete', function (env) {
 	prompt.innerHTML = promptLines;
 
 	// Remove the prompt from the output lines. -- cwells
-	for (var i = 0; i < commandLine.outputLines.length; i++) {
+	for (var i = 0; i < (commandLine.outputLines.length || 0); i++) {
```
RunDevelopment:
```suggestion
	for (var i = 0, l = (commandLine.outputLines || []).length; i < l; i++) {
```

RunDevelopment:
For other reviewers:

This bug is most likely caused by [this line](https://github.com/PrismJS/prism/blob/d2089d83509c1bcc3a5e74cbfc896680b2f76560/components/prism-core.js#L239) in Core.
Because the [`before-highlight`](https://github.com/PrismJS/prism/blob/d2089d83509c1bcc3a5e74cbfc896680b2f76560/plugins/command-line/prism-command-line.js#L9) hook is not run before [`complete`](https://github.com/PrismJS/prism/blob/d2089d83509c1bcc3a5e74cbfc896680b2f76560/plugins/command-line/prism-command-line.js#L90), `env.vars['command-line']` is `undefined`.

libantema:
> Thank you for providing this fix!
> 
> I left you a comment for with a minor improvement. (Don't forget to rebuild Prism.)
> Also, why did `package-lock.json` change?

I like your suggestions, I will commit the changes. Thanks!
I did rebuild Prism using gulp. I'm not sure why package-lock.json changed - it was simply altered after running gulp in my enviro.

RunDevelopment:
@libantema Sorry for the delay!

For now, it's probably the best idea to just revert the changes to `package-lock.json`, so we can merge this fix.

RunDevelopment:
Thank you very much for contributing!

Issue #1860, opened 2019-04-11
prism-line-highlight is slow on resize because it forces a reflow for every


justinbmeyer:
**Information:**
- Prism version: `1.16.0`.
- Environment: Chrome Browser

Does the latest version of Prism from the [download page](https://prismjs.com/download.html) also have this issue? I'm using it from `npm`.

**Description**

The line-highlight plugin is causing multiple forced reflows every time I resize a page.  I have about 29 highlighted code blocks on the page. You can see this and the warning here:

![image](https://user-images.githubusercontent.com/78602/55927725-b2f6e080-5bdb-11e9-8895-a02b281932f0.png)

Looking at the code, there is a lot of repeated __read layout and write layout__ loops around here:

https://github.com/PrismJS/prism/blob/d2089d83509c1bcc3a5e74cbfc896680b2f76560/plugins/line-highlight/prism-line-highlight.js#L70

This is in a for-loop for each highlighted range, which is in another for loop for each code block.

Could this be changed to read the layout all at once and write the layout all at once?  If interested, __I will submit a PR for this__.  Here's [an example](https://www.bitovi.com/blog/canjs-4.0#queues) that shows the difference (500ms to 5ms) if batched read/writes are used.

Thanks!

Golmote:
Hello! Thanks for opening this issue. If you're willing to submit a PR to fix this, please do, it is much appreciated!

justinbmeyer:
[Here's the code](https://gist.github.com/justinbmeyer/1230ae64918372ea9bf124f6dce74326) (a PR will be created tomorrowish).  It changes my highlight time from about 400ms to about 2ms on the slowest performing page.

In short, I added a little helper that allows me to store calling a function or setting a property:

```js
function makeActions() {
	var actions = [];

	return {
		actions: actions,
		apply: function(fn, context, args){
			actions.push(function(){
				fn.apply(context, args);
			});
		},
		set: function(obj, key, value) {
			actions.push(function(){
				obj[key] = value;
			});
		}
	};
}
```

Then, instead of mutating the DOM, I use the helper:

```js
// line.setAttribute('data-range', currentRange);
action.apply(line.setAttribute, line,['data-range', currentRange]);

// line.style.top = startNode.offsetTop + 'px';
action.set(line.style, "top",  startNode.offsetTop + 'px');
```

And then return these actions to be called later:

```js
		var actions = [];
		Array.prototype.forEach.call(preElements, function (pre) {
[...]

PR #1865, opened 2019-04-23
Batching read / writes to improve performance

Mattchewone:
This PR improves performance by batching the read / writes. 

Related to - https://github.com/PrismJS/prism/issues/1860#issuecomment-485933067


RunDevelopment:
First of all, thank you for committing this fix!

Now, please run gulp to update the minified files (that's why the build failed).

components/prism-json.js
```diff
@@ -8,6 +7,7 @@ Prism.languages.json = {
 		pattern: /"(?:\\.|[^\\"\r\n])*"(?!\s*:)/,
 		greedy: true
 	},
+	'comment': /\/\/.*|\/\*[\s\S]*?(?:\*\/|$)/,
 	'number': /-?\d+\.?\d*(e[+-]?\d+)?/i,
```
RunDevelopment: What happened here?

Mattchewone: Hmm

Seems I reverted to the incorrect branch and pushed up some changes not part of my commit. Sorry let me fix that.

Mattchewone:
@RunDevelopment Hopefully this resolves the issues, should now include the changes to the file and the updated min file.

RunDevelopment:
Yup, that works! Thank you!

Mattchewone:
That's ok, sorry for making a bit of a mess of this PR. But we got there in the end.

plugins/line-highlight/prism-line-highlight.js
```diff
@@ -1,181 +1,211 @@
...
+		for (var i=0, currentRange; currentRange = ranges[i++];) {
+			var range = currentRange.split('-');
+	
+			var start = +range[0],
+				end = +range[1] || start;
+	
+			var line = pre.querySelector('.line-highlight[data-range="' + currentRange + '"]') || document.createElement('div');
+	
+			action.apply(line.setAttribute, line,['aria-hidden', 'true']);
+			action.apply(line.setAttribute, line,['data-range', currentRange]);
```
RunDevelopment: I really don't know whether I like this style. It feels non-standard and kind of hacky.

That being said, I'm also unable to find a better solution because all of my approaches fail because of the scope of `var` variables. ES6 when?

RunDevelopment:
I will fully review your PR soon.

plugins/line-highlight/prism-line-highlight.js
justinbmeyer: Which style?

RunDevelopment: This function invocation style.

`action.apply(line.setAttribute, line,['aria-hidden', 'true']);` is exactly self-documenting IMO and it feels like it could be replaced by a simple lambda.

The lambda won't work because of the scoping of the captured variables, so we will most likely just go with this.

justinbmeyer: These could be replaced by a simple function like

```
action.add(function(){
  line.setAttribute('aria-hidden', 'true')
})
```

But this is a bit slower as it creates functions and adds extra function calls.

justinbmeyer: actually, there really isn't a performance gain.  I was pretending to be like CanJS's queues system, but didn't go all the way.  We can probably change it to the lambda style if you like that more.

RunDevelopment: I don't think it can because the scope of the `line` variable is declared using `var`, so function scope outside of the loop.
This means that all of the created function will capture the same variable causing all of the created functions to have the same value for `line` when executed.

This is the exact problem which ES6 `let` and `const` solve if my understanding is correct.

RunDevelopment: Also, you don't really need to be that concerned about the overhead of a single extra function call. It practically nonexistent compared to the performance bottle-nack it solves.

justinbmeyer: Gotcha. Well, good to go then? I’m looking forward to snappy resizes.

RunDevelopment:
I made you a [PR with a few improvements](https://github.com/Mattchewone/prism/pull/1).

Also, could you please provide an example which shows the reflow problem?
I can see the code that causes reflow but I cannot reproduce that this reflow significantly influences the overall performance. And certainly not as much as in your original issue.

justinbmeyer:
@RunDevelopment , depending on your hardware, you should be able to see the problem on this page: https://www.bitovi.com/academy/learn-angular/creating-components.html.  This was the page the original issue's performance graph was made from.  Since that time, I've split that page into another page to improve the performance.  

Despite this I can still visually see a delay when I resize the window.  You can see this in the following GIF. Notice that as I resize to see more of the screen, the "new" parts are white for a few ms.

Recording-9


With this fix, resizing feels instantaneous.

RunDevelopment:
@justinbmeyer Thanks for sharing this!
I can definitely see the issue. Resizing takes more than a second on my laptop.

Hopefully, the page-splitting won't be necessary anymore after this fix.

justinbmeyer:
When do you think this might be part of a release? Thanks!

RunDevelopment:
We don't have a fixed release schedule, so I can't say with certainty when a new version gets released. 
But maybe a month or so.

In the meantime, you can always get the latest version of Prism at our [download page](https://prismjs.com/download.html). This version includes this PR.

justinbmeyer:
Sorry to bother again, but any chance you'll be releasing soon?  I'd like to avoid changing our packages from using `npm` to a download (especially as we have multiple projects using prism and our build makes sure only one prism gets included).

RunDevelopment:
[Very soon!](https://github.com/PrismJS/prism/issues/1988)

PR #1866, opened 2019-04-25
Fix spelling mistake for copy-to-clipboard plugin page

featherbear:


RunDevelopment:
Thank you for contributing!

mAAdhaTTah:
@RunDevelopment @featherbear This should be pointed @ master, not `gh-pages`, which _should_ really be frozen.

RunDevelopment:
Ahhhh, indeed. My mistake.

Why do we still have `gh-pages` again?

PR #1868, opened 2019-04-25
Revert "Fix spelling mistake for copy-to-clipboard plugin page"

RunDevelopment:
Reverts PrismJS/prism#1866


mAAdhaTTah:
Since it was just a docs change, I don't think we need to revert it. We _do_ need to merge the fix into master though or it won't show up on our site.

RunDevelopment:
Agreed. Let me do that real quick.

PR #1873, opened 2019-04-26
Custom class: Implemented TODOs

RunDevelopment:
This PR allows for `map` to take functions as an argument.

This indirectly fixes the TODOs as both can be trivially implemented using a mapper function.


plugins/custom-class/prism-custom-class.js
```diff
@@ -7,24 +7,61 @@ if (
...
+/**
+ * @callback ClassMapper
+ * @param {string} className
+ * @param {string} language
+ * @returns {string}
+ */
+/**
+ * @typedef CustomClassOptions
+ * @property {ClassMapper} classMap
+ * @property {string} prefixString
```
mAAdhaTTah: I don't think this `typedef` is right: `classMap` can be either a function or an object, right?

plugins/custom-class/prism-custom-class.js
```diff
@@ -7,24 +7,61 @@ if (
...
+		}
 	},
+	/**
+	 * Adds the given prefix to all class names.
+	 *
+	 * @param {string} string
+	 */
 	prefix: function prefix(string) {
 		options.prefixString = string;
 	}
```
mAAdhaTTah: Thoughts on adding a `reset` method or something so users can revert the `classMap` / `prefix` changes?

plugins/custom-class/prism-custom-class.js
RunDevelopment: The `classMap` of the `map` function can be [either function or object](https://github.com/RunDevelopment/prism/blob/b2cfbd1afcad8468c5a5a3d85b01cc2f92ce26f5/plugins/custom-class/prism-custom-class.js#L37) but this is the type definition of the object storing the [internal state](https://github.com/RunDevelopment/prism/blob/b2cfbd1afcad8468c5a5a3d85b01cc2f92ce26f5/plugins/custom-class/prism-custom-class.js#L26) of the plugin.
`map` will [convert its given parameter into a function](https://github.com/RunDevelopment/prism/blob/b2cfbd1afcad8468c5a5a3d85b01cc2f92ce26f5/plugins/custom-class/prism-custom-class.js#L43), so the type definition is correct.

plugins/custom-class/prism-custom-class.js
RunDevelopment: Sure, sounds good.
But, we should make it so that the user can reset `map` and `prefix` separately. (The user can't read the internal state, so resetting one without the other would be quite a bit harder.)

Maybe a good way is to just provide the initial value via an `initial` property like `Prism.plugins.customClass.map.initial`?

Or maybe something like `Prism.plugins.customClass.map.reset()` so we don't expose any internal state?

mAAdhaTTah: Sorry, on second thought, I'm not sure it's worth it; I don't see it really being used that often. Let's pass on adding something like that unless we get a request for it.

plugins/custom-class/index.html
```diff
@@ -113,14 +123,6 @@ <h2>Output</h2>
 &lt;/code&gt;&lt;/pre&gt;</code></pre>
 </section>
 
-<section>
-	<h1>Todo</h1>
-	<ul>
-		<li>Take a function as the transformation (<a href="https://github.com/PrismJS/prism/pull/950#issuecomment-224289585">Comment in #950</a></li>
-		<li>Allow to custom tokens per language, using nested object (<a href="https://github.com/PrismJS/prism/issues/947#issuecomment-216979932">Comment in #947</a>)</li>
```
mAAdhaTTah: This isn't implemented, is it?

mAAdhaTTah: Maybe it doesn't matter... this can be implemented with a function.

RunDevelopment: That's the exact thought I also had.
It's really easy to implement this using a function, so I considered it done. Plus, we can still add it in the future.

PR #1883, opened 2019-05-07
doc: list 3rd party components on website

peterpeterparker:
Hi,

I absolutely don't mind at all if you would not like to merge that PR, it's all fine for me but I wrote an open source Web Component which encapsulate prism in order to highlight code and I'm happy to share it

Furthermore maybe other members of the community have written components for their favorite frameworks which they would like to share too?

Anyway, thx for the really cool library and good work

Best
David

### DIFF
--- a/index.html
+++ b/index.html
@@ -226,6 +226,16 @@ var html = Prism.highlight(code, Prism.languages.haml, 'haml');</code></pre>
 	</ul>
 </section>
 
+<section id="components">
+	<h1>Third-party components</h1>
+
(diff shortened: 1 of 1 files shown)


RunDevelopment:
@mAAdhaTTah @LeaVerou

mAAdhaTTah:
I'm on board with this. There are definitely e.g. React components that use Prism that would be cool to list here. Nice for "social proof" too as well as integrations users can rely on instead of rolling their own.

LeaVerou:
I would love this!

But I think we should list at least a few before we start having such a section, it would actually have the opposite effect if we just list one.

PR #1885, opened 2019-05-10
Add support for Falcon (Fix #1884)

### LINKED ISSUE #1884: Falcon Lang Support
**Falcon**
Falcon is a  is a powerful language for describing your structured data. It is easy for developer to read and write.

**Additional resources**
https://marketplace.visualstudio.com/items?itemName=FalconTeam.falcon

qfish:


components/prism-falcon.js
```diff
@@ -0,0 +1,26 @@
...
+});
+
+Prism.languages.falcon['comment'].push({
+	pattern: /(^|[^\\])#.*/,
+	lookbehind: true,
+	greedy: true
+});
+
+if (Prism.languages.markup) {
+	Prism.languages.markup.tag.addInlined('script', 'falcon');
```
RunDevelopment: This call to `addInlined` should cause JavaScript `<script>` tags to be overshadowed, right?

components/prism-falcon.js
```diff
@@ -0,0 +1,26 @@
+Prism.languages.falcon = Prism.languages.extend('javascript', {
+	'number': [
+		Prism.languages.javascript.number,
+		{
+			pattern: /\b(?:api)\b/
```
RunDevelopment: 1. Why is `api` a number?
2. You don't need the group.
3. You don't need to create an object with a `pattern` property. You can just inline the regular expression.

```diff
- {
-     pattern: /\b(?:api)\b/
- }
+ /\bapi\b/
```

components/prism-falcon.js
```diff
@@ -0,0 +1,26 @@
...
+		}
+	],
+	'builtin': /\b(?:get|post|put|delete|head|byte|int|int8|int16|int32|int64|uint|uint8|uint64|int_t|int8_t|int16_t|int32_t|int64_t|uint_t|uint8_t|uint64_t|long|Long|float|Float|double|Double|boolean|Boolean|bool|Bool|string|String|binary|string|Function|any|number|boolean|Array|symbol|console|Promise|unknown|never)\b/,
+});
+
+Prism.languages.falcon['class-name'][0].pattern = /(\b(?:enum|struct|class|interface|extends|implements|instanceof|new)\s+)[\w.\\]+/;
```
RunDevelopment: Please sort the keyword list.
```suggestion
Prism.languages.falcon['class-name'][0].pattern = /(\b(?:class|enum|extends|implements|instanceof|interface|new|struct)\s+)[\w.\\]+/;
```

components/prism-falcon.js
```diff
@@ -0,0 +1,26 @@
+Prism.languages.falcon = Prism.languages.extend('javascript', {
+	'number': [
+		Prism.languages.javascript.number,
+		{
+			pattern: /\b(?:api)\b/
+		}
+	],
+	'builtin': /\b(?:get|post|put|delete|head|byte|int|int8|int16|int32|int64|uint|uint8|uint64|int_t|int8_t|int16_t|int32_t|int64_t|uint_t|uint8_t|uint64_t|long|Long|float|Float|double|Double|boolean|Boolean|bool|Bool|string|String|binary|string|Function|any|number|boolean|Array|symbol|console|Promise|unknown|never)\b/,
```
RunDevelopment: 1. Please sort the keyword list.
2. The pattern `int|int8|int16|int32|int64|uint|uint8|uint64|int_t|int8_t|int16_t|int32_t|int64_t|uint_t|uint8_t|uint64_t` can be written as `int(?:8|16|32|64)?(?:_t)?|uint(?:8|64)?(?:_t)?`.
3. Are the 16 and 32bit variants of `unit` missing on purpose?

```suggestion
	'builtin': /\b(?:int(?:8|16|32|64)?(?:_t)?|uint(?:8|64)?(?:_t)?|Array|Bool|Boolean|Double|Float|Function|Long|Promise|String|any|binary|bool|boolean|boolean|byte|console|delete|double|float|get|head|long|never|number|post|put|string|string|symbol|unknown)\b/,
```

RunDevelopment:
I will now close this for the same reason as #1884: I have no idea what language this is supposed to implement.

PR #1844, opened 2019-03-28
Core: `Token.stringify` will now call `util.encode`

RunDevelopment:
This PR changes the behavior of `Token.stringify` so that the function will now encode strings itself rather than relying on pre-encoded token streams.

This also simplified `util.encode` because it doesn't need to be able to handle token streams anymore.

It's also more efficient like this because `encode` doesn't have to create a deep copy of the token stream.

---

Btw. No language or plugin uses `encode`, so this should be a purely internal change.

PR #1889, opened 2019-05-14
Diff: Added support for syntax highlighting inside diffs

### LINKED ISSUE #1878: Is there a way to use both language and diff highlights?
**Information**
- Language: YAML, diff
- Plugins: line-numbers

As the title says, I'm trying to use both language (in my case it is YAML) and diff highlights, but Prism highlights blocks only as diff. So, the questions is: Is there a way to use these both highlights?

RunDevelopment:
This PR adds support for syntax highlighting inside diffs.

Instead of using `language-diff`, it is now possible to use `language-diff-xxxx` highlight in a specific language. If you call the highlighting function directly use `Prism.highlight(code, Prism.languages.diff, 'diff-' + diffLanguage)`.

To be compatible with other plugins that need access to the language-specific diff grammar, a new alias for every diff-variant will be created during run-time. (See the `before-sanity-check` and `before-tokenize` hooks.)

### Before
image

### After
image

### Slightly tweaked CSS
image

## Known issues

1. ~~If the attributes of the highlighted embedded code contain line breaks, incorrect HTML code is generated.~~ Fixed
2. Tokens spaning diff blocks aren't highlighted correctly.
    E.g. a multiline comment spanning multiple lines, one of which has been modified.
    ```diff
     /*
    - * foo
    + * bar
      */
   ```
3. ~~If other plugins add tags to the HTML of a block, that block won't be highlighted. I.e. the show Invisibles plugin adds nested tokens which generate tokens (and later tags) inside the blocks.~~ Fixed
4. The `diff-xxxx`-pattern is incompatible with auto loader.

---

This PR resolves #1878.


components/prism-diff.js
```diff
@@ -1,20 +1,122 @@
...
+			diffGrammar = Prism.languages[diffLanguage];
+		}
+
+		// one of the diff tokens without any nested tokens
+		if (env.type in prefixes) {
+			/** @type {string} */
+			var content = env.content.replace(HTML_TAG, ''); // remove all HTML tags
+
+			/** @type {string} */
+			var decoded = content.replace(/&amp;/g, '&').replace(/&lt;/g, '<');
```
ExE-Boss: This should be:
```suggestion
			var decoded = content.replace(/&lt;/g, '<').replace(/&amp;/g, '&');
```
Otherwise, `&amp;lt;` would end up as `<` instead of `&lt;`.

RunDevelopment: Indeed! Thanks for pointing this out.

components/prism-diff.js
```diff
@@ -1,20 +1,122 @@
...
+			diffGrammar = Prism.languages[diffLanguage];
+		}
+
+		// one of the diff tokens without any nested tokens
+		if (env.type in prefixes) {
+			/** @type {string} */
+			var content = env.content.replace(HTML_TAG, ''); // remove all HTML tags
+
+			/** @type {string} */
+			var decoded = content.replace(/&lt;/g, '<').replace(/&amp;/g, '&');
```
RunDevelopment: I wrote the same code in another PR (can't remember which one), so maybe we should move this line to something like `Prism.util.decode` as an inverse function to `encode` (ignoring that `encode` can deal with more than only strings).

ExE-Boss: Well, after https://github.com/PrismJS/prism/pull/1844, `Prism.util.encode(…)` will only know how to deal with strings.

mAAdhaTTah: It's in #1898.

tests/languages/diff/diff_feature.test
```diff
@@ -9,13 +11,14 @@
...
+	["inserted-sign", "+    qt: core gui\r\n"],
+	["deleted-arrow", "< qt: core\r\n"],
+	["inserted-arrow", "> qt: core quick"]
 ]
 
 ----------------------------------------------------
 
-Checks for deleted, inserted and different lines.
\ No newline at end of file
+Checks for deleted, inserted and different lines.
```
mAAdhaTTah: Can we get tests for inside diff highlighting?

RunDevelopment: What do you mean with "inside diff highlighting"?
Do you mean a test where we actually highlight the code?

mAAdhaTTah: Yup!

ExE-Boss: I believe that means add&nbsp;a&nbsp;test like&nbsp;this, except&nbsp;for&nbsp;**Markdown**:
https://github.com/PrismJS/prism/blob/master/tests/languages/markup!%2Bjavascript/javascript_inclusion.test

RunDevelopment: Can I even [write a test](https://prismjs.com/test-suite.html#writing-specific-tests)?

I need to the test runner to load `diff` but use `diff-xxxx` to highlight some given test code. I don't think it's capable of doing that right now, is it?

RunDevelopment: You could even say that this is less a problem of the test runner and more a problem of my PR. I basically create language definitions on the fly. No part of Prism is expecting this.

mAAdhaTTah: Is it worth extracting to another language, like `diff-extras` or something? Maybe worth considering what our philosophy should be for what's included in the base language vs an extension.

RunDevelopment: I was thinking of maybe making this a plugin, so I would also be able to include the custom CSS I used for the [third screenshot](https://github.com/PrismJS/prism/pull/1889#issue-278797097).

mAAdhaTTah: Yeah, I like that idea. That addition would be really nice.

RunDevelopment: Alright, a plugin it is then!

components.json
```diff
@@ -1078,6 +1078,11 @@
 			"owner": "mAAdhaTTah",
 			"require": "toolbar",
 			"noCSS": true
+		},
+		"diff-highlight": {
+			"title": "Diff Highlight",
+			"owner": "RunDevelopment",
+			"require": "diff"
```
RunDevelopment: I'm surprised that this actually worked, well, at least it did on the download page.

As a more general question should a plugin be able to require a language like this?

plugins/diff-highlight/prism-diff-highlight.css
```diff
@@ -0,0 +1,13 @@
+pre.diff-highlight > code .token.deleted:not(.prefix),
+pre > code.diff-highlight .token.deleted:not(.prefix) {
+	background-color: rgba(255, 0, 0, .1);
+	color: inherit;
+	display: block;
```
RunDevelopment: Because of the `display: block`, it's compatible with the Funky theme. I already added a note for this on the plugin page but I would like to fix it. Any ideas?

<details>

image

</details>

components.json
mAAdhaTTah: I don't see why not. It already works, it has a use case, why disallow it?

RunDevelopment: Because you don't know whether it requires a plugin named `diff` or a language.

RunDevelopment: Don't get me wrong. It's super useful that it works but I don't like that it's ambiguous.

tests/languages/diff/diff_feature.test
RunDevelopment: Regarding the testing: I'll just add it to the list in #1970 when it's merged.

components.json
mAAdhaTTah: I think the likelihood that we have an plugin and language with the same name is basically nil. I don't think I mind that ambiguity. Maybe it should be documented somewhere?

plugins/diff-highlight/prism-diff-highlight.css
mAAdhaTTah: Funky can do its "own" implementation of these styles, if they need to be different than the global ones.

components.json
RunDevelopment: I'm planning a big dependency update for #1992. I'll document it then.

plugins/diff-highlight/prism-diff-highlight.css
RunDevelopment: Done.
Not beautiful but ok, I guess

<details>

image

</details>

themes/prism-funky.css
```diff
@@ -115,3 +115,16 @@ code[class*="language-"] {
...
+pre.diff-highlight > code .token.deleted:not(.prefix),
+pre > code.diff-highlight .token.deleted:not(.prefix) {
+	background-color: rgba(255, 0, 0, .3) !important;
+	display: inline !important;
+}
+
+pre.diff-highlight > code .token.inserted:not(.prefix),
+pre > code.diff-highlight .token.inserted:not(.prefix) {
+	background-color: rgba(0, 255, 128, .3) !important;
+	display: inline !important;
```
mAAdhaTTah: Instead of `!important`, we can double the class names for extra specificity like [I did here](https://github.com/PrismJS/prism/blob/master/themes/prism-coy.css#L202-L212).  That gives consumers more flexibility if they need to modify the styles.

RunDevelopment: Ooof. Very well then.

RunDevelopment: Done.

PR #1898, opened 2019-05-18
Markdown: Added support for auto-loading code block languages

### LINKED ISSUE #1644: Markdown should auto load languages
Markdown should use the Autoloader plugin to load the languages of code blocks, if not loaded already.

More details [here](https://github.com/PrismJS/prism/pull/1562#issuecomment-443451153).

RunDevelopment:
This resolves #1644.

Autoloader now also exposes a `loadLanguages` function which, in contrast to the private `loadLanguages`, guarantees that callbacks are executed asynchronously.


plugins/autoloader/prism-autoloader.js
```diff
@@ -170,7 +170,28 @@
...
+		 * Loads the given language(s) and calls the given callback(s) asynchronously.
+		 *
+		 * The error callback will be called once for each language failing to load.
+		 *
+		 * @param {string|string[]} languages
+		 * @param {(languages: string[]) => void} [success]
+		 * @param {(language: string) => void} [error]
+		 */
+		loadLanguages: function (languages, success, error) {
+			// loadLanguages might call the callbacks synchronously, so we guarantee that this is not the case
```
ExE-Boss: Maybe also put this comment into the&nbsp;description of&nbsp;the&nbsp;internal `loadLanguages(…)` function JSDoc.

mAAdhaTTah:
Thought: are we better off adding extra code so the public API so it's always async or fix the internal implementation to be always async?

Issue #1904, opened 2019-05-22
fileHighlight is included in Prism.js and as a plugin


JakeChampion:
**Information:**
- Prism version: master branch (latest)
- Environment: Browser

> Does the latest version of Prism from the [download page](https://prismjs.com/download.html) also have this issue?

Yes

~If you're using Prism with Node or Webpack, please also see the [usage section](https://prismjs.com/index.html#basic-usage).~

**Description**
The fileHighlight functionality is included as a [plugin](https://github.com/PrismJS/prism/blob/ce9616a0ce643a4be19abf5442f34712242df386/plugins/file-highlight/prism-file-highlight.js#L1) but it is also included in [Prism.js](https://github.com/PrismJS/prism/blob/6068bf18e53222a1ee712b073a9864413fefeef3/prism.js#L856-L964) at the end of the file. Including it at the end of the file means that users of Prism always end up having fileHighlight, even if they do not want that functionality.

I'm hoping we can remove it from the main Prism.js file and keep it only as a plugin, this is likely a major version change as it does remove default functionality of Prism.js.

RunDevelopment:
> Including it at the end of the file means that users of Prism always end up having `fileHighlight`

Not really.
The `prism.js` generated by the [download page](https://prismjs.com/download.html) doesn't have this issue.

In node, this is a little more tricky. Instead of [importing `prism` directly](https://github.com/Financial-Times/o-syntax-highlight/blob/7d6c0a7970f7ded7a1de4796673f0a521f51f8b5/src/js/syntax-highlight.js#L3), you can import its parts. 
`prism.js` is a generated file which is the result of the concatenation of [`prism-core.js`, `prism-markup.js`, `prism-css.js`, `prism-clike.js`, `prism-javascript.js`, and `file-highlight/prism-file-highlight.js`](https://github.com/PrismJS/prism/blob/6068bf18e53222a1ee712b073a9864413fefeef3/gulpfile.js#L18).
So if you don't want any of the parts of `prism.js`, then you can import its parts instead and leave out the ones you don't need.

To exclude `fileHighlight`, do the following:

```diff
-import prism from 'prism/prism.js';
+import prism from 'prism/components/prism-core.js';
+import 'prism/components/prism-markup.js';
+import 'prism/components/prism-css.js';
+import 'prism/components/prism-clike.js';
+import 'prism/components/prism-javascript.js';
```

[...]

PR #1905, opened 2019-05-22
Remove file-highlight plugin from default usage of prism

JakeChampion:
Solves #1904


RunDevelopment:
As mentioned in [my answer](https://github.com/PrismJS/prism/issues/1904#issuecomment-494797149), `prism.js` is generated, so please don't modify it yourself.

Also, `prism.js` is used by the [prismjs.com](https://prismjs.com/index.html) website, so removing the File Highlight plugin will break the website.

PR #1907, opened 2019-05-24
Download page: Improved performance for chromium based browsers

RunDevelopment:
I have this issue that when I selected all languages on the [download page](https://prismjs.com/download.html#themes=prism), Chrome freezes for about 12 seconds (!!) while using FireFox/Edge this operation takes an instant.

Btw. CPU: Intel i7-8700K @ 4.2GHz. (On my laptop, Chrome takes about 30sec)

This PR fixes this issue and after the fix, Chome takes less than a 1sec (not ideal but ok). FF and Edge stay fast.

<details>
<summary>
Here are some profiling results (Chrome):
</summary>

### Before

image

TLDR;
Scripting: 177ms
Rendering: 11,485ms

### After

image

TLDR;
Scripting: 121ms _// no idea why this is suddenly faster_
Rendering: 592ms

</details>


scripts/download.js
```diff
@@ -552,9 +552,18 @@ function generateCode(){
...
-				Prism.highlightElement(codeElement, true);
+				var newCode = document.createElement('CODE');
+				newCode.className = codeElement.className;
+				newCode.textContent = text;
+
+				Prism.highlightElement(newCode, true, function () {
+					while (pre.firstChild) {
+						pre.removeChild(pre.firstChild);
+					}
+					pre.appendChild(newCode);
```
mAAdhaTTah: Instead of clearing everything & appending, can we use `replaceChild` to swap the old with the new?

RunDevelopment: Good idea!

PR #1911, opened 2019-05-26
Test: Added test for nice token names

RunDevelopment:
This adds a test for nice token names. (This also affects aliases.)

Token names and aliases are used as CSS class names and as such should play nicely with CSS.
The new test matches all names against a strict name pattern to identify unsuitable names.


RunDevelopment:
TIL that this check might actually be necessary because of the [ES6 spec on property enumeration](http://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys).

[Here](https://www.stefanjudis.com/today-i-learned/property-order-is-predictable-in-javascript-objects-since-es2015/)'s a good article with some examples on this topic.

PR #1914, opened 2019-05-28
cpp extend clike

jlice:
Fix: prism-cpp load failed.

If cpp extend c, it will fallback to default language.

### DIFF
--- a/components/prism-cpp.js
+++ b/components/prism-cpp.js
@@ -1,4 +1,4 @@
-Prism.languages.cpp = Prism.languages.extend('c', {
+Prism.languages.cpp = Prism.languages.extend('clike', {
 	'class-name': {
 		pattern: /(\b(?:class|enum|struct)\s+)\w+/,
 		lookbehind: true


jlice:
```javascript
const md = require('markdown-it')();
const prism = require('markdown-it-prism');
md.use(prism);
const markdown = "```cpp\n" +
                "class Demo { };\n"+
                "```";
const ret = md.render(markdown);
console.log(ret);
```

`markdown-it-prism` will use `prism` as backend to render codeblocks. However, if cpp extend c, it will fail to load `prism-cpp.js`, and it render the markdown text as follow:

```html
<pre class="language-cpp"><code class="language-cpp">class Demo { };
</code></pre>
```

While if cpp extend clike, it is:

```html
<pre class="language-cpp"><code class="language-cpp"><span class="token keyword">class</span> <span class="token class-name">Demo</span> <span class="token punctuation">{</span> <span class="token punctuation">}</span><span class="token punctuation">;</span>
</code></pre>
```

Maybe there is some mistakes in it.

RunDevelopment:
This issue is caused by [markdown-it-prism](https://github.com/jGleitz/markdown-it-prism), not PrismJS. See the issue [here](https://github.com/jGleitz/markdown-it-prism/issues/22).

PR #1921, opened 2019-05-28
Julia: Highlight ‘Inf’ and ‘NaN’ as constants

dvdhansen:


RunDevelopment:
Thank you for contributing!

Issue #1922, opened 2019-05-30
Add CDN configurator to homepage


foxycode:
**Motivation**
When I hit https://prismjs.com/download.html there is nice configurator for files to download. It would be nice to also show all files I need to include if I want to use CDN. I tool me really long time to find that I need also `prism-markup-templating.js` too to whole thing to work.

**Description**
I'd like to have a list of CDN files I need to include as a second option to download configured source from https://prismjs.com/download.html

RunDevelopment:
I don't think that we need this.

IMO the end user shouldn't have to worry about which files have to be loaded in what order so that syntax highlighting for e.g. TypeScript or PHP (since you mentioned `markup-templating`) works. Not only is this annoying but it's also error-prone because languages have quite a few dependencies on each other which not only determine what has to be loaded but also in which order.

Another downside of loading languages and their dependencies ourselves is that we usually load languages by declaring `script` tags like this:

```html
<!-- I just want to highlight PHP and TypeScript code -->
<script src="https://myCDN.com/prism/components/prism-core.js"></script>
<script src="https://myCDN.com/prism/components/prism-markup.js"></script>
<script src="https://myCDN.com/prism/components/prism-markup-templating.js"></script>
<script src="https://myCDN.com/prism/components/prism-php.js"></script>
<script src="https://myCDN.com/prism/components/prism-clike.js"></script>
<script src="https://myCDN.com/prism/components/prism-javascript.js"></script>
<script src="https://myCDN.com/prism/components/prism-typescript.js"></script>
```

All of them are loaded sequentially one after another when a few of these could be loaded in parallel.

__The solution__

The [Autoloader plugin](https://prismjs.com/plugins/autoloader/) can do all of this language loading and dependency handling for you.

The above code then becomes:

```html
<!-- I just want to highlight PHP and TypeScript code -->
<script src="https://myCDN.com/prism/components/prism-core.js"></script>
<script src="https://myCDN.com/prism/plugins/autoloader/prism-autoloader.js"></script>
<script>Prism.plugins.autoloader.languages_path = 'https://myCDN.com/prism/components/'</script>
```

Please note that the Autoloader loads languages asynchronously as needed, so won't ever load languages, you don't have to. 
But this also means that only the languages you have already used are loaded, so keep that in mind if you use the Prism API more directly.

I hope that the Autoloader works for you. If not, then we can reopen this.

foxycode:
[...]

Issue #158, opened 2013-08-28
Suggestion: Release on CDNJS.com


wmh:
CDNJS is a free Content Delivery Network hosted by CloudFlare for open source, community-driven projects. It would be great to release prism on CDNJS for all to use.

Instructions to add a project to CDNJS:
https://github.com/cdnjs/cdnjs

PS. Or you can simply add a tag to current version, and I'll do the rest.
```
git tag 1.0.0
git push --tags
```

loganfranken:
It looks like someone did this (wahoo!):
https://cdnjs.com/libraries/prism

I think this can be closed.

Golmote:
The version is outdated. I guess since we now have [release 1.0.0](https://github.com/PrismJS/prism/releases), something can be done to update CDNJS.

thednp:
Please change/add CDN provider. I use <a href="https://github.com/jsdelivr/jsdelivr">jsDeliver</a>, it automatically adds new release tags to clouds.
The 0.0.1 version on cdnjs is pretty buggy with JS.

Golmote:
@thednp I configured the autopupdate on jsdelivr. (https://github.com/jsdelivr/jsdelivr/tree/master/files/prism)

thednp:
Thanks, as I can see the jsdeliver repository, there is no prism.min.js and prism.min.css, so you better check that too.

Golmote:
No, there are all temes, all components (minified), all plugins (minified) and a default prism.js file that is mainly used for Prism's website, currently. We can argue about the minification of the themes, but well, I think we're ok for now.

Golmote:
Both CDNJS and jsDeliver look up to date. I'm closing this issue.

wmh:
:+1:

NullVoxPopuli:
Is there a simple example / html file that demonstrates how to use the cdn?

Golmote:
@NullVoxPopuli Unfortunately, using the CDN, you need to handle language dependencies yourself.

Handlebars requires Markup-templating, so a working example would be: https://codesandbox.io/s/3vx76wq76m

hades200082:
Why not put a "Complete" js on CDN that includes all languages... and each theme css as individual CDN files.  that way the developer only has two files to include from CDN.

If anyone doesn't want everything included they can still build their own version.

mAAdhaTTah:
[...]

PR #1924, opened 2019-05-31
Website: Added basic usage for CDNs

RunDevelopment:
Per [suggestion](https://github.com/PrismJS/prism/issues/1922#issuecomment-497681398) by @foxycode, I added an example for the usage of Prism with CDNs.

### DIFF
--- a/index.html
+++ b/index.html
@@ -152,6 +152,27 @@
 	<p>If you want to prevent any elements from being automatically highlighted, you can use the attribute <code>data-manual</code> on the <code>&lt;script></code> element you used for prism and use the <a href="extending.html#api">API</a>.
 	Example:</p>
 	<pre><code>&lt;script src="prism.js" data-manual>&lt;/script></code></pre>
+	<h2>Usage with CDNs</h2>
+
+	<p>In combination with CDNs, we recommend using the <a href="plugins/autoloader">Autoloader plugin</a> which automatically loads languages when necessary.</p>
+
+	<p>The setup of the Autoloader, will look like the following. You can also your own themes of course.</p>
+
+	<pre><code>&lt;!DOCTYPE html>
+&lt;html>
+&lt;head>
+	...</code>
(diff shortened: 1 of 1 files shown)


foxycode:
@RunDevelopment There is still missing CSS I needed to add.

Also, I believe `cdnjs.cloudflare.com` is your official mirror, isn't it? It would be nice to add concrete example.

RunDevelopment:
> There is still missing CSS

Do we even need to add this? It's dead simple to change the CSS link and many people use custom CSS.
So do we really need this? Just asking, if you think it is then I'll add it.

> I believe `cdnjs.cloudflare.com` is your official mirror

[We also support jsDelivr apparently.](https://github.com/PrismJS/prism/issues/158#issuecomment-145876048)

> It would be nice to add a concrete example.

Prism is available on quite a few CDNs, so it feels like we would disproportionally promote a single provider. Maybe a list of CDNs would also work?

foxycode:
@RunDevelopment I believe it is needed. Or people may ask as I did why it isn't working. You can always choose not to include CSS and use your own, but it takes work to figure out why you don't see code colors and how to fix it.

CDN list would be nice. Problem is with file naming on different CDN's. You are referring `prism-core.js` in your example, but there is no such file on cloudflare, it is named differently.

What I simply want is, that with zero knowledge about prism I can open your page, copy fully functional code without more googling to my page and it will work. Otherwise, you may discourage not so experienced users and they will go try another plugin.

RunDevelopment:
> What I simply want is, that with zero knowledge about prism I can open your page, copy fully functional code without more googling to my page and it will work.

That's a really good point. Will do.

> You are referring prism-core.js in your example, but there is no such file on cloudflare.

[There is](https://cdnjs.cloudflare.com/ajax/libs/prism/1.16.0/components/prism-core.js), don't worry.

foxycode:
@RunDevelopment I see. So I should include `prism-core.js` when using autoloader I suppose.

Also, it would be good to link `min` files in your example when using CDN.

Many thanks for your effort.

Issue #1930, opened 2019-06-07
Highlight HTML, CSS, Markdown, GraphQL in JS tagged template literals


karlhorky:
**Motivation**

Prism doesn't highlight supported languages when they are in a tagged template literal with common tags.

<img width="457" alt="Screen Shot 2019-06-07 at 11 25 17" src="https://user-images.githubusercontent.com/1935696/59095008-bd9dcf80-8917-11e9-8622-b62b77089441.png">

**Description**

Support for common tagged template literals such as:

```js
// HTML
html`...`

// CSS
css`...`

// React Components CSS
styled.foo``
Component.extend``
styled.foo.attr({})``
Component.extend.attr({)``
styled(Component).attr({})``
css.global``
css.resolve``

// Markdown
md`...`
markdown`...`

// GraphQL
gql`...`
graphql`...`
graphql.experimental`...`
```

To allow this highlighting behavior without adding a tag call, one option is to also support comments in the place of tags:

```js
/* HTML */`...`
/* CSS */`...`
```

There are also other non-template literal syntaxes which are potentially desired:

```js
// Angular Components HTML + CSS
@Component({
  template: `<div>...</div>`,
  styles: [`h1 { color: blue; }`]
})

// styled-jsx CSS
<style jsx>{`div{color:red}`}</style>
```

---

Prettier has some AST tests for these things which could probably be repurposed here (or at least, could inspire a solution):

HTML tagged template literals:
- [isHtml](https://github.com/prettier/prettier/blob/e4f0df5bed7448e012c50e9e9c91904d8b9cab88/src/language-js/embed.js#L568-L585)

CSS tagged template literals:
- [isCss](https://github.com/prettier/prettier/blob/e4f0df5bed7448e012c50e9e9c91904d8b9cab88/src/language-js/embed.js#L26-L31) (overall test for all types of CSS tagged template literals: normal CSS, Angular, React Components)
- [isStyledJSX](https://github.com/prettier/prettier/blob/e4f0df5bed7448e012c50e9e9c91904d8b9cab88/src/language-js/embed.js#L345-L376)
- [isStyledComponent](https://github.com/prettier/prettier/blob/e4f0df5bed7448e012c50e9e9c91904d8b9cab88/src/language-js/embed.js#L434-L474)
- [isCssProp](https://github.com/prettier/prettier/blob/e4f0df5bed7448e012c50e9e9c91904d8b9cab88/src/language-js/embed.js#L479-L489)
[...]

PR #1931, opened 2019-06-08
Highlight JS (tagged) template literals

RunDevelopment:
This PR adds support for arbitrary syntax highlighting within JS template literals via a new JS Templates language/extension.

Currently implemented are most ideas from #1930:

- Markup
    - ✅ <code>html\`&lt;p>&lt;/p>\`</code>
    - ✅ <code>div.innerHTML = \`&lt;p>&lt;/p>\`</code>
- CSS
    - ✅ <code>css\`a { color: red; }\`</code>
    - ✅ <code>styled.foo\`color: red;\`</code>
    - ⛔ <code>Component.extend\`color: red;\`</code> Since this is [deprecated](https://github.com/styled-components/styled-components/issues/1546), I don't think we have to support it.
    - ❓ <code>&lt;style jsx>{\`div{color:red}\`}&lt;/style></code>
        Implementing this will likely be a bit harder because of the way JSX matches text. But this implementation won't be part of JS Templates, so it's not included in this PR.
- Markdown
    - ✅ <code>md\`# title`</code>
- GraphQL
    - ✅ <code>gql\`{ foo }`</code>
- Other
    - ⛔ Angular Components HTML + CSS:
        ```
        @Component({
          template: `<div>...</div>`,
          styles: [`h1 { color: blue; }`]
        })
        ```
        Implementing this in regexes is next to impossible. Prettier implemented this using the created AST which is something Prism doesn't have.
    - ⛔ <code>/* HTML */\`...\`</code> Not included because I don't want to introduce Prsim-specific syntax for the purpose of highlighting.

<details>
<summary>Example</summary>

image

</details>

---

This is based on #1714.


tests/languages/javascript!+js-templates/css_inclusion.test
```diff
@@ -0,0 +1,134 @@
...
+css.global`a:hover { color: blue; }`;
+css.resolve`a:hover { color: blue; }`;
+
+// styled-components
+const Button = styled.button`
+	color: blue;
+	background: red;
+`;
+styled.button.attr({})`color: blue;`;
+styled(Component).attr({})`color: blue;`;
```
mAAdhaTTah: Also need:

```js
styled(Component)`color: blue;`
```

RunDevelopment: Done.

luwes:
Will be awesome to have this in Prism. Nice work!

Could the `svg` tag possibly be also added?

`createTemplate('svg', /\bsvg|\.\s*(?:inner|outer)HTML\s*=/.source, 'markup'),`

Quite a few libs have this tag separately from the `html` tag.

luwes:
Thanks @RunDevelopment! I already put your code to use, https://sinuous.netlify.com/examples/clock/

There seems to be one issue with the html/svg tags that have an expression in them, they have other colors.

Screen Shot 2019-07-17 at 8 10 34 PM

luwes:
Found a fix I think. The issue was that the markup pattern regex was too strict w/ the included interpolations. The following is a snippet of the working code.

If it's `markup`, take the extended lang `template-string-markup`.
Here's the diff of the regex.

Screen Shot 2019-07-18 at 11 12 40 AM


```js
    const lang = language === 'markup' ? 'template-string-markup' : language;
    patternObj.inside[tokenName] = !language ? /[\s\S]+/ : {
      pattern: /[\s\S]+/,
      alias: 'language-' + language,
      inside: Prism.languages[lang]
    };

    patternObj.inside.rest = Prism.languages[lang];

    return patternObj;
  }

});

const templateMarkup = Prism.languages['template-string-markup'] = Prism.languages.extend('markup', {});

templateMarkup.tag.pattern = /<\/?(?!\d)[^\s>\/=$<%]+(?:\s(?:\s*[^\s>=]+(?:\s*=\s*(?:"[^"]*"|'[^']*'|[^\s'">=]+(?=[\s>])|$)|(?=[\s/>])))+)?\s*\/?/i;

Prism.languages.javascript['template-string'].push(Prism.languages.javascript['template-string'].createTemplate('string'));
```

RunDevelopment:
The solution is really not optimal. Instead of relaxing the HTML tag pattern we should do something similar to Markup templating (MT).
MT first searches for all parts matching a pattern (in our case that the string interpolation) and cuts them out (put simply). Then MT highlights the markup and inserts the cut-out pieces again. This approach is very reliable and works great for embedded languages (e.g. PHP, EJS, ...) but my problem is that it only does this for Markup languages.

I also want to do the same for all other embedded languages, so I could either adjust MT to support more than just Markup (which kind of invalidates the name) or I could write my own MT without the Markup focus.
I'm currently contemplating as to what I'll do but I'm leaning towards the first option.

---

As to why relaxing the pattern isn't the best solution: A relaxed pattern will also match syntactically invalid code and are likely to break in more complex situations. Also, you relaxed just one pattern when you actually had to relax ALL patterns to make the string interpolation for other tokens as well. This is a whole lot of work and had to be done for every embedded language.

luwes:
Thanks for the reply, that makes a lot of sense. I'll have a look at MT.
I also just noticed my solution was breaking for markup after the interpolations :s.

RunDevelopment:
> I'll have a look at MT

Don't worry. I'll do it.
I might as well fix the bugs of my own PR 😉

RunDevelopment:
Alright, this was a lot more work than I expected but it finally works.

<details>
<summary>Example</summary>

image
[Source](https://sinuous.netlify.com/examples/clock/)

</details>

The whole setup was inspired by Markup templating but has support for recursive tokenization. This enables us to correctly highlight code like this:

image

The basic idea of the templating algorithm is this:

1. Tokenize JavaScript
2. Find all tagged-template tokens
3. Replace all interpolation expressions with placeholders
4. Tokenize the embedded code
5. Tokenize the interpolation expressions
    (this is the recursive step as we go to step 1 with different code)
6. In-insert the tokenized interpolation expressions

---

This also fixes the bug that escaped interpolations were highlighted. E.g. ``var a `\${42}`;``.

luwes:
Thank you @RunDevelopment! Looks amazing and working great on my site. 
Hope it can be merged soon.

mAAdhaTTah:
Looking over the PRs, one thing that occurs to me is we've got a few implementations of "highlighting one language in another." I'm going to take another pass over them individually, but I wanted to ask: is there any logic that could be shared between them, a la `markup-templating`?

RunDevelopment:
> We've got a few implementations of "highlighting one language in another."

Have we? I only know all the languages which use Markup templating and the ones which embed languages literally like HTML embeds CSS & JS.
The excessive amount of code I wrote is only necessary because of the string interpolations which require tokenizing the code in multiple passes.

Even though I'll admit that I'm guilty of copying a [few lines](https://github.com/PrismJS/prism/blob/9c203477f6c73506e151e6cf5f4bc639041a2a7e/components/prism-js-templates.js#L200-L227) from MT, I don't think that there is a whole lot to share. At least, not now.

mAAdhaTTah:
We have markdown, js template strings, & diff, which do highlighting of one language in another. They don't have to use `markup-templating`, but are there enough similarities that they can use shared logic. Sounds like no, that's fine, just wanted to put the question out there.

mAAdhaTTah:
What if we made the dependencies peerDependencies so you wouldn't need to include `graphql` if you aren't using it?

RunDevelopment:
> Sounds like no

Sorry, that's not what I wanted to say. Diff, JS Templating, and Markup Templating don't share a lot of logic because they have different solutions for the same problem.

I absolutely think that we could make a more general templating engine which would be able to make the above three obsolete. I actually tried just that in the process of making this but MT is very different from the solution I employed here. I would most likely have to adjust every language using MT, so I didn't want to include this here.
That being said, I do think that a more general templating engine would be a great addition to Prism.

RunDevelopment:
Regarding the peer dependencies:

Should the templating be optional as well, so we don't even add the template if the embedded language isn't loaded? 

<details><summary>RN, missing grammar look like this.</summary>

image

</details>

mAAdhaTTah:
> Should the templating be optional as well, so we don't even add the template if the embedded language isn't loaded?

Yeah, that sounds good to me.

RunDevelopment:
I want to cry a little: The test suite can't handle peer dependencies.
I hope modifying the path works...

mAAdhaTTah:
Don't we have a syntax to load languages in the tests? We've got these folders like `django!+css`, which I believe loads both those languages.

RunDevelopment:
Yup, that's what I meant with "modifying the path".
The solution is actually quite simple now that I look at it more closely.

components/prism-js-templates.js
```diff
@@ -0,0 +1,332 @@
...
+
+		// md`# h1`, markdown`## h2`
+		createTemplate('markdown', /\b(?:md|markdown)/.source),
+
+		// gql`...`, graphql`...`, graphql.experimental`...`
+		createTemplate('graphql', /\b(?:gql|graphql(?:\s*\.\s*experimental)?)/.source),
+
+		// vanilla template string
+		templateString
+	].filter(function (x) { return x; });
```
mAAdhaTTah:
```suggestion
	].filter(Boolean);
```

components/prism-js-templates.js
```diff
@@ -0,0 +1,332 @@
...
+		// replace all interpolations with a placeholder which is not in the code already
+		var placeholderCounter = 0;
+		/** @type {Object<string, string>} */
+		var placeholderMap = {};
+		var embeddedCode = _tokens.map(function (token) {
+			if (typeof token === 'string') {
+				return token;
+			} else {
+				var interpolationExpression = token.content;
+				if (typeof interpolationExpression !== 'string') {
```
mAAdhaTTah: If this shouldn't happen, do we need this check?

components/prism-js-templates.js
```diff
@@ -0,0 +1,332 @@
...
+
+			return tokens
+		}
+		walkTokens(embeddedTokens);
+
+		return new Prism.Token(language, embeddedTokens, 'language-' + language, code);
+	}
+
+	/**
+	 * A list of languages which support the
```
mAAdhaTTah: What's this supposed to say?

components/prism-js-templates.js
```diff
@@ -0,0 +1,332 @@
...
+	Prism.hooks.add('after-tokenize', function (env) {
+		if (!(env.language in supportedLanguages)) {
+			return;
+		}
+
+		/**
+		 *
+		 * @param {(Token | string)[]} tokens
+		 */
+		function walkTokens(tokens) {
```
mAAdhaTTah: Can we call this something else? We've got 2 functions named `walkTokens`.

components/prism-js-templates.js
RunDevelopment: It isn't supposed to happen but it can because plugins like Show Invisibles modify the grammar.

Actually, now that I think about it again, it should just convert the content of the token into a string.

components/prism-js-templates.js
RunDevelopment: Thumbs up from me. I would have never thought about this.

components/prism-js-templates.js
RunDevelopment: I'll take that back. Wrong function. I'll just remove it.

components/prism-js-templates.js
RunDevelopment: That it was 3 am when I wrote this. Will fix.

RunDevelopment:
I renamed one of the functions. Wasn't that what you meant?

mAAdhaTTah:
Yeah, you renamed the second function, which I missed when I was looking over the code. My b.

PR #1932, opened 2019-06-08
Core: Fixed greedy targeting bug

RunDevelopment:
This PR fixes a bug of greedy matching.

The `target` parameter defines the token which causes rematching. Rematching is done by recursively calling `matchGrammar` where `target` will be the current token name. The rematching is done if either a match was found or we reached the target token, whatever comes first.

The bug is that a token is identified by its name. But the name alone doesn't uniquely identify a token in a grammar because many tokens can have the same name:

```js
{
    'token': [
        /foo/,
        /bar/i
    ]
}
```

The fix is quite obvious. `target` will no be a combination of the token name and its index.

_Note:_ This is a technically a __breaking change__ because `matchGrammar` is publicly exposed. But the function is also not documented, so its behavior is undefined anyway.

---

Edit:
An example of the bug is the following JS code:

```js
// /*

/* const
```

__Before:__
image

__After:__
image


mAAdhaTTah:
Possible to get tests that demonstrates the fixed behavior?

PR #1935, opened 2019-06-09
Autoloader: Fixed and improved callbacks

RunDevelopment:
This PR fixes and improves the internal Autoloader callbacks.

### Bug

Callbacks were stored in `success_callbacks` and `error_callbacks` and not removed after execution meaning once added, a callback was executed every time the language for which the callback was added for was requested (either directly or as a dependency).

This bug wasn't very visible because of how the Autoloader is usually used. Usually, the Autoloader is used to load the languages for the static code in a page. This means that the Autoloader gets asked for a bunch of languages, loads them asynchronously and then executes the callbacks with this [line](https://github.com/PrismJS/prism/blob/d4373f3a282f4c09e4fb4dcbd8885c7848b61073/plugins/autoloader/prism-autoloader.js#L187) (`success` is [this](https://github.com/PrismJS/prism/blob/d4373f3a282f4c09e4fb4dcbd8885c7848b61073/plugins/autoloader/prism-autoloader.js#L306) function). Because it takes a little while until the language is loaded (certainly until the next tick), all callbacks will be registered before the `onload` of the script is called meaning that [this](https://github.com/PrismJS/prism/blob/d4373f3a282f4c09e4fb4dcbd8885c7848b61073/plugins/autoloader/prism-autoloader.js#L300) never gets executed.
Thus the bug doesn't show up. (And even if it does, the success callback is usually just a `Prism.highlightElement` call.)

I discovered this bug when I used Autoloader to highlight C like code and then after that was done (some ticks later), I used Autoloader to highlight JavaScript code. What followed was that the synchronously called callbacks caused a stack overflow. Just a bit of good ol' infinite recursion.

I could have also fixed it by deferring callbacks to the next tick but I decided to rework the callback structure because this change of behavior might break some code. Autoloader invokes the callbacks of already loaded languages (and failed ones) synchronously while all other callbacks are called asynchronously. This fix keeps this behavior.
(It's another question as to whether we should call all callbacks asynchronously. <sub><sup>Yes, we should IMO.</sup></sub>)

### Improvements

1. [This](https://github.com/PrismJS/prism/blob/d4373f3a282f4c09e4fb4dcbd8885c7848b61073/plugins/autoloader/prism-autoloader.js#L218) declaration and usage of `data` is completely unnecessary and was removed.
1. Typings for `lang_data`.


plugins/autoloader/prism-autoloader.js
```diff
@@ -152,6 +152,13 @@
 		"yml": "yaml"
 	}/*]*/;
 
+	/**
+	 * @typedef LangDataItem
+	 * @property {{ success: function, error: function }[]} callbacks
```
mAAdhaTTah: The way the code checks them below, each of these are optional?

RunDevelopment: Yes, they are!

PR #1936, opened 2019-06-09
Update test and example pages to use Autoloader

RunDevelopment:
This PR makes the test and example pages use the Autoloader to load language definitions.

__Motivation__

I wanted to get rid of [this](https://github.com/PrismJS/prism/blob/d4373f3a282f4c09e4fb4dcbd8885c7848b61073/test.html#L206) [copypasted](https://github.com/PrismJS/prism/blob/d4373f3a282f4c09e4fb4dcbd8885c7848b61073/scripts/examples.js#L171) code altogether.

---

This is based on #1935.


scripts/examples.js
```diff
@@ -152,64 +152,23 @@ function update(id) {
...
-
-			loadLanguage(id).then(function () {
-				Prism.highlightAllUnder(examples[id]);
+			/** @type {HTMLElement} */
+			var container = examples[id];
+			container.innerHTML = buildContentsHeader(id) + contents;
+
+			// the current language might be an extension of a language
+			// so to be safe, we explicitly add a dependency to the current language
+			$$('pre', container).forEach(/** @param {HTMLElement} code */function (pre) {
```
mAAdhaTTah:
```suggestion
			$$('pre', container).forEach(/** @param {HTMLElement} pre */function (pre) {
```

RunDevelopment:
I should have communicated this earlier but this PR has one major issue: Extensions.
When you select an extension like CSS Extras on the test page, the Autoloader will go into an infinite loop loaden `prism-css-extras.js` about 100/s (localhost). This is because of the way Autoloader decides whether a language has been loaded yet by checking `Prism.languages['css-extras']`. 
The problem: For extensions, this property remains undefined.

If you assume that languages won't be uploaded again after they are loaded, Autoloader could just keep an internal record of every loaded language as an addition to `Prism.languages`.
This is a breaking change because you won't be able to load a language again after unloading it, but how unloads languages on their website?

I'll implement this soon, so please re-review after that.
@mAAdhaTTah @Golmote

RunDevelopment:
> soon

±2 years.

---

Since the extension issue has since been solved and the PR has already been approved, I'll just merge this now.

PR #1938, opened 2019-06-10
SCSS: Fixed exp time failing of url pattern

### LINKED ISSUE #1937: scss specific text lasts forever to highlight
Example:
/* 16:44:48: -------------------------------------------------- */

in definition you only need to change url regex
FROM: 'url': /(?:[-a-z]+-)*url(?=\()/i
TO: 'url': /(?:[-a-z]+-)?url(?=\()/i,

It should have the same result, just it is much faster in processing dashes.

Thanks

RunDevelopment:
This fixes #1937.

Thank you @mhosna.


RunDevelopment:
I will now merge this because of the severity of the problem and the simplicity of the fix.

PR #1943, opened 2019-06-14
Add `return` to recognised Perl keywords

leonerd:


RunDevelopment:
@leonerd Thank you for this PR!

However, before we can merge this there are two things which have to be done:

1. Rebuild Prism. This can be done by using the `npx gulp` command.
2. Add the new keyword to the [Perl keyword test](https://github.com/PrismJS/prism/blob/master/tests/languages/perl/keyword_feature.test).

leonerd:
> 1. Rebuild Prism. This can be done by using the npx gulp command.
That looks like it's going to be a little nontrivial - I don't have such a command:

```$ npx gulp
bash: npx: command not found```

Any pointer to how to grab that? And/or I'll just allow others to push a new commit to this branch if someone else is better set-up to manage it.

> 2. Add the new keyword to the Perl keyword test.
Done in 4a0ed18d

**Comment by RunDevelopment** [comment]
You need to have `npm` installed for this to work. Install any recent-ish version of [NodeJS](https://nodejs.org/en/download/) and your good to go.

**Comment by leonerd** [comment]
Hrm. I'm not sure I'm doing very well so far:

```
$ apt-get install nodejs
$ npm install npx
$ npm install gulp
```

but then

```
$ node_modules/.bin/npx gulp
internal/modules/cjs/loader.js:583
    throw err;
    ^

Error: Cannot find module 'gulp-rename'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
    at Function.Module._load (internal/modules/cjs/loader.js:507:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/home/leo/src/prism/gulpfile.js:3:16)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
```

**Comment by RunDevelopment** [comment]
Ahhhh right.
You have to install node's dependencies first. So first `npm install` to install all dependencies, then `npx gulp` to rebuild Prism.

**Comment by leonerd** [comment]
Ah:

```
$ npm install gulp-rename
$ npm install gulp-uglify
$ npm install gulp-header
$ npm install gulp-concat
$ npm install gulp-replace
$ npm install pump
$ npm install simple-git
$ npm install shelljs
```

and now it appears to run. It seems to have modified a fair few files more than just the perl one I was expecting though:

```
$ git status
On branch master
Your branch is up to date with 'leonerd/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   components/prism-abnf.min.js
        modified:   components/prism-batch.min.js
        modified:   components/prism-core.min.js
        modified:   components/prism-perl.min.js
        modified:   components/prism-powershell.min.js
        modified:   components/prism-toml.min.js
        modified:   plugins/autoloader/prism-autoloader.min.js
        modified:   plugins/command-line/prism-command-line.min.js
        modified:   plugins/custom-class/prism-custom-class.min.js
        modified:   plugins/line-numbers/prism-line-numbers.min.js
        modified:   plugins/previewers/prism-previewers.min.js
        modified:   plugins/toolbar/prism-toolbar.min.js

```

**Comment by leonerd** [comment]
b52a16b4 contains just the modified `prism-perl.min.js`, and I'll ignore the others for now.

**Comment by RunDevelopment** [comment]
I'm not sure, why it changed this many files. It didn't do this on my machine. No idea why it does this.

---

Small note: You just have to run the following to install all dependencies:

```
$ npm install
```

That's it.

**Comment by RunDevelopment** [comment]
Nice. Tests passed.

Thank you for contributing!

**Comment by leonerd** [comment]
While I'm here, more of a question: I was actually looking at the file, in order to work out how to add some new keywords provided by some of my CPAN modules. These keywords are `async`, `await`, `try`, `catch`, `finally`. I have managed to get those added by the use of a cunning `<script>` tag in the page:

```
<script type="text/javascript">
    Prism.languages.perl.keyword = [
        Prism.languages.perl.keyword,
        /\b(async|await|catch|finally|return|try)\b/,
    ];
</script>
```

but I was wondering, would it be possible to just store the toplevel `keywords` entry already as a list? That way I could use a somewhat simpler:

```
    Prism.languages.perl.keyword.push(/\b(async|await|catch|finally|try)\b/);
```

mAAdhaTTah:
I think for most use cases, having an array with a single item is marginally to be less ergonomic.

PR #1944, opened 2019-06-15
New Match Braces plugin

RunDevelopment:
This adds a new plugin for highlighting matching braces to Prism.

Sometimes it's hard to find the partner of a brace which is why most IDEs highlight the brace pair when the text cursor is next to one of the braces.
Since Prism isn't a text editor, I opted for highlighting brace pair on hover and on click.

I also added a simple rainbow braces mode.

### Examples

image

image


mAAdhaTTah:
This is sweeeeeeeeeeeeeet. I am working on an editor with prism (it's tied up in another plugin) so this would be cool to integrate. Will need to take a pass through.

plugins/match-braces/prism-match-braces.js
```diff
@@ -0,0 +1,182 @@
...
+	var pairIdCounter = 0;
+
+	/**
+	 * Returns the brace partner given one brace of a brace pair.
+	 *
+	 * @param {HTMLElement} brace
+	 * @returns {HTMLElement}
+	 */
+	function getPartnerBrace(brace) {
+		var match = /^(pair-\d+-)(open|close)$/.exec(brace.id);
```
mAAdhaTTah: Can we lift this regex outside of this function?

plugins/match-braces/prism-match-braces.js
```diff
@@ -0,0 +1,182 @@
...
+				var openEle = punctuation[pair[0]];
+				var closeEle = punctuation[pair[1]];
+
+				openEle.id = pairId + 'open';
+				closeEle.id = pairId + 'close';
+
+				[openEle, closeEle].forEach(function (ele) {
+					ele.addEventListener('mouseenter', hoverBrace);
+					ele.addEventListener('mouseleave', leaveBrace);
+					ele.addEventListener('click', clickBrace);
```
mAAdhaTTah: If this can be rehighlighted, should these have checks as well?

RunDevelopment: What do you with _rehighlighted_ and what other checks are there already?

mAAdhaTTah: This check: https://github.com/PrismJS/prism/pull/1944/files#diff-9d8e1aa772fc2efe8149a426bc770d93R106-R116

RunDevelopment: Ahh, ok. We don't need the checks here because when we rehighlight, all of the original spans get removed and replaced with new ones (the Prism generated HTML).
So we only have to check the `pre` and `code` (which isn't being used).

mAAdhaTTah: I _think_ that may produce a memory leak, as we're removing DOM nodes but not removing their associated event handlers, which would prevent both the DOM node & the handler from being garbage collected. Not sure how to solve that though.

RunDevelopment: I'm pretty sure that's not the case because for the listener to prevent the garbage collection of the element, the listener would have to reference the element which it doesn't.

mAAdhaTTah: Ok, yeah, I confirmed this with some colleagues, so this looks good to me.

PR #1947, opened 2019-06-21
pacaligo: adding support for pascaligo, http://ligolang.org

DefinitelyNotAGoat:
**Summary**
This PR adds the language definition for pascaligo, a layer 2 smart contract language for the Tezos blockchain. 

http://ligolang.org


RunDevelopment:
@DefinitelyNotAGoat Thank you for making this PR.

However, I have a little question: Isn't this language definition basically the combination of PascalLIGO and CameLIGO which are two languages with different syntax?

And a small note: the tests failed because `components/prism-ligo.min.js` isn't up to date. Rebuilding Prism should fix this issue.

I'll do the full review soon.

DefinitelyNotAGoat:
Thanks @RunDevelopment! So this is actually the pascaligo version so I will edit to make sure that's noted, rebuild and push.

components.json
```diff
@@ -632,6 +632,14 @@
 			},
 			"owner": "Golmote"
 		},
+		"pascaligo": {
+			"title": "Pascaligo",
+			"owner": "DefinitelyNotAGoat",
+			"alias": "pascaligo",
+			"aliasTitles": {
```
RunDevelopment: You don't need `alias` and `aliasTitles` anymore.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,58 @@
...
+		// Decimal
+		/\b\d+(?:\.\d+)?(?:e[+-]?\d+)?/i
+	],
+	'operator': [
+		/\.\.|\*\*|:=|<[<=>]?|>[>=]?|[+\-*\/]=?|[@^=]/i,
+	],
+	'punctuation': /\(\.|\.\)|[()\[\]:;,.]/
+};
+
+Prism.languages.pascaligo = Prism.languages.pascaligo;
```
RunDevelopment: Same here with the alias.

DefinitelyNotAGoat:
Rebuilt, and pushed name changes to accurately reflect pascaligo.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,58 @@
...
+        pattern: /\w+ ?(?= is ?)/i,
+    },
+	'number': [
+		// Hexadecimal, octal and binary
+		/(?:[&%]\d+|\$[a-f\d]+)/i,
+		// Decimal
+		/\b\d+(?:\.\d+)?(?:e[+-]?\d+)?/i
+	],
+	'operator': [
+		/\.\.|\*\*|:=|<[<=>]?|>[>=]?|[+\-*\/]=?|[@^=]/i,
```
RunDevelopment: Unnecessary `i` flag and array. Just `'operator': /\.\.|\*\*|:=|<[<=>]?|>[>=]?|[+\-*\/]=?|[@^=]/` is enough.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,58 @@
+// Pascaligo is a layer 2 smart contract language for the tezos blockchain
+
+Prism.languages.pascaligo = {
+	'comment': [
+		/\(\*[\s\S]+?\*\)/,
+		/\/\/.*/
```
RunDevelopment: I think it would better to combine these two patterns so that code like this `// (* foo *)` will be matched correctly.

So `/\(\*[\s\S]+?\*\)|\/\/.*/`

components/prism-pascaligo.js
```diff
@@ -0,0 +1,58 @@
...
+	],
+	'string': [
+		{
+			pattern: /(?:'(?:''|[^'\r\n])*'|#[&$%]?[a-f\d]+)+|\^[a-z]/i,
+			greedy: true
+		},
+		{
+			pattern: /(?:"(?:""|[^"\r\n])*"|#[&$%]?[a-f\d]+)+|\^[a-z]/i,
```
RunDevelopment: Does this mean that `"foo"#7` is also a valid string?

components/prism-pascaligo.js
```diff
@@ -0,0 +1,58 @@
...
+			greedy: true
+		},
+	],
+	'keyword': [
+		{
+			pattern: /(^|[^&])\b(?:if|while|for|return|nil|remove|from|else|then|skip|with|is|function|var|const|type|end|begin|block)\b/i,
+			lookbehind: true
+		},
+		{
+			pattern: /(^|[^&])\b(?:True|False)\b/i,
```
RunDevelopment: Could you please make this a `boolean` pattern for consistency with other languages.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,58 @@
...
+			greedy: true
+		},
+		{
+			pattern: /(?:"(?:""|[^"\r\n])*"|#[&$%]?[a-f\d]+)+|\^[a-z]/i,
+			greedy: true
+		},
+	],
+	'keyword': [
+		{
+			pattern: /(^|[^&])\b(?:if|while|for|return|nil|remove|from|else|then|skip|with|is|function|var|const|type|end|begin|block)\b/i,
```
RunDevelopment: Please sort the keyword list.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,58 @@
...
+			pattern: /(^|[^&])\b(?:True|False)\b/i,
+			lookbehind: true
+		},
+    ],
+    'builtin': {
+        pattern: /(^|[^&])\b(?:int|unit|string|nat|map)\b/i,
+		lookbehind: true
+    },
+    'function': {
+        pattern: /(?!with\b)\b\w+ ?(?=\()/i,
```
RunDevelopment: Shouldn't `with` be matched by `keyword` already? Can you give an example where you need the negative lookahead?

components/prism-pascaligo.js
```diff
@@ -0,0 +1,58 @@
...
+    },
+    'variable': [
+        { 
+            pattern: /\w+ ?(?=\:\=)/i,
```
RunDevelopment: This pattern is a subset of the below pattern and therefore unnecessary. It can be removed.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,58 @@
...
+        },
+        { 
+            pattern: /\w+ ?(?=\:)/i,
```
RunDevelopment: Unnecessary `i` flag.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,58 @@
...
+        },
+    ],
+    'class-name': {
+        pattern: /\w+ ?(?= is ?)/i,
```
RunDevelopment: A word about spaces: We generally opt to not included leading or trailing spaces inside tokens. Also, just matching normal spaces is usually not enough which is why we usually use `\s` (which also includes tabs and newlines) for spaces in C style languages.
This also applies to the `function` and `variable` patterns.

```suggestion
        pattern: /\w+(?=\s+is\b)/i,
```

tests/languages/pascaligo/builtin_feature.test
```diff
@@ -0,0 +1,12 @@
...
+----------------------------------------------------
+
+[
+	["builtin", "int"], ["builtin", "unit"],
+	["builtin", "string"], ["builtin", "nat"], ["builtin", "map"]
+]
+
+----------------------------------------------------
+
+Checks for comments.
```
RunDevelopment:
```suggestion
Checks for builtins.
```

tests/languages/pascaligo/function_feature.test
```diff
@@ -0,0 +1,23 @@
...
+	["punctuation", ")"],
+
+	["function", "somefunc42"],
+	["punctuation", "("],
+	["punctuation", ")"]
+]
+
+----------------------------------------------------
+
+Checks for comments.
```
RunDevelopment:
```suggestion
Checks for functions.
```

components/prism-pascaligo.js
DefinitelyNotAGoat: Yeah this is mainly functions that look like:
```
function pop (const h : heap) : (heap * heap_element * nat) is
  begin
   const result : heap_element = get_top (h) ;
   var s : nat := size(h) ;
   const last : heap_element = get_force(s, h) ;
   remove s from map h ;
   h[1n] := last ;
   s := size(h) ;
   var i : nat := 0n ;
   var largest : nat := 1n ;
   var left : nat := 0n ;
   var right : nat := 0n ;
   var c : nat := 0n ;
   while (largest =/= i) block {
     c := c + 1n ;
     i := largest ;
     left := 2n * i ;
     right := left + 1n ;
     if (left <= s) then begin
       if (heap_element_lt(get_force(left , h) , get_force(i , h))) then begin
        largest := left ;
        const tmp : heap_element = get_force(i , h) ;
        h[i] := get_force(left , h) ;
        h[left] := tmp ;
       end else skip ;
     end else if (right <= s) then begin
       if (heap_element_lt(get_force(right , h) , get_force(i , h))) then begin
         largest := right ;
         const tmp : heap_element = get_force(i , h) ;
         h[i] := get_force(right , h) ;
         h[left] := tmp ;
       end else skip ;
     end else skip ;
  }
  end with (h , result , c)
```

specifically the end of the function  `with (h , result , c)`. This is a return statement and not to be identified as a function.

RunDevelopment: Seems like I missed that `with` is a keyword already. In this case, you don't need the assertion at all.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,58 @@
...
+	'number': [
+		// Hexadecimal, octal and binary
+		/(?:[&%]\d+|\$[a-f\d]+)/i,
+		// Decimal
+		/\b\d+(?:\.\d+)?(?:e[+-]?\d+)?/i
+	],
+	'operator': [
+		/\.\.|\*\*|:=|<[<=>]?|>[>=]?|[+\-*\/]=?|[@^=]/i,
+	],
+	'punctuation': /\(\.|\.\)|[()\[\]:;,.]/
```
RunDevelopment: For [your code example](https://github.com/PrismJS/prism/pull/1947/files#r301338524) I noticed that `{` and `}` should also be punctuation.

DefinitelyNotAGoat:
@RunDevelopment thanks for the review. 

I added some new changes to correct the issues. Let me know if you need anything else!

examples/prism-pascaligo.html
```diff
@@ -0,0 +1,63 @@
...
+
+<h2>Functions</h2>
+<pre><code>foo()
+Bar()
+_456()
+</code></pre>
+
+<h2>Full Example</h2>
+<pre><code>
+    function pop (const h : heap) : (heap * heap_element * nat) is
```
RunDevelopment: I really like the example but its indentation is kind of awkward. Sometimes one space is used, sometimes two. If there is no preferred indentation style set by the language (or its community) then please opt for tabs.
Also, the... well... global indentation of 4 spaces should be removed.

tests/languages/pascaligo/boolean_feature.test
```diff
@@ -0,0 +1,11 @@
+True False
+
+----------------------------------------------------
+
+[
+["boolean", "True"], ["boolean", "False"]
```
RunDevelopment: Indent this, please.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,45 @@
+// Pascaligo is a layer 2 smart contract language for the tezos blockchain
+
+Prism.languages.pascaligo = {
+	'comment': /\(\*[\s\S]+?\*\)|\/\/.*/,
+	'string': [
+		{
+			pattern: /(?:'(?:''|[^'\r\n])*')+|\^[a-z]/i,
```
RunDevelopment: This pattern `(?:'(?:''|[^'\r\n])*')+` as well as the below `"`-version can potentially show exponential backtracking because of the `+`.
Luckily, this is really easy to fix as the patterns `(?:'(?:''|[^'\r\n])*')+` and `'(?:''|[^'\r\n])*'` are equivalent.

```suggestion
			pattern: /'(?:''|[^'\r\n])*'|\^[a-z]/i,
```

components/prism-pascaligo.js
```diff
@@ -0,0 +1,45 @@
...
+			greedy: true
+		},
+		{
+			pattern: /(?:"(?:""|[^"\r\n])*")+|\^[a-z]/i,
```
RunDevelopment: Couldn't we combine this pattern with the above?
This would get rid of the duplicate `\^[a-z]`.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,45 @@
...
+		pattern: /\w+ ?(?=\:)/,
+	},
+    'class-name': {
+        pattern: /\w+(?=\s+is\b)/i,
+    },
+	'number': [
+		// Hexadecimal, octal and binary
+		/(?:[&%]\d+|\$[a-f\d]+)/i,
+		// Decimal
+		/\b\d+(?:\.\d+)?(?:e[+-]?\d+)?/i
```
RunDevelopment: Please also add support for the [natural number syntax](https://ligolang.org/docs/next/language-basics/cheat-sheet/).

components/prism-pascaligo.js
```diff
@@ -0,0 +1,45 @@
...
+			pattern: /(?:'(?:''|[^'\r\n])*')+|\^[a-z]/i,
+			greedy: true
+		},
+		{
+			pattern: /(?:"(?:""|[^"\r\n])*")+|\^[a-z]/i,
+			greedy: true
+		},
+	],
+	'keyword': {
+		pattern: /(^|[^&])\b(?:begin|block|const|else|end|for|from|function|if|is|nil|remove|return|skip|then|type|var|while|with)\b/i,
```
RunDevelopment: The `case` and `of` keywords are missing.
Example:
```pascaligo
    case p of
    | Increment(n) -> add(s, n)
    | Decrement(n) -> subtract(s, n)
    end
```

components/prism-pascaligo.js
```diff
@@ -0,0 +1,45 @@
...
+    'class-name': {
+        pattern: /\w+(?=\s+is\b)/i,
+    },
+	'number': [
+		// Hexadecimal, octal and binary
+		/(?:[&%]\d+|\$[a-f\d]+)/i,
+		// Decimal
+		/\b\d+(?:\.\d+)?(?:e[+-]?\d+)?/i
+	],
+	'operator': /\.\.|\*\*|:=|<[<=>]?|>[>=]?|[+\-*\/]=?|[@^=]/,
```
RunDevelopment: Missing operators `->`, `=/=`, and `|`.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,45 @@
...
+    },
+    'function': {
+        pattern: /(?!with\b)\b\w+ ?(?=\()/i,
+    },
+    'variable': 
+	{ 
+		pattern: /\w+ ?(?=\:)/,
+	},
+    'class-name': {
+        pattern: /\w+(?=\s+is\b)/i,
```
RunDevelopment: This pattern can't match anything because `is` already matched by `keyword`. You have to move the `class-name` pattern before the `keyword` pattern for it to work.

components/prism-pascaligo.js
```diff
@@ -0,0 +1,45 @@
...
+	'keyword': {
+		pattern: /(^|[^&])\b(?:begin|block|const|else|end|for|from|function|if|is|nil|remove|return|skip|then|type|var|while|with)\b/i,
+		lookbehind: true
+	},
+	'boolean': {
+		pattern: /(^|[^&])\b(?:True|False)\b/i,
+		lookbehind: true
+	},
+    'builtin': {
+        pattern: /(^|[^&])\b(?:int|unit|string|nat|map)\b/i,
```
RunDevelopment: Shouldn't this also include [records](https://ligolang.org/docs/next/language-basics/cheat-sheet/)?
Example:
```
type person is record
  age: int ;
  name: string ;
end
```

DefinitelyNotAGoat:
Thanks for the tips @RunDevelopment! I re-pushed new changes to reflect the change requests.

RunDevelopment:
Nice!
Could please also take a look at [the PR](https://github.com/DefinitelyNotAGoat/prism/pull/1) I send you?

RunDevelopment:
I think this is ready.

Thank you for contributing @DefinitelyNotAGoat!

PR #1950, opened 2019-06-22
JS: Private fields

RunDevelopment:
This PR adds support for JS's new [private field](https://github.com/tc39/proposal-class-fields#private-fields) by highlighting the `#` as an operator.

### Example

image


RunDevelopment:
I made it an operator because I wanted _some_ highlighting but looking at [this](https://github.com/tc39/proposal-class-fields/issues/250#issuecomment-506012450) it might have been the wrong choice.

> \# is part of the name

The [spec](https://github.com/tc39/proposal-class-fields/blob/ac34be11094553690210977d660e3fd7cf87589d/spec.html#L92).

Maybe a new token (e.g. `sigil`) is the best solution?

Anyhow, I have to update this PR because I didn't consider private async function.

mAAdhaTTah:
If it's part of the name, then maybe it shouldn't be highlighted separately? Especially if it's supposed to be part of the name as the "mental model", we'd be do a disservice to that way of thinking by highlighting it separately.

Not in love with a new token, unless you're suggesting aliasing it to `operator`?

RunDevelopment:
> we'd be doing a disservice to that way of thinking by highlighting it separately.

That's true.

Since it's not clear yet as to whether and how we should highlight the `#`, I'll remove the highlighting for now. We can always add it in the future may the need arise.

The PR is not closed because the function and function variables tokens have to be adjusted (see new test).

RunDevelopment:
Re-reviewed and adjusted JS Extras. I think it's ready.

PR #1951, opened 2019-06-25
Add className option for toolbar items

mkaz:
Allows for passing in a className when registering a toolbar button to allow to target via CSS.

An example:
```js
   Prism.plugins.toolbar.registerButton( 'my-button', {
        text: 'My Button',
        className: 'my-button',
    });
```
This will add the span element `<span class="my-button">My Button</span>`


mkaz:
@RunDevelopment Updated my PR per review, thanks!

plugins/toolbar/prism-toolbar.js
```diff
@@ -14,6 +14,7 @@
 	 * @property {string} text The text displayed.
 	 * @property {string} [url] The URL of the link which will be created.
 	 * @property {Function} [onClick] The event listener for the `click` event of the created button.
+	 * @property {string} className The class attribute to include with element.
```
mAAdhaTTah: Mark as optional:

```suggestion
	 * @property {string} [className] The class attribute to include with element.
```

mAAdhaTTah:
One minor nit but lgtm!

mAAdhaTTah:
Thank you for the contribution!

PR #1956, opened 2019-07-01
F#: Improved char literals

RunDevelopment:
This adds the missing `\xFF` and `\uFFFF` character escapes to character literals.

I also removed the incorrect "string" from the F# example page.


RunDevelopment:
Finally found [a page which lists all escape sequences](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/strings#remarks). I think this is ready.

PR #1957, opened 2019-07-01
Doubly check the `manual` flag

TimWolla:
Currently the `manual` is only checked before attaching the event
listener, making it hard to reliably set the flag without jumping
through hoops and special cases if Prism is bundled together with
other scripts into a single script file.
Change to check the flag once before attaching the event listener
and once again before actually highlighting to support setting it
after the Prism initialization has run.

see discussion starting at:
https://github.com/PrismJS/prism/pull/1087#issuecomment-274959529

----

Quoting from the discussion in the issue:

> We can; we would break BC for people using the old document.removeEventListener('DOMContentLoaded', Prism.highlightAll) method of removing auto-highlighting, as now the callback we're attaching won't be just Prism.highlightAll.

and

> Fine by me; if we do that, we should document that somewhere though. Is the CHANGELOG enough or should we include a "how to disable auto-highlighting" somewhere?

Should I make some other adjustments anywhere? /cc @mAAdhaTTah


TimWolla:
Misunderstood the requirement to not touch anything outside of `components/` and did not run gulp. Now fixed and force pushed.

TimWolla:
This pull request is the only PR that is not labeled at all, so it looks like it might have slipped through the cracks. Can you take a look, please?

components/prism-core.js
```diff
@@ -524,15 +524,21 @@ if (script) {
 	_.filename = script.src;
 
 	if (!_.manual && !script.hasAttribute('data-manual')) {
+		_.highlightAutomaticallyCallback = function () {
```
RunDevelopment: I'm not a fan of exposing this function.
I guess that it's so you can still do `document.removeEventListener('DOMContentLoaded', Prism.highlightAutomaticallyCallback)`  but I don't think that this will help anyone. It doesn't preserve backward-compatibility and if we break things, we might as well force people to use `Prism.manual`.

components/prism-core.js
```diff
@@ -524,15 +524,21 @@ if (script) {
 	_.filename = script.src;
 
 	if (!_.manual && !script.hasAttribute('data-manual')) {
```
RunDevelopment: This is only related to the intent of this PR but shouldn't `data-manual` set `Prism.manual`?

I think that it's really strange that you can include Prism via

```html
<script src="prism.js" data-manual></script>
```

And it will switch to manual highlighting without, well, being set to manual.

components/prism-core.js
TimWolla: Yes, I exposed it for that reason. I agree that it would be redundant with either `Prism.manual` or `Prism.highlightAll()` and could be removed.

components/prism-core.js
TimWolla: It probably makes sense to set this flag when the HTML attribute is set to allow reading it from other code.

TimWolla:
> I think we definetly need the doc and should put this on your front page in the basic usage section (preferably above the CDN section).

This can simply go into `index.html`?

RunDevelopment:
> This can simply go into index.html?

Yes.

I didn't notice it already mentioned `data-manual`. What an attentive reader I am...
In that case, just `Prism.manual` and how `data-manual` affects `Prism.manual` is enough.

TimWolla:
> In that case, just `Prism.manual` and how `data-manual` affects `Prism.manual` is enough.

Okay, I've added the docs as well. Let me know if there's something else I forgot!

index.html
```diff
@@ -149,9 +149,16 @@ <h1>Basic usage</h1>
 	<pre><code>&lt;pre>&lt;code class="language-css">p { color: red }&lt;/code>&lt;/pre></code></pre>
 	<p>If you use that pattern, the <code>&lt;pre></code> will automatically get the <code>language-xxxx</code> class (if it doesn’t already have it) and will be styled as a code block.</p>
 
-	<p>If you want to prevent any elements from being automatically highlighted, you can use the attribute <code>data-manual</code> on the <code>&lt;script></code> element you used for prism and use the <a href="extending.html#api">API</a>.
+	<p>If you want to prevent any elements from being automatically highlighted and instead use the <a href="extending.html#api">API</a>, you can set <code>Prism.manual = true;</code> before <code>DOMContentLoaded</code> is fired. By setting the attribute <code>data-manual</code> on the <code>&lt;script></code> element you used for prism this will be done automatically.
```
RunDevelopment: > before <code>DOMContentLoaded</code> is fired

Not only that but also before the timeouts we use otherwise. I wouldn't mention any specific DOM events.

index.html
```diff
@@ -149,9 +149,16 @@ <h1>Basic usage</h1>
...
 	Example:</p>
 	<pre><code>&lt;script src="prism.js" data-manual>&lt;/script></code></pre>
+	<p>or</p>
+	<pre><code>&lt;script>
+window.Prism = window.Prism || {};
+window.Prism.manual = true;
```
RunDevelopment: I know that this is what was suggested in the issue thread you linked to but we know that `highlightAutomaticallyCallback` is always called synchronously, so why don't we use this fact.

```html
<script>Prism.manual = true;</script>
```

Shouldn't this work as well?

index.html
TimWolla: `DOMContentLoaded` is the latest point you can _rely_ on. If the page is already loaded and Prism is dynamically embedded one is not able to reliably set manual (unless the file containing Prism also sets the flag), because you cannot ensure that your code runs before the timers run.

So while technically it's not 100% precise I believe the information is valuable for consumers that cannot simply set `data-manual`.

index.html
TimWolla: Yes, I believe it should for a simple script tag. However I prefer the current version, because your suggestion will break if e.g. `async` is put onto the script tag or Prism is loaded by a loader such as require.js or head.js.

RunDevelopment: That's true.

index.html
RunDevelopment: > one is not able to reliably set manual

> However I prefer the current version, because your suggestion will break if e.g. async is put onto the script tag or Prism is loaded by a loader such as require.js or head.js.

As you implied, it's possible to reliably set `manual` using the example you added to the site.

That aside, You convinced me that we should mention the DOM event.

Also, here is a small suggestion to make the doc a bit more specific:

```diff
-, you can set <code>Prism.manual = true;</code> before <code>DOMContentLoaded</code> is fired.
+, you can set <code>Prism.manual</code> to <code>true</code> before the <code>DOMContentLoaded</code> event is fired.
```

And

```diff
-By setting the attribute <code>data-manual</code> on the <code>&lt;script></code> element you used for prism this will be done automatically.
+By setting the <code>data-manual</code> attribute on the <code>&lt;script></code> element containing Prism core, this will be done automatically.
```

TimWolla: Changed to your suggestion.

index.html
```diff
@@ -149,9 +149,16 @@ <h1>Basic usage</h1>
 	<pre><code>&lt;pre>&lt;code class="language-css">p { color: red }&lt;/code>&lt;/pre></code></pre>
 	<p>If you use that pattern, the <code>&lt;pre></code> will automatically get the <code>language-xxxx</code> class (if it doesn’t already have it) and will be styled as a code block.</p>
 
-	<p>If you want to prevent any elements from being automatically highlighted, you can use the attribute <code>data-manual</code> on the <code>&lt;script></code> element you used for prism and use the <a href="extending.html#api">API</a>.
+	<p>If you want to prevent any elements from being automatically highlighted and instead use the <a href="extending.html#api">API</a>, you can set <code>Prism.manual</code> to <code>true</code> before the <code>DOMContentLoaded</code> event is fired. By setting the <code>data-manual</code> attribute on the <code>&lt;script></code> element containing Prism core, this will be done automatically.
```
RunDevelopment: One last suggestion from my side. After this one, I think this is ready to merge.

```suggestion
	<p>If you want to prevent any elements from being automatically highlighted and instead use the <a href="extending.html#api">API</a>, you can set <code class="language-javascript">Prism.manual</code> to <code class="language-javascript">true</code> before the <code>DOMContentLoaded</code> event is fired. By setting the <code>data-manual</code> attribute on the <code>&lt;script></code> element containing Prism core, this will be done automatically.
```

TimWolla: Makes sense. Amended my previous doc commit and force pushed.

components/prism-core.js
```diff
@@ -522,19 +522,29 @@ var script = document.currentScript || [].slice.call(document.getElementsByTagNa
...
+	if(document.readyState !== "loading") {
+		if (window.requestAnimationFrame) {
+			window.requestAnimationFrame(highlightAutomaticallyCallback);
+		} else {
+			window.setTimeout(highlightAutomaticallyCallback, 16);
 		}
 	}
+	else {
+		document.addEventListener('DOMContentLoaded', highlightAutomaticallyCallback);
+	}
```
mAAdhaTTah: Idea: add a function to the `Prism` object here to unhook the automatic highlighting. Thoughts?

TimWolla: I'm not sure whether I understand correctly, but please see this thread: https://github.com/PrismJS/prism/pull/1957#discussion_r307058234

I don't think that adding a function is going to make a difference to setting the `manual` property to `true`.

mAAdhaTTah:
Just need to resolve the conflicts.

TimWolla:
> Just need to resolve the conflicts.

@mAAdhaTTah I've rebased my branch onto PrismJS:master and resolved the conflicts.

RunDevelopment:
Thank you for contributing @TimWolla!

components/prism-core.js
```diff
@@ -533,21 +533,21 @@ if (script) {
 	_.filename = script.src;
 
 	if (!_.manual && !script.hasAttribute('data-manual')) {
-		_.highlightAutomaticallyCallback = function () {
+		function highlightAutomaticallyCallback() {
```
piotr-cz: After this change it's not possible to minify script using uglify-js and `"use strict"` declaration:

```
  13971 |
  13972 |       if (!_.manual) {
> 13973 |               function highlightAutomaticallyCallback() {
        |       ^ In strict mode code, functions can only be declared at top level or immediately within another function.
  13974 |                       if (!_.manual) {
  13975 |                               _.highlightAll();
  13976 |                       }
```

Functions shouldn't be declared inside the if code block.

PR #1961, opened 2019-07-02
Added support for DNS zone files

### LINKED ISSUE #1959: support for DNS zone file syntax
**Language**
DNS zone file

**Additional resources**
https://en.wikipedia.org/wiki/Zone_file

**Example**

Github uses highlight with dns-zone language, see below

```dns-zone
$TTL 3d
@ IN SOA root.localhost. root.sneaky.net. (
    2015050503 ; serial
    12h        ; refresh
    15m        ; retry
    3w         ; expire
    3h         ; negative response TTL
  )
  IN NS root.localhost.
  IN NS localhost. ; secondary name server is preferably externally maintained

www IN A 3.141.59.26
ww1 IN CNAME www
```

RunDevelopment:
This adds support for DNS zone files.

The highlighting is quite simple as every type and class is highlighted as `keyword`. This is intentional as other token names (e.g. `class-name`, `function`, `builtin`, ...) are not highlighted by every theme resulting in large portions of unstyled text for some themes.

This resolves #1959.

### Example

image

Code by @BnitoBzh.

---

@BnitoBzh Can I use your code as the language example? Your code will be visible on [this page](https://prismjs.com/examples.html) after this PR is merged.


BnitoBzh:
@RunDevelopment Yes you can ! And thank you for your reactivity !

PR #1966, opened 2019-07-07
Toolbar: Fixed overflow issue

RunDevelopment:
This fixed the overflow issue of Toolbar with Prism's Theme selection. The fixed bug can be seen on the plugin pages of [Toolbar](https://prismjs.com/plugins/toolbar/), [Copy to Clipboard](https://prismjs.com/plugins/copy-to-clipboard/), and [Show language](https://prismjs.com/plugins/show-language/).

I'm aware that the artificial transfer of the `pre` margin to the wrapper isn't an elegant solution, so I'm open for suggestions.


plugins/toolbar/prism-toolbar.js
```diff
@@ -86,6 +86,10 @@
 		pre.parentNode.insertBefore(wrapper, pre);
 		wrapper.appendChild(pre);
 
+		// transfer margin of pre to wrapper
+		wrapper.style.margin = window.getComputedStyle(pre).margin;
+		pre.style.margin = "0";
```
mAAdhaTTah: Not really a fan of this. Why are we doing this?

RunDevelopment: > Not really a fan of this.

Me neither.

> Why are we doing this?

Because of the new `overflow: auto;`, the margin of the `pre` will no longer extend beyond the `div`. Because of this, the `div` is larger than before because its height is then `pre.height + pre.marginTop + pre.marginBottom`. This will cause the toolbar to be too high because the top of the toolbar is aligned at the `div`.
This is why I introduced this strange margin-transfer.

(My explanation isn't very good. Just remove these lines and see for yourself. The problem is kinda obvious when you look at it.)

mAAdhaTTah: Is this a problem only for the Prism site or is this likely to effect other users?

RunDevelopment: Only the Prism site. It might affect other users but if so only a small percentage.
It's just that it doesn't give off a good image when your plugin doesn't work on your own site. Some examples are even unusable (see Copy to Clipboard).

mAAdhaTTah: Can we move that change to the website code itself & hook into Prism to fix this?

RunDevelopment: Good idea.

RunDevelopment:
Alright, I changed the implementation.
The "margin transfer" code is now in `scripts/code.js`. It'll add a hook to get every highlighted element and applies the fix to elements which have been given a toolbar.

scripts/code.js
```diff
@@ -210,4 +210,28 @@ function listPlugins(ul) {
...
+// small polyfill for Element#matches
+if (!Element.prototype.matches) {
+	Element.prototype.matches = Element.prototype.msMatchesSelector ||
+		Element.prototype.webkitMatchesSelector;
+}
+
+Prism && Prism.hooks.add('complete', function (env) {
+	var element = env.element;
+
+	setTimeout(function () {
```
mAAdhaTTah: Nit: Can we use `rAF` here?

RunDevelopment: Sure

Issue #1965, opened 2019-07-05
Plugins incompatibility: File Highlight and Copy to Clipboard


majal:
**Information:**
- Prism version: [1.16.0]
- Environment: [Browser]

Does the latest version of Prism from the [download page](https://prismjs.com/download.html) also have this issue? **Yes.**

**Description**
The plugin `Copy to Clipboard` works well if the code is sourced between the `<pre><code>` tags. On the other hand, if the code was sourced using the `File Highlight` plugin, e.g.

`<pre data-src="/path/to/asset" class="language-shell" data-download-link="" data-download-link-label="download"></pre>`

... then `Copy to Clipboard` fails. Instead of copying the code, it simply copies `Loading…`, which interestingly, is what initially appears when `File Highlight` is still loading the code from the file.

**Example**
Here is a sample webpage that shows this incompatibility: [https://www.majlovesreg.one/adding-brotli-to-a-built-nginx-instance](https://www.majlovesreg.one/adding-brotli-to-a-built-nginx-instance).

RunDevelopment:
Alright, after a few hours of debugging your website, I finally found the answer! <sup><sub>Yaaaaay</sub></sup>

__TLDR;__

Don't load the script containing Prism asynchronously until this is fixed.

__The long version__

File Highlight and Copy to Clipboard are not incompatible. In fact, the plugin page of [Copy to Clipboard](https://prismjs.com/plugins/copy-to-clipboard/) uses File Highlight to load the code which is then can then be copied. 
(You can't see the copy button because of an issue of Toolbar and the Prism page design (will be fixed soon). In the meantime, setting `right: 10em` for `div.code-toolbar > .toolbar` in the dev tools of your browser will make the button visible.)

The problem is that File Highlight has to executed __after__ the initial `Prism.highlightAll` because the FH will insert a `code` element with the text _Loading..._ inside all `pre`s with a `data-src` attribute. If FH were to be executed before `Prism.highlightAll`, all of the _Loading..._-`code`-elements would be highlighted as well (which is not intended) only to highlighted again after the code has been loaded.
One of the consequences of this double-highlighting is that all plugins are run twice on the `code` element. Most plugins check for multiple executions on the same element, so this usually isn't a problem but it is in this case because the text content of the `code` elements changed from _Loading..._ to whatever code has been loaded. This is the reason why the copied text was _Loading..._: The Copy to Clipboard plugin also detects these multiple executions on the same element.

[...]

PR #1969, opened 2019-07-11
Added more plugin tests

RunDevelopment:
This adds more plugin tests and a new way to create Prism instances inside a JSDOM VM.

It also fixes that Copy to Clipboard did not update the copied text if the text of the code element changed.
This can technically be seen as a fix for #1965.

---

Also, I just named the tests files `basic-functionality` for lack of a better name. Suggestions for a better pattern are very welcome.


plugins/copy-to-clipboard/prism-copy-to-clipboard.js
```diff
@@ -50,7 +52,7 @@
 		function registerClipboard() {
 			var clip = new ClipboardJS(linkCopy, {
 				'text': function () {
-					return env.code;
+					return element.textContent;
```
mAAdhaTTah: Can we move this out of this PR and come back to it later in #1974.

tests/helper/prism-loader.js
```diff
@@ -26,20 +52,81 @@ module.exports = {
...
+						continue;
+					}
+					loadedLanguages.add(language);
+
+					const require = languagesCatalog[language].require;
+					if (require) {
+						loadLanguages(require);
+					}
+
+					window.eval(loadLanguageSource(language));
```
mAAdhaTTah: This function can also be used with the dependency loader,  ya? We should really get that in.

tests/plugins/copy-to-clipboard/basic-functionality.js
```diff
@@ -0,0 +1,103 @@
...
+		switch (event) {
+			case 'success':
+				this.onSuccess.push(callback);
+				break;
+			case 'error':
+				this.onError.push(callback);
+				break;
+
+			default:
+				throw new Error(`Unknown event. Please fix this make-believe.`);
```
mAAdhaTTah:
```suggestion
				throw new Error(`Unknown event: ${event}. Must be one of: 'success', 'error'`);
```

tests/plugins/copy-to-clipboard/basic-functionality.js
```diff
@@ -0,0 +1,103 @@
...
+
+const { Prism, document } = dom.window;
+
+// A very simple class pretending to by ClipboardJS v2.
+// This is our "virtual clipboard".
+let ClipboardText = null;
+let ClipboardCopyWillSucceed = true;
+dom.window.ClipboardJS = class {
+	/**
+	 *
```
mAAdhaTTah: Remove line.

tests/plugins/custom-class/basic-functionality.js
```diff
@@ -0,0 +1,73 @@
...
+			code: `var a = true;`,
+			expected: `<span class="token keyword">var</span> a <span class="token operator">=</span> <span class="token boolean">true</span><span class="token punctuation">;</span>`
+		});
+	});
+
+});
+
+after(() => {
+	dom.window.close();
+});
```
mAAdhaTTah: I think these need to go in the `describe` block. Or maybe run this when you create the `dom` object? If you do that, we should probably move the logic at the top within `describe` as well.

mAAdhaTTah: Yeah, adding the `after` globally runs after everything, rather than after the suite.

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
RunDevelopment: Sure but why would this be part of #1974?

mAAdhaTTah: Because #1974 changes File Highlight and I'd like to look at fixing the issue by updating `env.code` when the file gets highlighted.

RunDevelopment:
Shit, I forgot about this.

I resolved the merge conflicts and addressed all review comments.

I fixed the usage of `after` by adding a util function called `reusablePrismDom`. The idea is that this function creates a Prism DOM and registers a cleanup function for the DOM. It's called reusable because the DOM instance is intended to be used by multiple `it` tests.

I also updated JSDom. JSDom 16 added support for `createRange` which drastically simplifies Keep Markup tests.

RunDevelopment:
I will merge this now because I want to implement update #2640 and this PR will cause conflicts.

PR #1971, opened 2019-07-11
Add string syntax $"..." to visual-basic

dovisutu:
This makes string like $"Foo bar" highlight *with* $.


RunDevelopment:
Thank you for contributing!

PR #1973, opened 2019-07-12
Core: Minor improvements

RunDevelopment:
A few minor improvements and bug fixes here and there.


components/prism-core.js
```diff
@@ -189,25 +208,17 @@ var _ = {
...
 		// Set language on the element, if not present
 		element.className = element.className.replace(lang, '').replace(/\s+/g, ' ') + ' language-' + language;
 
 		if (element.parentNode) {
 			// Set language on the parent, for styling
-			parent = element.parentNode;
+			var parent = element.parentNode;
 
-			if (/pre/i.test(parent.nodeName)) {
+			if (/^pre$/i.test(parent.nodeName)) {
```
Golmote:
```suggestion
			if (parent.nodeName.toLowerCase() === 'pre') {
```
Wouldn't something like this be more efficient?

RunDevelopment: Good question. Transforming & allocating a string + compare or traversing what is effectively a DFA.
(Inline patterns are cached, so that's not an issue.)

RunDevelopment: Ok. I benchmarked it and the regex approach is slightly slower.

PR #1974, opened 2019-07-12
 File Highlight & JSONP Highlight update

### LINKED ISSUE #1965: Plugins incompatibility: File Highlight and Copy to Clipboard

RunDevelopment:
This implements the changes to File Highlight discussed [here](https://github.com/PrismJS/prism/issues/1965#issuecomment-509700854) and also applies them to JSONP Highlight making both APIs very similar.

I took the chance to basically rewrite both plugins, so here's a list of changes:

- Both now add special attributes to the `pre` element depending on the current status. The attributes are `data-{src|jsonp}-status="{loading|loaded|failed}"` depending on the plugin and the current status. These also prevent multiple (or even concurrent) executions for the same element.
    _Edit: changed attributes_
- ~It adds a new restriction `:empty` for all `pre` elements being processed by the plugins.
  _Why?_ It doesn't make any sense for the `pre` to contain anything, so enforcing it seems like a good idea (and a breaking change). 
  The previous behavior was that FH just delete the text and JH just appended the new `code` element without doing anything.~
     _Edit:_ Removed for backward compatibility.
- Deprecated `Prism.fileHighlight` in favor of `Prism.plugins.fileHighlight.highlight` analog to `Prism.plugins.jsonphighlight.highlight`.
- Added a `timeout` config property for JSONP Highlight.
- Updated error messages of both plugins to be more consistent.
- ~FH preloads~ Both preload languages using the Autoloader if available.
- Resolves the plugin incompatibility between FH and Copy to clipboard.

---

<details>

A few things I would like to further discuss and potentially change:

__1. Preloading language.__ _Edit:_ Done.
Both plugins know which language the code is before the file is received. This can be used to preload the language definition using the Autoloader which has to fetch it anyway.
This requires exposing Autoloader's language loading API (already done in #1898). Alternatively, we could also create a fake element to highlight which will cause Autoloader to load the languages.

__2. Fixing callbacks.__ _Edit:_ The fix in #1973 is good enough
Right now, the callback function for the `pre`s [won't be executed at all](https://github.com/PrismJS/prism/blob/9d9e2ca47e9461f0b3a5b69e94c1f9a5f3d167e6/components/prism-core.js#L240). #1973 fixes this but only calls it synchronously.
To fix this I would like to add a new hook: `delegate`. The purpose of the hook is to allow `highlightElement` to hand off the responsibility of highlighting the given element to some other instance. All registered hook callbacks will then be asked whether they take the element and the first to say _yes_ has the responsibility to handle highlighting the element (including calling the `callback`).
The environment of the hook I imagine like this:

```ts
interface DelegateHookEnvironment {
    element: Element;
    async: boolean;
    callback?: (element: Element) => void;
    language: string;
    delegate: boolean;
}
```

Hook callbacks will set the `delegate` property to `true` to say _yes_.
The hooks will be run before any DOM mutation, so [here](https://github.com/PrismJS/prism/blob/9d9e2ca47e9461f0b3a5b69e94c1f9a5f3d167e6/components/prism-core.js#L202) and the code to do so could look like this:

```js
const env = { element, async, callback, language, delegate: false };
for (const f of Prism.hooks.all['delegate']) {
    f(env);
    if (env.delegate) return;
}
```

Thoughts?

__3. Asynchronous highlighting.__ _Edit:_ Maybe later.
Right now, both FH & JH ignore the `async` parameter and highlight code synchronously.
_Note:_ The proposed solution for 2) fixes this issue as well.

</details>

---

This resolves #1965.


RunDevelopment:
I added [this](https://github.com/PrismJS/prism/pull/1969/files/0891516b423e07dd2e4bec2ae344085ec421f011#r320054362) to the PR.

I think it's pretty much ready as-is aside from one detail:
I added the `:empty` pseudo-class to the selector which is a breaking change. I understand why it makes sense to enforce this but is it worth a breaking change?

RunDevelopment:
@mAAdhaTTah 
I showed this PR some love again and would like to merge this soon (because both #1813 and #1969 are blocked by this one).

As a note:
The `before-sanity-check` hook of both plugins is quite similar because I wanted them to have a somewhat similar API. Despite that, I don't think that there is much to gain from trying to abstract and moving the common elements into a separate plugin.

plugins/file-highlight/prism-file-highlight.js
```diff
@@ -1,89 +1,141 @@
...
+			// set language classes
+			setLanguageClass(code, language);
+			setLanguageClass(pre, language);
 
-			var xhr = new XMLHttpRequest();
+			// preload the language
+			var autoloader = Prism.plugins.autoloader;
+			if (autoloader) {
+				autoloader.loadLanguages(language);
+			}
```
mAAdhaTTah, line 83: I keep coming back to this. Just to check my understanding: we don't need to coordinate the autoloader with this plugin because it doesn't matter which finishes first, both will just rehighlight after they're done. Is this right?

RunDevelopment: No. Both work by 1) loading the file/data, and 2) highlighting it. 
This is just an optimization of step 2). If Autoloader was active, then it might have to download some languages in step 2). But we already know the language in step 1), so we preload them using `autoloader.loadLanguages(language)` (This doesn't do any highlighting). This means that both the language and the file/data can be downloaded in parallel. When Autoloader highlights the code in step 2), the language will already be loaded and highlighting can start right away.

RunDevelopment: TBH, Autoloader might try to highlight the code block in step 1) but it should short-circuit because there isn't any code and if it doesn't then the "highlighted code" (whatever it is) will be overwritten in step 2).

So it's really no problem at all.

RunDevelopment: Oh, I guess your understanding was correct. Just my understanding of your understanding wasn't...

But can humans truly ever understand each other? kappa

PR #1977, opened 2019-07-13
Website: FAQ update

### LINKED ISSUE #1975: Update the little script on the FAQ page supposedly showing all tokens
We really need to update [this script](https://prismjs.com/faq.html#how-do-i-know-which-tokens-i-can-style-for). It doesn't support:

- Recursive language definitions ([example](https://github.com/PrismJS/prism/blob/b9ec6fd8d1477ddea4a93e4e85c2af410f0d2ac3/components/prism-javascript.js#L74)).
- Aliases.
- Array tokens, e.g. `'comment': [ /\/\/.*/, /#.*/ ]`
- [Rest](https://github.com/PrismJS/prism/blob/b9ec6fd8d1477ddea4a93e4e85c2af410f0d2ac3/components/prism-javascript.js#L74) properties.

RunDevelopment:
This removes the little script to show all tokens of a language and replaces it with a little UI to do the task on the Prism website. I use Autoloader in a kind of hacky way to load selected languages.

Everything works as intended in Chrome, Firefox, Edge, and IE11.

### Screenshot
image

---

This resolves #1975.


faq.html
```diff
@@ -175,6 +144,111 @@ <h1>How can I use different highlighting for tokens with the same name in differ
...
+				lines.push(line);
+			}
+		}
+
+		var languageMap = new Map();
+		Object.keys(Prism.languages).filter(function (l) { return l in components.languages; }).forEach(function (l) {
+			languageMap.set(Prism.languages[l], 'Prism.languages["' + l + '"]');
+		});
+
+		var stack = new Map();
```
mAAdhaTTah: If this project supports IE11, I think the website should too.

RunDevelopment: I agree. This is why I only used the part of the Map API which is [supported by IE11](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility).

As I said:
> Everything works as intended in Chrome, Firefox, Edge, and IE11.

mAAdhaTTah: Oh, I didn't realize Map was supported in IE11.

PR #1978, opened 2019-07-14
Rewrote Autoloader using Promises

RunDevelopment:
This re-implements the Autoloader using Promises. As the number of addition/deletions suggests, the code is a lot simpler and easier to understand.

The only one holding us back right now is IE11, so #1578 when?


RunDevelopment:
We will most likely want to rewrite Autoloader completely to support the new dependency system in v2.0. Therefore this PR will have little benefit.

Feel free to reopen in case I'm wrong.

PR #1981, opened 2019-07-15
Substantial improvements to language-sas syntax highlighting

cedporter:
* Added numerous language elements
* Created more granular keyword objects, for finer grained styling if desired
* Expanded keyword regex list
* Special identification of PROC SQL and the SQL that follows
* Test coverage for all added language elements


components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
+	'macrodeclaration': {
+		pattern: /((^|\n)(%macro)((\s)+([\w\s\S]+?))+);/im,
+		inside: {
+			keyword: /%macro/im,
+		}
+	},
+	'macroend': {
+		pattern: /((^|\n)(%mend)((\s)+([\w\s\S]+?))+);/im,
+		inside: {
+			keyword: /%mend/im,
```
RunDevelopment:
```suggestion
			keyword: /%mend/i,
```

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
+		}
+	},
+	'macrodeclaration': {
+		pattern: /((^|\n)(%macro)((\s)+([\w\s\S]+?))+);/im,
+		inside: {
+			keyword: /%macro/im,
+		}
+	},
+	'macroend': {
+		pattern: /((^|\n)(%mend)((\s)+([\w\s\S]+?))+);/im,
```
RunDevelopment: Multiple issues:

1. `(^|\n)` because of the `m` flag `^` is equivalent to `(?<!.)`, or roughly `(?<=[\r\n])`. So I don't quite understand your intention here.
2. `((\s)+([\w\s\S]+?))+` with a few braces less `(\s+[\w\s\S]+?)+`: Because of this part your pattern will show exponential backtracking. 
  You can see for yourself using tools like [regexr](https://regexr.com/) using the pattern `(\s+[\w\s\S]+?)+;`. The red `ERROR` means that your pattern took longer than 250ms to execute.
  Don't worry, this is easy to fix because `(\s+[\w\s\S]+?)+` is equivalent to `[\s\S]+`, but I guess you really meant `[\s\S]+?`.
3. `[\w\s\S]` === `[\s\S]`

```suggestion
		pattern: /^%mend[\s\S]+?;/im,
```

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
+			'parentheses': {
+				pattern: /(\()([A-Z]+)(\))/i,
+			},
+			'arg': {
+				pattern: /([A-Z]+)/im,
+				alias: 'keyword'
+			},
+			'number': /\b(?:[\da-f]+x(?!\w+)|\d+(?:\.\d+)?(?:e[+-]?\d+)?)/,
+			'string': {
+				pattern: /(["'])(?:\1\1|(?!\1)[\s\S])*\1/,
```
RunDevelopment: This string pattern is used multiple times. Please make a variable for it.

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
+			},
+			'equals': {
+				pattern: /=/,
+				alias: 'operator'
+			},
+			'parentheses': {
+				pattern: /(\()([A-Z]+)(\))/i,
+			},
+			'arg': {
+				pattern: /([A-Z]+)/im,
```
RunDevelopment:
```suggestion
				pattern: /[A-Z]+/i,
```

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
 			lookbehind: true
 		},
 		/\/\*[\s\S]+?\*\//
 	],
+	'options': {
+		pattern: /(^options)(?:((?:\s+?)(?:\r?\n|\r)*[\'\-|\"\/\\<>\*\w+=\(\)]*)*;)/im,
+		inside: {
+			'options': {
+				alias: 'keyword',
+				pattern: /(^options)/i,
```
RunDevelopment:
```suggestion
				pattern: /^options/i,
```

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
 	'comment': [
 		{
-			pattern: /(^\s*|;\s*)\*.*;/m,
+			pattern: /(^\s*|;\s*)\*(.|[\r\n])*?;/m,
 			lookbehind: true
 		},
 		/\/\*[\s\S]+?\*\//
 	],
+	'options': {
+		pattern: /(^options)(?:((?:\s+?)(?:\r?\n|\r)*[\'\-|\"\/\\<>\*\w+=\(\)]*)*;)/im,
```
RunDevelopment:
```suggestion
		pattern: /^options[-'"|/\\<>*+=()\w\s]*/im,
```

I'm 90% sure that my suggested change and the original regex are equivalent but please make sure that all your tests pass.

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
+				{
+					pattern: /(^\s*|;\s*)\*.*;/m,
+					lookbehind: true
+				},
+				/\/\*[\s\S]+?\*\//
+			],
+			'number': /\b(?:[\da-f]+x(?!\w+)|\d+(?:\.\d+)?(?:e[+-]?\d+)?)/i,
+		}
+	},
 	'comment': [
```
RunDevelopment: Please also make a variable for this one.

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
 		{
-			pattern: /(^\s*|;\s*)\*.*;/m,
+			pattern: /(^\s*|;\s*)\*(.|[\r\n])*?;/m,
```
RunDevelopment: `(.|[\r\n])*?` Isn't this a single line comment?

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
+			},
+			'step': {
+				pattern: /(?:proc\s\w+|quit|run|data(?!\=))\b/i,
+				alias: 'keyword'
+			},
+			rest: Prism.languages.sql
+		}
+	},
+	'macrodeclaration': {
+		pattern: /((^|\n)(%macro)((\s)+([\w\s\S]+?))+);/im,
```
RunDevelopment: Same as what I said about the below `macroend`.

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
+					alias: 'number',
+				}
+		}
+	},
+	'altformat': {
+		pattern: /(\b)(?:((format)|(put))(?:\s)+)([\w\'])+((\s)+([\$\.\w]+))+;/im,
+		inside: {
+			'keyword': /(format|put)/im,
+			'format': {
+					pattern: /(\w|\$)+\.\d?/im,
```
RunDevelopment:
```suggestion
					pattern: /(\w|\$)+\.\d?/,
```

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
+			'format': {
+					pattern: /(\w|\$\d)+\.\d?/im,
+					alias: 'number',
+				}
+		}
+	},
+	'altformat': {
+		pattern: /(\b)(?:((format)|(put))(?:\s)+)([\w\'])+((\s)+([\$\.\w]+))+;/im,
+		inside: {
+			'keyword': /(format|put)/im,
```
RunDevelopment:
```suggestion
			'keyword': /format|put/i,
```

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
+						alias: 'operator'
+					},
+			'format': {
+					pattern: /(\w|\$\d)+\.\d?/im,
+					alias: 'number',
+				}
+		}
+	},
+	'altformat': {
+		pattern: /(\b)(?:((format)|(put))(?:\s)+)([\w\'])+((\s)+([\$\.\w]+))+;/im,
```
RunDevelopment:
```suggestion
		pattern: /\b(?:(?:format|put)\s+)[\w']+(\s+[$.\w]+)+;/i,
```

components/prism-sas.js
```diff
@@ -26,9 +145,21 @@ Prism.languages.sas = {
...
+	},
+	// In SAS Studio syntax highlighting, these operators are styled like keywords
+	'specialOperator':{
+		pattern: /\b(?:eq|ne|gt|lt|ge|le|in|not)\b/i,
+		alias: 'operator'
+	},
 	// Decimal (1.2e23), hexadecimal (0c1x)
-	'number': /\b(?:[\da-f]+x|\d+(?:\.\d+)?(?:e[+-]?\d+)?)/i,
-	'operator': /\*\*?|\|\|?|!!?|¦¦?|<[>=]?|>[<=]?|[-+\/=&]|[~¬^]=?|\b(?:eq|ne|gt|lt|ge|le|in|not)\b/i,
+	'number': /\b(?:[\da-f]+x(?!\w+)|\d+(?:\.\d+)?(?:e[+-]?\d+)?)/i,
```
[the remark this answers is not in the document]
cedporter: I was testing and revising these patterns based on SAS code, but I can't find a good, unique number at the moment. For now, I'll remove the (?!\w+) and wait to find better examples and improve it then.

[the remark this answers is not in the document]
RunDevelopment: Found [some doc](http://support.sas.com/documentation/cdl/en/lrcon/62955/HTML/default/viewer.htm#a000780334.htm) on this topic.

> A numeric constant that is expressed as a hexadecimal value starts with a numeric digit (usually 0), can be followed by more hexadecimal characters, and ends with the letter X.

So the hexadecimal part has to look like this `\d[\da-f]*x`. I'll update my suggestion.

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
+			'format': {
+					pattern: /(\w|\$)+\.\d?/im,
+					alias: 'number',
+				},
+				'punctuation': /;/
+		}
+	},
 	'datetime': {
 		// '1jan2013'd, '9:25:19pm't, '18jan2003:9:27:05am'dt
 		pattern: /'[^']+'(?:dt?|t)\b/i,
```
RunDevelopment: Datetimes can also be wrapped in double-quotes.
I know this isn't one of your changes but we might as well fix it while we're at it.

[Doc.](http://support.sas.com/documentation/cdl/en/lrcon/62955/HTML/default/viewer.htm#a000780334.htm)

components/prism-sas.js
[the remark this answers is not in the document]
RunDevelopment: Do we also want to add support for hexadecimal (e.g. `'534153'x`) and binary (e.g. `'10000000'b`) literals?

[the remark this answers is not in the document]
cedporter: Good idea. I'm going to replace datetime with a numericconstants pattern
`/'\d+'((dt)|[bxt])\b/i`. This will cover time, date, bit, hex.

components/prism-sas.js
cedporter: Once you've started a comment in SAS, it can run multiple lines, regardless of line breaks.

For instance:
`* This is
a long comment;`

That would be interpreted as a single comment.

components/prism-sas.js
cedporter: Yep. that looks right. Thanks.

components/prism-sas.js
cedporter: Same applies to the pattern above it.

components/prism-sas.js
```diff
@@ -10,13 +10,132 @@ Prism.languages.sas = {
...
+			keyword: /%mend/im,
+		}
+	},
+	/*%_zscore(headcir, _lhc, _mhc, _shc, headcz, headcpct, _Fheadcz); */
+	'macro': {
+		pattern: /%_\w+(?=\()/,
+		alias: 'keyword'
+	},
+	'input': {
+		pattern: /(\b)+(?:input(?!\=))\s+(?:\w|\$|\&|\s|\-|\.|\/|\*)+;/im,
```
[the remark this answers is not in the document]
cedporter: Thanks. Yep.

components/prism-sas.js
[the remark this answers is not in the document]
RunDevelopment: Datetime is more complex than just that. I was thinking of making something like this after `datetime`.

```js
'numeric-constant' {
    pattern: /(["'])(?:\1\1|\\.|(!?\1)[^\\])\1[bx]/,
    alias: 'number'
}
```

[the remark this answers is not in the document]
cedporter: Yep, I realized that as soon as I started testing. I'll leave datetime as is. I can use your pattern for numeric-constant.

components/prism-sas.js
cedporter: Will revise.

components/prism-sas.js
cedporter: Will revise.

components/prism-sas.js
cedporter: Will revise.

components/prism-sas.js
cedporter: Will revise.

components/prism-sas.js
cedporter: Will do.

components/prism-sas.js
RunDevelopment: Ahhh, ok. But please use `[\s\S]` instead of `(.|[\r\n])`.

components/prism-sas.js
```diff
@@ -1,3 +1,16 @@
+(function (Prism) {
+
+	numberpattern = /\b(?:\d[\da-f]*x|\d+(?:\.\d+)?(?:e[+-]?\d+)?)\b/i;
```
RunDevelopment: Please make it `var numberPattern = ...`.

We are using the IEFE because we don't want global variables and camel case greatly improves readability.

components/prism-sas.js
```diff
@@ -10,25 +23,152 @@ Prism.languages.sas = {
...
+	'step': {
+		pattern: /(^|\n)(?:proc\s\w+|quit|run|data(?!\=))\b/i,
+		alias: 'keyword'
+	},
+	'keyword': {
+		pattern: /((^|[\s])=?)(?:action|after|analysis|and|array|barchart|barwidth|begingraph|by|cas|cbarline|cfill|close|column|compute(d)?|contains|data(?=\=)|define|document|do\s+over|do|dol|drop|dul|end|entryTitle|else|endcomp|fill(attrs)?|filename|group(by)?|headline|headskip|histogram|if|infile|keep|label|layout|legendlabel|length|libname|merge|midpoints|name|noobs|nowd|ods|or|out(put)?|overlay|ranexp|rannor|rbreak|retain|set|session|sessref|statgraph|sum|summarize|table|temp|then\sdo|then|title|to|var|where|xaxisopts|yaxisopts|y2axisopts)\b/i,
+		lookbehind: true,
+	},
+	// In SAS Studio syntax highlighting, these operators are styled like keywords
+	'specialOperator':{
```
RunDevelopment: Please use snake-case for token names.
```suggestion
	'special-operator': {
```

RunDevelopment: Just an idea but maybe `operator-keyword` is a better name of this token?

components/prism-sas.js
```diff
@@ -10,25 +23,152 @@ Prism.languages.sas = {
...
 		}
 	},
+	'procSQL': {
+		pattern: /(^|\n)(?:proc\s+sql(?:\s+[\w|=]+)?);([\s\S]+(?:\r?\n|\r)(?=((?:proc\s\w+|quit|run|data(?!\=));))((?:proc\s\w+|quit|run|data(?!\=));)|(?:\r?\n|\r)[\s\S]+;)/im,
+		inside: {
+			'fullStep': {
+				pattern: /(^|\n)(?:proc\s+sql(?:\s+[\w|=]+)?);/i,
+				inside: {
+					'step': {
+						pattern: /(?:proc\s\w+|quit|run|data(?!\=))\b/i,
```
RunDevelopment: Aren't __all__ `proc\s\w+` supposed to be `proc\s+\w+`?

components/prism-sas.js
```diff
@@ -10,25 +23,152 @@ Prism.languages.sas = {
...
+				}
+			},
+			'step': {
+				pattern: /(?:proc\s\w+|quit|run|data(?!\=))\b/i,
+				alias: 'keyword'
+			},
+			rest: Prism.languages.sql
+		}
+	},
+	'macrodeclaration': {
```
RunDevelopment: Please use snake-case here as well.
```suggestion
	'macro-declaration': {
```

Same for `macroend`.

components/prism-sas.js
```diff
@@ -10,25 +23,152 @@ Prism.languages.sas = {
...
+		pattern: /((^|\n)(%macro)((\s)+([\s\S]+?))+);/im,
+		inside: {
+			keyword: /%macro/im,
+		}
+	},
+	'macroend': {
+		pattern: /^%mend[\s\S]+?;/i,
```
RunDevelopment: You accidentally delete the `m` flag.
```suggestion
		pattern: /^%mend[\s\S]+?;/im,
```

components/prism-sas.js
```diff
@@ -10,25 +23,152 @@ Prism.languages.sas = {
...
+			},
+			'step': {
+				pattern: /(?:proc\s\w+|quit|run|data(?!\=))\b/i,
+				alias: 'keyword'
+			},
+			rest: Prism.languages.sql
+		}
+	},
+	'macrodeclaration': {
+		pattern: /((^|\n)(%macro)((\s)+([\s\S]+?))+);/im,
```
RunDevelopment: Same problem that `macroend` used to have.
```suggestion
		pattern: /^%macro[\s\S]+?;/im,
```

components/prism-sas.js
```diff
@@ -10,25 +23,152 @@ Prism.languages.sas = {
...
+		inside: {
+			keyword: /%macro/im,
```
RunDevelopment: The `m` flag is unnecessary.

```suggestion
			keyword: /%macro/i,
```

components/prism-sas.js
cedporter: Will do.

components/prism-sas.js
cedporter: Thanks.

components/prism-sas.js
cedporter: I actually intended to remove that. the pattern is generally 

> %mend macroName;

I'm not sure that there's ever a line break.

components/prism-sas.js
cedporter: For comment? Done.

components/prism-sas.js
cedporter: Done.

components/prism-sas.js
cedporter: Thanks, I incorporated this, but readded the semicolon at the end of the expression. Tests pass.

tests/languages/sas/number_feature.test
```diff
@@ -13,9 +13,9 @@ BadFacex
 	["number", "3.2e10"],
 	["number", "0.4e-8"],
 	["number", "1.4E+2"],
-	["number", "BadFacex"]
+	"\r\nBadFacex"
```
RunDevelopment: Like this, the tests may pass but it doesn't really test anything. Please add one or two __real__ hexadecimal numbers.

components/prism-sas.js
RunDevelopment: `^` with the `m` flag means start of line. But without the `m` flag it means start of string.
The `m` flag is essential.

Here is a macro highlighted with the current pattern without the `m` flag.
image
[Source](http://support.sas.com/documentation/cdl/en/mcrolref/61885/HTML/default/viewer.htm#macro-stmt.htm).

As you can see, it's highlighted incorrectly.

Please make a test for macros in general (you can use the above code if you want) and remove the ones for `macro-declaration` and `macro-end`.

Also as a bit of general advice: Try to get your hands on as much SAS code as possible (>1k lines) and see how it gets highlighted. This will make problems like the missing punctuation visible.

RunDevelopment:
@cedporter Could you please resolve the conflicts with master?

This will allow our CI to check your code as well.

Also, there are still comments left to address.

components/prism-sas.js
cedporter: Yes, thanks. I am testing against SAS code, and directly comparing the results of Prism.JS highlighting vs. SAS Studio.

components/prism-sas.js
cedporter: Will do.

components/prism-sas.js
cedporter: Will change to operator-keyword

components/prism-sas.js
cedporter: Accepted.

components/prism-sas.js
cedporter: Will do.

components/prism-sas.js
cedporter: Yep. Will fix.

RunDevelopment:
@cedporter I'm sorry but could you please only mark comments as resolved after you pushed the changes which resolve them? From my perspective, comments are marked as resolved without actually being resolved which is a little confusing.

For example, I don't know whether you accidentally resolved [this one](https://github.com/PrismJS/prism/pull/1981#discussion_r306037157) or whether the commit just isn't pushed yet.

Thank you.

components/prism-sas.js
```diff
@@ -10,25 +36,137 @@ Prism.languages.sas = {
...
+					alias: 'number',
+				},
+				'punctuation': /;/
+		}
+	},
+	'numeric-constant': numericConstant,
 	'datetime': {
 		// '1jan2013'd, '9:25:19pm't, '18jan2003:9:27:05am'dt
-		pattern: /'[^']+'(?:dt?|t)\b/i,
+		pattern: /(["'])(?:.+)\1(?:dt?|t)/,
```
RunDevelopment: `["']` ⊂ `.`. As a consequense, your pattern will match two dates after another as one date.
```suggestion
		pattern: /(["'])(?:(?!\1).)+\1(?:dt?|t)/,
```

RunDevelopment: Also, why did you remove the `i` flag?

components/prism-sas.js
RunDevelopment: You forgot about this one.

components/prism-sas.js
```diff
@@ -1,3 +1,29 @@
...
+	};
+
+	step = {
+		pattern: /(?:proc\s+\w+|quit|run|data(?!\=))\b/i,
+		alias: 'keyword'
+	};
+
+	comment = [
+		{
+			pattern: /(^\s*|;\s*)\*(.|[\s\S]])*?;/m,
```
RunDevelopment: Typo. There's one `]` too much. But `(.|[\s\S])*?;` itself is an even bigger problem because it fails in exponential time.

I encourage you to try the pattern `(.|[\s\S])*?;` on a site like [regexr.com](https://regexr.com). You'll see that for a string which doesn't contain any `;`s, this pattern fails in O(2^n) (n being the length of the input string). After only 12 characters, this pattern takes >100ms to fail.

I won't go into the finer details but when you have quantified groups, make sure that the alternatives of the group are disjoint (meaning that there is no word which can be matched by more than one alternative of the group). Otherwise, the pattern will fail in O(2^n) on certain input strings.
This is also known under the name [ReDoS](https://en.wikipedia.org/wiki/ReDoS) and is what happened to [Cloudflare](https://blog.securityinnovation.com/cloudflare-dos-blackout) not too long ago although there the cause for the exponential behavior was slightly different.

In this case, `.` is a subset of `[\s\S]` which is both the problem and the reason why it's easy to fix. We can just make it `[\s\S]*?;` and it will still behave the same but it will fail in linear time.
But we can go even simpler because we don't even need the lazy matching.

```suggestion
			pattern: /(^\s*|;\s*)\*[^;]*;/m,
```

components/prism-sas.js
cedporter: "dt" and "t" are case sensitive in SAS. I tested in SAS Studio.

RunDevelopment: Ok.

components/prism-sas.js
cedporter: Thanks, I was not familiar with this site. WIll use it now instead of Regex101.

components/prism-sas.js
cedporter: Committed now.

tests/languages/sas/format.test
```diff
@@ -0,0 +1,35 @@
...
+----------------------------------------------------
+
+[
+  ["altformat",
+    [
+      ["keyword", "put"],
+      " x ",
+      ["format", "$hex16."],
+      ["punctuation", ";"]
+    ]
```
RunDevelopment: Please fix the formatting of your tests. The JSON should also be indented using tabs.

Btw. You can have the JSON generated for you. I a test fails, it will show you the full token stream. 
With the command `npm run test:languages -- --pretty` this token stream will be formatted. You can then just copy and past the token stream and done. Well, almost done; you still have to adjust the indentation from spaces to tabs.

components/prism-sas.js
```diff
@@ -1,3 +1,29 @@
+(function (Prism) {
+
+	numberPattern = /\b(?:\d[\da-f]*x|\d+(?:\.\d+)?(?:e[+-]?\d+)?)\b/i;
```
RunDevelopment:
```suggestion
	var numberPattern = /\b(?:\d[\da-f]*x|\d+(?:\.\d+)?(?:e[+-]?\d+)?)\b/i;
```

Same for the other variables.

components/prism-sas.js
```diff
@@ -1,3 +1,29 @@
...
+		},
+		/\/\*[\s\S]+?\*\//
+	];
+
+	string = {
+		pattern: /(["'])(?:\1\1|(?!\1)[\s\S])*\1/,
+		greedy: true
+	};
+
 Prism.languages.sas = {
```
RunDevelopment: Please indent everything. Not just the new variables.

tests/languages/sas/macrodefinition_feature.test
```diff
@@ -0,0 +1,44 @@
...
+  ["operator", "&"],
+  "sum",
+  ["punctuation", ";"],
+  ["step", "run"],
+  ["punctuation", ";"],
+  [
+    "macro-end",
+    [
+      ["keyword", "%mend"],
+      " prnt;"
```
RunDevelopment: The `;` is supposed to be punctuation, right?

components/prism-sas.js
```diff
@@ -10,25 +36,137 @@ Prism.languages.sas = {
...
+			'options': {
+				alias: 'keyword',
+				pattern: /^options/i,
+			},
+			'equals': {
+				pattern: /=/,
+				alias: 'operator'
+			},
+			'parentheses': {
+				pattern: /(\()([A-Z]+)(\))/i,
```
RunDevelopment: Please add an example for this being used in the `options` test,

tests/languages/sas/options.test
```diff
@@ -0,0 +1,34 @@
...
+      ";"
+    ]
+  ],
+  [
+    "options",
+    [
+      ["options", "options"],
+      ["arg", "validmemname"],
+      ["equals", "="],
+      ["arg", "extend"],
```
RunDevelopment: Isn't it incorrect that `extend` is matched as `arg`?

tests/languages/sas/macrodefinition_feature.test
cedporter: Yes, I'll add a punctuation rule to macro-end and macro-declaration.

RunDevelopment: There are quite a few tokens which have the same problem. Instead of including the `;` in the token, we could just use a lookahead to not include it and use the global punctuation to match the `;`.

tests/languages/sas/options.test
cedporter: It is. I've added an argValue pattern.

components/prism-sas.js
cedporter: Removed. Replaced with punctuation pattern.

tests/languages/sas/macrodefinition_feature.test
cedporter: I've addressed this.

components/prism-sas.js
```diff
@@ -1,34 +1,173 @@
...
+
+	Prism.languages.sas = {
+		'datalines': {
+			pattern: /^(\s*)(?:(?:data)?lines|cards);[\s\S]+?^;/im,
+			lookbehind: true,
+			alias: 'string',
+			inside: {
+				'keyword': {
+					pattern: /^(?:(?:data)?lines|cards)/i,
+					lookbehind: true
```
RunDevelopment: Alright, the reason the build failed is that I forgot to remove this lookbehind. Sorry about that.

components/prism-sas.js
```diff
@@ -1,34 +1,173 @@
...
+				},
+				'punctuation': /;/
+			}
+		},
+		'proc-sql': {
+			pattern: /^proc\s+sql(?:\s+[\w|=]+)?;(?:[\s\S]+?^(?:proc\s+\w+|quit|run|data);|(?:\r?\n|\r)[\s\S]+;)/im,
```
RunDevelopment: Could you please this pattern to me? I don't get this part:

```
(?:
    [\s\S]+?^(?:proc\s+\w+|quit|run|data);
|
    (?:\r?\n|\r)[\s\S]+;
)
```

Does this mean that a `proc sql` statement can be terminated by any other `proc *` statement?
From the doc, I thought that a `proc sql` statement has to be ended with a `quit;`.

cedporter: Yep. That's what it means. From the [doc](https://go.documentation.sas.com/?cdcId=pgmsascdc&cdcVersion=9.4_3.4&docsetId=sqlproc&docsetTarget=n0w2pkrm208upln11i9r4ogwyvow.htm&locale=en)
> When a PROC SQL statement is executed, PROC SQL continues to run until a QUIT statement, a DATA step, or another SAS procedure is executed.

components/prism-sas.js
cedporter: Yep. I just fixed this in my latest commit.

components/prism-sas.js
RunDevelopment: And what's the second part then?

cedporter: If it’s a code fragment, and it’s not ended, the regex should still recognize the proc.

RunDevelopment: Could you give an example of a code fragment?

---

Code fragments aside, I usually put embedded languages (like SQL) in their token to not mix languages. Since we already a `step` token in SAS, couldn't we match SQL Code like this?

```js
'sql': {
	// TODO: Support for code fragments
	pattern: /(^proc\s+sql(?:\s+[\w|=]+)?;)[\s\S]+?(?=^(?:proc\s+\w+|quit|run|data);)/im,
	lookbehind: true,
	alias: 'language-sql',
	inside: Prism.languages.sql
}
```

(This requires minor adjustments to `step` to also include `proc\s+sql(?:\s+[\w|=]+)?`.)

cedporter: `proc sql outobs=10;
   title 'Proclib.Payroll';
   title2 'First 10 Rows Only';
   select * from proclib.payroll;
   title;`

[Doc](https://go.documentation.sas.com/?cdcId=pgmsascdc&cdcVersion=9.4_3.4&docsetId=sqlproc&docsetTarget=n1lhnlzhrmrqggn1rz570u89oq2m.htm&locale=en). And actually, I mispoke. SAS does not require that you end your program with "run" or "quit". The quoted code would execute just fine.

cedporter: And sure, we can match SQL that way if it's preferred.

RunDevelopment: > The quoted code would execute just fine.

Which quoted code?

RunDevelopment: Question: What is `title` and `title2`. It's not part of regular SQL but right now we use SQL to highlight it.

We should only highlight SQL statements with SQL, so is it possible to exclude any SAS constructs? Maybe this is an impossible task when looking at code like this.

```sas
proc sql;
   title  'Stores Table';
   title2 'Coordinates of Stores';
   select * from stores;
   title  'Houses Table';
   title2 'Coordinates of Houses';
   select * from houses;
title;
```

([Source](https://go.documentation.sas.com/?cdcId=pgmsascdc&cdcVersion=9.4_3.4&docsetId=sqlproc&docsetTarget=n02w9rre6jk3n5n10zot9n4uqrzj.htm&locale=en))

cedporter: > Which quoted code?

The code you just quoted would run. See also:

```
proc sql outobs=10;
   title 'Proclib.Payroll';
   title2 'First 10 Rows Only';
   select * from proclib.payroll;
   title;
```
You do not need to end a proc statement with run or quit. SAS will operate fine just encountering EOF.

RunDevelopment: Ok. I thought you meant SAS code. Sorry about that.
I added the EOF condition to my [proposed pattern](https://github.com/PrismJS/prism/pull/1981#discussion_r307689351).

cedporter: title and title2 are [global statements](https://go.documentation.sas.com/?cdcId=pgmsascdc&cdcVersion=9.4_3.4&docsetId=lestmtsglobal&docsetTarget=n12y6ohua6d9nxn1q1v8kr63t0e5.htm&locale=en).  

Would it be acceptable to declare Prism.languages.sql as a variable within the IIFE and add a global-statements object to it? Then, in the sql object within SAS, use sasSQL rather than Prism.languages.sql?

```
	var globalStatements = {
		pattern: /((?:^|[\s])=?)(?:catname|checkpoint execute_always|dm|endsas|filename|footnote|%include|libname|%list|lock|missing|options|page|resetline|%run|sasfile|skip|sysecho|title\d?)\b/i,
		lookbehind: true,
		alias: 'keyword'
	};

	var sasSQL = Prism.languages.sql;


	sasSQL['global-statements'] = globalStatements;
```

RunDevelopment: Here you modify the SQL language itself.

How about this approach: If my understanding is correct then only a [handful of SQL statements](https://go.documentation.sas.com/?cdcId=pgmsascdc&cdcVersion=9.4_3.4&docsetId=sqlproc&docsetTarget=n1oihmdy7om5rmn1aorxui3kxizl.htm&locale=en) can be used. We can use this because we now that every line which starts which looks like this `/^[ \t]*(?:select|alter table|...)(?:<string>|[^;"'])+;/im`.

This should be able to find us all SQL statements in a `proc sql` statement. Everything has to be regular SAS or more specific global statements.

So like this, I suppose:

```js
'proc-sql': {
	pattern: /^proc\s+sql(?:\s+[\w|=]+)?;[\s\S]+?(?=^(?:proc\s+\w+|quit|run|data);|(?![\s\S]))/im,
	inside: {
		'sql': {
			pattern: RegExp(/^[ \t]*(?:select|alter table|...)(?:<str>|[^;"'])+;/.source.replace(/<str>/g, stringPattern), 'im'),
			alias: 'language-sql',
			inside: Prism.languages.sql
		},
		...regular_SAS
	}
}
```

cedporter: The documentation might make it seem as though those statements are the only that work, but you can write[ fairly complex SQL queries](https://go.documentation.sas.com/?cdcId=pgmsascdc&cdcVersion=9.4_3.4&docsetId=sqlproc&docsetTarget=n0340vzqamcsopn1oo4lvwkbkxuq.htm&locale=en#n0340vzqamcsopn1oo4lvwkbkxuq) within proc sql:

```
libname proclib 'SAS-library';
proc sql;
   create view proclib.jobs(pw=red) as
   select Jobcode,
          count(jobcode) as number label='Number',
          avg(int((today()-birth)/365.25)) as avgage
             format=2. label='Average Age',
          avg(salary) as avgsal
             format=dollar8. label='Average Salary'
   from payroll
   group by jobcode
   having avgage ge 30;
   title1 'Current Summary Information for Each Job Category';
   title2 'Average Age Greater Than or Equal to 30';
   select * from proclib.jobs(pw=red);
title;
```

I've been working on this and the other two proc that allow programmers to execute code from another programming language ([proc groovy](https://go.documentation.sas.com/?cdcId=pgmsascdc&cdcVersion=9.4_3.4&docsetId=proc&docsetTarget=n12njw2j3tuptnn1bmmr0cl857d2.htm&locale=en) and [proc lua](https://go.documentation.sas.com/?cdcId=pgmsascdc&cdcVersion=9.4_3.4&docsetId=proc&docsetTarget=n1w8nl91tml15dn1mw9p5l8oj6hy.htm&locale=en)) using a similar approach to what I described above. You're right, however, I was neglecting the fact I was pointing to and adding to the SQL language support itself. With a bit of experimentation, I hit on:

`var sasSQL = Object.assign({}, Prism.languages.sql);`

That will copy the original language, rather than simply pointing to it. Would that be acceptable?

RunDevelopment: [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) is ES6, so we can't use it. But you can use `Prism.util.clone` which will make a deep copy of a given object.

However, I still believe that my approach is more desirable. Also, your example of a fairly complex SQL query is matched correctly by my pattern.

---

Groovy and Lua seem to be quite easy to add but let's please do that after SQL.

cedporter: This is what I've got right now based on your suggestion above, but I'm definitely not getting full highlighting
```
		'proc-sql': {
			pattern: /^proc\s+(fed)?sql(?:\s+[\w|=]+)?;[\s\S]+?(?=^(?:proc\s+\w+|quit|run|data);|(?![\s\S]))/im,
			inside: {
				'sql': {
					pattern: RegExp(/^[ \t]*(?:select(?:\s+(distinct|unique))?|alter\s+table|(create|describe|drop)\s+(index|table|view)|insert\s+into|update)(?:<str>|[^;"'])+;/.source.replace(/<str>/g, stringPattern), 'im'),
					alias: 'language-sql',
					inside: Prism.languages.sql
				},
				'global-statements': globalStatements,
				'sqlStatements': {
					pattern: /((?:^|[\s]))(?:disconnect\s+from|exec(ute)?|reset|validate)\b/i,
					lookbehind: true,
					alias: 'keyword'
				}
			}
		},
```
image

I suspect it's due to the fact that "create view" does not end with a semicolon for several lines?

RunDevelopment: Don't use capturing groups if not necessary.

The inserted string pattern references the first capturing group which in your case is `(distinct|unique)`. Making all groups non-capturing fixes the issue.
Also: You can remove `(?:\s+(distinct|unique))?` because it can also be matched by the following `(?:<str>|[^;"'])+`.

All in all, this gives us:

```js
'sql': {
	pattern: RegExp(/^[ \t]*(?:select|alter\s+table|(?:create|describe|drop)\s+(?:index|table|view)|create\s+unique\s+index|insert\s+into|update)(?:<str>|[^;"'])+;/.source.replace(/<str>/g, stringPattern), 'im'),
	alias: 'language-sql',
	inside: Prism.languages.sql
},
```

(I also added support for `CREATE UNIQUE INDEX ...`)

(Also, if you show an example, please provide a link to the example or the code itself because I have to copy the whole thing by hand if you don't...)

cedporter: Sorry about the screenshot of the source. In any case, I've committed changes, and I included that code in the sql test.

gulpfile.js/index.js
```diff
@@ -190,8 +190,8 @@ const plugins = series(languagePlugins, minifyPlugins);
 
 module.exports = {
 	watch: watchComponentsAndPlugins,
-	default: parallel(components, plugins, componentsJsonToJs, build),
 	premerge,
 	linkify,
-	changes
+	changes,
```
RunDevelopment: What happened here?

cedporter: For some reason, after a gulp update, I gulp would not work for me without this modification.

RunDevelopment: What? Changing the order of exported items shouldn't have any effect.
What error message did you get? 
Is this also the reason you down-graded Mocha?

Changes to `gulpfile.js/*`, `package.json`, and `package-lock.json` should not be part of a language PR.

cedporter: I'm honestly not sure what happened. I just replaced the gulp.js files with what's in master, and gulp ran fine. I'll commit those to revert.

cedporter: I'm not sure how to fix this travis build failure. Any suggestions?

RunDevelopment: The reason the build failed is that you forgot to revert the `package-lock.json` and now they are out of sync which causes `npm ci` to fail.

> cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.

components.json
```diff
@@ -816,6 +816,7 @@
 		},
 		"sas": {
 			"title": "SAS",
+			"require": ["sql", "lua", "groovy"],
```
RunDevelopment: I know that you plan to add support for Lua and Groovy but maybe this should be its own PR. I really want to merge this soon.

gulpfile.js/index.js
RunDevelopment: > I'm honestly not sure what happened. I just replaced the gulp.js files with what's in master, and gulp ran fine.

Good that it works now.

components/prism-sas.js
```diff
@@ -1,34 +1,184 @@
...
+			inside: {
+				'keyword': {
+					pattern: /^(?:(?:data)?lines|cards)/i
+				},
+				'punctuation': /;/
+			}
+		},
+
+		'proc-sql': {
+			pattern: /(^proc\s+(fed)?sql(?:\s+[\w|=]+)?;)[\s\S]+?(?=^(?:proc\s+\w+|quit|run|data);|(?![\s\S]))/im,
```
RunDevelopment: Unnecessary capturing group.

```suggestion
			pattern: /(^proc\s+(?:fed)?sql(?:\s+[\w|=]+)?;)[\s\S]+?(?=^(?:proc\s+\w+|quit|run|data);|(?![\s\S]))/im,
```

components/prism-sas.js
```diff
@@ -1,34 +1,184 @@
...
+					pattern: /((?:^|[\s]))(?:disconnect\s+from|exec(ute)?|begin|commit|rollback|reset|validate)\b/i,
+					lookbehind: true,
+					alias: 'keyword'
+				},
+				'number': number,
+				'numeric-constant': numericConstant,
+				'punctuation': punctuation,
+				'string': string
+			},
+			lookbehind: true
```
RunDevelopment: Please move this property before `inside`. It's really hard to see here.

I usually use the order `pattern`, `lookbehind`, `greedy`, `alias`, `inside`.

components/prism-sas.js
```diff
@@ -1,34 +1,184 @@
...
+		'proc-sql': {
+			pattern: /(^proc\s+(fed)?sql(?:\s+[\w|=]+)?;)[\s\S]+?(?=^(?:proc\s+\w+|quit|run|data);|(?![\s\S]))/im,
+			inside: {
+				'sql': {
+					pattern: RegExp(/^[ \t]*(?:select|alter\s+table|(?:create|describe|drop)(?:\s+)(?:index|table(?:\s+constraints)?|view)|create\s+unique\s+index|insert\s+into|update)(?:<str>|[^;"'])+;/.source.replace(/<str>/g, stringPattern), 'im'),
+					alias: 'language-sql',
+					inside: Prism.languages.sql
+				},
+				'global-statements': globalStatements,
+				'sqlStatements': {
```
RunDevelopment: Please use snake-case
```suggestion
				'sql-statements': {
```

components/prism-sas.js
```diff
@@ -1,34 +1,184 @@
...
+					pattern: /((?:^|[\s]))(?:disconnect\s+from|exec(ute)?|begin|commit|rollback|reset|validate)\b/i,
```
RunDevelopment:
```suggestion
					pattern: /(^|\s)(?:disconnect\s+from|exec(?:ute)?|begin|commit|rollback|reset|validate)\b/i,
```

components/prism-sas.js
```diff
@@ -1,34 +1,184 @@
...
+				},
+				'punctuation': /;/
+			}
+		},
+
+		'proc-sql': {
+			pattern: /(^proc\s+(fed)?sql(?:\s+[\w|=]+)?;)[\s\S]+?(?=^(?:proc\s+\w+|quit|run|data);|(?![\s\S]))/im,
+			inside: {
+				'sql': {
+					pattern: RegExp(/^[ \t]*(?:select|alter\s+table|(?:create|describe|drop)(?:\s+)(?:index|table(?:\s+constraints)?|view)|create\s+unique\s+index|insert\s+into|update)(?:<str>|[^;"'])+;/.source.replace(/<str>/g, stringPattern), 'im'),
```
RunDevelopment:
```suggestion
					pattern: RegExp(/^[ \t]*(?:select|alter\s+table|(?:create|describe|drop)\s+(?:index|table(?:\s+constraints)?|view)|create\s+unique\s+index|insert\s+into|update)(?:<str>|[^;"'])+;/.source.replace(/<str>/g, stringPattern), 'im'),
```

components/prism-sas.js
```diff
@@ -1,34 +1,184 @@
...
+				'number': number,
+				'numeric-constant': numericConstant,
+				'punctuation': punctuation,
+				'string': string
+			},
+			lookbehind: true
+		},
+
+		'proc-args': {
+			pattern: /(^proc\s+\w+)(\s+[\w="'']+);/im,
```
RunDevelopment: Removed trailing space. I also made it more accepting.
The quotes are for strings, right? I that case, we should add proper string support similar to the `sql` pattern.

```suggestion
			pattern: /(^proc\s+\w+\s+)(?!\s)[^;]+;/im,
```

components/prism-sas.js
```diff
@@ -1,34 +1,184 @@
...
+			lookbehind: true,
+			inside: args
+		},
+
+		/*Special keywords within macros*/
+		'macro-keyword': {
+			pattern: /((?:^|[\s])=?)%(?:ABORT|BQUOTE|BY|CMS|COPY|DISPLAY|DO|ELSE|END|EVAL|GLOBAL|GO|GOTO|IF|INC|INCLUDE|INDEX|INPUT|KTRIM|LENGTH|LET|LIST|LOCAL|NRBQUOTE|NRQUOTE|NRSTR|PUT|QKTRIM|QSCAN|QSUBSTR|QSYSFUNC|QUOTE|QUPCASE|RETURN|RUN|SCAN|STR|SUBSTR|SUPERQ|SYMDEL|SYMGLOBL|SYMLOCAL|SYMEXIST|SYSCALL|SYSEVALF|SYSEXEC|SYSFUNC|SYSGET|SYSRPUT|THEN|TO|TSO|UNQUOTE|UNTIL|UPCASE|WHILE|WINDOW)\b/i,
```
RunDevelopment:
```suggestion
			pattern: /((?:^|\s)=?)%(?:ABORT|BQUOTE|BY|CMS|COPY|DISPLAY|DO|ELSE|END|EVAL|GLOBAL|GO|GOTO|IF|INC|INCLUDE|INDEX|INPUT|KTRIM|LENGTH|LET|LIST|LOCAL|NRBQUOTE|NRQUOTE|NRSTR|PUT|QKTRIM|QSCAN|QSUBSTR|QSYSFUNC|QUOTE|QUPCASE|RETURN|RUN|SCAN|STR|SUBSTR|SUPERQ|SYMDEL|SYMGLOBL|SYMLOCAL|SYMEXIST|SYSCALL|SYSEVALF|SYSEXEC|SYSFUNC|SYSGET|SYSRPUT|THEN|TO|TSO|UNQUOTE|UNTIL|UPCASE|WHILE|WINDOW)\b/i,
```

components/prism-sas.js
```diff
@@ -1,34 +1,184 @@
...
+		'numeric-constant': numericConstant,
+		'datetime': {
+			// '1jan2013'd, '9:25:19pm't, '18jan2003:9:27:05am'dt
+			pattern: RegExp(stringPattern + '(?:dt?|t)'),
+			alias: 'number'
+		},
+		'string': string,
+		'step': step,
+		'keyword': {
+			pattern: /((?:^|[\s])=?)(?:action|after|analysis|and|array|barchart|barwidth|begingraph|by|cas|cbarline|cfill|close|column|computed?|contains|data(?=\=)|define|document|do\s+over|do|dol|drop|dul|end|entryTitle|else|endcomp|fill(?:attrs)?|filename|group(?:by)?|headline|headskip|histogram|if|infile|keep|label|layout|legendlabel|length|libname|merge|midpoints|name|noobs|nowd|ods|options|or|out(?:put)?|overlay|plot|ranexp|rannor|rbreak|retain|set|session|sessref|statgraph|sum|summarize|table|temp|then\sdo|then|title\d?|to|var|where|xaxisopts|yaxisopts|y2axisopts)\b/i,
```
RunDevelopment:
```suggestion
			pattern: /((?:^|\s)=?)(?:action|after|analysis|and|array|barchart|barwidth|begingraph|by|cas|cbarline|cfill|close|column|computed?|contains|data(?=\=)|define|document|do\s+over|do|dol|drop|dul|end|entryTitle|else|endcomp|fill(?:attrs)?|filename|group(?:by)?|headline|headskip|histogram|if|infile|keep|label|layout|legendlabel|length|libname|merge|midpoints|name|noobs|nowd|ods|options|or|out(?:put)?|overlay|plot|ranexp|rannor|rbreak|retain|set|session|sessref|statgraph|sum|summarize|table|temp|then\sdo|then|title\d?|to|var|where|xaxisopts|yaxisopts|y2axisopts)\b/i,
```

components.json
cedporter: Ok, i'll cover lua and groovy in the next pull request after we merge this one.

gulpfile.js/index.js
cedporter: > The reason the build failed is that you forgot to revert the `package-lock.json` and now they are out of sync which causes `npm ci` to fail.
> 
> > cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.

Once I run npm install, I get the gulp error I got before:

```
[09:46:55] Using gulpfile ~\git\prism\gulpfile.js
[09:46:55] Task never defined: default
[09:46:55] To list available tasks, try running: gulp --tasks
````

Comment (RunDevelopment) [reply gulpfile.js/index.js]: Cannot reproduce. I clone your repo and everything works on my side.
What version of Node are you using? (I'm using v10.4.0)

<details>

This is the Windows CMD log. Commands are the same on bash.

```
C:\Users\Michael\Desktop>git clone https://github.com/cedporter/prism.git
Cloning into 'prism'...
remote: Enumerating objects: 2, done.
remote: Counting objects: 100% (2/2), done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 17571 (delta 0), reused 1 (delta 0), pack-reused 17569Receiving objects: 100% (17571/17571), 3.77 MiB | 3
Receiving objects: 100% (17571/17571), 5.85 MiB | 4.26 MiB/s, done.
Resolving deltas: 100% (11993/11993), done.

C:\Users\Michael\Desktop>cd prism

C:\Users\Michael\Desktop\prism>git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean

C:\Users\Michael\Desktop\prism>git checkout sas-improvements
Switched to a new branch 'sas-improvements'
Branch 'sas-improvements' set up to track remote branch 'sas-improvements' from 'origin'.

C:\Users\Michael\Desktop\prism>git status
On branch sas-improvements
Your branch is up to date with 'origin/sas-improvements'.

nothing to commit, working tree clean

C:\Users\Michael\Desktop\prism>npm i
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.7 (node_modules\fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.7: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})

added 544 packages from 816 contributors and audited 6938 packages in 8.839s
found 95 high severity vulnerabilities
  run `npm audit fix` to fix them, or `npm audit` for details

C:\Users\Michael\Desktop\prism>npx gulp
[16:56:30] Using gulpfile ~\Desktop\prism\gulpfile.js
[16:56:30] Starting 'default'...
[16:56:30] Starting 'minifyComponents'...
[16:56:30] Starting 'componentsJsonToJs'...
[16:56:30] Starting 'build'...
[16:56:30] Starting 'languagePlugins'...
[16:56:30] Finished 'componentsJsonToJs' after 106 ms
[16:56:30] Finished 'languagePlugins' after 127 ms
[16:56:30] Starting 'minifyPlugins'...
[16:56:30] Finished 'build' after 203 ms
[16:56:31] Finished 'minifyPlugins' after 591 ms
[16:56:32] Finished 'minifyComponents' after 1.32 s
[16:56:32] Finished 'default' after 1.33 s
```

</details>

File: components.json (hunk 35 above)
Comment (RunDevelopment) [reply components.json]: Nice. Please remove the unnecessary requires in the meantime.

Also, instead of `require`, maybe we could use `peerDependencies` instead. This will make SQL and the others optional so that users can make smaller Prism builds and aren't forced to include SQL (and others) which they might not even use. 
You don't even need to change your language definition fo this. The only real change will be for tests, where you will have to [manually include SQL](https://prismjs.com/test-suite.html#writing-tests) and the test page where you have to manually load SQL before SAS (simply click on SQL before clicking on SAS).

File: gulpfile.js/index.js (hunk 34 above)
Comment (cedporter) [reply gulpfile.js/index.js]: Once I ran gulp with "npx gulp" it worked. Running simply "gulp" appears to have been the culprit.

File: components.json (hunk 35 above)
Comment (cedporter) [reply components.json]: The commit I just made removes the other requirements. I didn't swap to peerDependencies. But I could if you really wanted. PROC SQL is part of base sas, so it's fairly commonly used.

File: components/prism-sas.js (hunk 43 above)
Comment (RunDevelopment) [reply components/prism-sas.js]: This isn't resolved yet.

File: components.json (hunk 35 above)
Comment (RunDevelopment) [reply components.json]: If it's common, then we'll leave it as `require`.

File: gulpfile.js/index.js (hunk 34 above)
Comment (RunDevelopment) [reply gulpfile.js/index.js]: Just a guess but could it be that your global `gulp` and your local `npx gulp` are different versions?
Well, as long as it works for you, it's fine I guess.

File: components/prism-sas.js [hunk 44]
```diff
@@ -1,34 +1,184 @@
...
+			lookbehind: true,
+			inside: {
+				'sql': {
+					pattern: RegExp(/^[ \t]*(?:select|alter\s+table|(?:create|describe|drop)\s+(?:index|table(?:\s+constraints)?|view)|create\s+unique\s+index|insert\s+into|update)(?:<str>|[^;"'])+;/.source.replace(/<str>/g, stringPattern), 'im'),
+					alias: 'language-sql',
+					inside: Prism.languages.sql
+				},
+				'global-statements': globalStatements,
+				'sql-statements': {
+					pattern: /(^|[\s])(?:disconnect\s+from|exec(ute)?|begin|commit|rollback|reset|validate)\b/i,
```
RunDevelopment:
```suggestion
					pattern: /(^|\s)(?:disconnect\s+from|exec(?:ute)?|begin|commit|rollback|reset|validate)\b/i,
```

RunDevelopment:

PR #1986, opened 2019-07-17
reST: Fixed exponential pattern

### LINKED ISSUE #1984: reST (reStructuredText) language can crash the browser
**Information:**
- Prism version: 1.16.0
- Environment: Browser, reproducible in Chrome (Version 75.0.3770.142 (Official Build) (64-bit)) & Firefox v67.0.4 - Build: 20190619235627

Does the latest version of Prism from the [download page](https://prismjs.com/download.html) also have this issue? Yes

**Description**
On https://prismjs.com/test.html
Using the input string of 
```
12345678901234567890123456897987987689567858756785765765675:a_a___
```

Then selecting the language `reST (reStructuredText)` causes the browser tab to hang, which I think is caused by:
https://github.com/PrismJS/prism/blob/bab744a6d5e27f3c0ff37c205d749de6f2076e73/components/prism-rest.js#L189

RunDevelopment:
This fixes #1984.

---

Thank you @nathf for the detailed report. Thanks to that, I was able to quickly fix the issue.


mAAdhaTTah:
If this fixes the issue, I'm ok with it. I tagged @Golmote as a reviewer, but if he doesn't have time, we should merge before the weekend.

PR #1996, opened 2019-07-23
Added test for unused capturing groups

RunDevelopment:
This adds a test which checks for whether there are unused capturing groups in a pattern.
_Edit:_ I added another test which ensures that lookbehind groups are always the first thing being matched in a pattern.

The new test does this by parsing every pattern using [regexpp](https://github.com/mysticatea/regexpp). This is a quite costly operation but it only doubles the test time from 0.5s to 1s one my machine so it's not even close to the 7s for the >1.3k language tests.

@Golmote once went on a killing spree to [eliminate every last unused capturing group](https://github.com/PrismJS/prism/blob/master/CHANGELOG.md#updated-components-10) and this test makes sure that no new ones get added.


mAAdhaTTah:
Would we be able to use the new dependency loader for testing the peer dependencies?

RunDevelopment:
What do you mean by _new dependency loader_? Do you mean #1998?
If so, then yes of course! #1998 is able to do everything and more what our current logic is able to do.

mAAdhaTTah:
Yes, #1998, which I want to land because I'm getting bugs on the babel plugin that I want to verify is fixed by this.

Issue #1490, opened 2018-07-20
What are peer dependencies?


RunDevelopment:
In `components.json` I came across `peerDependencies`. What are they?

(Is there a doc for `components.json`?)

mAAdhaTTah:
`peerDependencies` were added here: https://github.com/PrismJS/prism/pull/1417 – They're components that are modified by the component when it's added. e.g. if `markup` is present when `css` gets added, `markup` is modified by `css` to highlight `<style>` tag contents.

There is no doc for `components.json` at this time.

RunDevelopment:
>  They're components that are modified by the component when it's added.

That being the case, one would expect, that e.g. `markup` has to be loaded before `css`, right? Because otherwise `css` cannot modify `markup`.
But when you move `css` above `markup` in `components.json`, `css` will be loaded first when downloading `prism.js`. No nice `<style>` tag.

Peer dependencies don't do anything for the download page, so what do they do?

mAAdhaTTah:
They're currently used by `loadComponents` in Node to ensure the require.cache is cleared as expected. That's a good idea, use peerDeps for download page as well as the babel plugin.

RunDevelopment:
I'm currently implementing #1475, so I have a lot of peer dependencies.
When will this be ready, or should I try implementing it myself?  (I don't know about babel though.)

(Just out of curiosity: Is there anything checking that there are no circular dependencies? ...apart from a stack overflow?)

RunDevelopment:
I went ahead and fixed the download page. 
Babel is still an open issue but I don't have any experience working with it.

RunDevelopment:
While looking into it, I found that plugins have an `after` property which is functionally equivalent to how peer dependencies work in Prism.

Maybe we should rename it?

RunDevelopment:
I found an other issue related to peer dependencies:

The Test Drive page doesn't handle peer dependencies at all. This is very similar to the problem #1417 solved.

RunDevelopment:
On an unrelated note: 
I really have to say that renaming `components` to `config` is very confusing...

mAAdhaTTah:
If you mean in the babel plugin, you can change that if you'd like. You can post your questions in that issue if you need help.

RunDevelopment:
Thank you very much.
ES6 isn't a very common sight for me so it takes some time to get used to, but I get it now.

However, I have to say that I am shocked by the lack of strict indentation conventions ;)

PR #1998, opened 2019-07-24
Standardized dependency logic implementation

### LINKED ISSUE #1427: `loadLanguages()` should work with aliases
The `loadLanguages()` function defined in [`components/index.js`](https://github.com/PrismJS/prism/blob/gh-pages/components/index.js) should resolve aliases.

```js
loadLanguages('js');
```

### LINKED ISSUE #1992: Standardize dependency logic
**Motivation**
We have how many different implementations for loading and handling dependencies? 
The following components all have their own distict logic:

- Download page
- Examples & test page (#1936)
- Autoloader
- Testing suite
- [loadLanguages](https://github.com/PrismJS/prism/blob/master/components/index.js)
- [babel-plugin-prismjs](https://github.com/mAAdhaTTah/babel-plugin-prismjs) (external)

**Description**
Standardize the dependency logic. 

It should be able to handle both languages and plugins so that we can use it at the download page. This also includes reloading and support for required CSS files.

It would also be nice if it could deal with aliases, but that's optional.

**Info**
I already implemented something similar [here](https://github.com/RunDevelopment/prism/blob/deps/component-manager.js) so that should be a good starting point.

RunDevelopment:
This is my first take to implement a standardized dependency logic to resolve #1992.
Edit: This now also resolves #1427.

To keep the interface simple, it exports just one function: `getLoad` (better names welcome).
This function takes a components map (basically `components.json`), a list of components to load, and a list of already loaded components and returns two things:

1. The __set__ of components which have to be loaded.
2. A function to load the components in order. This function supports both synchronous and asynchronous loading.

This is the first draft and by no means finished.

I also renamed `peerDependencies` to `modify`. This is a more accurate description as [npm peer dependencies](https://dev.to/btdev/what-are-those-peerdepenedencies-in-a-nodejs-project-51jo) have a different meaning from [Prism peer dependencies](https://github.com/PrismJS/prism/issues/1490#issuecomment-406656175).

There is still one major problem which is that I don't know how to make this logic available to the website, the loadLanguage function, the Autoloader, and externals.

---

There are a few changes I made to `components.json`:

I also renamed `peerDependencies` to `modify`. This is a more accurate description as [npm peer dependencies](https://dev.to/btdev/what-are-those-peerdepenedencies-in-a-nodejs-project-51jo) have a different meaning from [Prism peer dependencies](https://github.com/PrismJS/prism/issues/1490#issuecomment-406656175).

The meaning of `require` changed. Right now, every `require` is an implicit `modify`. This is a bit of a problem because it significantly increases the number of reloads as every dependency of a component to load has to be reloaded and every loaded component depending on a component to reload has to be reloaded as well.
By changing the meaning of `require` to: _Load this component before me, I won't change it._ The number of reloads decreases.

There are now 3 types of dependencies:
1. `require`: As mentioned above it mean: _Load this component before me. I won't change it._
    (Classic dependencies. Readonly.)
1. `after`: _Load this component before me if it's loaded at all. I won't change it._
    (Optional dependencies. Readonly.)
1. `modify`: _Load this component before me if it's loaded at all. I will modify it._
    (Optional dependencies. Read/Write.)

The old behavior of `require` can be achieved by using both a `require` and `modify` dependency for the same component.

---

Now a few details about the loading process:

`getLoad` takes two component sets: `load` (components to load) and `loaded` (components which are already loaded).
It generally tries to minimize the number of components which have to be (re)loaded but all components in `load` are guaranteed to be loaded. Only the dependencies of the components to load might not appear in the final set of components to load. I.e. If we were to load JavaScript and had C like loaded already, we don't need to reload C like.


mAAdhaTTah:
I'm going to take a deeper review of this later, but first and foremost: we're gonna need tests for this 😄

RunDevelopment:
I added a few tests for basic functionality which should cover every feature of `getLoad`.
The only things I didn't test are the `series` and `parallel` functions which are kind of hard to test and should behave correctly if the general load order is correct.

RunDevelopment:
`getLoad` now supports aliases.
Aliases in the `load` and `loaded` will now be resolved. E.g. `js` will be resolved to `javascript` and `html` to `markup`.

This means that `getLoad(['html', 'svg', 'js'])` will now be equivalent to `getLoad(['markup', 'markup', 'javascript'])`.

When merged, this will take care of #1427.

Dependency aliases are not supported meaning that you e.g. can't require `js` for `javascript`. I don't think alias support for dependencies is necessary because we just don't use aliases in dependencies.

dependencies.js
```diff
@@ -0,0 +1,364 @@
...
+
+	/**
+	 * Returns a new set for the given string array.
+	 *
+	 * @param {string[]} array
+	 * @returns {StringSet}
+	 * @typedef {Object<string, true>} StringSet
+	 */
+	function toSet(array) {
+		/** @type {StringSet} */
```
mAAdhaTTah: Should this be a `Hash` vs. a `Set`? Set is a collection of unique values vs a keyed object of booleans.

dependencies.js
```diff
@@ -0,0 +1,364 @@
...
+		var set = {};
+		for (var i = 0, l = array.length; i < l; i++) {
+			set[array[i]] = true;
+		}
+		return set;
+	}
+
+	/**
+	 *
```
mAAdhaTTah: Few places: add a docblock or remove this extra line.

tests/dependency-test.js
```diff
@@ -0,0 +1,116 @@
...
+	 * Returns the load order of `getLoad`.
+	 *
+	 * @param {string[]} load
+	 * @param {string[]} [loaded]
+	 * @returns {string[]}
+	 */
+	function loadOrder(load, loaded) {
+		const order = [];
+		getLoad(components, load, loaded).load(id => order.push(id));
+		return order;
```
mAAdhaTTah: This function would be useful exported rather than just as a test function.

mAAdhaTTah:
I dropped this into the Babel plugin and it works! Some minor adjustments but should be good.

dependencies.js
RunDevelopment: Because we still target ES5 for browsers, I can't use the ES6 Set, so I made a small Set pseudo-type based on JS objects. It's supposed to my substitute for ES6 sets, so I named it that way.

Also, it's not really a _keyed object of booleans_ since there is only one value. The Java HashSet, for example, is implemented as a HashMap where all values are the same object.

tests/dependency-test.js
RunDevelopment: Instead, we could also replace the `ids` of the returned object with a `getIds` function which returns the ids in `ids` in order. I just want to keep the API minimal.

Also, where is this useful? The functional interface is more powerful.

dependencies.js
RunDevelopment: I'll add a few lines of doc in the next few days.

dependencies.js
mAAdhaTTah: I'm just talking of the naming of the type here, to be clear, but that makes sense.

RunDevelopment: Ahh, ok. 
I'll respond to the rest of your comments tomorrow. It's almost 2 am where I live.

tests/dependency-test.js
mAAdhaTTah: Oh, I see. Can this function be shortened to `return getLoad(...).load(id => id)`?

RunDevelopment: No, it can't. 
I changed the API, so we now have a `getIds` function which returns all ids in topological order.

RunDevelopment:
> minor adjustments

Well, since this update is also going to change `components.json`, we are going to have to change everything related to dependencies. This necessarily includes:

- Download page
- Testing suite
- loadLanguages

The Autoloader can come later but these 3 have to be updated to use the new dependency logic as side from the Babel plugin.

The only reason I haven't done this yet is that I don't want to resolve any merge conflicts in `components.json`.

mAAdhaTTah:
Oh I see. Ok, I need to spend some time thinking about this then.

mAAdhaTTah:
I'm a bit concerned that changing the structure & meaning of `components.json` is a breaking change though.

RunDevelopment:
Oh yes, this definitely is a breaking change for people who the dependencies from `components.json` but that is intentional as the current meaning and usage of `require` and `peerDependencies` is suboptimal. Both dependencies cause more reloads than necessary (not a huuuge issue but still) and for `peerDependencies` the name doesn't even describe their purpose (see #1490 and [npm peer dependencies](https://nodejs.org/es/blog/npm/peer-dependencies/)).

I always intended this to be a breaking change.

That being said, most users probably won't even notice this because who uses `components.json` directly beside us?

mAAdhaTTah:
Which PRs do you want to get landed ahead of this one?

RunDevelopment:
#1944 would be great.

mAAdhaTTah:
Any other PRs you want to land before we work on this? Otherwise, please go ahead and make any remaining changes and I'll give this another review.

RunDevelopment:
I'm actually making the changes right now, so I'll be done soon.

RunDevelopment:
Alright, here are the changes.

The only I haven't touched yet is the test suite which oddly enough works? I'll look into it tomorrow.

A couple of additional points:

- I haven't tested `loadLanguages` yet. Maybe we want to include this in our test suite somehow?

RunDevelopment:
I added a few checks to validate the given components. Maybe these checks should be a test instead of being part of `dependencies.js`?

components.json
```diff
@@ -432,14 +435,17 @@
 		"javadoc": {
 			"title": "JavaDoc",
 			"require": ["markup", "java", "javadoclike"],
-			"peerDependencies": [
+			"modify": [
+				"java"
+			],
```
mAAdhaTTah: Do we need `modify` here since java is already in `require`?

components.json
```diff
@@ -85,7 +85,7 @@
 		"actionscript": {
 			"title": "ActionScript",
 			"require": "javascript",
-			"peerDependencies": "markup",
+			"modify": "markup",
```
mAAdhaTTah: Maybe this is too much but what if we changed `modify` -> `modifies` & `require` -> `dependencies`? That has some nice symmetry internally as well as with other tooks (`dependencies` in pkg.json). Thoughts?

components.json
```diff
@@ -351,7 +352,7 @@
 		"haml": {
 			"title": "Haml",
 			"require": "ruby",
-			"peerDependencies": [
+			"after": [
```
mAAdhaTTah: I know I'm coming back to this after a while but `after` doesn't really say what this is supposed to do. Maybe this should be `optionalDependencies`? Is that what this means?

RunDevelopment: Yes, it's an optional dependency. I used `after` because it was already used by some plugins.

Also, regarding `optionalDependencies`: I'm not a fan of the `Dependencies` suffix. This is why I choose simple one-word names.

dependencies.js
```diff
@@ -0,0 +1,412 @@
...
+ * @property {string | string[]} [alias] An optional list of aliases for the id of the component.
+ * @property {Object<string, string>} [aliasTitles] An optional map from an alias to its title.
+ *
+ * Aliases which are not in this map will the get title of the component.
+ * @property {string | string[]} [require]
+ * @property {string | string[]} [modify]
+ * @property {string | string[]} [after]
+ */
+
+var getLoad = (function () {
```
mAAdhaTTah: Definitely need to bikeshed this function name lol. Maybe `getRequiredLanguages`? Something else?

components.json
RunDevelopment: `require` assumes that the required language isn't changed. It's a dependency where the required resource is readonly.

The meaning of `require` before this PR was different (see the initial comment of this PR), so the `modify` is necessary.

components.json
mAAdhaTTah: Well, I wouldn't combine `modify` + `require`. Let's step back for a second: we have these three keys currently:

1. `modify` - languages that, if loaded before this language, will be modified by it.
2. `require` - languages that need to be loaded before this language.
3. `after` - languages that if loaded, should appear before this language (? - not sure about this)

Are these descriptions correct?

components.json
mAAdhaTTah: Gonna resolve this convo and focus the bikeshedding the names in the other thread.

components.json
mAAdhaTTah: Is that distinction necessary to capture here? For our purposes (generating a list of dependencies in the correct order), we don't need to know whether it modifies the dependency or not. Are there use cases where the distinction is significant?

dependencies.js
RunDevelopment: The thing is: This function doesn't only return the required languages. It also returns all languages which have to the reloaded. Because of the reloads, even something as generic as `getDependencies` isn't a fitting name.

Also, this function can also be used for plugins.

TIL: [Bikeshedding](https://en.wiktionary.org/wiki/bikeshedding) is a thing.

components/index.js
```diff
@@ -1,82 +1,49 @@
...
-	if (arr && !arr.length) {
-		return;
-	}
-
-	if (!Array.isArray(arr)) {
-		arr = [arr];
-	}
+	// the user might have loaded languages via some other way or used `prism.js` which already includes some
+	// we don't need to validate the ids because `getLoad` will ignore invalid ones
+	const loaded = [...loadedLanguages, ...Object.keys(Prism.languages)];
```
mAAdhaTTah: If we can get the current list of loaded languages from `Prism.langauges`, do we need to track it internally?

dependencies.js
```diff
@@ -0,0 +1,412 @@
...
+		return set;
+	}
+
+	/**
+	 * Creates a map of every components id to its entry.
+	 *
+	 * @param {Components} components
+	 * @returns {EntryMap}
+	 *
+	 * @typedef {{ [id: string]: Readonly<ComponentEntry> | undefined }} EntryMap
```
mAAdhaTTah: Can the value actually end up `undefined`?

dependencies.js
```diff
@@ -0,0 +1,412 @@
...
+				throw new Error('Circular dependency: ' + stack.slice(firstIndex).join(' -> '));
+			}
+
+			/** @type {StringSet} */
+			var dependencies = {};
+
+			var entry = entryMap[id];
+			if (entry) {
+				/** @type {string[]} */
+				var deps = (/** @type {any[]} */([]).concat(entry.require, entry.modify, entry.after).filter(Boolean));
```
mAAdhaTTah: Is this weird `@type any[]` thing necessary? Makes the code hard to read (esp w/ the extra parens).

components.json
RunDevelopment: This is distinction is very important when it comes to reloading. We have to reload resources when loading languages not in batch but incrementally. Examples of incremental language loading are [`loadLanguages`](https://github.com/PrismJS/prism/blob/master/components/index.js) and the AutoLoader.

The problem here is that with the old meaning of `require`, every language reachable from the languages being added has to be reloaded. 
For example, if `c-like`, `js`, `css`, and `markup` are loaded and now we want to add `java`, all 4 languages have to be reloaded. (This is the current implementation of `loadLanguages`)
(A full example as to why is kinda lengthy. Should I do one?)

So, to minimize the number of reloaded languages, I introduced readonly dependencies (do they have a real name?).
For example, we know that Java doesn't modify C-like, so we don't have to reload C-like (and all of the other languages because of C-like).

dependencies.js
RunDevelopment: I know it sucks but I couldn't find another way to make TypeScript happy.

Let's keep this open here. I'll try to find a better solution in a bit.

components.json
mAAdhaTTah: No, I get it, thanks.

dependencies.js
```diff
@@ -0,0 +1,412 @@
...
+	 * @typedef {<T> (loadComponent: (id: string) => T, series?: (before: T, after: T) => T, parallel?: (values: T[]) => T) => T} LoadFunction
+	 * A functional interface to load components.
+	 *
+	 * The `loadComponent` function will be called for every component in the order in which they have to be loaded.
+	 *
+	 * `series` and `parallel` are useful for asynchronous loading and can be thought of as
+	 * `Promise#then` and `Promise.all`.
+	 *
+	 * _Note:_ Even though, both `series` and `parallel` are optional, they have to both defined or both
+	 * undefined together. It's not valid for just one to be defined while the other is undefined.
```
mAAdhaTTah: If this is the case, can we instead require an object (`loaders`, maybe?) that needs these two properties rather than passing them in as separate functions? That would make it clearer the object is optional but if you pass in the object, it needs both properties.

dependencies.js
```diff
@@ -0,0 +1,412 @@
...
+	 *
+	 * @example
+	 * load(id => { loadComponent(id); }); // returns undefined
+	 *
+	 * await load(
+	 *     id => loadComponentAsync(id), // returns a Promise for each id
+	 *     (before, after) => before.then(() => after),
```
mAAdhaTTah: I'm not entirely sure this will work the way you want (although maybe it does). Promises are eager, so every time we call the first function, the request to load the language is kicked off. At the point `series` is called, both `before` & `after` are executing promises. Chaining them this way won't defer the load of `after`, unless there's something internally managing that correctly.

dependencies.js
```diff
@@ -0,0 +1,412 @@
...
+			if (dependsOn.length === 0) {
+				value = loadComponent(id);
+			} else {
+				var depsValue = parallel(dependsOn.map(function (depId) {
+					var value = handleId(depId);
+					// none of the dependencies can be ends
+					delete ends[depId];
+					return value;
+				}));
+				value = series(depsValue, loadComponent(id));
```
mAAdhaTTah: Yeah, this won't ensure all the `depsValue` promises are resolved before kicking off the request to the `id`.

dependencies.js
RunDevelopment: Correction: Seems like they fixed this. I'll just remove the `/** @type {any[]} */` plus parens.

dependencies.js
RunDevelopment: Not really. This is just so the type of the returned value from when you index from the map (e.g. `let a = map[x]`) is optional (potentially undefined).

Just a little trick, since TS assumes that you know what you're doing when accessing a map.

components/index.js
RunDevelopment: Yes, we do because of extensions like CSS extras.

dependencies.js
RunDevelopment: Ahh, right it won't work. TBH, I didn't test it.
It fixed now and I also added a test.

dependencies.js
RunDevelopment: Fixed.

dependencies.js
RunDevelopment: I opted for Chainer. They actually build a tree but it's more specific than Loaders. Better names are always welcome. I suck at naming things.

components.json
RunDevelopment: Yes, they are correct. Both `modify` and `after` describe optional dependencies. Both `after` and `require` describe dependencies where we assume that the depended resource is only read and never changed.

| | normal dependency | optional dependency |
| --- | --- | --- |
| **readonly** | `require` | `after` |
| **read-write** | --- | `modify` |

mAAdhaTTah:
@RunDevelopment I tagged this on a new milestone, 1.18.0. If there are other PRs you want to get into the next release, tag them in the milestone as well. I'm going to get back to reviewing this in a bit, but I'll want to get this released as soon as it's ready so I wanna start thinking about what else we want to land there before we do.

RunDevelopment:
Nice. I'll tag my languages and plugins.

components.json
mAAdhaTTah: What about this minor adjustment then:

| | normal dependency | optional dependency |
| --- | --- | --- |
| **readonly** | `require` | `optional` |
| **read-write** | --- | `modify` |

dependencies.js
mAAdhaTTah: Maybe `getLoader`? Then the type of the returned object is a `Loader`. Then we'd have `Loader#load()` & `Loader#getIds()`. Or maybe `#loadIds()` for consistency? Whatever you like better.

mAAdhaTTah:
Let's just bikeshed those last two nits and get this in. I think the API is good, if we get bugs, we'll address them when they get reported.

components.json
RunDevelopment: Totally for it. The only reason why I choose `after` is because some plugins call it that right now.

dependencies.js
RunDevelopment: Really like it.
But I'll keep it `getIds()` because it doesn't load anything. (I know that you want the "load" in "loadIds" to refer to the "load" function but I (and prob anybody else) read this as a verb.) 

Maybe `getLoadIds()` ?

components.json
mAAdhaTTah: Were those plugins updated as well?

dependencies.js
mAAdhaTTah: `getIds` is good, let's leave it.

mAAdhaTTah:
And we're ok with the renaming of these properties not being "breaking" from an API perspective? Is this considered a public API? Will we see any real world impact of this?

components.json
RunDevelopment: [Yes.](https://github.com/PrismJS/prism/blob/39ff76e3ba7300ef762551a9c8fd8e030a272323/components.json#L1185)

RunDevelopment:
From my experience, most people don't even know our languages have dependencies and just assume that every language file is a self-contained package. (I believe that I remember a library that didn't even know that `require` dependencies are a thing... and most libraries that did know `require`, didn't know what `modify` is.)

So I can't say that I feel too bad about changing a detail almost nobody seems to be aware of. And those that are aware of dependencies usually only know our `require` which didn't change much. Old dependency loaders that only consider `require` will still work (see Autoloader).

So I don't think that we'll have any negative impact (but only because we don't really tell people how dependencies work). And those that are affected can switch to the new API and (hopefully) be safe for the future.

That being said, I consider dependencies in their current form mostly non-public. The only thing which is vaguely explained is `require` which didn't change too much. But after this PR we definitely should make this part of the public API and add corresponding documentation.

mAAdhaTTah:
That explanation works well for me, although I think we should decide instead that components.js{on} are not considered public but are solely to be used by our internal tools.

RunDevelopment:
> we should decide instead that components.js{on} are not considered public but are solely to be used by our internal tools.

Yes but the whole `getLoader` should be public API.
I agree that the implementation of dependencies should be just that: an implementation detail. But I still think that we should document how our dependency system works.

RunDevelopment:
@mAAdhaTTah With that 1.18.0 should be ready. I'll finish the changelog.

mAAdhaTTah:
@RunDevelopment Are you working on the changelog currently? Once that's in I can do a publish. I'm leaving for vacation Friday, so I'm hoping to get this out before then.

RunDevelopment:
Yes, sorry. For some reason the changelog script returned an error, so I saw that and stopped for the day. I'll do it now.

mAAdhaTTah:
@RunDevelopment No need to apologize, just wanted to check in case I needed to pick that up myself.

PR #1999, opened 2019-07-24
Core: `insertBefore` will now update all references

RunDevelopment:
This removes a condition the reference update logic in `insertBefore`. All references will now be updated, no matter the name of the key.


mAAdhaTTah:
Why is this needed?

RunDevelopment:
Sorry, I forget to reply. 

I removed this line because it's unnecessary. `value === old && key != inside` means "all values which key is `inside` will not be replaced". 
This doesn't make sense for two reasons: Firstly, the entry in `Prism.languages` is already replaced in the line about (`root[inside] = ret;`) and secondly, why wouldn't you want to replace an old reference just because of its key?

mAAdhaTTah:
> Firstly, the entry in Prism.languages is already replaced in the line about (`root[inside] = ret;`)

My guess is this is why the check exists. We're already replacing `inside`, so we don't need to do it again.

RunDevelopment:
> My guess is this is why the check exists. We're already replacing inside, so we don't need to do it again.

That's also what I interpreted its purpose to be but it really doesn't matter where we set a reference again or not. 

Btw. We can't remove `root[inside] = ret;` because we can't guarantee that `root` is reachable from `Prism.languages`.

mAAdhaTTah:
So if it's fine as-is, I'm going to lean on "less churn" as my preference here and say this isn't really needed.

PR #2002, opened 2019-07-26
Standardized templating logic implementation

### LINKED ISSUE #1991: Standardize templating logic
**Motivation**
See #1931 – we have a couple of ways that we handle highlighting languages inside of other languages. It would be awesome if there was a single, unified way to do this.

**Description**
Simplify the code of the current implementations and make adding future embedding languages easier. Code weight reduction is a hard requirement, imo, otherwise, the individual implementations can stay the way they are.

**Alternatives**
Leave the code alone. If it ain't broke, don't fix it.

RunDevelopment:
This implements a standardized templating logic and resolves #1991.

The new "language" exposes two functions: `createTemplate` and `tokenizeWithHooks`.
While `tokenizeWithHooks` is just utility which is exposed for convenience, `createTemplate` is the heart of the PR.

Edit: It now also exposes a `replaceWithTemplate` function which is basically a generalization of Markup templating (MT). MT is just a thin wrapper around this templating library now.


RunDevelopment:
Can we merge this?

I can't implement #2403 with Markup Templating alone. In Balde templates, you can escape statements and directives by [adding an `@` symbol](https://laravel.com/docs/master/blade#blade-and-javascript-frameworks) in front of it. To catch this, I need lookbehinds that I don't have in Markup Templating.
While impossible in the current Markup templating, it's trivial to do in the Markup templating of this PR.

@mAAdhaTTah @Golmote

RunDevelopment:
Closed in favor of #3541.

Issue #1987, opened 2019-07-17
New inline color plugin


RunDevelopment:
**Motivation**
Most editors these days show you an inline preview for colors. E.g.

![image](https://user-images.githubusercontent.com/20878432/61371261-a5b15800-a894-11e9-988b-3b5fa4ece794.png)
(VSCode)

**Description**
For now, the new plugin should show an inline preview for every CSS color value like in the above screenshot. The box should be non-selectable and has to be exactly 2 characters wide to keep all characters of monospaced fonts aligned.

This can then be expanded to other styling languages (Less, Sass, Stylus, ...).

**Alternatives**
This could also be part of the [Previewers plugin](https://prismjs.com/plugins/previewers/) but I think that the Previewers plugin is complex enough already and I don't like the idea of having the load the heavy Previewers plugin just to get some inline colors.

PR #2007, opened 2019-07-31
Added Inline color plugin

RunDevelopment:
This implements #1987.

The new plugin uses `color` and `hexcode` tokens to detect colors and work for all languages which detect these token. Right now, this is only CSS and SCSS.

### Example

image


plugins/inline-color/prism-inline-color.js
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	/**
+	 * Validates the given Color using the current browser's internal implementation.
+	 *
+	 * @param {string} color
+	 * @returns {string | undefined}
+	 */
+	function validateColor(color) {
+		var s = new Option().style;
```
mAAdhaTTah: Why `Option`?

plugins/inline-color/prism-inline-color.js
```diff
@@ -0,0 +1,88 @@
...
+	 * If the format of the given string is invalid, `undefined` will be returned.
+	 * Valid formats are: `RGB`, `RGBA`, `RRGGBB`, and `RRGGBBAA`.
+	 *
+	 * @param {string} hex
+	 * @returns {string | undefined}
+	 */
+	function parseHexColor(hex) {
+		// the regex explained: In first lookahead we check whether the string is valid and then we use the
+		// capturing groups to split the string into its components.
+		var match = /^#?(?=(?:[\da-f]{1,2}){3,4}$)([\da-f][\da-f]?)([\da-f][\da-f]?)([\da-f][\da-f]?)([\da-f][\da-f]?)?$/i.exec(hex);
```
mAAdhaTTah: Can this regex be lifted out of this function?

plugins/inline-color/prism-inline-color.js
```diff
@@ -0,0 +1,88 @@
...
+		if (match) {
+			var rgb = match.slice(1, 4).map(function (c) {
+				return String(Math.round(parseInt(c, 16) * scale * 255));
+			}).join(',');
+
+			var alpha = match[4] === undefined ? '1' : (parseInt(match[4], 16) * scale).toFixed(3);
+
+			return 'rgba(' + rgb + ',' + alpha + ')';
+		}
+		return undefined;
```
mAAdhaTTah: Can we reverse the `if` logic so:

```js
if (!match) return undefined;
// ... logic
```

plugins/inline-color/prism-inline-color.js
```diff
@@ -0,0 +1,88 @@
...
+	 * @param {string} hex
+	 * @returns {string | undefined}
+	 */
+	function parseHexColor(hex) {
+		// the regex explained: In first lookahead we check whether the string is valid and then we use the
+		// capturing groups to split the string into its components.
+		var match = /^#?(?=(?:[\da-f]{1,2}){3,4}$)([\da-f][\da-f]?)([\da-f][\da-f]?)([\da-f][\da-f]?)([\da-f][\da-f]?)?$/i.exec(hex);
+
+		// This is used to scale normalize 4bit and 8bit values
+		var scale = hex.length <= 4 ? 1 / 15 : 1 / 255;
```
mAAdhaTTah: This can go after the `if` check.

plugins/inline-color/prism-inline-color.js
RunDevelopment: Short: Copied from StackOverFlow.
Long: `new Option().style` is the shortest sequence of characters which create a new CSSStyleDeclaration.

plugins/inline-color/prism-inline-color.js
RunDevelopment: Sure

plugins/inline-color/prism-inline-color.js
```diff
@@ -0,0 +1,90 @@
...
+	 *
+	 * If the format of the given string is invalid, `undefined` will be returned.
+	 * Valid formats are: `RGB`, `RGBA`, `RRGGBB`, and `RRGGBBAA`.
+	 *
+	 * @param {string} hex
+	 * @returns {string | undefined}
+	 */
+	function parseHexColor(hex) {
+		// the regex explained: In first lookahead we check whether the string is valid and then we use the
+		// capturing groups to split the string into its components.
```
mAAdhaTTah: This comment should probs go with it :D

plugins/inline-color/prism-inline-color.js
```diff
@@ -0,0 +1,90 @@
...
+				color = parsers[i](rawText);
+			}
+
+			if (!color) {
+				return;
+			}
+
+			var previewElement = '<span class="inline-color" style="background-color:' + color + ';"></span>';
+
+			env.attributes['data-color'] = '';
```
mAAdhaTTah: Why are we clearing this?

RunDevelopment: We don't clear it, we add it. 
I just need something to identify the elements with inline color preview and didn't want it to be a class because some of our plugins change these.

mAAdhaTTah: A class like `color-preview` would be an issue? I know the prefixer plugin might change the class names but what else impacts this?

RunDevelopment: It can do a little more than just adding a prefix, but yes, that's my concern.

mAAdhaTTah: Other plugins create their own elements without running through that plugin (e.g. the Toolbar plugin). I would use a class here & file an issue for running plugin-created elements through that plugin, cuz that's more widespread than just this.

RunDevelopment: So you like an ignore filter for special plugin classes?
This will work but then either all plugins have to communicate their classes in one form or another. What control do you suggest?

---

Also, Toolbar isn't affected because its classes are outside the reach of the `wrap` hook which is the hook [Custom class uses](https://github.com/PrismJS/prism/blob/010a0157b8976ec544b4931f09dd06c1daa89c45/plugins/custom-class/prism-custom-class.js#L58).

mAAdhaTTah: Oh, I got my understanding backwards–I thought this wasn't run through custom classes & should be, but the issue is it _is_ through custom classes and... shouldn't be? I'm not sure why it shouldn't be though. When using custom classes, you have to modify the theme yourself anyway, as we have no mechanism for producing themes with the customization added. All that would mean is if you use custom classes with this plugin, you have the modify the CSS for this plugin manually as well.

RunDevelopment: Ahhh, let's not make this bigger than it is. Let's use a class. It's only used in the CSS file anyway and I guess you have to expect that using CC will screw up the CSS.

RunDevelopment: Sorry, I didn't see your comment soon enough and I agree.

RunDevelopment: I removed the attribute. Why did I even add it to begin with?

PR #2013, opened 2019-08-08
SAS: String pattern without backreferences to fix `proc-args` token

RunDevelopment:
While rebasing #1996 I noticed a small mistake I overlooked in #1981.

Because the `proc-args` pattern has a lookbehind capturing group, the `\1` in the string pattern will refer to the lookbehind group. This PR fixes this bug by replacing the old string pattern with old that doesn't use backreferences.


RunDevelopment:
I'll immediately merge this because it's a small mistake with an easy fix.

PR #2031, opened 2019-08-29
Solidity support revisited revisited

### LINKED ISSUE #1404: Support for Solidity?

RunDevelopment:
I completely rewrote the Solidity language. It now uses C like as a base because it's less likely to change compared to JS and because JS contains a lot of features which Solidity doesn't.

I also added a few tests.

### Example

image

---

This closes #1843 and resolves #1404.


tests/languages/solidity/keyword_feature.test
```diff
@@ -0,0 +1,128 @@
...
+event
+external
+for
+from
+function
+if
+import
+indexed
+inherited
+interface;
```
mAAdhaTTah: Is the semicolon needed in the test?

RunDevelopment: Yes because of the class name detection. We use the same trick in many other languages a well, e.g. JS and C++.

cliffhall:
@davidmurdoch When will you be tagging a version including solidity.min.js and pushing to unpkg.com?

mAAdhaTTah:
We have 3 PRs slated for the next release, so whenever those are in, we'll cut the next one.

cliffhall:
Cool. I'm just serving a local copy for now. Thanks!

PR #2012, opened 2019-08-05
Add RDF Turtle/TriG language

jakubklimek:

PR #2033, opened 2019-08-30
Added support for SPARQL language

thomasdegroot18:
This is the support for SPARQL based on the W3C SPARQL specifications
https://www.w3.org/TR/sparql11-query/

All of the queries from the specifications are included in the tests and as examples.


RunDevelopment:
Thank you for this PR @thomasdegroot18!

Before I'll review this, a few minor things:

1. Please move `sparql` below `soy`. The list is sorted alphabetically.
2. We use tabs for indentation. This is true for all files: `.html`, `.js`, `.test`.
3. Please remove the `g` flags from all patterns. Prism will add a `g` if it has to.
4. Travis failed because you forgot to rebuild Prism. Run `npx gulp` to rebuild Prism.

I'll properly review your PR soon.

thomasdegroot18:
Thanks for the quick comments.

I've updated the location, changed everything in tabs now, removed the g flags and reran npx gulp. 
Thanks for the update.

tests/languages/sparql/function_feature.test
```diff
@@ -13,17 +13,17 @@ foo.-:
...
-    ["PrefixedName", "foo:ba1"],
-    ["PrefixedName", "foo:1ba"],
-    ["PrefixedName", "foo:bar"],
-    ["PrefixedName", ":bar"],
-    ["PrefixedName", "foo:bar%20bar"],
-    ["PrefixedName", "foo:bar.bar"],
-    ["PrefixedName", "foo-2:"],
-    ["PrefixedName", "foo.-:"]
+		["PrefixedName", "foo:bar"],
+		["PrefixedName", "foo:"],
```
RunDevelopment: Please correct the indentation to one tab. (same for the other tests except `keyword_feature.test`)

thomasdegroot18:
After checking it, the SPARQL spec and the turtle spec are a bit to diverged to extend one onto the other. Due to most of the changes to both of the specs, both are similar but differ quite a lot in the basics. 

I've updated the code with the comments about the quotes and the array.

RunDevelopment:
While it's true that SPARQL and Turtle don't share a lot when it comes to semantics, their syntax is quite similar on purpose:

> Most forms of SPARQL query contain a set of triple patterns called a basic graph pattern. Triple patterns are like RDF triples except that each of the subject, predicate and object may be a variable. ([source](https://www.w3.org/TR/sparql11-query/#basicpatterns))

> Turtle provides levels of compatibility with the N-Triples [N-TRIPLES] format as well as the triple pattern syntax of the SPARQL W3C Recommendation. ([source](https://www.w3.org/TR/turtle/#h2_abstract))

Also, extending in Prism doesn't have any strict definitions or restrictions. It's just a way to share common patterns and structures between language definitions. I suggested extending from Turtle because both language definitions seem very similar and because extending generally reduces redundancy across language definitions and the overall bundle size of Prism.

The only differences between Turtle and SPARQL that I see are the actual syntactical differences (new (e.g. `variable`) and drastically changed (e.g. `keyword`) patterns) and small differences where you abided to spec more closely than I and the author of #2012 did (e.g. `iri`) and where you didn't quite adhere to the spec (e.g. `string`).

I'll make an update to make Turtle adhere to its spec more closely and to fix a few small mistakes I found.

thomasdegroot18:
Thanks for the update, I will update the code accordingly and update the pull request.

components.json
```diff
@@ -118,7 +118,7 @@
 			"owner": "Golmote"
 		},
 		"asciidoc": {
- 			"alias": "adoc",
+			"alias": "adoc",
```
RunDevelopment: Good catch! Thanks!

components.json
```diff
@@ -883,6 +883,12 @@
 			"require": "markup-templating",
 			"owner": "Golmote"
 		},
+		"sparql": {
+			"title": "SPARQL",
+			"require": "turtle",
+			"owner": "Triply",
```
RunDevelopment: Shouldn't this be your GitHub handle?

components/prism-sparql.js
```diff
@@ -0,0 +1,17 @@
+Prism.languages.sparql = Prism.languages.extend('turtle', {
+		'variable': {pattern: /(\?|\$)[\w]+/,
```
RunDevelopment: Is the first group supposed to be a lookbehind group?

If not, please you: `/[?$]\w+/`

components/prism-sparql.js
```diff
@@ -0,0 +1,17 @@
...
+								 greedy: true
```
RunDevelopment: Interesting indentation.

components.json
thomasdegroot18: I've updated the handle to match the company handle we use on github. It would be me most of the time doing bugfixes, but we are better reachable this way, we thought.

components/prism-sparql.js
thomasdegroot18: True, I've updated it.

components/prism-sparql.js
thomasdegroot18: Improved the indentation, would look a better now.

components.json
RunDevelopment: Ah, ok. I was just a little confused because usually, it's the author itself, but that also ok.

tests/languages/sparql/keyword_feature.test
```diff
@@ -0,0 +1,271 @@
...
+TZ(
+UCASE(
+URI(
+YEAR(
+
+----------------------------------------------------
+
+[
+		["keyword", "A"],
+		["keyword", "ADD"],
```
RunDevelopment: I forgot to say this last time: Your editor seems to have incorrectly detected the indentation as 2 spaces before converting to tabs. Please fix the indentation in, I think, all test files.

package-lock.json
```diff
@@ -1535,7 +1535,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
```
RunDevelopment: Please revert the changes to package-lock.json.

thomasdegroot18: Reverted the package-lock back to previous commit.

tests/languages/sparql/keyword_feature.test
thomasdegroot18: No problem. Thanks, updated the testfiles that had wrong indentation.

RunDevelopment:
Thank you for contributing!

PR #2034, opened 2019-08-30
Added Robot Framework plain text format

### LINKED ISSUE #1990: Robot Framework
**Language**

Robot Framework is a Python-based, extensible keyword-driven automation framework for acceptance testing, acceptance test driven development (ATDD), behavior driven development (BDD) and robotic process automation (RPA). It can be used in distributed, heterogeneous environments, where automation requires using different technologies and interfaces.

This is request to support Robot Framework "plain text format" (usually files ending with .robot): https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#plain-text-format

**Additional resources**

"specification": https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#plain-text-format

JS/RegExp based CodeMirror mode definition: https://github.com/gtri/irobotframework/blob/master/src/packages/jupyterlab-robotframework/src/mode.ts

RunDevelopment:
This resolves #1990.

I only added support for the space-separated format because the pipe-separated format is even harder to parse than the space-separated one. Both are very flexible and context depended which makes parsing them using regexes a nightmare (maybe there's an obvious and simple way to do it but I at least couldn't find it).

### Example

image


components/prism-robot-framework.js
```diff
@@ -0,0 +1,104 @@
...
+			lookbehind: true,
+			inside: {
+				'punctuation': /\[|\]/
+			}
+		};
+		extendecInside['variable'] = variable;
+		extendecInside['comment'] = comment;
+
+		return {
+			pattern: RegExp(/^ ?\*{3}[ \t]*<name>[ \t]*\*{3}(?:.|[\r\n](?!\*{3}))*/.source.replace(/<name>/g, name), 'im'),
```
mAAdhaTTah: Will this work with the gulp task we have in the build that does the `.source` replace?

RunDevelopment: Yes. Here's the minified version of that line (pulled from `components/prism-robot-framework.min.js`):

```js
pattern:RegExp("^ ?\\*{3}[ \t]*<name>[ \t]*\\*{3}(?:.|[\r\\n](?!\\*{3}))*".replace(/<name>/g,t),"im")
```

What made you think it wouldn't work?

mAAdhaTTah: I didn't remember how it worked and I was wondering if the dynamic rewrite was an issue.

RunDevelopment: > I didn't remember how it worked

Barely. I used a regex to find all RegExp literals which end with `/.source` and replace those with the optimized source value. I say optimized but it's really just a few things like replace `\\t` with `\t` to save a few characters. It's technically incorrect as it's not the real source value but it works well for Prism where we never leave the world of regex patterns.

I'm planning to drop this as soon as our minifier supports this. With regard to #1578, I opened [an issue](https://github.com/terser/terser/issues/400) at [Terser](https://github.com/terser/terser) which can minify ES6+ code.

PR #2038, opened 2019-09-02
Turtle: Minor improvements

RunDevelopment:
This makes a few small changes to Turtle:

- Multiline strings were too greedy and matched `"""a"""""` whole.
- URLs (IRI) were too accepting.
- The keyword patterns are now disjoint.


RunDevelopment:
I'll merge this now so that these changes can potentially be used in #2033.

PR #2041, opened 2019-09-04
single-quote-character not a string, fixes #2039

### LINKED ISSUE #2039: Smalltalk: single-quote-character is not a string
**Information**
- Language: Smalltalk
- Plugins: none

**Description**
The current prism test site is operating correctly but I noticed a missing test.
The single-quote character used to define literal-strings...  e.g. 'foobar' 
can also be defined as a literal-character... e.g.  $'
This causes a problem on the Exercism site https://github.com/exercism/exercism/issues/5011
I presume this was incidentally fixed by changes since the version they are using.
It would still be good to explicitly capture this as a test case.

**Code snippet**

Running...
 **npm run test:languages -- --language=smalltalk --pretty**

<details>
<summary> The following passes... </summary> 

string_character_feature.test
```
$a
'foobar'                                                                                                                
[
        ["character", "$a"],
        ["string", "'foobar'"],
]
Checks string parsing not confused after string-character-literal. 
```
</details>
<details>
<summary>The following doesn't... </summary>

string_character_feature.test
```
$'
'foobar'                                                                                                                
[
        ["character", "$'"],
        ["string", "'foobar'"],
]
Checks string parsing not confused after string-character-literal.  
```

</details>

bencoman:
For Smalltalk, a single quote character... $'
does not start a string e.g.  'foobar'.
Fixes #2039.


RunDevelopment:
Thank you for the fix @bencoman!

Before we can merge this, you have to rebuild Prism (this is also the reason why the build failed). To do this run the `npx gulp` command. This will automatically minify all languages definitions.

bencoman:
I'm not familiar with this error "cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing."

RunDevelopment:
Unfortunately, neither am I, but I still have a fix.
Please revert your changes to `package.json` and `package-lock.json`. Then use `npm ci` instead of `npm install`. After that, `npx gulp` again. This should fix the build and also revert the ~10 other minified files that changed.

mAAdhaTTah:
`cipm` was the name of the `npm ci` command before it was merged into npm core. `npm ci` uses the lockfile only to install the npm dependencies. It's warning you that because the lockfile & the pkg.json out of sync, relying on the lockfile alone won't produce the expected results, so it errors.

That said, @RunDevelopment's solution should work.

tests/languages/smalltalk/string_character_feature.test
```diff
@@ -0,0 +1,13 @@
...
+----------------------------------------------------
+
+[
+	["character", "$x"],
+	["string", "'foobar'"]
+]
+
+----------------------------------------------------
+
+xChecks for strings.
```
mAAdhaTTah:
```suggestion
Checks for strings.
```

bencoman: thx.  I added that 'x' while experimenting trying to work out how the build/test worked, and forgot.

RunDevelopment:
Thank you for contributing @bencoman!

PR #2044, opened 2019-09-05
https link to alistapart.com

coliff:
saves a HTTP redirect


RunDevelopment:
Thank you for contributing!

PR #2046, opened 2019-09-06
Added missing keywords

ifcologne:
K_SHORTEST_PATHS
SEARCH


RunDevelopment:
Thank you for adding the keywords!

However, you have to add them to `components/prism-aql.js` and then run `npm ci && npx gulp` which will generate `components/prism-aql.min.js` for you.

Also, the `K_SHORTEST_PATHS` keyword is not mentioned [here](https://www.arangodb.com/docs/stable/aql/fundamentals-syntax.html#keywords). Are there different variants of AQL?

RunDevelopment:
I found it in the [example section](https://www.arangodb.com/arangodb-training-center/graphs/k-shortest-paths-queries-in-aql/). On that note, shouldn't `TO` and `OPTIONS` also be keywords?

image

ifcologne:
Correct, `TO` and `OPTIONS` are missing as well.
And there are pseudo-values `OLD`and `NEW`. Perhaps they might need a different highlighting color.

```
FOR u IN users
   REMOVE u IN users 
   LET removed = OLD 
   RETURN removed._key
```
I've checked the _syntax documentation_ on the ArangoDB website and might suggest the related changes there as well.

RunDevelopment:
Why did you close this?

> And there are pseudo-values `OLD` and `NEW`. Perhaps they might need a different highlighting color.

Maybe `constant`?

PR #2047, opened 2019-09-06
AQL: Update keyword list

Simran-B:
Add K_SHORTEST_PATHS, OPTIONS, SEARCH, TO

Technically, some of the words are not keywords / reserved words in AQL (you may use them as attribute names for instance without having to wrap them in backticks). Should they be removed from this list and implemented in a different way? If no, then e.g. `doc.search` will have `search` colored like a keyword even though ArangoDB processes it like a regular attribute access. If yes, how to do that in prism?

True keywords:
- `AGGREGATE` (keyword)
- `ALL` (modifier)
- `AND` (operator)
- `ANY` (modifier)
- `ASC` (keyword)
- `COLLECT` (declaration)
- `DESC` (keyword)
- `DISTINCT` (modifier)
- `FILTER` (declaration)
- `FOR` (declaration)
- `GRAPH` (keyword)
- `IN` (keyword)
- `INBOUND` (modifier)
- `INSERT` (command)
- `INTO` (keyword)
- `LET` (declaration)
- `LIMIT` (declaration)
- `NONE` (modifier)
- `NOT` (operator)
- `OR` (operator)
- `OUTBOUND` (modifier)
- `REMOVE` (command)
- `REPLACE` (command)
- `RETURN` (declaration)
- `SHORTEST_PATH` (keyword)
- `SORT` (declaration)
- `UPDATE` (command)
- `UPSERT` (command)
- `WITH` (keyword)

Keyword-like:
- `true`
- `false`
- `null`

Not keywords but used in language constructs:
- `OPTIONS`
- `SEARCH`
- `TO`


RunDevelopment:
Thank you for the detailed explanation!

I added `NEW` and `OLD` to `keywords` (case-sensitive).

> If no, then e.g. `doc.search` will have `search` colored like a keyword even though ArangoDB processes it like a regular attribute access.

This will simply be a false positive. Many other languages also suffer from those (e.g. JS with `from`). You could tokenize every property access but that will generate a lot of extra tokens which aren't styled in any way.

How common will this be?
Depending on how many users will experience this issue, the overhead might be worth it.

Simran-B:
> I added `NEW` and `OLD` to `keywords` (case-sensitive).

Very nice, thanks!

> How common will this be?
> Depending on how many users will experience this issue, the overhead might be worth it.

The most common occurrence will probably be attribute access using the dot notation. It depends on what the user calls his attributes. If the attribute name is any of `NEW`, `OLD`, `options`, `search` or `to` then there will be a false-positive, e.g. `FOR doc IN coll RETURN doc.options`.

The bracket notation is largely unaffected (string literal: `FOR doc IN coll RETURN doc["options"]`), but users may also call collections and variables after one of these pseudo-keywords (`FOR doc IN options` / `LET search = "xxx" RETURN doc[search]`).

Overall, it shouldn't be too bad and I doubt that it's worth to go to great length to avoid false-positives. I wonder however if there are some low hanging fruits.

For example, to prevent _search_  from being identified as keyword in `doc.search`, we could use something other than `\b` which excludes the dot. My first thought was `\s`, but it isn't actually guaranteed to be surrounded by whitespace:

    FOR doc IN someView SEARCH(doc.attr == "foo") RETURN doc

Eventhough this _should_ be a function call to an unknown function SEARCH, it is actually interpreted as `SEARCH doc.attr == "foo"`. I need to think and test a bit more to be able to tell if there's always a space in front of SEARCH at least.

RunDevelopment:
I made a small adjustment to get rid of _most_ of the false positives:

<details>

image

</details>

Code like `foo. search` or `foo[ to]` will still lead to false positives, so let's hope that AQL coders like formatted code.

PR #2055, opened 2019-09-13
AQL: More pseudo keywords

Simran-B:
As a follow up to #2047, I want to add some more pseudo-keywords:
- **COUNT**: `COLLECT … WITH COUNT INTO …`
- **CURRENT**: `array[* FILTER CURRENT.value == 1 RETURN CURRENT]`
- **KEEP**: `COLLECT … INTO … KEEP …`
- **PRUNE**: `FOR v IN 1..5 OUTBOUND "verts/1" edges PRUNE conditions …`

I tried to avoid false-positives by detecting `COUNT` based on the surrounding keywords, but this is in exchange for false-negatives (it won't be detected if comments are inserted in between, which is unlikely however).

For `OPTIONS` I attempted something similar (it must be followed by an object literal, variables/expressions are not excepted in this position) , but I'm not sure if it's worth it. Let me know what you think.


RunDevelopment:
> it won't be detected if comments are inserted in between, which is unlikely however

This is one of the most common types of false negatives in Prism, so I wouldn't worry too much about it. As you said, it's unlikely to occur.

Regarding `OPTIONS`: I think your solution is fine.

I think that language definition is ready to merge but we'll have to adjust the tests before that. Please add your new keywords to the [keyword test](https://github.com/PrismJS/prism/blob/master/tests/languages/aql/keyword_feature.test). ([More infos on test files](https://prismjs.com/test-suite.html#writing-tests).)

Simran-B:
Cool, will add the tests, probably tomorrow.

Simran-B:
@RunDevelopment I added some more tests

Issue #1041, opened 2016-10-26
How to combine plugins


donnieflorence:
Hello, 

I'm wondering how does one combine several plugins together like `line-numbers` & `line-highlight`? Do you just do something like this in the class="line-numbers line-highlight" and put both plugin logic in one function?

Golmote:
You just need to include both plugins and use both classes just as you said.

PR #2056, opened 2019-09-16
Tests: Let runTestCase() pretty-print expected TokenStream too

Simran-B:
Only the actual token stream was pretty-printed. This makes the helper function pretty-print the expected token stream as well (and the diff too). Also added some `\n` to separate it from the mocha output.

### DIFF
--- a/tests/helper/test-case.js
+++ b/tests/helper/test-case.js
@@ -64,14 +64,27 @@ module.exports = {
 		// the first language is the main language to highlight
 		const simplifiedTokenStream = this.simpleTokenize(Prism, testCase.testSource, usedLanguages.mainLanguage);
 
-		const tzd = JSON.stringify(simplifiedTokenStream);
+		const actual = JSON.stringify(simplifiedTokenStream);
+		const expected = JSON.stringify(testCase.expectedTokenStream);
 
-		const tokenStreamStr = pretty ? TokenStreamTransformer.prettyprint(simplifiedTokenStream) : tzd;
+		if (actual === expected) {
+			// no difference
+			return;
+		}
+
+		// The index of the first difference between the expected token stream and the actual token stream.
+		// The index is in the raw expected token stream JSON of the test case.
+		const diffIndex = translateIndexIgnoreSpaces(testCase.expectedJson, expected, firstDiff(expected, actual));
+		const expectedJsonLines = testCase.expectedJson.substr(0, diffIndex).split(/\r\n?|\n/g);
+		const columnNumber = expectedJsonLines.pop().length + 1;
+		const lineNumber = testCase.expectedLineOffset + expectedJsonLines.length;
+
+		const tokenStreamStr = pretty ? TokenStreamTransformer.prettyprint(simplifiedTokenStream) : actual;
 		const message = "\n\nActual Token Stream:" +
 			"\n-----------------------------------------\n" +
 			tokenStreamStr +
 			"\n-----------------------------------------\n" +
-			"File: " + filePath + "\n\n";
+			"File: " + filePath + ":" + lineNumber + ":" + columnNumber + "\n\n";
 
 		assert.deepEqual(simplifiedTokenStream, testCase.expectedTokenStream, testCase.comment + message);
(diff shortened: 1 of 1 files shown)


RunDevelopment:
Could you please explain why you did this?

I don't see the benefit. You can see the (usually) formatted token stream in the test file itself. And the calculated difference between two formatted token streams and two non-formatted token streams should be the same (apart from some whitespaces).

Also, a potential problem I see is that the mocha output will be longer which is a bit of a problem because consoles usually only store a fixed number of lines. The output of some more complex test cases can easily exceed 1000 lines because of the (current) diff. (For reference: The number of stored lines in VS Code's terminal is 1000 by default.)
Won't this be even longer with is PR?

Simran-B:
Well, it already formatted the actual token stream. Having it display the actual data pretty-printed but not the expected data seems inconsistent. If you want the terse representation then you can just leave out the `--pretty` flag. For longer tests I would for sure redirect the output to a file for further inspection.

Based on your argument that you can see the formatted token stream in the test file itself, it shouldn't display the expected data at all, but just the actual data.

Not sure about the usefulness of the diff, it's really just the actual data from the first difference on - if there's an extra character in the very first line of a test then you get the full actual data twice. And mocha actually does a better job at showing a mismatch:

image

RunDevelopment:
> Having it display the actual data pretty-printed but not the expected data seems inconsistent.

True, but do need to have the expected token stream formatted?
I only added formatting for the actual token stream because I want to be able to copypaste it.

> Based on your argument that you can see the formatted token stream in the test file itself, it shouldn't display the expected data at all [...]

Well, yes kinda. I don't think we _gain_ anything from displaying the expected token stream. After all, you can see it in the test file, so you know exactly what's expected.
The only reason I see for its existence is consistency. Test errors usually have this expected vs actual style, so I guess it makes sense to display the expected token stream. I'm not too sure about this myself.

> Not sure about the usefulness of the diff, it's really just the actual data from the first difference on - if there's an extra character in the very first line of a test then you get the full actual data twice. And mocha actually does a better job at showing a mismatch:

(I never gave any thought to our diff...) You're exactly right. Mocha does a way better job. We should just remove our diff.

---

I spent ~way too much~some time digging through the git blames and found out that the message was added by [this commit](https://github.com/PrismJS/prism/commit/dfc19413d35919ebadd8514cb04b956c775354ba) from #1041.
But strangely, in the PR it was added by a [merge commit](https://github.com/PrismJS/prism/pull/1014/commits/5cbbe998a84100ac168ac35b4577328061bb890f#diff-eb28194298ee269030517fa2fa223411).

If I had to guess, @Golmote didn't notice this change before merging (@Golmote [wouldn't have accepted](https://github.com/PrismJS/prism/commit/46e6f0c08e4164d253def6ec2d4fc3f524879aab) code indented this badly) and it just stuck around.

Knowing this: What more than the actual token stream and a small diff from mocha do we _really_ need?
Might as well take this chance to make the errors a little simpler.

Simran-B:
To be honest, I would scratch the diff and the expected output and just print the actual token stream. This could even be behind another flag, as mocha tells you pretty much everything you need to know. But I guess it doesn't hurt as long as it is printed above what mocha outputs.

Not sure if there is anything else needed. The position of the first difference / error? But that's tricky, because the raw as well as the reformatted token streams differ from the test file, so there's no straightforward way to compute that for the file. What could be helpful though is the location of the test in question, which mocha doesn't tell you.

I committed these changes, let me know what you think:

image

(without --pretty)

RunDevelopment:
I really like the simplified format.

I added a bit code to also have the line and column number of the first difference at the end of the file path. Most IDEs understand this format, so you can jump directly into the test case at the position of the difference.

<details>

image

</details>

Please tell me what you think! 
@Simran-B @mAAdhaTTah

Simran-B:
I tried it with different white-space in the test file and it seems to point reliably at or behind the position of the deviation. Looks very good to me!

mAAdhaTTah:
I dig it. :shipit:

PR #2058, opened 2019-09-18
AQL: Unclosed multi-line comments are allowed

Simran-B:
AQL does not require multi-line comments to be closed:

```
/* unterminated
multi-line
comment
RETURN 1
```

ArangoDB simply treats everything until the end of the query string as comment in that case. This PR adjusts Prism to highlight such comments correctly.

Note that the test for this needs to be last, as everything past it will be treated as part of this comment.


RunDevelopment:
Thank you for contributing!

PR #2061, opened 2019-09-24
Add multiline strings to clojure, lisp style

sooheon:
The following should be legal in clojure:
```
"multiline
string"
```


RunDevelopment:
Thank you for contributing @sooheon!

I'll review in a sec, but before that: The build fails because `components/prism-clojure.min.js` isn't up to date. If I had to guess I'd say you rebuilt Prism and then made another change to `prism-/clojure.js` (happens to me all the time). 
So please rebuild Prism again, so the build passes.

sooheon:
Thanks for the tip, that indeed was the problem. Rebuild and pushed!

RunDevelopment:
Thank you for contributing!

PR #2070, opened 2019-09-26
Added extra newlines in pretty tokenstream

RunDevelopment:
This changes the pretty simplified token stream to have an extra newline if the test code also has a double newline there.

<details>
<summary>Example</summary>

```
+ - * / %
> >= <= < == != =
!~ =~

! && ||

[**] [***] [****]

..
1..100

-------------------------------------------------------------

[
    ["operator", "+"],
    ["operator", "-"],
    ["operator", "*"],
    ["operator", "/"],
    ["operator", "%"],
    ["operator", ">"],
    ["operator", ">="],
    ["operator", "<="],
    ["operator", "<"],
    ["operator", "=="],
    ["operator", "!="],
    ["operator", "="],
    ["operator", "!~"],
    ["operator", "=~"],

    ["operator", "!"],
    ["operator", "&&"],
    ["operator", "||"],

    ["punctuation", "["],
    ["operator", "**"],
    ["punctuation", "]"],
    ["punctuation", "["],
    ["operator", "***"],
    ["punctuation", "]"],
    ["punctuation", "["],
    ["operator", "****"],
    ["punctuation", "]"],

    ["range", ".."],
    ["number", "1"],
    ["range", ".."],
    ["number", "100"]
]

------------------------------------------------------

This is the output pretty token stream as is.
The different operator groups are neatly separated by an extra newline.
```

</details>

This makes it easier to read long token streams without having to do any manual formatting.
I wanted this feature because this is how I usually format pretty token streams.


tests/helper/test-case.js
```diff
@@ -97,7 +97,6 @@ module.exports = {
 	 * @param {any} Prism The Prism instance which will tokenize `code`.
 	 * @param {string} code The code to tokenize.
 	 * @param {string} language The language id.
-	 * @returns {Array<string|Array<string|any[]>>}
```
mAAdhaTTah, line 100: Do we still want a return type here?

RunDevelopment: I think I removed it because TS can infer the type.

mAAdhaTTah: I think this would be better to be explicit and leave it there. Then another developer wouldn't need to read the file to know what the return types are.

RunDevelopment: Ok.

RunDevelopment: Actually, this brings me right back to my question in #2071: How do we define a token stream?

mAAdhaTTah: Can we leave the doc as-is & resolve it with that issue?

RunDevelopment: Alright.

RunDevelopment:
I can't merge this without a definition of a token stream. (These tests don't fail for nothing.)

Sorry, if I haven't made this clear that #2071 blocks this PR.

mAAdhaTTah:
Oh, I see, my bad. So as I commented on #2071, I would be in favor of bringing the types into alignment with the actual runtime behavior. Whatever properties of `TokenStream` `tokenize` actually produces at runtime should be reflected in the `TokenStream` type. If our current types are suggesting it returns things it doesn't, then I'd go with the simpler types, especially if your code has made assumptions about what `tokenize` produces and functional correctly with those assumptions.

RunDevelopment:
Ok yeah. I just looked at Markup templating and it also assumes the narrower definition. The minor difference here: While this code _might_ break for other definition, MT is going to break, no questions. (See `walkTokens` in MT).

I'm gonna change the definition in #1782.

PR #2073, opened 2019-09-26
JS: Tokenize `:` as an operator

### LINKED ISSUE #2072: PrismJS fails to tokenize operators, calls them "punctuation" (reproducable on prismjs.com)
This can be reproduced on Prism.js main site: 

https://prismjs.com/test.html#language=javascript

Simply this snippet here fails to be properly highlighted by prismJs
```js
const a = b ? 0 : 1
```
The colon should have the token "operator", but instead gets the token "punctuation"

PrismJS will find the `=` to be an operator, and the `?` to be an operator, but fails to see `:` as an operator and says it is "punctuation" - which causes a different color to be present on the `:` 

Imgur

There are other cases as well, I'll see if I can dig them up.

RunDevelopment:
This resolves #2072.

This will make JS and JSON consistent in how they tokenize `:`.


tests/languages/coffeescript/property_feature.test
```diff
@@ -5,11 +5,11 @@ foo_bar :
...
 	["property", "foo_bar"],
-	["punctuation", ":"]
+	["operator", ":"]
 ]
 
 ----------------------------------------------------
 
-Checks for object properties.
\ No newline at end of file
+Checks for object properties.
```
mAAdhaTTah: This is the thing though: it's not an operator here. It's tough to tokenize `:` in a way that distinguishes between its usage in ternary & properties. It's not clear that really solves the issue the original opener had either.

RunDevelopment: I don't know Coffeescript, so what is the `:` here?

Because if it's the same as JS properties (`{ foo: bar }`), then `:` is also an operator in CoffeeScript.

mAAdhaTTah: Sorry, didn't realize this was coffeescript. I think conceptually, my issue is still the same:

```
const value = { hello: 'world' };
```

The colon here is not an operator, but with this change, will be highlighted as one.

mAAdhaTTah:
To make this broader: I'm not in love with changing all `:` from punctuation -> operator because of ternaries. Their usage in object literals is far more common than in ternaries, so punctuation is more likely to be correct.

RunDevelopment:
I am for changing `:` to `operator` because of consistency. JSON also uses `operator` and it ternaries look a little strange. 
There is also the point that other highlighters also use something akin to `operator`, e.g. GitHub:

```js
const value = { hello: 'world', foo: (bar ? 1 : 0) };
```

Golmote:
It's true that the JSON component already has it tagged as an operator for some reason.

I'm only worried that in the end, someone will open another issue asking for the exact opposite of this PR.

MattMcFarland:
I agree that `:` should not be an operator when it is punctuation.  

I'm not sure how PrismJS is tokenizing things, but,  I believe that inorder to fix this, you would have to not broadly assume `:` is punctuation, but take in prior tokenized items as context and apply the the correct token type "punctuation" or "operator"

RunDevelopment:
Well, IMO the underlying issue here is that there is no clear-cut definition of what differentiates an operator from mere punctuation.
I always felt that this divide is kind of artificial but it's quite useful for highlighting.

So my question: What defines an operator vs punctuation?
The JS language spec doesn't.

MattMcFarland:
> It's true that the JSON component already has it tagged as an operator for some reason.
> 
> I'm only worried that in the end, someone will open another issue asking for the exact opposite of this PR.

I am afraid of this happening too.

The reason why I've brought this up in an issue is because our UX designer is looking at our code snippets on our site and not seeing parity github (we are using github's theme) - the first thing they pointed out to us was that our colons were not matching.  Since they are going over this with a fine-toothed comb, I'm sure if it does happen, it will happen much sooner rather than later lol.

Golmote:
@RunDevelopment The JS spec does define it, I believe. The colon is part of the Conditional Operator (12.14) which is an operator as stated, but also part of the PropertyName of an ObjectLiteral (12.2.6) in which case nothing says it's an operator.
But the spec also defines a comma operator, which we do not handle. x')

@MattMcFarland The thing is Prism is regex-based, it has no notion of context. Absolute correctness is not a goal for this project, as Lea [once said](https://github.com/PrismJS/prism/issues/412#issuecomment-66247927).

If we can handle some cases of colons used as operator, without adding too much complexity, that's good. But we most likely won't be able to handle the generic case.

If your UX designer expects your code snippets to be highlighted with absolute correctness, then I'm afraid Prism is not the right tool for the job.

MattMcFarland:
> Well, IMO the underlying issue here is that there is no clear-cut definition of what differentiates an operator from mere punctuation.
> I always felt that this divide is kind of artificial but it's quite useful for highlighting.
> 
> So my question: What defines an operator vs punctuation?
> The JS language spec doesn't.

Having not looking at the spec myself, I believe you.  I think you have a very convincing argument for broadly assuming `:` is an operator.   Even if someone comes back saying they think it should be punctuation, it looks like `:` is used as an operator for github's syntax highlighting, if you look at this snippet:

```js
const foo = bar ? 0 : 1
const baz = {
  stuff: things,
  things: 'stuff',
  doStuff: () => MyThing.doStuff()
}

**Comment by MattMcFarland** [comment]
@Golmote Please.. Say it ain't so!  I'd much rather stick with prism if I can.  :) I do think that if there are no negative consequences for switching `:` to operator from punctuation, then you should do it.  It does look like it is treated like operator when looking at github snippets at least.

**Comment by RunDevelopment** [comment]
I also agree that we shouldn't change our tokenization to fit a specific theme.
I implemented this because of consistency.

---

@Golmote 
The spec even has [this punctuator section](https://www.ecma-international.org/ecma-262/6.0/#sec-punctuators).

**Comment by RunDevelopment** [comment]
@Golmote @mAAdhaTTah 
Do we have some guidelines to distinguish between punctuation and operator?

Because when I create a language, I just choose whatever I think looks better in Prism's themes. Even though this kinda goes against '_we shouldn't change our tokenization to fit a specific theme_'.

**Comment by MattMcFarland** [comment]
I agree you shouldnt change tokenization to fit a specific theme as well. 

When I bring up github's theme and am looking for parity, it's only to show my use case.  

As for the issue, I made it a point to leave theme's out of it and just stick with the facts, where an operator was being misread as punctuation.  That's what I was hoping a fix for :)

As for parity with Github, that's something my UX designer is using as a benchmark.  If they can give us some leeway that would be great, if not, it would just create more work on our board that we would probably have to defer.

**Comment by Golmote** [comment]
Consistency is a good enough argument to me, honestly. So I won't fight against this PR.

@RunDevelopment No we don't have guidelines for this. I personnally tried to stick to the specs, whenever possible, but I might have made aesthetic choices too, from time to time. Again, given Prism is no linter, that's perfectly fine.

**Comment by mAAdhaTTah** [comment]
I'm not sure that spec reference is that helpful, cuz it shows some tokens that are definitely operators, like spread `...`. Broadly, what is an operator is defined the spec. It sounds like punctuation is... made up? for the benefit of syntax highlighting?

@RunDevelopment So no, I don't think we have guidelines for "punctuation".

---

> It does look like it is treated like operator when looking at github snippets at least.

It looks this way because the GitHub theme highlights both with the same color. The `?` is also highlighted in that color, so hypothetically, if whatever highlighter GH uses highlighted them as different things but applied the same color, the result would be the same while "treating" them as different.

**Comment by mAAdhaTTah** [comment]
If you need to satisfy your designer, you could easily just change the color of punctuation & operator to be the same in your theme. No conflict 😄

**Comment by MattMcFarland** [comment]
@mAAdhaTTah I tried that but it didn't work.  It changes commas and dots to red (github treats them differently than operator)

**Comment by RunDevelopment** [comment]
@MattMcFarland Maybe we could add a bit functionality similar to [Highlight keywords](https://prismjs.com/plugins/highlight-keywords/) for general tokens to satisfy your UX designer? I'm thinking about the ability to modify the classes of tokens on a per-language level.
Anyway, we shouldn't discuss this here.

---

@Golmote 
> I personally tried to stick to the specs, whenever possible

Me too but most specs just call everything an operator that's not very applicable. Also, if present, what is punctuation vs operators varies from spec to spec.

**Comment by mAAdhaTTah** [comment]
Ah so maybe GH's highlighter is treating `:` as an operator.

**Comment by RunDevelopment** [comment]
> Ah so maybe GH's highlighter is treating : as an operator.

Seems like it. It's highlighted like `=` but differently from `(){}`.

**Comment by mAAdhaTTah** [comment]
Inconsistency between JS & JSON is a bit weird. We should probably bring them into alignment. The issue for me is it seems weird to apply "`:` is an operator" to JSON because it's definitely _not_ an operator there. I know this flies in the face of the original request, but my inclination is to make them both punctuation.

**Comment by Golmote** [comment]
Funnily enough, the Rouge syntax highlighter switched from having `?` and `:` tagged as operators to having them both tagged as punctuation in 2012.

**Comment by mAAdhaTTah** [comment]
lol there's your consistency: make `?` punctuation!

**Comment by RunDevelopment** [comment]
> make ? punctuation!

We can also do that. 
But with that, I have another question: What about [optional chaining](https://github.com/tc39/proposal-optional-chaining) and [nullish coalescing](https://github.com/tc39/proposal-nullish-coalescing)?

`??` is an operator for sure but `?.` ?

**Comment by mAAdhaTTah** [comment]
sorry sorry i was kidding 😄

**Comment by RunDevelopment** [comment]
Seems like I didn't get the joke... <sup>Well, happens.<sup>

Anyway, so what do want to do?
About `??` and `?.`: I'm planning to add both with either this PR or another one, so I just wanted some input because the result of this will determine their fate (as tokens).

**Comment by Golmote** [comment]
To me, `??` is an operator while `?.` is punctuation. But I'll admit I have not much to back it up.

Regarding `:`, consistency with JSON seems relevant. We can either change the JSON one to punctuation or the JS one to operator. I'm fine with both solutions, but I think @mAAdhaTTah has a point saying that it's semantically not an operator in JSON at all.

**Comment by Golmote** [comment]
Also, @RunDevelopment I like your idea that this could be solved using some smart plugin, that would give all the liberty they want to the users.

Another solution _could_ be to add two classes on just the colon symbol, since we consider it special. Like `punctuation colon`. This is a bit far-fetched, but it would allow users to override only its style.

**Comment by RunDevelopment** [comment]
> I think @mAAdhaTTah has a point saying that it's semantically not an operator in JSON at all.

`?` and `:` and being punctuation is JS and JSON it is?

Just want make sure. I want them to be consistent with both JSON and each other.

**Comment by Golmote** [comment]
No, I meant only change `:` to a punctuation in JSON. No need to mess with `?` at all.

**Comment by RunDevelopment** [comment]
Seems like I won't get my consistent `?:`. 
I'll cry in a corner then. 😢

**Comment by Golmote** [comment]
> both JSON and each other

Oh, I missed that part of your sentence indeed, sorry about that.

@mAAdhaTTah Is it _that bad_ to have `:` tagged as an operator solely for aesthetic purpose?

**Comment by RunDevelopment** [comment]
Seems like the discussion isn't over. 😄 

But on the point of `:` being an operator: When we then consider other C style languages such as C/C++ where `::` is an op but `:` is punctuation even though `:` can only be part of `?:` to my knowledge. 
And regarding JSON: In Yaml, a superset of JSON, `:` is being highlighted as punctuation.

Point is: If we really want to make it consistent, we have a looot of languages to go through where we can have the same discussion. Maybe we should create guidelines so we can make all languages roughly consistent with each other and not just JS and JSON.

**Comment by Golmote** [comment]
In C/C++, the colon isn't only used in ternary. (see https://stackoverflow.com/questions/1711990/what-is-this-weird-colon-member-syntax-in-the-constructor for example)

Guidelines would be nice, but what would they be based of? What kind of generic property can be used to definitively make a choice here (and one that hopefully doesn't require a real deep knowledge of each language)?

**Comment by mAAdhaTTah** [comment]
> Seems like the discussion isn't over. 😄

Hah yeah I didn't mean to imply I was making a definitive answer here.

---

Part of me was thinking guidelines would be helpful here, but I actually don't know we could come up with guidelines that apply broadly enough to be useful.

---

> @mAAdhaTTah Is it that bad to have : tagged as an operator solely for aesthetic purpose?

No. The goal of Prism has never been correctness, so I don't really mind doing it for that reason either. A goal of consistency leads me to suggest punctuation for `:`, but if the goal is aesthetics then operator may be the way to go here.

Y'all are the regex experts, but I'm assuming it's impossible to highlight the ternary's colon as an operator separately from an object literal's colon as punctuation, correct?

---

TIL YAML is a superset of JSON.

**Comment by Golmote** [comment]
> Y'all are the regex experts, but I'm assuming it's impossible to highlight the ternary's colon as an operator separately from an object literal's colon as punctuation, correct?

We can certainly handle simple cases of ternary, like literals. But we most likely won't be able to handle every complex expression, especially expressions containing other colons. Furthermore ternaries can be infinitely nested, which we can't handle either.

**Comment by Golmote** [comment]
Suff like
```js
a ? {b: function() { label: return 1; }, c: /d:/ ? e : "f:g"} : h;
```
Ugh.

RunDevelopment:
> I'm assuming it's impossible to highlight the ternary's colon as an operator separately from an object literal's colon as punctuation, correct?

The problem, like @Golmote said, is that there can be any valid JS expression between  `?` and `:`.
The infinite nesting problem can be solved by just stopping after a fixed level of nesting. 
I actually wrote a little function which could transform a CF grammar into a RE with that idea. The pattern of a JS expression with a max nesting level 3 was >10kB if I remember correctly. Nesting with multiple branches means that our pattern grows exponentially. It's not fun and it becomes even worse because when I tested this with Chrome, it just wouldn't work if patterns were too long.

Point is: While possible, please don't.

> A goal of consistency leads me to suggest punctuation for `:`

@mAAdhaTTah Why is punctuation consistent for `:`? Do you mean consistent with other languages? Because most C style languages have the same problem.

This is why I said that making `?` punctuation isn't such a bad idea. And we could even change this in C like to apply it to all C style languages.
It's just a little... strange. Because `?:` is definitely an operator. 

---

Regarding guidelines:
I don't think we can do much better than trivial stuff like `()[]{},;` is punctuation. 
But then there are languages like EBNF where `,` is the concatenation operator... (I tokenized it as punctuation despite that.)

The only thing we might be able to do is to go for aesthetics and utility (aka not too ugly and not all the same token type).

mAAdhaTTah:
I think I've got a principle we can apply that gets us `:` as an operator:

> Don't make any token "punctuation" if it fits into another token type. In other words, "punctuation" is a token of last resort.

If we accept this principle, we have a guideline for deciding what "punctuation" is in the future while giving us the outcome we want on JS & JSON (this should stay as an operator to be consistent).

MattMcFarland:
I like that principle @mAAdhaTTah

RunDevelopment:
@mAAdhaTTah Your principle sounds good.

I'll make an issue, so we can collect these guidelines and put them on the website once we have enough.

mAAdhaTTah:
If we have docs on making a new language, that's where they should live. If not, then... well, we need that lol.

Golmote:
We only have https://prismjs.com/extending.html AFAIK.

@mAAdhaTTah's idea for the guidelines sounds good to me. This way we have the best of all worlds: this issue can be fixed, we have JS/JSON consistency and a rule to follow that "makes sense".

I still think a plugin that extends the possibilities for the users would be a nice addition.

Golmote:
Speaking of guidelines, @RunDevelopment what rule are you following to order the operator regexp parts, if any?

RunDevelopment:
> what rule are you following to order the operator regexp parts, if any?

To order? You mean why `a|b` instead of `b|a`?
Well, above all: correctness. E.g. `-|--` doesn't match `--`, so it has to be `--|-`. After correctness is ensured, I generally order them to group together as many as possible to make the regex as small as possible. E.g. `[!=]=|-` instead of `!=|-|==`.
After that is done, nothing really. I just choose whatever order as long as it's correct and fairly small.

RunDevelopment:
@Golmote 
> I still think a plugin that extends the possibilities for the users would be a nice addition.

#2075.

RunDevelopment:
I'll merge this now and will make another PR for `??` and `?.`.

We can continue the guideline discussion in #2083.

Golmote:
> To order? You mean why a|b instead of b|a?
Well, above all: correctness. E.g. -|-- doesn't match --, so it has to be --|-. After correctness is ensured, I generally order them to group together as many as possible to make the regex as small as possible. E.g. [!=]=|- instead of !=|-|==.
After that is done, nothing really. I just choose whatever order as long as it's correct and fairly small.

Fair enough. At one point in time, I tried to group them by their starting character(s), because that would theoretically be more optimized (least amount of backtracking needed when testing each alternative).

> #2075.

You're far too quick and I'm far too blind! :joy:

RunDevelopment:
> that would theoretically be more optimized (least amount of backtracking needed when testing each alternative).

I would really like to know how they optimize this under the hood. Most regexes are fairly simple and could easily be transformed into a DFA.
Time to look into the V8 source I guess.

> You're far too quick

Gotta go fast! *blue hedgehog noises*

Golmote:
> I would really like to know how they optimize this under the hood. Most regexes are fairly simple and could easily be transformed into a DFA.
Time to look into the V8 source I guess.

From "Mastering Regular Expressions (2nd edition)", by Jeffrey E. F. Friedl:

OkWruU3
[...]
LtoeaGj

I don't know how true are those statements but this is definitely interesting.

MattMcFarland:
I now know where to find regex ninjas ;)

mAAdhaTTah:
@MattMcFarland Yeah dude, these two are _so good_ at this. I'm mostly just here to make sure things move along.

If you wanna learn regex, look over any of the PRs with extensive comments. Lotta good insights.

RunDevelopment:
@Golmote 
I did some perf tests.

```md
sample        | func          | comp          | avg        | dev        | min        | max        | samples
------------- | ------------- | ------------- | ---------- | ---------- | ---------- | ---------- | ----------
prism-core.js | old operator  | 100% +-21%    | 0.07933    | 0.01674    | 0.04493    | 0.5169     | 1318
              | new operator  | 102% +-13%    | 0.08099    | 0.009933   | 0.04720    | 0.2069     | 1335
              | flat operator | 101% +-15%    | 0.08043    | 0.01168    | 0.04720    | 0.2379     | 1329
prism.js      | old operator  | 100% +-9%     | 3.470      | 0.3251     | 3.247      | 5.950      | 396
              | new operator  | 101% +-9%     | 3.503      | 0.3267     | 3.248      | 5.839      | 394
              | flat operator | 101% +-9%     | 3.495      | 0.3012     | 3.274      | 5.523      | 394
```

`avg` (average), `dev` (standard deviation), `min`, and `max` are all in ms. `comp` is the relative average and standard deviation compared to the lowest average.
`old operator` is the operator pattern before this PR, `new operator` is the operator pattern after this PR sans `:`, and `flat operator` is just all operators as a keyword list of sorts.
`prism-core.js` is what it sounds like from master, and `prism.js` is the current Prism version from the download page will all languages included.

<details>
<summary>Testing code</summary>

```js
const fs = require("fs");
const { performance } = require("perf_hooks");
// force the lazy init to happen
performance.now();


const samples = fs.readdirSync('./samples').filter(f => !/^_/.test(f)).map(f => {
	return {
		title: f,
		value: fs.readFileSync('./samples/' + f, 'utf8')
	}
});

benchmark(samples, [
	toTestCase(
		"old operator",
		/-[-=]?|\+[+=]?|!=?=?|<<?=?|>>?>?=?|=(?:==?|>)?|&[&=]?|\|[|=]?|\*\*?=?|\/=?|~|\^=?|%=?|\?|\.{3}/g
	),
	toTestCase(
		"new operator",
		/--|\+\+|\*\*=?|=>|&&|\|\||[!=]==|<<=?|>>>?=?|[-+*/%&|^!=<>]=?|\.{3}|[~?]/g
	),
	toTestCase(
		"flat operator",
		/--|-=|-|\+\+|\+=|\+|!==|!=|!|<<=|<<|<=|<|>>>=|>>>|>>=|>>|>=|>|===|==|=>|=|&&|&=|&|\|\||\|=|\||\*\*=|\*\*|\*=|\*|\/=|\/|~|\^=|\^|%=|%|\?|\.\.\./g
	),
]);

/**
 *
 * @param {string} title
 * @param {RegExp} regex
 */
function toTestCase(title, regex) {
	return {
		title,
		value: text => allMatches(regex, text),
	};
}


/**
 *
 * @param {{ title: string; value: T }[]} testSamples
 * @param {{ title: string; value: (value: T) => void}[]} testFunctions
 * @template T
 */
function benchmark(testSamples, testFunctions) {
	const warmupIterations = 1_000;
	const warmupMaxTime = 200;

	const warmupSample = testSamples[0].value;
	testFunctions.forEach(({ value }) => {
		const start = performance.now();
		for (let i = 0; i < warmupIterations; i++) {
			if (performance.now() - start > warmupMaxTime) {
				break;
			}
			value(warmupSample);
		}
	});

	const maxSamples = 10_000; // samples
	const maxTimePerFunc = 2_000; // ms


	const columnWidth = [
		Math.max(...testSamples.map(s => s.title.length)),
		Math.max(...testFunctions.map(s => s.title.length)),
	];

	function printLine(...cells) {
		let s = "";
		for (let i = 0; i < cells.length; i++) {
			let cell = String(cells[i]);
			if (i < columnWidth.length) {
				cell = cell.padEnd(columnWidth[i], " ");
			}
			if (s) {
				s += " | "
			}
			s += cell;
		}
		console.log(s);
	}

	let headerPrinted = false;

	/**
	 * @param {{ title: string; value: T }} sample
	 * @param {{ title: string; value: (value: T) => void}} func
	 * @param {Object<string, any>} result
	 */
	function printResult(sample, func, result) {
		if (!headerPrinted) {
			columnWidth.push(...Object.keys(result).map(k =>
				Math.max(10, k.length, Math.ceil(String(result[k]).length * 1.25))
			));
			printLine("sample", "func", ...Object.keys(result));
			printLine(...columnWidth.map(i => "-".repeat(i)));
			headerPrinted = true;
		}
		const cells = [
			func === testFunctions[0] ? sample.title : "",
			func.title
		];

		for (const key in result) {
			cells.push(result[key]);
		}
		printLine(...cells);
	}

	/**
	 * @param {Object<string, any>[]} results
	 */
	function printResults(sample, results) {
		const min = Math.min(...results.map(r => r.comp[0]));
		results.forEach(r => {
			const [avg, dev] = r.comp;
			r.comp = `${Math.round(100 * avg / min)}% +-${Math.round(100 * dev / min)}%`;
		});
		results.forEach((r, i) => {
			printResult(sample, testFunctions[i], r);
		})
	}

	for (const sample of testSamples) {
		const results = [];
		for (const func of testFunctions) {
			/** @type {number[]} */
			let samples = [];
			const samplingStart = performance.now();

			for (let i = 0; i < maxSamples; i++) {
				if (performance.now() - samplingStart > maxTimePerFunc) {
					break;
				}
				global.gc && global.gc();

				let start = performance.now();
				func.value(sample.value);
				samples.push(performance.now() - start);
			}

			const min = Math.min(...samples);
			const max = Math.max(...samples);
			const avg = samples.reduce((x, y) => x + y) / samples.length;
			const dev = Math.sqrt(samples.reduce((s, x) => s + (x - avg) ** 2) / (samples.length - 1))


			const prec = (n = 0) => n.toPrecision(4);

			results.push({
				comp: [avg, dev],
				avg: prec(avg),
				dev: prec(dev),
				min: prec(min),
				max: prec(max),
				samples: samples.length,
			});
		}
		printResults(sample, results);
	}
}

function allMatches(re, text) {
	if (!re.global) {
		throw new Error(`RegExp has to be global! ${re}`);
	}
	re.lastIndex = 0;
	while (re.exec(text)) { }
}
```

Run using Node.js v12.11.0 with the following command: `node --expose-gc bench.js`.
I take control over the GC (as much as I can) to avoid minor GCs during testing.

[bench.zip](https://github.com/PrismJS/prism/files/3678211/bench.zip)

</details>

Backtracking optimization or not doesn't seem to matter. 
While your pattern generally seems to be just a tat faster than the other ones (like 1% faster), it's hard to tell because of the noise.

To reduce this noise, I experimented a little with v8's GC and got it down by ~10% but there's still too much to conclusively say that your pattern is faster.

Golmote:
@RunDevelopment Nice! :open_mouth:

The results do not confirm my theory that much. I guess the engine optimizes those patterns indeed!

The conclusion seems to be the same in Firefox too.

PR #2074, opened 2019-09-26
Added new Not all plugin

### LINKED ISSUE #2068: Exclude certain languages from highlighting altogether
**Motivation**
Prism's current behavior makes it impossible to use Prism and Mermaid on the same page.

**Description**
I'd like to have what the title in #558 says ("Ignore some html tags") and not what the provided solution does ("Preserve markup even after Prism has applied syntax highlighting"). The issue I'm facing is having both [Mermaid](https://github.com/knsv/mermaid) and Prism on the same page and Prism just destroys Mermaid code blocks so Mermaid can't do anything sensible with it. Instead of preserving markup (which I suppose will also "fix" the problem), I'd like Prism to just not touch `.language-mermaid` elements. Is this not possible?

**Alternatives**
I've searched around and found gaearon/gitbook-plugin-prism#28 which proves that there's a need for this feature and I can't think of a way to work around this. Had there only been a `Prism.highlightElements()` method that took a list of elements I could select and filter beforehand manually, I could have solved this myself. But since this method doesn't exist, I'm unable to fix this without learning how to develop a Prism plugin of some sort. And I find the Prism API documentation to be very sparse (the available `hooks` don't seem to be documented, for instance), I'm struggling to do so. Please help! 🙏 😃

RunDevelopment:
This adds a new plugin Not all with allows to filter out certain elements the `highlightAll` and `highlightAllUnder` methods query.

This adds a new `filter` property to the env of `before-highlightall` which can be used to add filters. I named it `filter` because it very much behaves like `Array#filter`.

This also exposes `getLanguage` as `Prism.util.getLanguage`.

---

@asbjornu
My attempt to fix #2068.


asbjornu:
Thanks for this, @RunDevelopment! Much appreciated! 🙏

asbjornu:
I just thought that it would be completely awesome if this was possible to configure in a `data-` attribute on the `<script>` element, so no custom JavaScript was necessary and the `data-manual` attribute could be removed.

RunDevelopment:
Done. Now you can do the following:

```html
<script src="prism.js" data-exclude-css=".no-highlight, .no-highlight *"></script>
```
This will exclude all element with a `.no-highlighting` class and all elements in a element with a `.no-highlighting` class.

As for the original issue: You can either use `filterKnown`:

```html
<script src="prism.js" data-filter-known></script>
```

Or define a filter function:

```html
<script>
Prism.plugins.notAll.filter(({ language }) => language !== 'mermaid');
</script>
```

asbjornu:
@RunDevelopment, Fantastic! 🎉  Thank you so much! 🙏

components.json
```diff
@@ -1196,6 +1196,12 @@
 			"description": "Highlights the code inside diff blocks.",
 			"owner": "RunDevelopment",
 			"require": "diff"
+		},
+		"not-all": {
```
mAAdhaTTah: Some thoughts on the name:

* `filter-elements`
* `modify-highlighted`
* `filter-highlighted`

Thoughts?

components/prism-core.js
```diff
@@ -234,23 +234,28 @@ var _ = {
...
 		_.hooks.run('before-highlightall', env);
 
 		var elements = container.querySelectorAll(env.selector);
 
-		for (var i=0, element; element = elements[i++];) {
-			_.highlightElement(element, async === true, env.callback);
+		for (var i = 0, element; element = elements[i++];) {
+			if (filters.every(function (f) { return f(element); })) {
+				_.highlightElement(element, async === true, env.callback);
+			}
```
mAAdhaTTah: Instead of modifying the internal logic of Prism core to enable this, I think we should take advantage of Prism's built-in hooks instead. Couple of ideas for how to implement this:

1. Modify the actual API/goal of the plugin to be about modifying the `selector` to `not()` classes we want to ignore from the query.
  * Pro: smaller, more focused plugin, no internal changes to core.
  * Con: less flexibility
2. Move the `before-highlight` to after the query, include `elements` on `env`, and hook the plugin there to modify the elements.
  * Pro: take advantage of what's there, minimal changes
  * Con: changes the behavior of this hook, as people could use it to modify `selector`, could be breaking.
3. Add a new hook after the query & modify `elements`
  * Pro: take advantage of what's there, minimal changes
  * Con: two similar hooks in one part of the code that are only marginally different

Thoughts?

plugins/not-all/prism-not-all.js
```diff
@@ -0,0 +1,92 @@
...
+			config.filterCss(attr);
+		}
+		if (attr = script.getAttribute('data-exclude-css')) {
+			config.excludeCss(attr);
+		}
+	}
+
+	Prism.hooks.add('before-highlightall', function (env) {
+		for (var i = 0, l = filters.length; i < l; i++) {
+			env.filter(filters[i]);
```
mAAdhaTTah: Looking at the API from this perspective, `filter` doesn't really do what I expect it to do. Since I associate it with filtering an array, not seeing something get filtered here is a bit odd.

mAAdhaTTah:
Sorry I took so long to get to this. Thanks for your patience!

plugins/not-all/prism-not-all.js
RunDevelopment: That's true. I wanted to name it after Array methods, but `addFilter` would be a better name.

components.json
RunDevelopment: I want to have _all_ in the name, so people associate it with `highlightAll`.

- `filter-all-elements`
- `filter-highlight-all`

(Also, no past tense (`filter-highlighted`). They're not highlighted yet ;) )

components/prism-core.js
RunDevelopment: Maybe a bit of explanation what my goals are for this plugin:

Its purpose is to prevent elements from being highlighted by `highlightAll` or `highlightAllUnder`. So it's a mechanism to enable more fine grained control over which elements are being highlighted by Prism's `highlightAll*` methods.
This is plugin is supposed to have zero effect on elements passed to `highlightElement`. So even elements which are excluded by `highlightAll` and `highlightAllUnder`, will be highlighted. 

__Option 1__

I suppose you mean to use the CSS `:not()` selector? That's not possible because the `:not` is a component selector and has to be applied on a component basis. So we would have to _parse_ the selector and apply the `not`. It's also very limited as something like the inheritance Prism sometimes uses (e.g. language classes) can't be implemented.

Examples: Assume that the selector is `code[class*="language-"], [class*="language-"] code` and we want to exclude all elements which have the class `.exclude`.
The correct selector would be 

    code[class*="language-"]:not(.exclude), [class*="language-"] code:not(.exclude)

This is not trivial and requires that we parse the selector to some extend. Even worse: some operation are simply impossible.
Let's say that we want to exclude all elements have are descendants of an element with the class `no-highlight`. It's impossible to do because of CSS's greedy selector matching.

- `:not(.no-highlight) code` doesn't work
- `code(:not(.no-highlight) *)` not a valid component selector

The basic problem with the "modify the selector" approach is that CSS selectors only provide a logical OR for selectors and logical NOT and AND for components. We would have to implement the selector AND and NOT ourselves which, in the case of the selector NOT, is not possible as a CSS selector.

__Option 2__

It is [breaking](https://github.com/PrismJS/prism/blob/dab7998e5d9a8b7acbff5be05058a9829a0d084a/plugins/unescaped-markup/prism-unescaped-markup.js#L9-L12).

__Option 3__

Something like this, right?

<details>

```js
highlightAllUnder: function(container, async, callback) {
	var env = {
		callback: callback,
		selector: 'code[class*="language-"], [class*="language-"] code, code[class*="lang-"], [class*="lang-"] code'
	};

	_.hooks.run('before-highlightall', env);

	env.elements = [...container.querySelectorAll(env.selector)];

	_.hooks.run('before-highlightall-name-pending', env);

	for (var i = 0, element; element = env.elements[i++];) {
		_.highlightElement(element, async === true, env.callback);
	}
},
```

</details>

The only problem I see is that the `selector` is basically useless if plugins can just add elements as they like in the second hook. This also means that filtering elements becomes a pain because you have to keep the order of the hook listeners in mind (because other listeners can just add elements). But we can't enforce that element array after the hook has to be a subset of the element array before the hook, can we?

Simple filter functions don't have this problem. But I understand that the filter chaining I implemented comes quite close to a hook

mAAdhaTTah: > So we would have to parse the selector and apply the `not`.

Yeah, at the time, I was just looking at the main selector, thinking it might not be too difficult to `split(',')` then map over & append `not` as needed. But thinking about it now, that's definitely not doable.

---

If it's breaking, then option 2 is out as well.

---

Yeah, the option 3 example is definitely what I was thinking.

>  is basically useless if plugins can just add elements as they like in the second hook.

I'd almost consider this a feature? If people want to be able to add their own elements, they'll now be able to without modifying `selector`, which _could_ be easier. I don't know that we need to do that enforcement–the responsibility is on the plugin (developer) to ensure that the elements in the array are what's desired after the hooks runs. You do have the issue of ordering, if one plugin is adding and another is removing elements, but that seems like a problem we could address if it comes up (e.g. hook priorities?)

Ultimately, my thought is just that using a hook here, rather than modifying core with the specific logic, is more flexible for people and seems more like "the Prism way" of doing things.

RunDevelopment: > the responsibility is on the plugin (developer) to ensure that the elements in the array are what's desired after the hooks runs

That I can get behind.

I guess we can just ignore the ordering issue for now as there's only going to be one plugin that uses this hook in the near future. If someone really wants to get their elements filtered by this plugin, then they can just give their elements ids and add those ids to the selector in `before-highlightall`.

It just feels weird that you can add whatever elements you want in the new hook. 
On that note: We should expose the container.

RunDevelopment: I just need a name for the new hook...

`before-highlightall-elements`?

Thoughts?

plugins/not-all/prism-not-all.js
RunDevelopment: I'll close this now because of the new hook.

components.json
mAAdhaTTah: `filter-highlight-all` works for me!

components/prism-core.js
mAAdhaTTah: > It just feels weird that you can add whatever elements you want in the new hook.

Honestly, this doesn't really bother me.

> On that note: We should expose the container.

The "container" being `document`?

RunDevelopment: The container being whatever you give to `highlightAllUnder`.

components/prism-core.js
```diff
@@ -236,21 +236,24 @@ var _ = {
...
+			container: container,
 			selector: 'code[class*="language-"], [class*="language-"] code, code[class*="lang-"], [class*="lang-"] code'
 		};
 
 		_.hooks.run('before-highlightall', env);
 
-		var elements = container.querySelectorAll(env.selector);
+		env.elements = Array.prototype.slice.apply(container.querySelectorAll(env.selector));
+
+		_.hooks.run('before-highlightall-name-pending', env);
```
mAAdhaTTah: `after-elements-query`?

components/prism-core.js
```diff
@@ -236,21 +236,24 @@ var _ = {
...
 		var env = {
 			callback: callback,
+			container: container,
 			selector: 'code[class*="language-"], [class*="language-"] code, code[class*="lang-"], [class*="lang-"] code'
 		};
 
 		_.hooks.run('before-highlightall', env);
 
-		var elements = container.querySelectorAll(env.selector);
+		env.elements = Array.prototype.slice.apply(container.querySelectorAll(env.selector));
```
mAAdhaTTah: Presumably, we're making `container` available on `env` so users can modify it via this hook.

```suggestion
		env.elements = Array.prototype.slice.apply(env.container.querySelectorAll(env.selector));
```

plugins/filter-highlight-all/prism-filter-highlight-all.js
```diff
@@ -0,0 +1,106 @@
...
+
+	var config = Prism.plugins.filterHighlightAll = {
+
+		/**
+		 * Filters the elements of `highlightAll` and `highlightAllUnder` such that only element for which the given
+		 * function returns `true` will be highlighted.
+		 *
+		 * @param {(value: { element: HTMLElement, language: string }) => boolean} condition
+		 */
+		filter: function (condition) {
```
mAAdhaTTah: `addFilter`?

plugins/filter-highlight-all/prism-filter-highlight-all.js
```diff
@@ -0,0 +1,106 @@
...
+			});
+		},
+
+		/**
+		 * Filters the elements of `highlightAll` and `highlightAllUnder` such that only element which match the given
+		 * CSS selection will be highlighted.
+		 *
+		 * @param {string} selector
+		 */
+		filterCss: function (selector) {
```
mAAdhaTTah: `addCssFilter`?

mAAdhaTTah: Hmmm, actually, this name is fine.

mAAdhaTTah: Or maybe `filterSelector`?

plugins/filter-highlight-all/prism-filter-highlight-all.js
```diff
@@ -0,0 +1,106 @@
...
+			});
+		},
+
+		/**
+		 * Filters the elements of `highlightAll` and `highlightAllUnder` such that only element which do not match the
+		 * given CSS selection will be highlighted.
+		 *
+		 * @param {string} selector
+		 */
+		excludeCss: function (selector) {
```
mAAdhaTTah: Then this one could be `rejectSelector`? And/or make the above one `acceptSelector`? `filterSelector` would be ok, cuz that does match the semantics of the array method `filter`.

mAAdhaTTah:
Just need to bikeshed some naming here but the code itself is great! Thanks for your patience.

components/prism-core.js
RunDevelopment: Still want that `all` in there.

`after-highlightall-elements-query`?

(Seems like [Java enterprise](https://projects.haykranen.nl/java/) has left its marks on me.)

plugins/filter-highlight-all/prism-filter-highlight-all.js
RunDevelopment: I don't like the "accept"/"reject"-pair because it reminds me of network requests.
I want to keep the "filter" but you're right that "filter" and "exclude" aren't exactly opposites.

Or what about "matchesSelector" and "notMatchesSelector"? Quite verbose but clear.

The only thing now, that isn't clear from the name is that all filters and combined with a logical AND. But that should be in the doc.

mAAdhaTTah: I was thinking about it as the array functions:

https://lodash.com/docs/4.17.15#filter
https://lodash.com/docs/4.17.15#reject

Hence `filterSelector` & `rejectSelector`.

`includeSelector` & `excludeSelector` could also work.

RunDevelopment: `filterSelector` and `rejectSelector` sounds good to me.

plugins/filter-highlight-all/prism-filter-highlight-all.js
RunDevelopment: I think it's enough to know that this function adds a filter. If the name should also reflect this behaviour, then we'd have to rename the selector functions as well.

components/prism-core.js
mAAdhaTTah: Technically, the hook is in `highlightUnder`... `after-highlight-elements-query` would work for me. I don't think it needs the `all`, tbh, as it's not specific to the plugin, so doesn't need the name to match the same way.

plugins/filter-highlight-all/prism-filter-highlight-all.js
mAAdhaTTah: I'd almost go the other way and suggest then we should make all of them `addX`. I could also conceivably see `Prism.plugins.filterHighlightAll.add.{filter,selector,etc.}` as the API, making the `add` a namespace object, but that might be a bit...much.

components/prism-core.js
RunDevelopment: It's in `highlightAllUnder`. Also, we have the method `highlightElement` which makes the name `after-highlight-elements-query` just a tad confusing IMO.

Instead, we could choose a name that describes that this hook is the last one executed before all elements are iteratively highlighted.

`before-all-elements` ? (but this is kinda missing the verb/action that is done afterward)

plugins/filter-highlight-all/prism-filter-highlight-all.js
RunDevelopment: The naming problem I see with the namespace idea is that current verbs will become nouns, e.g. `filterSelector` -> `add.selectorFilter`. This is fine for "filter" but for "reject"... `rejectSelector` -> `add.selectorRejecter` ??

Also, I think it's enough to just say that these functions add a new filter.

components/prism-core.js
mAAdhaTTah: `before-all-elements-highlighted`? Naming things is hard...

plugins/filter-highlight-all/prism-filter-highlight-all.js
mAAdhaTTah: I was thinking `addFilterSelector` and `addRejectSelector` for both of those, as we're adding a "reject selector" (a selector that rejects). If I'm being pedantic, the function it creates is the "rejector".

My overall issue with the naming here is I'd prefer these method names not match array methods. The `add` prefix indicates that we're adding a thing that the plugin is holding onto and running later, rather than running the filter immediately.

components/prism-core.js
RunDevelopment: Let's just roll with this for now. 
~Nothing is more permanent than a temporary solution.~ Maybe we'll find something better later on.

RunDevelopment: `*-highlighted` or `*-highlight` ?

plugins/filter-highlight-all/prism-filter-highlight-all.js
RunDevelopment: How about `add` (current `filter`), `addSelector`, `reject.add` (new), `reject.addSelector` ?
"Filter" is in the plugin's name so maybe we don't have to repeat it?

I would still keep the current `data-filter-selector` and `data-reject-selector` attributes.

mAAdhaTTah: This works for me!

components/prism-core.js
mAAdhaTTah: `*-highlight` is more in line with the naming conventions of other hooks. None of them use past tense.

RunDevelopment:
Should be good to go now.

PR #2075, opened 2019-09-27
Custom Class: New class adder feature

RunDevelopment:
This adds a function to let the user define custom classes for specific tokens. Example:

```js
Prism.plugins.customClass.add(function ({content, type, language}) {
	if (content === ':' && type === 'punctuation' && language === 'javascript') {
		return 'colon';
	}
});
```

(You can also define multiple such functions.)

The whole thing is implemented via the `wrap` hook, so `content` is the innerHTML of the token. I choose this over `before-tokenize` because for users innerHTML is easier to understand than the Prism token stream.

### Further considerations

This feature is kind of a superset of [Highlight keywords](https://prismjs.com/plugins/highlight-keywords/). 
So maybe we could ~combine the two~ _edit:_ merge HK into CC deprecate HK?

---

@MattMcFarland I hope this helps.


MattMcFarland:
Looks promising!

mAAdhaTTah:
My apologies–could you remind me what sparked this PR?

RunDevelopment:
The idea was some people want to selectively highlight certain tokens differently from what we do.

In the above example (which was pre #2073), the colon `:` should be highlighted like an operator in JS. With this PR, people can do it without changing the language definition.
Another example is how #1743 includes keyword groupings and special names for certain operators for more granular highlighting.

The main purpose is to give the designers of themes more freedom, so they aren't limited to the way we choose to tokenize. Of course, you can also achieve this by modifying the grammar but that's a lot harder.

plugins/custom-class/index.html
```diff
@@ -61,6 +64,21 @@ <h2>2. Replace some Prism classes with your defined ones via an object</h2>
 	}
 });</code></pre>
 
+	<h2>Add new classes</h2>
+
+	<p>You can add new classes with per-token and per-language precision.</p>
+
+	<pre class="language-js"><code>Prism.plugins.customClass.add(function (content, type, language) {
```
mAAdhaTTah: Looking at this API, I might suggest content go last, cuz its the least likely to be used by any particular implementation.

plugins/custom-class/index.html
```diff
@@ -96,26 +114,61 @@ <h2>CSS Modules Usage:</h2>
...
+	<p>Input</p>
+	<pre class="language-markup"><code>&lt;pre class=&quot;language-css&quot;&gt;&lt;code&gt;
+a::after {
+	content: '\2b00 ';
+	opacity: .7;
+}
+&lt;/code&gt;&lt;/pre&gt;</code></pre>
+
+	<p>Options</p>
+	<pre class="language-js"><code>Prism.plugins.customClass.add(function (language, type, content) {
```
mAAdhaTTah: `content` is last here...?

plugins/custom-class/index.html
mAAdhaTTah: To be clear, I think `type, language, content` is the preferred order.

RunDevelopment: I actually changed the API from `language, type, content` to `content, type, language` in the last commit because I think it's the most useful one. Everything you can do with only `language` and `type`, you can also do in CSS (styling-wise), so content is the feature that enables you to do more.
Why do you think it's the least useful one?

> we could pass it in as an object with the three properties [from your other comment]

I like that. Makes it feel like a hook.

RunDevelopment:
Also, regarding the API:
I don't like that `prefix` and `map` _set_ the current prefix or map but `add` _adds_ another mapper. So either we change the name to indicate this difference or we change the behaviour. 
Thoughts?

mAAdhaTTah:
Yeah, it kinda feels like `map` is a subset of what `add` does. Can we change `map` to use `add` such that it always adds a new mapper?

RunDevelopment:
No, because:

1) rn, `add` can't modify existing classes. It doesn't even see them.
2) there is no method to remove the functions passed to `add`, so we can't overwrite the mapper.

mAAdhaTTah:
Ok... maybe it makes more sense for `add` to just take a single function then, and overwrite the previous one? Since we didn't see a need for making the first 2 take multiple functions, and we haven't gotten complaints about it, it makes sense to keep them aligned by just having the single `add` function.

RunDevelopment:
Sounds good, but this won't make `add` a superset of `map`. Is that still ok?

mAAdhaTTah:
Yeah, I don't think that's a necessary goal at this point. At least this makes the behavior consistent.

RunDevelopment:
Sorry for the late responses. I've been a little busy lately.

The only things stopping v1.18 are #1998 and #2074. TBH, #2074 modifies and isn't as important as #1998, so I'm fine with leaving it out of this release if it'll mean that we can release faster.

I'll do the changelog in the meantime.

plugins/custom-class/prism-custom-class.js
```diff
@@ -72,31 +69,26 @@
...
 		}
 
-		if (classMapper === defaultClassMap && !prefixString) {
+		if (!mapper && !prefixString) {
 			return;
 		}
 
 		env.classes = env.classes.map(function (c) {
-			return prefixString + classMapper(c, env.language);
+			return prefixString + mapper(c, env.language);
```
mAAdhaTTah: I think this still needs a check for `mapper`'s existence. The check above only short-circuits if _both_ are undefined. But we could have no mapper but a prefix and we'll need to check this here.

Also, in reverse, if `prefixString` is undefined when we do this, `undefined` will get added as a prefix, so it also needs to be checked here.

RunDevelopment: Ahh, right. My bad.

PR #2079, opened 2019-09-29
Added SQF language

### LINKED ISSUE #2067: SQF
**Language**
SQF is used for modding the game Arma 3.

**Additional resources**
https://github.com/highlightjs/highlight.js/blob/master/src/languages/sqf.js

RunDevelopment:
This resolves #2067.

There is no spec for this language, but there is a [wiki](https://community.bistudio.com/wiki/SQF_syntax) from the Amra 3 devs.
Here are the most relevant pages: [syntax](https://community.bistudio.com/wiki/SQF_syntax), [numbers](https://community.bistudio.com/wiki/Number), [strings](https://community.bistudio.com/wiki/String), [control structures](https://community.bistudio.com/wiki/Control_Structures), [commands](https://community.bistudio.com/wiki/All_Arma_Commands_Pages), [functions](https://community.bistudio.com/wiki/Function), [operators](https://community.bistudio.com/wiki/Operators), [data types](https://community.bistudio.com/wiki/Data_Types), [preprocessing](https://community.bistudio.com/wiki/PreProcessor_Commands), and [examples](https://community.bistudio.com/wiki/Category:Example_Code).

I wasn't able to find any copyright-free SQF code and from my understanding, the entire wiki is owned by Bohemia Interactive. So if somebody has an example, I can use, please tell me.

### Example

<details>

image

[source](https://community.bistudio.com/wiki/Example_Code:_Reveal_Units_Temporarily)

</details>


tests/languages/sqf/magic-variable_feature.test
```diff
@@ -0,0 +1,35 @@
+_exception
+_fnc_scriptName
+_fnc_scriptNameParent
+_forEachIndex
+_this
+_thisEventHandler
+_thisFSM
+_thisScript
+_time
```
BaerMitUmlaut: `_time` is not a magic variable in SQF, only in SQS.

jonpas:
You may use anything from https://github.com/CBATeam/CBA_A3 or https://github.com/acemod/ACE3 for examples. :)

In addition, GitHub's linguist syntax for SQF may be taken as a reference to keep it consistent. Source for it can be found at https://github.com/JonBons/Sublime-SQF-Language.

Very nice to see this! It was brought up in ACE Slack and I am sure some developers will comment on here and help out!

RunDevelopment:
@jonpas and @BaerMitUmlaut Thank you for reviewing my PR and given me such fine example code!

I actually used https://github.com/JonBons/Sublime-SQF-Language as reference, so this shouldn't stray too far from it.

components/prism-sqf.js
```diff
@@ -0,0 +1,33 @@
+Prism.languages.sqf = Prism.languages.extend('clike', {
+	'string': {
+		pattern: /"(?:(?:"")?[^"])*"(?!")|'(?:[^'])*'/,
+		greedy: true
+	},
+	'keyword': /\b(?:call|case|catch|default|do|echo|else|endsWith|for|forEach|from|goto|if|nil|private|step|switch|then|to|try|while)\b/,
```
mharis001:
```suggestion
	'keyword': /\b(?:call|case|catch|default|do|echo|else|exitWith|for|forEach|from|goto|if|nil|private|step|switch|then|to|try|while)\b/,
```

Should also be removed from the "function" list below as I think this should be a keyword (like JonBons/Sublime-SQF-Language).

mharis001: Should also move `with` to "keyword" list (from "function" list). Sublime-SQF-Language considers it a keyword.

RunDevelopment: Yes, definitely. Seems like I simply mistyped `exitWith`.

RunDevelopment: Agreed.

RunDevelopment:
I have a few questions regarding keywords:

1. Keywords are case-insensitive `(?i:`. Is that true?
2. What are all these keywords of the form `DIK(?:_[A-Z\d])+` like `DIK_0|DIK_1|DIK_APPS|DIK_BACKSLASH|DIK_CAPITAL`?
3. I went through the [whole keyword list](https://github.com/JonBons/Sublime-SQF-Language/blob/36585fa5e5279df5b8e13bad6ddafc1221b20d07/SQF.tmLanguage#L22) and apart from point 2., these are the only ones not being keywords right now: `forEachMember|forEachMemberAgent|forEachMemberTeam|spawn|preprocessFile|preprocessFileLineNumbers|execVM|execFSM`. None of them seem to function as keywords, so we might as well let them be `function`s. Thoughts?

BaerMitUmlaut:
1. Yes, most of SQF is case insensitive. There are a couple exceptions with string comparisons, but that's irrelevant for highlighting.
2. Keycode macros. DIK stands for direct input keycode.
3. In reality, SQF only has one keyword, which is `private`. `if`, `then`, `for` etc. are actually commands just like `setPos`. However, commands that are used for control flow are usually highlighted as keywords for readability and because they're treated as keywords by most SQF scripters.

jonpas:
3. What @BaerMitUmlaut said, I think we should just keep consistent with https://github.com/JonBons/Sublime-SQF-Language on keywords/functions.

RunDevelopment:
1. Ok. It wasn't specified in the wiki, so I assumed it's case-sensitive.
2. Added as `constant`.
3. @jonpas I would actually like to keep the current keyword list and I don't think we should have https://github.com/JonBons/Sublime-SQF-Language make the final decision here. It's not even sure whether `execFSM` should be a keyword or a function (it's included in both lists). (Also, if `execVM` and `execFSM` are keywords, then why isn't `execEditorScript`? Btw. `exec` is missing completely.)
    My point is that we should highlight functions as keywords because they function as keywords and not because they are included in [this list](https://github.com/JonBons/Sublime-SQF-Language/blob/36585fa5e5279df5b8e13bad6ddafc1221b20d07/SQF.tmLanguage#L22).
    Also, some of them seem oddly specific to be keyword... `preprocessFileLineNumbers`.
    Then again, I never even wrote a single line of code in SQF, so maybe I shouldn't decide this. 
    Further input would be very welcome.

jonpas:
It is a grey area, neither of those are actually functions nor keywords (macros, which are all upper-case are), but are scripting commands, functions is something you can write with SQF, such as `ace_common_fnc_banana` or `BIS_fnc_camera`.

I agree that Sublime-SQF-Language is not perfect either, maybe @JonBons can pitch in and we can streamline both.

BaerMitUmlaut:
>Also, if `execVM` and `execFSM` are keywords, then why isn't `execEditorScript`?
Because that command doesn't do anything.

>Btw. `exec` is missing completely.
That one is used to run SQS scripts, which are basically deprecated since 2002. That one does actually do something though.

>I never even wrote a single line of code in SQF
You must think we're maniacs.

RunDevelopment:
Alright, I added the remaining "keywords".
Since, as @jonpas said, it's a grey area, so it shouldn't matter too much whether they are or not, so I'll just make them what you want.

mharis001:
The current list looks good to me 👍.

JonBons:
As for my repo, I generated the first version of it programmatically and I've never actually spent time cleaning it up but at the moment I don't have any time to take a look into it and I won't for the foreseeable future.

PR #2085, opened 2019-10-04
Minor improvements to SAS language support

cedporter:
* Reorder comment array to fix edge case added to comment_feature.test
* Adjust datalines to allow for possible preceding spaces before closing semicolon
* Move comment in sas object for correct syntax matching precedence
* Add several more keywords


components/prism-sas.js
```diff
@@ -15,11 +15,11 @@
 	};
 
 	var comment = [
+		/\/\*[\s\S]+?\*\//,
```
RunDevelopment: I just noticed this but isn't `/**/` also a valid comment?

components/prism-sas.js
```diff
@@ -167,7 +167,7 @@
...
 		'keyword': {
-			pattern: /((?:^|\s)=?)(?:action|after|analysis|and|array|barchart|barwidth|begingraph|by|cas|cbarline|cfill|close|column|computed?|contains|data(?=\=)|define|document|do\s+over|do|dol|drop|dul|end|entryTitle|else|endcomp|fill(?:attrs)?|filename|group(?:by)?|headline|headskip|histogram|if|infile|keep|label|layout|legendlabel|length|libname|merge|midpoints|name|noobs|nowd|ods|options|or|out(?:put)?|overlay|plot|ranexp|rannor|rbreak|retain|set|session|sessref|statgraph|sum|summarize|table|temp|then\sdo|then|title\d?|to|var|where|xaxisopts|yaxisopts|y2axisopts)\b/i,
+			pattern: /((?:^|\s)=?)(?:action|after|analysis|and|array|barchart|barwidth|begingraph|by|cas|cbarline|cfill|class(?:lev)?|close|column|computed?|contains|data(?=\=)|define|document|do\s+over|do|dol|drop|dul|end|entryTitle|else|endcomp|fill(?:attrs)?|filename|group(?:by)?|headline|headskip|histogram|if|infile|keep|keylabel|keyword|label|layout|legendlabel|length|libname|merge|midpoints|name|noobs|nowd|ods|options|or|out(?:put)?|overlay|plot|ranexp|rannor|rbreak|retain|set|session|sessref|statgraph|sum|summarize|table|temp|then\sdo|then|title\d?|to|var|where|xaxisopts|yaxisopts|y2axisopts)\b/i,
```
RunDevelopment: Shouldn't `then\sdo` be `then\s+do` like `do\s+over`?

components/prism-sas.js
```diff
@@ -53,7 +53,7 @@
 
 	Prism.languages.sas = {
 		'datalines': {
-			pattern: /^(\s*)(?:(?:data)?lines|cards);[\s\S]+?^;/im,
+			pattern: /^(\s*)(?:(?:data)?lines|cards);[\s\S]+?^(\s*);/im,
```
RunDevelopment: 1. Please don't use capturing groups.
2. You don't need the group.

```suggestion
			pattern: /^(\s*)(?:(?:data)?lines|cards);[\s\S]+?^\s*;/im,
```

components/prism-sas.js
cedporter: Agreed. Fixed in latest commit.

components/prism-sas.js
cedporter: Fixed in latest commit.

components/prism-sas.js
cedporter: Fixed in latest commit.

PR #2091, opened 2019-10-16
Add support for Groovy and Lua code within PROC GROOVY and PROC LUA

cedporter:
Like PROC SQL, PROC GROOVY and PROC LUA both contain syntax that should be treated as the language specified by the PROC, not SAS. This pull request implements that highlighting.


RunDevelopment:
Thank you for this PR.
A few things before the review:

1. Please rebase this. I recently added a new test that points out all unnecessary capturing groups.
2. Should we make the dependencies optional? Some users might not need SQL, Lua, and Groovy highlighting.
   The only problem with optional dependencies is that the languages might be present when the SAS language is created (duh, optional deps), so the `rest` properties will cause trouble. So the Lua and Groovy `summit`s have to be replaced with something like this:
    ```js
    {
        'groovy': {
            pattern: RegExp(/(^[ \t]*submit(?:\s+(?:load|parseonly|norun))?)(?:<str>|[^"'])+?(?=endsubmit;)/.source.replace(/<str>/g, stringPattern), 'im'),
            lookbehind: true,
            alias: 'language-groovy',
            inside: Prism.languages.groovy
        },
        'submit-statement': submitStatement,
        'global-statements': globalStatements,
        'number': number,
        'numeric-constant': numericConstant,
        'punctuation': punctuation,
        'string': string
    }
    ```
    Similarlly for Lua.
    Also, I assume that it has to be `(?:<str>|[^"'])+?` instead of `(?:<str>|[^"'])+`, isn't it?

cedporter:
I made the changes specified above and dealt with the unnecessary capturing groups. How would I make Lua and Groovy optional?

RunDevelopment:
To make them optional, you change this in `components.json`:

```diff
-	"require": ["groovy", "lua", "sql"],
+	"peerDependencies": ["groovy", "lua", "sql"],
```

I think SQL should be optional as well or is it used very often so that including it by default is justified?

RunDevelopment:
Also, with Lua and Groovy (and maybe also SQL?) being optional, you have to change your test files.

To tell Prism's testing suite to load Lua/Groovy/SQL, you create a new folder in `test/languages/` and call it `lua+sas`. This will load first Lua and then SAS. Also, we usually call these kinds of test _inclusion_  tests ([Example](https://github.com/PrismJS/prism/blob/master/tests/languages/css%2Bhttp/css_inclusion.test) of CSS in HTTP).

cedporter:
I've made all three languages optional and updated the tests to reflect it.

components.json
```diff
@@ -851,7 +851,7 @@
 		},
 		"sas": {
 			"title": "SAS",
-			"require": "sql",
+			"peerDependencies": ["groovy","lua","sql"],
```
RunDevelopment: One last minor nit.

```suggestion
			"peerDependencies": ["groovy", "lua", "sql"],
```

PR #2092, opened 2019-10-17
Add include/extend/prepend to Ruby keywords

janko:
In Ruby, `include`, `extend` and `prepend` are treated as keywords (e.g. GitHub's syntax highlighter has that behaviour), so I thought it would be nice if Prism did this as well.

By the way, I'm generally impressed with Prism's Ruby tokenization, especially coming from highlight.js.


RunDevelopment:
Thank you for submitting this PR!

Please run `gulp` to generate the minified files and please add the new keywords to Ruby's [keyword test](https://github.com/PrismJS/prism/blob/master/tests/languages/ruby/keyword_feature.test).

RunDevelopment:
Ooof. Seems like npm installed a more recent version of UglifyJS for you.
Could you please tell me which version because npm only installs v3.5.2 for me. You can check by opening `path/to/prims/node_modules/uglify-js/package.json`. At the end of the file, there will a `version` property.

mAAdhaTTah:
@RunDevelopment `npm ls uglify-js` would be easier.

@janko Use `npm ci` to make sure you get the exact versions in the lockfile.

janko:
Thanks for the instructions. I've actually switched to Atom's [highlights](https://github.com/atom/highlights) package, which uses the "official" [Ruby grammar](https://github.com/textmate/ruby.tmbundle/blob/master/Syntaxes/Ruby.plist) (it's used by GitHub, Atom and VSCode). This grammar is more robust and highlights all Ruby code correctly for me, and is up-to-date with these new keywords.

PR #2096, opened 2019-10-18
Added support for BrightScript

### LINKED ISSUE #2037: Brightscript Syntax highlighting
**Brightscript**
BrightScript is a scripting language to build media and networked applications for embedded devices. Used by Roku devices.

**Refereces:**
https://developer.roku.com/es-mx/docs/references/brightscript/language/brightscript-language-reference.md

RunDevelopment:
This resolves #2037.

### Screenshot

image


tests/languages/brightscript/directive_feature.test
```diff
@@ -0,0 +1,67 @@
...
+
+[
+	["directive", [
+		["directive", "#const"],
+		["expression", [
+			" someFlag ",
+			["operator", "="],
+			["boolean", "true"]
+		]]
+	]],
```
mAAdhaTTah: Are these nested directives expected/necessary?

mAAdhaTTah:
Example HTML page needed.

tests/languages/brightscript/directive_feature.test
RunDevelopment: I could rename the inner one to `keyword` or do you mean that the wrapping is kinda useless because it contains a BightScript expression?

mAAdhaTTah: Well, it just seemed odd that a directive is nested inside a directive. It's even weirder further down where the only element inside the directive is another directive. I'm not sure whether that makes sense syntactically for the language, but maybe just dropping the outer wrapping directive would do it?

RunDevelopment: I did this to be consistent with other languages that have compiler directives like C/C++ (even though the tokens are called `macro` there).

mAAdhaTTah: Ok, that makes some sense, although if the languages aren't really related to each other, then it's not really necessary to make them consistent with each other.

RunDevelopment: It's usually nice to have a wrapper around these semantical units. The wrapper makes it easier to style thinks (some like to have compiler stuff in a different color) and right now, removing the wrapper will change the style because of the `property` alias.

Also, yes. They don't have to be consistent but it's nice that they are.

mAAdhaTTah: Ok, so we can leave the wrapper if you think that's helpful but maybe the wrapper should be a different name?

RunDevelopment: A different name sounds good. Any suggestions?

btw [this](https://developer.roku.com/es-mx/docs/references/brightscript/language/conditional-compilation.md) is the relevant part of the spec.

mAAdhaTTah: `directive-expression` or `directive-statement`? Based on the spec, they seem somewhat similar to `if` statements, but for the compiler.

RunDevelopment: I like `directive-statement`. Let's go with that.

Issue #2102, opened 2019-10-23
Defer breaks autoload (when loading page without having resources cached)


BrainStone:
**Information:**
- Prism version: 1.17.1
- Plugins: autoloader
- Environment: Browser

**Description**
When the page is loaded without the JS resources being cached and the scripts are marked with `defer` then the syntax highlighting is not applied.

**Example**
Without `defer` (workking): https://jsfiddle.net/f5vj903e/  
With `defer` (not working): https://jsfiddle.net/f5vj903e/1/

RunDevelopment:
Cannot reproduce.
What browser are you using?

BrainStone:
Very weird. I now also can't reproduce in a minimal example.  
Would it help if I linked you a live page where the issue appears?

Also I am using FireFox 70.0 (64-Bit)

BrainStone:
Well, in any case this is the page: https://dev.aura-dev.team/documentation/test  
Reloading without the cache breaks the page in funny new ways every time.

RunDevelopment:
Can we even fix this?
Right now, we use [these lines](https://github.com/PrismJS/prism/blob/3640b3f238a0b18eea8ab899526718e7dec4f162/plugins/autoloader/prism-autoloader.js#L185-L186) to get the script element which contains the Autoloader. Because of `defer`, the Autoloader script might get executed after some additional scripts have been loaded meaning that it isn't the last script element in the collection anymore.

We _could_ use `document.currentScript` but that isn't supported by IE11 and I don't want Autoloader behave differently depending on the browser. 
Any ideas @mAAdhaTTah @Golmote ?

RunDevelopment:
This is probably because all the different Prism files have to be loaded in a specific order and the `defer` creates a race condition. Unfortunately, you can't specify defer-dependencies, so you have to either bundle them into one file or load them synchronously.

BrainStone:
`defer` gurantuees that the scripts are executed in order.  
See https://www.growingwiththeweb.com/2014/02/async-vs-defer-attributes.html#script-defer for example

RunDevelopment:
> `defer` guarantees that the scripts are executed in order.

I didn't know that. That's for pointing that out. 

I can reproduce the issue you describe now albeit each with a low probability of occurrence. Everything seems to be working fine in Chrome and Edge, so maybe it's a FireFox issue?
Since the issues occur non-deterministically, I'm pretty sure that we have ourselves a race condition but I don't know where.

[...]

PR #2103, opened 2019-10-24
Core: Fixed race condition caused by deferring the script

RunDevelopment:
As discussed in #2102.

I'll make a different PR for the `currentScript` issue.


RunDevelopment:
You can verify that this indeed fixes the race condition by replacing `prism.min.js` in the [test page](https://github.com/PrismJS/prism/issues/2102#issuecomment-546022055) by @BrainStone with `<script defer src="http://localhost/prism.js"></script>` (localhost being this PR).

You'll see that the race condition is no more.

mAAdhaTTah:
Def going to want a code comment in there explaining a bit of why this is necessary.

RunDevelopment:
Comment as requested. 
I also added `readyState === 'interactive'` to the condition because if the `readyState === 'complete'` then the script can be deferred all it wants, `DOMContentLoaded` has already been fired.

PR #2104, opened 2019-10-24
Core: IE11 workaround for currentScript

### LINKED ISSUE #2102: Defer breaks autoload (when loading page without having resources cached)

RunDevelopment:
This fixes #2102.

The IE11 workaround was inspired by [this](https://github.com/amiller-gh/currentScript-polyfill/blob/master/currentScript.js).


BrainStone:
Are you sure #2103 needs fixing and not #2102? ;)

RunDevelopment:
Thanks for pointing that out 😅

components/prism-core.js
```diff
@@ -101,6 +101,46 @@ var _ = {
...
+			if ('currentScript' in document) {
+				return document.currentScript;
+			}
+
+			// IE11 workaround
+			// we'll get the src of the current script by parsing IE11's error stack trace
+			// this will not work for inline scripts
+
+			try {
+				throw new Error();
```
mAAdhaTTah: Do we need to throw the Error to get a stack trace?

components/prism-core.js
```diff
@@ -101,6 +101,46 @@ var _ = {
...
+				// Get file src url from stack. Specifically works with the format of stack traces in IE.
+				// A stack will look like this:
+				//
+				// Error
+				//    at _.util.currentScript (http://localhost/components/prism-core.js:119:5)
+				//    at Global code (http://localhost/components/prism-core.js:606:1)
+
+				var src = (/at [^(\r\n]*\((.*):.+:.+\)$/i.exec(err.stack) || [])[1];
+				var scripts = document.getElementsByTagName('script');
+				for (var i in scripts) {
```
mAAdhaTTah: We should probs wrap this and bail if `src` is undefined so we don't bother looping over all the scripts for no reason.

components/prism-core.js
```diff
@@ -527,11 +567,11 @@ if (!_self.document) {
 }
 
 //Get current script and highlight
-var script = document.currentScript || [].slice.call(document.getElementsByTagName('script')).pop();
```
mAAdhaTTah: Is it worth using this `pop` fallback if we don't find the current script in IE11?

components/prism-core.js
RunDevelopment: How else do you wanna do it?
Also, the stack trace is non-standard, so it's IE11 specific.

mAAdhaTTah: Oh, I just couldn't remember if just creating the error object captures the stack trace or if it needs to be thrown.

components/prism-core.js
RunDevelopment: The whole point of this PR is that this method doesn't work.

I think it's better to say that we couldn't find a `script` instead of returning an incorrect one.

components/prism-core.js
RunDevelopment: That's what you meant. I tested this and IE11 adds the `stack` property after its thrown.
Don't know how other browsers handle this though.

mAAdhaTTah: Fair enough. This is fine, since it's an IE11-specific workaround.

components/prism-core.js
mAAdhaTTah: Ok, I agree with that.

plugins/autoloader/prism-autoloader.js
```diff
@@ -182,8 +182,7 @@
 
 	var ignored_language = 'none';
 
-	var scripts = document.getElementsByTagName('script');
-	var script = scripts[scripts.length - 1];
+	var script = Prism.util.currentScript();
```
mAAdhaTTah: We need to check `script` exists in the usages below.

RunDevelopment: Is there a case where it doesn't exist?

mAAdhaTTah: Inline scripts? We check if `script` exists first in our other usages as well.

RunDevelopment: Well, true. My bad.

mAAdhaTTah: Maybe I've been working in TS for too long, but my thought is just that we know this function can return `null` so we should check before we interact with it.

RunDevelopment: Nah, it's alright.
Actually, my first through when changing this to `Prism.util.currentScript()` was: "TS won't approve of this".

muenzpraeger:
Hey - I wanted to use Prism embedded into a web component, but ran into the issue that t[his IE11 hack](https://github.com/PrismJS/prism/blob/master/prism.js#L210) is used. Specifically it is that in my environment `scripts[i]` can be `undefined`.

Any thoughts on accepting a PR that extends the `if` check on `undefined`? Happy to put that in.

RunDevelopment:
@muenzpraeger How can `scripts[i]` be `undefined`? `scripts` is a [HTMLCollection](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection), so iterating with `for in` should iterate over all indexes.

muenzpraeger:
It runs within a [secure context ](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/security_locker_service)of that environment.

RunDevelopment:
Ok, after a dive into the documentation, I found the issue. The "global" `document` object is a mock object that is an instance of the [`SecureDocument`](https://developer.salesforce.com/docs/component-library/tools/locker-service-viewer) interface (class?). This explains why the `currentScript` property is absent (it's listed as "not supported") but it doesn't explain why `getElementsByTagName` doesn't return an HTMLCollection since the function is listed as supported since v46. 

My conclusion is that this is a bug with `SecureDocument`. The `getElementsByTagName` function is supposed to return a value that fulfills the HTMLCollection interface but it doesn't. Please report this bug, so the authors of `SecureDocument` can fix it.

PR #2113, opened 2019-11-15
Rename robot-framework to robotframework

datakurre:
Support for highlighting Robot Framework plain text syntax was added in https://github.com/PrismJS/prism/pull/2034 and I'm so grateful for that.

Unfortunately, as my mistake, I forgot to mention that the canonical name for the language in various context has used to be ``robotframework`` withoud dash, not ``robot-framework``.

Since there is has not been release yet for prism since the addition of the language, it should be safe to still do the rename. I tried to the rename in this pull.


RunDevelopment:
LGTM;

Just the CI. I don't get why it failed. I'll look into it.

datakurre:
I might have been naive in just sed’ing and renaming files...

RunDevelopment:
Found it. Please run gulp. Uglify found some things to optimize.

Just run the following command and everything should be fine.
`npm ci && npx gulp` (this require NodeJS >= 8)

datakurre:
Thanks. I also found the tests and there is still places to fix...

datakurre:
Better now. Thanks for help.

RunDevelopment:
Thank you for contributing!

PR #2116, opened 2019-11-19
Add causejs to language definitions

bob2517:


bob2517:
I have no idea why the checks failed. All I did was add new language files (I think I did anyway - my second time using github ever). If anyone can shed any light on what is going on that would be awesome, thanks. Sorry to waste anyone's time if I've made a fundamental error.

bob2517:
Ignore last - failing on "1) – should pass test case 'all_features'". Sorting out.

bob2517:
Hooray! All checks passed. Travis is really good - not seen this in action before.

bob2517:
For some reason, the package-lock.json file has these extra lines which have been added in, eg:
+          "optional": true

I don't know enough to know what happened here. I have a feeling I shouldn't have checked this in. Will try and work out how to remove this change.

bob2517:
That should be it.

bob2517:
Note this language hasn't been released yet - it will be in December. It already works and the syntax is stable now, possibly :). I need the syntax highlighting for the docs website and for external use. I could write my own easily enough, but it's better to do something collaborative for this.

bob2517:
I'll be putting up some examples on codesandbox.io this weekend so production Cause syntax can be seen in action, and will leave a link or two.

bob2517:
Example:
https://codesandbox.io/s/cause-js-wack-a-mole-obd6e?module=%2Fcausejs-config.txt

bob2517:
Is there any way to get a download for this language so I can put it on the docs website, either in my fork or otherwise? I can't seem to work it out. There is now a live documentation site behind an nginx login waiting for the syntax highlighting to be slotted in for the examples and then it's a wrap. Do let me know if access needs approval through seeing documentation and I'll dish out the login details. Ta muchly.

bob2517:
Docs website is finally up. https://causejs.org - not many people have seen it yet and no one has shown any interest whatsoever, so there's no rush on this. I had to roll-my-own prism in the end as I needed to get the website live and ended up deciding to convert dynamic php pages to static with syntax already formatted, and it was quicker just to knock up some regexes myself rather than try to work out how to do it with prism offline. I've got new regexes which will need converting to JavaScript, which I'll put in here at some point. If anyone does show any interest in getting this language into prism then I'll sort out a new commit with the updated regexes. Otherwise, I'll just crack on with the project as it's good fun and we'll leave it at that.

RunDevelopment:
I'll close this now because CauseJS has been replaced by ActiveCSS.

PR #2125, opened 2019-11-29
Line Numbers: Improved resize performance

### LINKED ISSUE #2123: Line Numbers resize event going crazy when scrolling on mobile, freezing screen for more than a second
**Information:**
- Prism version: PrismJS 1.17.1
- Plugins: Line Numbers
- Environment: Grunt, BrowserSync, Ubuntu.
- ~2800 DOM elements, some long highlighted codes, no less than 12 `pre>code` highlighted.

**Description**
This [line](https://github.com/PrismJS/prism/blob/dab7998e5d9a8b7acbff5be05058a9829a0d084a/plugins/line-numbers/prism-line-numbers.js#L65):

```
window.addEventListener('resize', function () {
    Array.prototype.forEach.call(document.querySelectorAll('pre.' + PLUGIN_NAME), _resizeElement);
});
```

It is firing aggressively on mobile (Real, !emulated) browser scroll events, freezing the screen for a few seconds. Without this event listener, scrolling became smooth again, but sometimes because of the large DOM I can see some lag when scrolling up and down, but not so bad like with this event listener.

With event listener 'resize':
With/without web workers the rendering lag is crazy long. At least with web workers I was able to scroll, couldn't see anything because of the slow rendering, but was able to scroll.

Without event listener 'resize':
With/without web workers there was no difference on the smoothness of scrolling.

Why this event?

RunDevelopment:
This significantly improves the code which resizes an element by batching all DOM reads/writes.

This resolves #2123.


plugins/line-numbers/prism-line-numbers.js
```diff
@@ -16,17 +16,69 @@
...
+		 * Whether the plugin can assume that the units font sizes and margins are not depended on the size of
+		 * the current viewport.
+		 *
+		 * Setting this to `true` will allow the plugin to do certain optimizations for better performance.
+		 *
+		 * Set this to `false` if you use any of the following CSS units: `vh`, `vw`, `vmin`, `vmax`.
+		 *
+		 * @type {boolean}
+		 */
+		assumeViewportIndependence: false
```
mAAdhaTTah, line 61: Can we default to the optimizations and let people opt out if it breaks things for them? I think there's going to be a much larger group that can benefit from these optimizations and a much smaller group that is negatively impacted, so I think it's worth it to get the performance benefit to the largest number of them. Thoughts?

RunDevelopment: Alright then.

bkdotcom:
speaking of the window resize listener... 
see #2192

`Prism.manual = true;`
in 1.22 it tries to "refresh line numbers" of all `$$('pre[data-line]')`  regardless of whether `Prism.highlightElement()` has been called yet... which leads to "Uncaught TypeError: Cannot read property 'children' of null"

PR #2138, opened 2019-12-18
Toolbar: Adjusted the way the menu is positioned

### LINKED ISSUE #2134: Show Language Overlay shows over Vertical Scroll Bar
**Information:**
- Prism version: PrismJS 1.17.1
- Plugins: toolbar+show-language
- Environment: Browser

Does the latest version of Prism from the [download page](https://prismjs.com/download.html) also have this issue?
Yes, demonstrated on [Show Language](https://prismjs.com/plugins/show-language/) using Firefox 71.0 64bit.

**Description**
When using the Show Language Plugin, and there is a Vertical Scroll bar, the hover language overlay appears over the vertical scrollbar. This actually makes the name of the language difficult to read. 

When there is no scroll bar it can be read fine:
image

But as soon as there is one, seeing the language is far harder:
image

**Example**
[Show Language](https://prismjs.com/plugins/show-language/) demonstrates this.

**Expected behaviour**
I would expect the overlay to move slightly to the left when there is a scroll bar, or perhaps there could be a (class) option to denote the site the overlay appears on? This would not be a problem, for example if the overlay appears on the right (but could obscure code below).

For example:
image

RunDevelopment:
This fixes #2134 by introducing a new positioner element between `div.code-toolbar` and `div.toolbar`.

However, this PR has two problems:

1. To be independent of the size of the pre, I basically set the size of the positioner based on the width of the pre's scrollbar (can be 0). But this won't work if e.g. the visibility/width of the scrollbar changes.
    We could just re-adjust the size of the positioner like 60 times/sec but maybe there's something better. Ideas?
2. This is a breaking change. You'll notice that I had to change quite a bit of the CSS of Toolbar. People who have some custom styles for their toolbar will also have to adjust their styles.

### Screenshot

image


peterbud:
@RunDevelopment : for your question 1. could [this solution ](https://weblog.west-wind.com/posts/2014/Oct/20/A-jquerywatch-Plugin-for-watching-CSS-styles-and-Attributes#Watch-on) work?

mAAdhaTTah:
I know this solves a real problem but this feels kinda hack-y and inelegant. Seems like it would be less work for the consumer to add padding on their own if it scrolls.

Or maybe we structure things a little differently so the scrollbar doesn't overlap? Or position it on the left side instead? Just throwing ideas out there.

RunDevelopment:
> Or maybe we structure things a little differently so the scrollbar doesn't overlap?

How? You can either have the toolbar outside the `pre` or inside. We currently have it outside because even if we move it inside, the same issue persists. I don't think we can solve this in pure CSS.

<details>

image

image

```css
div.foo {
	position: absolute;
	width: 100px;
	height: 20px;
	background-color: red;
	top: 0;
	right: 0;
}
```

</details>

> Or position it on the left side instead?

image

The nice thing about having the toolbar on the right side is that it usually doesn't overlap with the code.

---

How about this: Same solution as for Line numbers: We refresh the layout when the element is highlighted and when the window resizes. Since that might not cover all cases, we expose a `resize` function that users can call.

PR #2140, opened 2019-12-19
added support for Latte & NEON

dg:
- Added support for templating system Latte https://latte.nette.org
- Added support for NEON https://ne-on.org


components/prism-neon.js
```diff
@@ -0,0 +1,39 @@
+Prism.languages.neon = {
+	'comment': {
+		pattern: /#.*/,
+		greedy: true
+	},
+	'datetime': {
+		pattern: /(^|[[{(=,\s])\d\d\d\d-\d\d?-\d\d?(?:(?:[Tt]| +)\d\d?:\d\d:\d\d(?:\.\d*)? *(?:Z|[-+]\d\d?(?::?\d\d)?)?)?(?=$|[\]}),\s])/,
```
RunDevelopment: Upon researching this, I found a very similar pattern [here](https://github.com/matej21/neon-js/blob/master/src/decoder.js#L298). Could be please explain why the last `:` is optional in your version? This is the only real difference between the two patterns.

components/prism-neon.js
```diff
@@ -0,0 +1,39 @@
...
+	},
+	'number': {
+		pattern: /(^|[[{(=,\s])[+-]?(?:0x[\da-fA-F]+|0o[0-7]+|0b[01]+|(?:\d+\.?\d*|\.?\d+)(?:[eE][+-]?\d+)?)(?=$|[\]}),\s])/,
+		lookbehind: true
+	},
+	'boolean': {
+		pattern: /(^|[[{(=,\s])(?:true|false|yes|no)(?=$|[\]}),\s])/i,
+		lookbehind: true
+	},
+	'null': {
```
RunDevelopment: Please add an alias `keyword` for null.

components/prism-neon.js
```diff
@@ -0,0 +1,39 @@
...
+		pattern: /(^|[[{(=,\s])(?:null)(?=$|[\]}),\s])/i,
+		lookbehind: true
+	},
+	'string': {
+		pattern: /(^|[[{(=,\s])(?:'''\r?\n(?:(?:[^\r\n]|\r?\n(?![\t\ ]*'''))*\r?\n)?[\t\ ]*'''|"""\r?\n(?:(?:[^\r\n]|\r?\n(?![\t\ ]*"""))*\r?\n)?[\t\ ]*"""|'[^'\n]*'|"(?:\\.|[^"\n])*")(?=$|[\]}),\s])/m,
```
RunDevelopment: 1. Why the `m` flag? All other patterns don't seem to need it.
1. The pattern will be shorter if we combine `'''\r?\n(?:(?:[^\r\n]|\r?\n(?![\t\ ]*'''))*\r?\n)?[\t\ ]*'''` and `"""\r?\n(?:(?:[^\r\n]|\r?\n(?![\t\ ]*"""))*\r?\n)?[\t\ ]*"""` using backreferences.
1. The single line strings are missing a `\r` in their character set.

```suggestion
		pattern: /(^|[[{(=,\s])(?:('''|""")\r?\n(?:(?:[^\r\n]|\r?\n(?![\t ]*\2))*\r?\n)?[\t ]*\2|'[^'\r\n]*'|"(?:\\.|[^"\r\n])*")(?=$|[\]}),\s])/,
```

components/prism-neon.js
```diff
@@ -0,0 +1,39 @@
...
+		lookbehind: true,
+		greedy: true
+	},
+	'literal': {
+		pattern: /(?:[^#"\',:=[\]{}()\x00-\x20`-]|[:-][^"\',\]})\s])(?:[^,:=\]})(\x00-\x20]+|:(?![\s,\]})]|$)|[\ \t]+[^#,:=\]})(\x00-\x20])*/,
```
RunDevelopment: This doesn't handle [JSON properties](https://doc.nette.org/en/3.0/neon#toc-syntax) correctly. 

Could you please explain the idea behind this pattern?

dg:
> P.S. I saw that you're already using these languages on the Nette website and they look great!

Yes, Prism is really great tool! And it's fun to create new languages. Thanks for feedback.

components/prism-latte.js
```diff
@@ -0,0 +1,68 @@
...
+				},
+			}
+		},
+	}, markupLatte.tag);
+
+	Prism.hooks.add('before-tokenize', function(env) {
+		if (env.language !== 'latte') {
+			return;
+		}
+		var lattePattern = /\{\*[\s\S]*?\*\}|\{[^'"\s{}*](?:[^"'/]|\/(?![*/])|("|')(?:\\[\s\S]|(?!\1)[^\\])*\1|\/\*[\s\S]*?(?:\*\/|$))*?\}/ig;
```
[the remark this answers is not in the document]
dg: 1) removed
2) yes, multiline comments are allowed inside Latte tags `{var $x = /* comment */ 10}`, while single line are forbidden
3) it is nonsense
4) in fact, support for nested brackets was removed from Latte, so I'll add `{}` to not allowed chars

components/prism-latte.js
```diff
@@ -0,0 +1,68 @@
...
+	Prism.languages.insertBefore('inside', 'attr-value', {
+		'n-attr': {
+			pattern: /n:[\w-]+(?:\s*=\s*(?:"[^"]*"|'[^']*'|[^\s'">=]+))?/,
+			inside: {
+				'attr-name': {
+					pattern: /^[^\s=]+/,
+					alias: 'important'
+				},
+				'attr-value': {
+					pattern: /=.+/i,
```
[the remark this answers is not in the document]
dg: Good point

components/prism-neon.js
dg: It is bug in that library, because regexp should match `'2016-06-03 19:00:00 +0200'`. I'll send PR.

components/prism-neon.js
dg: Yeah, I agree with everything, thanks.

components/prism-neon.js
dg: This is for "unquoted strings". I modified it, now it should work better.

dg:
Thanks Michael!

PR #2142, opened 2019-12-19
Tests: Adds a check redundant dependency declarations

RunDevelopment:
Since it's possible to have redundant dependency declarations (e.g. `require` + `optional` for the same component), I decided to add a test for it.


RunDevelopment:
@mAAdhaTTah Should we really merge these before the new release?

PR #2144, opened 2019-12-20
fix attribute names with $ in their names not being picked up for jsx

ceymard:
* [x] Fixed the pattern (`$-]` instead of `-$]`)
* [x] Ran gulp
* [x] Added a test case


ceymard:
Alright, I force pushed after following your instructions which in effect only shows the jsx files as being modified.

Is travis gonna run on its own ?

PR #2147, opened 2019-12-25
Added silent option to `loadLanguages`

### LINKED ISSUE #2145: An option remove console.warn from prismjs/components/index.js
**Motivation**

I am implementing a syntax highlight plugin for [Hexo](https://hexo.io) which requires runnning PrismJS inside Node.js.

I have use this approach (adopted from example provided by PrismJS website) to load supported languages that are not loaded by default:

```js
const prismLoadLanguages = require('prismjs/components/');
```

However, when trying to load a language that doesn't supported by PrismJS will cause a `console.warn` output `Language does not exist [language name]`.

**Description**

A `silent` option for not to print error message when loading a language that hasn't been supported by PrismJS could be helpful.

**Alternatives**

An API to get all current supported languages, maybe something like `Prism.supportedLanguages()` so that I could determine whether to use `loadLanguages()` or not.

My current solution is:

```js
const prismComponents = require('prismjs/components.json');
const prismSupportedLanguages = Object.keys(prismComponents.languages);
```

RunDevelopment:
This resolves #2145.

### DIFF
--- a/components/index.js
+++ b/components/index.js
@@ -48,4 +48,9 @@ function loadLanguages(languages) {
 	});
 }
 
+/**
+ * Set this to `true` to prevent all warning messages `loadLanguages` logs.
+ */
+loadLanguages.silent = false;
+
 module.exports = loadLanguages;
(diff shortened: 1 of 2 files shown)


mAAdhaTTah:
Could we add documentation for how to use this?

RunDevelopment:
Done. Some inline doc and a note in `index.html`.

PR #2153, opened 2020-01-04
Added benchmark suite

RunDevelopment:
This adds a benchmark suite to Prism.

Since I already explained everything in `benchmark.html`, I won't repeat myself but the basic idea is that the whole benchmark configuration is in `benchmark/config.js`. 

### Points to discuss

- ~I added support for multiple remote branches because it was easy and useful for testing my implementation but maybe we don't need it?~ It was pretty easy to implement, so let's just keep it.
- ~rn, I only display the mean, margin of error, and sample count. Anything else we need?~ More information can be added in future PRs if needed.
- ~The summary is quite simple, so maybe we could add something?~ It displays enough information IMO.


benchmark/files/prism.js
```diff
@@ -0,0 +1,909 @@
+/* PrismJS 1.17.1
```
mAAdhaTTah: What's the deal with this file + min version? Do we need them duplicated here?

RunDevelopment: Just some test files. (I originally also had copies of other files here before adding support for relative file paths.)

Also, the `prism.js` here isn't a copy of the root `prism.js`. Both of these files were generated by the download page and are supposed to represent as decently sized JS library minified and unminified.

RunDevelopment: I removed the file and use the root `prism.js` instead.

JaKXz (commented):
This looks really interesting! What would it take to get across the line?

RunDevelopment:
I think this just needs a proper review.

@mAAdhaTTah I looked over it again and did some minor refactoring. I think it's good to go but it would be nice if you could review it.

JaKXz (approved):
FWIW I think it looks good 👍🏽

benchmark/benchmark.js
```diff
@@ -0,0 +1,453 @@
...
+	console.log(`Test ${candidates.length} candidates with Prism.${config.options.testFunction}`);
+	const estimate = candidates.length * totalNumberOfCaseFiles * config.options.maxTime;
+	console.log(`Estimated duration: ${Math.floor(estimate / 60)}m ${Math.floor(estimate % 60)}s`);
+
+	/**
+	 * @type {Summary[]}
+	 *
+	 * @typedef {{ best: number; worst: number, relative: number[], avgRelative?: number }} Summary
+	 */
+	const totalSummary = Array.from({ length: candidates.length }).map(() => ({ best: 0, worst: 0, relative: [] }));
```
mAAdhaTTah, line 32: Nit, but the second param to `Array.from` is a mapper function, so you don't need to call `.map` separately.

RunDevelopment: Done.

benchmark/benchmark.js
```diff
@@ -0,0 +1,453 @@
...
+	if (/^https:\/\//.test(uri)) {
+		// it's a URL, so let's download the file (if not downloaded already)
+		const downloadDir = path.join(__dirname, 'downloads');
+		await fs.promises.mkdir(downloadDir, { recursive: true });
+
+		// file path
+		const hash = crypto.createHash('md5').update(uri).digest('hex');
+		const localPath = path.resolve(downloadDir, hash + '-' + /[-\w\.]*$/.exec(uri)[0]);
+
+		if (!fs.existsSync(localPath)) {
```
mAAdhaTTah, line 273: All the other `fs` interactions are async except this one. Any particular reason?

benchmark/benchmark.js
```diff
@@ -0,0 +1,453 @@
...
+ * @param {Result[]} results
+ * @returns {Result | null}
+ */
+function getWorst(results) {
+	if (results.length >= 2) {
+		const sorted = [...results].sort((a, b) => b.stats.mean - a.stats.mean);
+		const worst = sorted[0].stats;
+		const secondWorst = sorted[1].stats;
+
+		// basically, it's only the best if the two means plus their moe are disjoint
```
mAAdhaTTah, line 341: ngl i have no idea what this means. what's a "moe"?

mAAdhaTTah (approved):

benchmark/benchmark.js
RunDevelopment: No idea. It's been like a year. I'll try to fix it.

I only know that it's not what Google thinks it is lol.

benchmark/benchmark.js
RunDevelopment: Because there is [no async version of `exists`](https://stackoverflow.com/questions/17699599/node-js-check-if-file-exists). Seems like we are [supposed to use `stat` or `access`](https://stackoverflow.com/a/57708635/7595472) which is less obvious than `existsSync`.

Want me to write a little async util function for it?

benchmark/benchmark.js
RunDevelopment: Now I remember. moe = [margin of error](https://benchmarkjs.com/docs#stats_moe).

JaKXz:
@RunDevelopment I think this can be merged :)

RunDevelopment:
There's still one open comment.

cc @mAAdhaTTah

benchmark/benchmark.js
mAAdhaTTah: Oh, no need for the complexity.

mAAdhaTTah:
@RunDevelopment sorry, I approved, I don't think that change is necessary.

RunDevelopment:
@mAAdhaTTah Ahh ok. I always read "approved + comments" as "approved if all comments resolved". I'll keep that in mind in the future.

mAAdhaTTah:
In this case, I thought I had commented and resolved so that's my bad.

PR #2159, opened 2020-01-05
JSX: Improved spread operator in tag attributes

0xc14m1z:
This is a small tweak of tag regexes that allow spaces within the brackets of spread operator.


0xc14m1z:
First of all, thanks for considering my PR.
Then, I followed your suggestion and allowed for any number of spaces between the dots and the expression.

Regarding your question... no. It doesn't take account for javascript expression after the `...`. This PR is intended as a tweak for the spaces.

Considered the README though, where small and focused PR are preferred to multi-feature ones, I think that making expressions to work should be another PR. I tried [here](https://regex101.com/r/iOxhLg/1) and it doesn't work.

RunDevelopment:
Thank you for contributing!

And thanks for your answer. I just wanted to know. Making a pattern which allows any JS expression is extremely tricky, so I'll do that myself :)

Issue #1055, opened 2016-11-10
[Brainstorming] Decouple token names and themes


LeaVerou:
I know this is a tough one, but I'm opening this thread for an initial discussion. How do other highlighters deal with this? We currently have over 100 languages, and it's becoming harder and harder to have meaningful names, which was one of Prism's advantages. For example, see #1044.

Perhaps languages could associate with a palette (color1, color2, etc) instead of token names, and then have the stylesheet built automatically? But then what do we do for npm?

Anyways, feel free to throw your ideas in here! Even if you don't think they're good ideas, they might give someone a good idea!

mAAdhaTTah:
I actually had on my todo list setting something like this up with Sass. Each theme could (primarily) be a declaration of variables. Combine that with a shared set of rules, and you could generate each theme with a gulp task by applying the set of variables to the shared ruleset. Then if you need additional token names, you would just have to update the main ruleset and the generated themes would get them with one of the styles/colors applied.

That said, I don't know how that would work with the separate themes directory, but it's an idea.

LeaVerou:
Keep in mind that themes do more than just change colors, e.g. we have themes that add striped backgrounds, shadows, borders etc. We don't want to limit what themes can do.

Also, the main issue with generating things with gulp is that every theme needs to have tokens for every single language, regardless whether the language is actually included or not, which is very counterintuitive (hence why we reuse token names now, which is also counterintuitive, but less so).

Golmote:
The build tool on the Download page could be smart enough to build its own stylesheet using only the styles required for the selected languages in the selected theme.
Full stylesheets would then only be released for npm users I guess...

mAAdhaTTah:
Yeah, you could still include those theme-specific rules in the individual theme Sass file. What the main ruleset would do is provide a base set of tokens with a variable to apply to them, so you wouldn't need all the tokens in each theme.

That said, it doesn't necessarily address your second concern, unless each language had a common base of its own to then apply to. But then you'd be generating a stylesheet for each theme/language combo... :-/

Golmote:
Or, still using the same idea, we could have one stylesheet per language, containing only references to the variables declared in the themes. It might be easier then to generate custom builds for a subset of languages.

LeaVerou:
If we use Sass, how will we build these custom builds? There is a JS port of Sass, but it's not exactly mature yet…

mAAdhaTTah:
[...]

PR #2162, opened 2020-01-06
Download page: Added options

RunDevelopment:
This adds a new 'Options' section to the download page where plugins and Prism itself can be configured.

image

---

The basic idea here is that there are _options_ which are a collection of _option items_. An option can require any number of components and is only enabled if all required components are enabled. Only options that are enabled will be visible.

Every _option item_ translates to one label + one input element. An item is responsible for validating inputs. I added support for 3 types of inputs: boolean (-> `input[type=checkbox]`), number (-> `input[type=number]`), and string (-> `input[type=text]`). But right now, I don't use the number type.

The _option_ will then be used to apply changes to the JS and CSS source code according to the values of its _items_.

The _item_ values of all enabled _options_ will be stored in the URL if the value differs from the default value.

---

I added 3 options:

1. A small general option to disable automatic highlighting.
2. The Custom Class prefix option. (#1055)
3. Options for Filter HighlightAll to set `filterKnown` and to add CSS selectors. (not enabled in the image)

To apply the prefix to the theme, I used Prism's CSS language to tokenize the CSS code and get all selectors and replaced all token names using a regular expression.

### DIFF
--- a/download.html
+++ b/download.html
@@ -8,7 +8,8 @@
 <link rel="stylesheet" href="style.css" />
 <link rel="stylesheet" href="themes/prism.css" data-noprefix />
 <style>
-form label {
+form #components label,
+form > p:first-of-type > label {
 	display: block;
 	padding: .2em;
 	padding-left: 1.9em;
@@ -17,16 +18,17 @@ form label {
 	-webkit-column-break-inside: avoid;
 }
 
-	form label .name {
+	form #components label .name {
 		margin-right: .3em;
 	}
 
-	form label a.owner {
+	form #components label a.owner {
 		margin-left: 0;
 		hyphens: none;
 	}
 
-	form label input {
+	form #components label input,
+	form > p:first-of-type > label input {
 		margin-right: .7em;
 		margin-left: -1.7em;
 	}
@@ -54,7 +56,8 @@ section.options {
 }
 
 section.options#category-languages,
-section.options#category-plugins {
+section.options#category-plugins,
+section#options {
 	width: 100%;
 	float: none;
 	column-count: 3;
@@ -63,11 +66,13 @@ section.options#category-plugins {
 }
 
 	section.options#category-languages label,
-	section.options#category-plugins label {
+	section.options#category-plugins label,
+	section#options #option-container > div {
 		break-inside: avoid;
 	}
 	section.options#category-languages > h1,
-	section.options#category-plugins > h1 {
+	section.options#category-plugins > h1,
+	section#options > h1 {
 		margin-top: 0;
 		column-span: all;
 	}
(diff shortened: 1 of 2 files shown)


LeaVerou:
This is fantastic work @RunDevelopment! I haven't looked at the code very closely, but the functionality is sorely needed!

RunDevelopment:
Thanks. 
Looking at it again, the only thing I'm unhappy about is that `download.js` now contains plugin logic. I really should have made the `getOptions` function into its own file (this will also help the already quite large `download.js`).

And I need to revisit the Custom Class implementation. I used Prism to parse CSS because it was convenient but come on. I should do this the right way.

LeaVerou:
> And I need to revisit the Custom Class implementation. I used Prism to parse CSS because it was convenient but come on. I should do this the right way.

I did that a few days ago 😊 (to parse JS though, not CSS)

RunDevelopment:
Ok. No CSS parser. I wasn't able to find a good CSS parser that just works on the web and preserves spaces and comments. Prism-parsing it is then.

I also found another major pain point for the Custom class prefix option: To which classes do we add the prefix?
Right now, I add the prefix to all classes that are in the same component selector as a `.token` class. This works for all themes but breaks some plugins. It breaks the CSS of plugins in two ways:
1. Some component selectors of tokens don't have a `.token` class. I think that this is a problem of the plugin and should be fixed for each respective plugin. I know that Tree view has this problem. I haven't check for any other plugins yet.
2. Some plugins might all classes to tokens _after_ the `warp` hook meaning that Custom Class can't add the prefix. The CSS will still get the prefix, so the style doesn't work (e.g. Match braces). How do we fix this? (Match braces is incompatible with Custom class right now anyway.)

LeaVerou:
> Ok. No CSS parser. I wasn't able to find a good CSS parser that just works on the web and preserves spaces and comments. Prism-parsing it is then.

You can also just do regex replace on the fetched CSS, theme CSS is pretty simple, so that should work quite well.

> I also found another major pain point for the Custom class prefix option: To which classes do we add the prefix?

Would not prefixing the token class itself alleviate any of these? I suspect it's not a very commonly used name...

Also, perhaps we need to bite the bullet and bake this functionality into the core instead of making custom classes a plugin. It seems like fairly important functionality… 

(Aaand this is where I wish we had usage stats for plugins...)

RunDevelopment:
> Would not prefixing the token class itself alleviate any of these?

Not really. Most plugins search for [specific tokens by their type](https://github.com/PrismJS/prism/blob/9782cfe6ab9c798eecc0ba390ffc29825533d07d/plugins/match-braces/prism-match-braces.js#L119). Given that Prism supports arbitrary prefixes and even remapping (via Custom Class), you could argue that the plugin (e.g. Match Braces) is at fault.

Right now, the only thing nobody modifies during the conversion from token stream to markup are attributes, so I could use marker attributes instead of marker classes to implement Match Braces. The plugin will then be implemented partly in `warp` and partly in `complete`. In `wrap` I'll add the marker attributes before Custom Class changes any classes and in `complete` I'll query by these attributes. Should work, I guess.

To make this more stable we could make a `pre-wrap` hook that isn't allowed to _change_ data in the environment but it can _append_ new data (of course, you can also read). I.e. you can't change any classes or attributes in `pre-wrap` but you can add new ones. 

> (Aaand this is where I wish we had usage stats for plugins...)

We don't have those? I always assumed that we did. I'm not very familiar with Google Analytics but if they also collect information on the hash of the URL of the download page, then we have these stats.

PR #2163, opened 2020-01-06
JS/JSX: Add support for recursive expressions

RunDevelopment:
So after #2159 I wanted to add support for arbitrary JS spread expressions and kinda overdid it. I also added support for deeply nested expressions in JS template string interpolations.

Because I need very similar patterns in #2139, I decided that JS is going to share its patterns and added a way for languages to share internal patterns. It's done via a non-enumerable property `$`. Because the property is non-enumerable, it won't be cloned (e.g. for `extend`) but it will be kept by `insertBefore`.
Btw. I'm not very attached to this version by variable sharing, so better ideas are very welcome.

This PR is a 3 in 1: JS Template strings, JSX spread expressions, and the `$` property.


mAAdhaTTah:
Another one worth refreshing for bundle size impacts.

JaKXz:
@RunDevelopment this looks really interesting, but I'm not sure how to help since a lot of it is out of date at this point. Do you think this is worth picking back up again?

RunDevelopment:
> Do you think this is worth picking back up again?

Not in its current form, no. This PR relies on a sorta hack to share regexes across languages. I'd rather wait for v2 where we can just directly import/export regexes via ESM.

PR #2167, opened 2020-01-09
Avoid TypeError when mapper is undefined

harigopal:
A ternary operation in the `prism-custom-class.js` file needs to be wrapped with parentheses for it to be evaluated properly. In it's current form, it's throwing a _TypeError_.

```js
let a = "foo";
let b = undefined;
let c = "default";
a + b ? b() : c;
// Uncaught TypeError: b is not a function
a + (b ? b() : c);
// "foodefault"
```


mAAdhaTTah:
Thanks for the PR! Make sure you run `gulp` to rebuild the files & commit them.

@RunDevelopment This seems like enough of an issue that we should consider cutting another release soon. Thoughts?

RunDevelopment:
@mAAdhaTTah I agree. I'll get the changelog ready.

Maybe we could set up a `tsconfig.json`, so TSC can catch these null/undefined errors, so it won't happen again. After all, it's the second time this happened to me (#2106).

PR #2169, opened 2020-01-09
Updated changelog for 1.19.0

RunDevelopment:
(Not linkified, no date)

This also includes a small update to the changelog generation to account for `dependencies.js`.


mAAdhaTTah:
I think this needs to be `1.19.0`. We're introducing new languages, which is a new feature, so minor version bump.

mAAdhaTTah:
Set the date to today before merging.

RunDevelopment:
Is that a request or a statement?

RunDevelopment:
(This might sound passive aggress, but it isn't meant to.)
I intentionally publish these changelog branches in the PrismJS repository and not in my fork, so other maintainers can easily commit their own changes (maybe even using GitHub's in-browser editor). I do this because I don't want to miss a message like this causing a release to be delayed and you to be annoyed.

mAAdhaTTah:
Hah, no worries. The thought at the time was to update the date, merge, then I'd be able to pull master & publish w/o needing to update anything, but I fell asleep before I saw this. Anyway, it's published, thanks for your help!

Issue #2164, opened 2020-01-07
Coy theme shadow doesn't correctly wrap around floats.


garretwilson:
Years ago I had turned off the Coy theme shadow as we discussed in #894, but I had forgotten exactly why. (I had thought it only had to do with scrollbars in generated PDFs in #859.) So when I updated to the latest Prism a few days ago and published my entire [software development course](https://www.globalmentor.com/courses/softdev/), I left the Coy theme shadow on.

Now I see that the Coy theme shadow is broken when wrapping around flows. Here are a couple of examples:

* https://www.globalmentor.com/courses/softdev/unit-tests (Look for the figure with the caption "Testing `Math.abs(int)`.") ![coy-shadow-wrap-right](https://user-images.githubusercontent.com/627125/71899205-f87c1600-3139-11ea-836d-c190065e11f6.png)
* https://www.globalmentor.com/courses/softdev/surf (Look in the section "URF, SURF, and TURF".) ![coy-shadow-wrap-left](https://user-images.githubusercontent.com/627125/71899165-ea2dfa00-3139-11ea-8a56-f8a3e22741c6.png)

garretwilson:
Note that this may be related to #965.

RunDevelopment:
What should the correct behavior be? The reason why we see the white background is that the `pre` has the dimensions of the parent `figure` which spans across the whole width of the page (sans some margin). The overlap with other blocks comes from the usage of `float` which usually doesn't far well with `display: block` from my experience.

So we could either want the block to have the same width as its parent, in which case the solution is quite simple. (`clear: both;` for the `pre`):

![image](https://user-images.githubusercontent.com/20878432/71994539-36ac2f00-3239-11ea-9400-fd2fbcef221e.png)

Or we might want the code block to behave like text, so it "breaks" early.

![image](https://user-images.githubusercontent.com/20878432/71994820-b508d100-3239-11ea-9478-4da6349be9d3.png)

I'd go with the former solution. If you still want the code block to "break", then you can still change the parent `figure`. (`display: grid;` works for some reason.)

---

> this may or may not be related to #965.

It doesn't seem to be related.

garretwilson:
> What should the correct behavior be?

[...]

Issue #894, opened 2016-02-15
How to disable shadows manually in coy theme.


garretwilson:
The coy theme comes with shadows. I would like to know how to turn them off manually from my own CSS.

Yes, I know that most of the time they look snazzy. Yes, I know that a lot of work has been put into them. Yes, I'm fine with them being turned on by default for screen media.

But for my purposes I just need to be able to disable them, even for screen. What is the easiest way for me to do this?

I've asked this question several times in Issue #859, but it must have got lost in the shuffle so I'm just filing a separate issue. Thanks.

zeitgeist87:
Sorry if we missed your question. This should do it:

```css
pre[class*="language-"]:before,
pre[class*="language-"]:after {
    display: none;
}
```

garretwilson:
Bingo! That seems to work. Thanks so much!

Issue #866, opened 2016-01-24
Prism coy theme overwrites floating elements


garretwilson:
Let's say I have a figure with an image, and with CSS I make that figure float right. If I then put a `<figure><pre><code>` section in the body after the floated `<figure>`, all is well; the code fragment figure is displayed to the left of the floated-right figure.

But the I then bring in Prism with the coy theme:

````html
<pre class="command-line"><code class="language-bash">foo bar</code></pre>
````

Now the non-floating figure completely ignores the floating figure, and appears on top of the floating element. Presumably the CSS tricks played by the coy theme are so brute force that the don't accommodate floating elements.

![prism-overwrite-float](https://cloud.githubusercontent.com/assets/627125/12537217/b88c2b04-c2a0-11e5-9bb3-91c66d0f4128.png)

(I can make a wild guess and say that maybe this is yet another repercussion of the shadows, because otherwise I don't see why the other styling would be so inflexible as to not play nicely with floats.)

garretwilson:
Hmmm... no, I was wrong. I don't think this is about the shadow, about coy, or even about Prism. It seems that my figure is taking up the entire width anyway --- I only notice it when I turn on Prism, as it gives my figure a background. I guess I'll just have to float the figure containing the `<pre><code>` as well. That seems to work.

OK, you can close this issue as invalid; I don't think Prism or coy is at fault here.

Issue #859, opened 2016-01-19
turn of height restriction with scrolling by default


garretwilson:
I turned Prism on for my code samples and it looks gorgeous!

But a longer code section suddenly got its height restricted with scrollbars added with the coy skin. I didn't ask for this! I don't want scrollbars.

The reason this is a blocker issue is that I need to convert this documentation to a PDF. Scrollbars in a PDF won't help my students, and they will be missing code.

At the very least, there should be CSS that indicates the restricted height with scrollbars is only for screen media, not for print media!!

garretwilson:
The offending code is a `max-height` set for `<pre>`. I don't know if it is part of Prism or part of the coy skin.

Here is the workaround for CSS after Prism is included:

    pre {
      max-height: none !important; /*workaround for Prism scrollbars*/
    }

zeitgeist87:
> I don't know if it is part of Prism or part of the coy skin.

I think it is only part of the `coy` theme and I agree that such a restriction should not be determined in the theme but left to the end user. Changing it now would probably break a lot of websites which expect this behavior.

garretwilson:
I hate this "we can never ever change this for backwards-compatibility, even though it sucks" argument that keeps coming up.

Do it the right way: turn it off for print. Nobody could every _want_ it for print, because you lose information. Leave it on temporarily for backwards-compatibility on web sites.

As for long-term, deprecate this behavior. Create a checkbox on the download page that you have to check to get the restricted height. Indicate that you will eventually remove it.

I shouldn't have to add workarounds in my clean code just to get things to work the right way.

zeitgeist87:
> As for long-term, deprecate this behavior. Create a checkbox on the download page that you have to check to get the restricted height. Indicate that you will eventually remove it.

I don't think there is any need for that. I think we can remove it immediately for the next release. If people want the old behavior, they can use an earlier version.

garretwilson:
Yay!! Thank you. Sorry, I guess I had misunderstood you earlier, thinking you were going to always leave it in.

zeitgeist87:
@LeaVerou maybe you would like to comment on this? Otherwise I would remove the `max-height: 30em` attribute from the `coy` theme.

LeaVerou:
I would put it inside `@media screen`. @garretwilson is right that nobody would want this for print.

zeitgeist87:
> I would put it inside @media screen {...}.

Done.

garretwilson:
[...]

PR #2170, opened 2020-01-09
Added a Coy bug to the known failures page.

RunDevelopment:
As discussed [here](https://github.com/PrismJS/prism/issues/2164#issuecomment-572558797).

image

---

I also quickly looked for other open theme issue but couldn't find any. If there are any I missed, please leave a comment.


garretwilson:
Actually @RunDevelopment I discovered another Coy shadow bug, this time vertically!. If you have two flexbox items (this is very important to me to include code snippets in flexbox children!), the code snippit will only extend vertically as much as it needs, but the shadow will extend vertically to fill the flexbox!

prism-coy-shadow-flex-bug

The overall problem is that you seem to be using two different approaches to sizing the code section and the shadow. It seems to be that if this keeps appearing, there's a fundamental flaw in the approach.

The shadow should match the code area, both horizontally and vertically.

As I mentioned before I'm going to turn the shadow off across all my pages, but it's a shame and I think you should revisit how you do the shadow.

RunDevelopment:
So we can go in both directions... At least it's consistent.

But this seems to be the same bug just cause by flex(ing) this time.

> It seems to be that if this keeps appearing, there's a fundamental flaw in the approach.

There is. For Coy, the `pre` is just a wrapper and all styling is done on the `code` element. This is basically the opposite of what all other themes are doing and one of the reasons why we have so many Coy-related issues.
It's actually quite simple to rewrite the themes to style the `pre` making it behave like all others but I can't get the shadows working. [More details](https://github.com/PrismJS/prism/issues/2164#issuecomment-572184378).

> I'm going to turn the shadow off 

I am curious about how you do this. Do you just disable the pseudo-elements and make the background transparent or do you actually fix the layout (if so, how)?

garretwilson:
> > I'm going to turn the shadow off

> I am curious about how you do this. Do you just disable the pseudo-elements and make the background transparent or do you actually fix the layout (if so, how)?

I'm just following the [instructions](https://github.com/PrismJS/prism/issues/894#issuecomment-184317382) in #894 (which I opened after all the Coy shadow problems in #859).

It had been so long that I had forgotten all the pain the shadows caused, and tried to turn them back on. Now they are on across all my sites and I'll have to go back and turn them off again. It will be wonderful to have the bugs publicly documented so I won't make that mistake again.

mAAdhaTTah:
If there's an alternate approach you have in mind, we could possibly put it in [`prism-themes`](https://github.com/PrismJS/prism-themes) as like a `coy-alt` or something. Alternatively, I would accept `coy-without-shadows` there as well (although preferably only one of those).

garretwilson:
Thanks, but frankly this is the least of my worries at the moment. As long as I have a way to turn it off, I can concentrate on other problems. I think `coy-without-shadows` is a hack, but it does point out that there may be variations needed to some of the themes. So you might want to think about that in a general sense going forward; maybe each of the themes have customizable options on the download page or something.

In any case the most important thing for the short terms is to document the problems in a visible way and explain how to turn the shadows off.

RunDevelopment:
> I'm just following the instructions in #894 (which I opened after all the Coy shadow problems in #859).

And that works? I see that it disables the shadows but it doesn't change the background, so it should still be able to overlap with other elements if I'm not mistaken. I'll look into it.

garretwilson:
> And that works?

I haven't tried it in a while. When I last tried it I was using an old version of Prism. I just assumed it still worked. If not we have a bigger problem.

If it doesn't work then maybe you should provide a version of Coy after all that uses the same technique as the other themes but without the shadow.

garretwilson:
> it should still be able to overlap with other elements if I'm not mistaken.

I think the Coy theme always overlapped other elements, correct? That's why usually I would float the Prism block e.g. to the left if I had something I wanted it to wrap around.

This problem with the shadow is not that the code block overlapped with a float (because that always happened, unfortunately), but that the shadow didn't match up with the code block dimensions and therefore didn't look like a shadow.

RunDevelopment:
Yes, they still overlap.

<details>

image

</details>

I'll make a PR for a version of Coy without shadows.
Only problem: Right now, the download page doesn't support themes from Prism themes... This will be a little harder to bring to the download page.

garretwilson:
> Yes, they still overlap.

I can't say 100%, but I'm pretty sure that two or three years ago the overlap would occur but _the code would also extend to fill that white space_!

So did something change in the past couple of years? If as you say the `<pre>` is a block element and shouldn't wrap around the floats, then why is the Coy code section cut of or pushed to the left? I would almost swear that it didn't use to work like that.

garretwilson:
See, look at #866. I'm pretty sure that the Coy theme would reach right through the float. Why doesn't it work that way anymore? One could argue that this is the correct behavior, even though I don't like it.

garretwilson:
In fact #866 wouldn't be so bad if I could change the Z order somehow and have my floats go _on top_ of Prism.

PR #2176, opened 2020-01-11
Coy: Minor improvement

RunDevelopment:
Two minor improvements for the Coy theme:

1. It now consistently uses `[class*="language-"]`.
2. Removed the `:not(pre) > code[class*="language-"]:after` selector because `code` element don't have an after pseudo-element.

### DIFF
--- a/themes/prism-coy.css
+++ b/themes/prism-coy.css
@@ -45,7 +45,7 @@ pre[class*="language-"]>code {
 	background-attachment: local;
 }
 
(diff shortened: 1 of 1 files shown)


RunDevelopment:
I also noticed that Coy has custom styles for the [Show Invisibles plugins](https://prismjs.com/plugins/show-invisibles/) which aren't applied.

image

Which may be for the better...

image

Should I fix the theme or remove the rule?

PR #2179, opened 2020-01-12
OCaml: Several improvements

SaswatPadhi:
+ added some missing keywords
+ added highlighting for modules, labels and constructors
+ split polymorphic variant types and type variables


SaswatPadhi:
Thank you for the quick feedback.

RunDevelopment:
Thank you for contributing!

Issue #1381, opened 2018-04-06
Add support to load/unload plugin


filips123:
Add support to load/unload plugin for specific code block. This will be useful, because currently you have to download and include Prism version without plugin if you don't want to it.

Users should be able to enable or disable plugin(s) for specific code block.

You could do this by adding option `autoload-plugins` to enable or disable autoloading.
```html
<script src="prism.js" data-autoload-plugins="true"></script>
<script src="prism.js" data-autoload-plugins="false"></script>
```
You should also add `enable-plugins` and `disable-plugins` to code block for enabling or disabling plugins for specific code block.
```html
<pre data-enable-plugins="line-highlight,keep-markup,..."></pre>
<pre data-disable-plugins="command-line,keep-markup,..."></pre>
```

Golmote:
What plugin(s) are you talking about? Most plugin should already have opt-in or opt-out options.

filips123:
Keep markup and Copy to Clipboard.
Maybe this plugins also have opt-in and opt-out options, but I don't find them, or there is some other plugin that also doean't have opt-in and opt-out options.

filips123:
Maybe there are also some other plugins with this issue.

navjotjsingh:
Any progress on this front?

Two plugins which can benefit from this immensely are 
Copy to Clipboard - Would really prefer to disable it on single line code blocks.
Autolinker - There are instances where I would rather have it disabled than enabled by default.

RunDevelopment:
@navjotjsingh I hope #2199 will solve your issue with Copy to Clipboard.
Autolinker will be very hard to disable because it modifies Prism's grammars, so we'd have to reset those somehow.

For all plugins other than those that modify Prism's grammars, couldn't we use a similar solution as what I did in #2199 to selectively enable/disable plugins? The idea is that the `env` of all hooks is extended with a function which given a name returns whether that plugin (or something) is enabled or not.
Thoughts?

PR #2199, opened 2020-02-05
Toolbar: Buttons can now be disable and enabled

RunDevelopment:
In response to [this](https://github.com/PrismJS/prism/issues/1381#issuecomment-582516029) comment, I added the ability to enable/disable specific buttons on specific code blocks via the `data-toolbar-enable` and `data-toolbar-disable` attributes. See `index.html` for more info.

I think that this PR addresses half of #1381. Even though the Copy to clipboard plugin can't be unloaded its effects can be disabled.


RunDevelopment:
Closed because #2205 provides a more powerful API. See [this comment](https://github.com/PrismJS/prism/pull/2205#issuecomment-583750630).

PR #2201, opened 2020-02-06
CSS: Added support for the selector function

### LINKED ISSUE #2127: [CSS Conditional 4] @supports selector(…) function
## Motivation
The&nbsp;[**CSS&nbsp;Conditional&nbsp;Rules Module&nbsp;Level&nbsp;4**](https://drafts.csswg.org/css-conditional-4/) extends the&nbsp;`@supports`&nbsp;at‑rule with&nbsp;support for&nbsp;testing selectors&nbsp;using the&nbsp;[`@supports selector(…)`](https://drafts.csswg.org/css-conditional-4/#at-supports-ext)&nbsp;function.

## Description
Currently, the&nbsp;`css‑extras` **Prism**&nbsp;language&nbsp;definition provides highlighting&nbsp;support for&nbsp;CSS&nbsp;selectors, but&nbsp;doesn’t do&nbsp;anything for&nbsp;at‑rules.

I&nbsp;propose&nbsp;that the&nbsp;`css‑extras`&nbsp;grammar be&nbsp;extended with&nbsp;the&nbsp;capability to&nbsp;handle selectors&nbsp;within the&nbsp;`@supports selector(…)` function.

## Example

```css
@supports selector(::-webkit-foo) {
	/* style to apply when the `::-webkit-foo` selector is actually supported,
	 * instead of just being parsed as matching nothing because of Selectors 4
	 * § Appendix B: https://drafts.csswg.org/selectors-4/#compat */
}
```

RunDevelopment:
This resolves #2127.


components/prism-css.js
```diff
@@ -7,7 +7,12 @@
 		'atrule': {
 			pattern: /@[\w-]+[\s\S]*?(?:;|(?=\s*\{))/,
 			inside: {
-				'rule': /@[\w-]+/
+				'rule': /^@[\w-]+/,
+				'selector-function-argument': {
+					pattern: /(\bselector\s*\((?!\s*\))\s*)(?:[^()]|\((?:[^()]|\([^()]*\))*\))+?(?=\s*\))/,
+					lookbehind: true,
+					alias: 'selector'
```
mAAdhaTTah: Should this be aliased to `argument`? Do we have a token for that?

tests/languages/css/atrule_selector-function-argument_feature.test
```diff
@@ -0,0 +1,24 @@
...
+}
+
+----------------------------------------------------
+
+[
+	["atrule", [
+		["rule", "@supports"],
+		["function", "selector"],
+		["punctuation", "("],
+		["selector-function-argument", "::-webkit-foo"],
```
mAAdhaTTah: Why does this produce a different AST?

mAAdhaTTah: Oh wait, nvm, that's cuz the other is css-extras.

RunDevelopment: Different from the CSS Extras versions? Because CSS Extras adds `inside`, so the selector is tokenized with CSS Extras but not without it.

components/prism-css.js
RunDevelopment: > Do we have a token for that?

No. Neither `argument` nor something like `parameter`.

Also, I initially wanted to call this just `selector` but this wasn't possible because of the `rest` property which includes CSS, including the actual `selector` token.

tests/languages/css/atrule_selector-function-argument_feature.test
mAAdhaTTah: i got it no worries

RunDevelopment: Yeah. I only commented because GitHub doesn't refresh the comments, so I couldn't see your comment while writing mine.

PR #2203, opened 2020-02-08
Add Factor

catb0t:
[Factor](https://factorcode.org) language

### Example

Screenshot from 2020-02-07 21-36-43


components/prism-factor.js
```diff
@@ -0,0 +1,277 @@
+(function (prism) {
```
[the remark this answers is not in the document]
catb0t: that would have shadowed the `Prism` with which the IIFE is called, and I avoid shadowing whenever possible. I'll change it since you point it out though.

components/prism-factor.js
```diff
@@ -0,0 +1,277 @@
...
+ 	};
+
+	var word_pattern = {
+		start: '(^|\\s)',
+		end: '(?=\\s|$)'
+	};
+
+	var arr_to_words_regex = function (arr) {
+		return new RegExp(
+			word_pattern.start + '(?:' + arr.map(quote).join('|') + ')' + word_pattern.end
```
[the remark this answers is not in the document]
catb0t: yep, that's a holdover from a previous version i overlooked

catb0t:
So many upstream projects insist that contributors squash before they will merge/rebase, I forget that anyone does differently. I would rather not force-push too.

components/prism-factor.js
```diff
@@ -0,0 +1,277 @@
+(function (prism) {
+
+	var factor = {
+
+		comment: { // ! single-line exclamation point comments with whitespace after/around the !
+			pattern: /(^|\s|(?!"\s*))(?:![\t ].*|!$)/,
```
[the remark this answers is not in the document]
catb0t: This fixes a problem where strings containing `!` would be tokenised as comments starting before the string is closed.

Removing the lookahead in question, the test fails like:

```
: expected [ Array(23) ] to deeply equal [ Array(21) ]
      + expected - actual

           "string"
           "\" ! \""
         ]
         [
      -    "conventionally-named-word"
      -    "\"!"
      -  ]
      -  [
           "string"
      -    "\"\r\n\r\n\""
      +    "\"! \""
         ]
      -  "\\"
         [
           "string"
      -    "\"\""
      +    "\"\\\"\""
         ]
         [
           "string"
           "\"'\""
```

It's hard to exactly understand what goes wrong here (it's related to the CRLF because it doesn't happen on localhost/test.html), but this lookahead fixes it and makes it behave correctly regardless of line endings, which is the point.

components/prism-factor.js
```diff
@@ -0,0 +1,277 @@
...
+				"The normalized hex form ±0x1.MMMMMMMMMMMMM[pP]±EEEE allows any floating-point number to be specified precisely.
+				The values of MMMMMMMMMMMMM and EEEE map directly to the mantissa and exponent fields of the binary IEEE 754 representation."
+				<https://docs.factorcode.org/content/article-syntax-floats.html>
+			*/
+			{ 	pattern: /(^|\s)[+-]?0(?:b1\.[01]*|o1\.[0-7]*|d1\.\d*|x1\.[\dA-F]*)p\d+(?=\s|$)/i,
+				lookbehind: true }
+		],
+
+		boolean: {
+			pattern: /(^|\s|\r\n)(?:t|f)(?=\s|$)/,
```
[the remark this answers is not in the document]
catb0t: This is also a holdover from a previous version. I will add a comment about why `\s` is technically wrong for Factor, but the correct (non-current) implementation is much harder to write and maintain.

components/prism-factor.js
```diff
@@ -0,0 +1,277 @@
...
+
+		'math-builtin': {
+			pattern: null,
+			lookbehind: true,
+			alias: 'variable'
+		},
+
+		'constructor-word': {
+			// <array> but not <=>
+			pattern: /(^|\s)<(?!=+>|-+>)[^\r\n\t ]+>(?=\s|$)/,
```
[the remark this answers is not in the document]
catb0t: `\f` and `\v` (which are in `\s`) may appear literally in a word name, as they are not treated as whitespace. A literal tab character is forbidden from appearing anywhere in a program. 

`\s` and `\S` to quantify whitespace in tokenising Factor is wrong compared to the reference implementation, but using `(?:\n|\r\n)` everywhere (which would be perfectly correct) is a pain. I've tried to allow `\f` and `\v` in word names where possible, to align with the reference implementation.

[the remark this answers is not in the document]
catb0t: I also thought that JavaScript disallowed `[^\s]`, i.e. a builtin class other than an escaped literal inside a negative class, but now I see it's allowed, I might as well change them...

components/prism-factor.js
```diff
@@ -0,0 +1,277 @@
...
+
+		'conventionally-named-word': {
+			// for certain word naming conventions like +lt+, sequence?, readers>>, >>setters, writers<<
+			pattern: /(^|\s)(?:set-[^\r\n\t ]+|change-[^\r\n\t ]+|>[^>\r\n\t ]+|[^:>\r\n\t ]+>|[^>\r\n\t ]+>[^>\r\n\t ]+|\+[^\+\r\n\t ]+\+|[^?\r\n\t ]+\?|[^>\r\n\t ]+>>|>>[^>\r\n\t ]+|[^<\r\n\t ]+<<|\?[^?\r\n\t ]+|\([^()\r\n\t ]+\)|[^!\r\n\t ]+!)(?=\s|$)/,
+			lookbehind: true,
+			alias: 'keyword'
+		},
+
+		'colon-syntax': {
+			pattern: /(^|\s)(?:[A-Z]*#?):{1,2}\s+(?:(?!;)|;\S+|[^\r\n\t; ]\S*)(?=\s|$)/,
```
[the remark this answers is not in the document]
catb0t: The goal of this is to disallow `: ;` (because semicolon `;` is a definition-ender, not a word name, and shouldn't be highlighted as valid), but to allow `: a;` and `: ;a` because  `a;` and `;a` are word names; `;` is only a definition-ender when it is alone as a token, not part of another token.

You are right, I didn't read what I had written and the correct regex is as you say.

components/prism-factor.js
```diff
@@ -0,0 +1,277 @@
...
+				lookbehind: true }
+		],
+
+		boolean: {
+			pattern: /(^|\s|\r\n)(?:t|f)(?=\s|$)/,
+			lookbehind: true
+		},
+
+		'special-string-literal': {
+			pattern: /(^|\s)[A-Z0-9\-]+"\s(?:\\"|[^"])*"/,
```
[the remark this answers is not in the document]
catb0t: Yes, whitespace after the opening double-quote is really intentional. 

The words `URL"`, `SBUF"`, and `P"` (path literal) words are the target of `special-string-literal`. They are normal parsing words and must be followed by whitespace, then the content of the string and a closing quote, i.e `URL" https://github.com"`;  ref: [`URL"`](https://docs.factorcode.org/content/word-URL__quo__,urls.html).

<details><summary> Long explanation </summary> 

The double-quote with whitespace (or another string) preceding it, is treated specially by the Factor parser. Usually, as you've seen, tokens must be surrounded by whitespace, but the builtin normal string literal `"like this"` does not have to have whitespace after the first quote-mark (as it would, if `"` was implemented as a normal parsing word that calls `skip-blank` and looks for a closing `"`). Otherwise, builtin strings would be written like `" this"` and the first space after the `"` would be required but dropped by the parser.  </details>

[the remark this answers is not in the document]
catb0t: It seems to only match the `" \\"` part of `" \\""` though, so that's good.

components/prism-factor.js
```diff
@@ -0,0 +1,277 @@
...
+			alias: 'function'
+		},
+
+		'semicolon-or-setlocal': {
+			pattern: /(\s)(?:;|:>)(?=\s|$)/,
+			lookbehind: true,
+			alias: 'function'
+		},
+
+		'curly-brace-literal-open': {
```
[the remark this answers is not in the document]
catb0t: Readability, simplicity and naming them differently were all I had in mind. I'll combine them however.

components/prism-factor.js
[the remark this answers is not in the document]
RunDevelopment: I don't see why any language would allow `\f` and `\v` in names... An interesting choice on factor's side. 

I can't believe people actually use this. Seems to be part of the technically-possible-but-please-don't category.

components/prism-factor.js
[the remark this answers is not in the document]
RunDevelopment: It's ok. Especially when rewriting patterns, these things can happen.

components/prism-factor.js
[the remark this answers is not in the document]
RunDevelopment: The test failed because you didn't just remove the lookahead.
Because the lookahead will always accept (aka. the pattern `"\s*`) cannot be matched, it's unnecessary and can be removed. So we get:

`(^|\s|)(?:![\t ].*|!$)` (the alternative is still there)

If the empty alternative is removed, the tests fail because the language of the RE changed, so let's keep it. But that means that our lookbehind can match the empty and will therefore always accept, so it can be removed.

```js
comment: {
	pattern: /(?:![\t ].*|!$)/,
	inside: ...
}
```

This will also pass all tests because the meaning of the pattern didn't change. But that's not what you intended, right?

---

So, why do the tests fail with: `(^|\s)(?:![\t ].*|!$)` ?

The simplest example which fails is:

```
"! "
```

In this example, `"!` is tokenized as `conventionally-named-word` because it matches the `[^!\r\n\t ]+!` part of the pattern. So what Prism does is the following (simplified):

1. It tries to match `comment` and fails because `"` doesn't match the lookbehind (`(^|\s)`).
2. It tries to match `conventionally-named-word` and succeeds. The text is now tokenized as `[["conventionally-named-word", "\"!"], " \""]`.
3. Prism tries to match string and fails because the `conventionally-named-word` has already claimed that part of the string.

So the simplest fix is to move `string` before `conventionally-named-word` or `conventionally-named-word` after `string`. Without going into the details, please prefer the first option.
You also have to add the usual lookbehind to string.

__Some additional information:__

So why does the current (or old) pattern work? Let's look at how Prism matches the above example:

1. It tries to match `comment` and succeeds because the lookbehind is essentially non-existent.   The text is now tokenized as `["\"", ["comment", "! \""]]`.
2. It tries to match `conventionally-named-word` and fails because it's partially claimed by `comment` (first come first served).
3. Prism tries to match `string`. Because `string` is [greedy](https://prismjs.com/extending.html#language-definitions) and the match of `string` starts before the comment match, it will remove the comment highlighting and set it as `string`. The text is now tokenized as `[["string", "\"! \""]]`.

So the reason why it worked is that `comment` prevented `conventionally-named-word` from matching some text before `string` could. The reason why `string` can't remove `conventionally-named-word` is that they start at the same character position in which case it's first come first served again.

components/prism-factor.js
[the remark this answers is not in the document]
RunDevelopment: Sorry, `" \\""` is a bad example. Let's take `P" as\\""` which is fully matched.

components/prism-factor.js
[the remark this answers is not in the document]
catb0t: > `/(?:![\t ].*|!$)/`

`a!` is simply a word named `a!`, and must not be tokenised as a comment. The `!` must be surrounded by whitespace in order for it to be a comment token.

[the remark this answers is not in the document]
RunDevelopment: Yup. But it is tokenized as a comment right now.

components/prism-factor.js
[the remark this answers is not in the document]
catb0t: I don't know why it's like this, and indeed nobody uses non-printing characters in word names, but the reference implementation recognises only  the space character as horizontal whitespace, and LF/CRLF as vertical whitespace.

components/prism-factor.js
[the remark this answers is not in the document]
catb0t: The `lookbehind: true` option behaves in an unexpected manner with `(?: (lookbehind 1)pattern | (lookbehind 2)pattern )` (i.e. `lookbehind 2` is not treated as lookbehind by Prism; apparently `lookbehind: true` effectively drops the first capturing group from the entire pattern, rather than dropping the first captured group in the match). 

Whatever the case, it's not possible to combine these patterns into a single pattern and still use `lookbehind: true` for each pattern's preceding whitespace, but I have combined them into arrays.

components/prism-factor.js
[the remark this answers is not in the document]
catb0t: The fact that `"!` is matched by `conventionally-named-word` for ending with `!` is a bug. 

Also a bug is that `conventionally-named-word` allows `"` to start a word name; that should be disallowed just like it is in `normal-word`. I'm going to fix  those. 

It's a little bit sad I have to add whitespace lookbehind to `string`, because `"1""2"` is `"1" "2"` in the reference implementation, which is an implementation detail of the special treatment of the builtin string literal. 

The biggest worry is moving `string` before `conventionally-named-word`  -- hopefully it won't be necessary when `conventionally-named-word` correctly disallows words to start with `"`.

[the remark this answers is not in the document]
RunDevelopment: Fixing `conventionally-named-word` will also work. I guess you mean like `(^|\s)(?!")(?:set-\S+|change-\S+|...)(?=\s|$)`, right?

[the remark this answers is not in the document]
catb0t: I should have used the negative lookahead at first instead of adding `"` to the classes that needed it, but yes, that's the fix now.

components/prism-factor.js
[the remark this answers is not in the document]
catb0t: I'm fine with this, because it's hard to do better here than not matching the last erroneous quote mark, without lookaround that I'd rather not use with strings.

On the test page for me, all but the last quote mark in `P" as\\""` is matched.

Screenshot from 2020-02-09 17-13-07

catb0t:
This should be ready for merge, unless there's other changes needed.

tests/languages/factor/comments_feature.test
```diff
@@ -0,0 +1,43 @@
...
+	[ "comment", [ "! ! leading comment-like token" ] ],
+	[ "comment", [ "! whitespace before" ] ],
+	[ "normal-word", "words" ],
+	[ "normal-word", "blah" ],
+	[ "comment", [ "! comment after code on a line" ] ]
+]
+
+----------------------------------------------------
+
+TODO: from the multiline vocab:
```
RunDevelopment: What about this TODO?

catb0t: alright, i'll finish it later today

components/prism-factor.js
```diff
@@ -0,0 +1,303 @@
...
+			}
+		],
+
+		'boolean': {
+			pattern: /(^|\s)(?:t|f)(?=\s|$)/,
+			lookbehind: true
+		},
+
+		'special-string-literal': {
+			pattern: /(^|\s)[A-Z0-9\-]+"\s(?:\\"|\\\S|[^"\\])*"/,
```
RunDevelopment: Please also apply the improvement to `string` here.
```suggestion
			pattern: /(^|\s)[A-Z0-9\-]+"\s(?:\\\S|[^"\\])*"/,
```

catb0t: Haha, I forgot about it because I committed in a hurry from my phone.

tests/languages/factor/comments_feature.test
catb0t: the tests are not written for this yet, i'm sorry it takes so long on weekdays

RunDevelopment: Don't worry about it. It's not like there's a deadline or anything.

components/prism-factor.js
```diff
@@ -0,0 +1,344 @@
...
+			/* from basis/multiline: */
+			{
+				// /* comment */, /* comment*/
+				pattern: /(^|\s)\/\*\s.*?\*\/(?=\s|$)/,
+				lookbehind: true,
+				inside: comment_inside
+			},
+			{
+				// ![[ comment ]] , ![===[ comment]===]
+				pattern: /(^|\s)!\[(?:\[\s.*?]|=\[\s.*?]=|==\[\s.*?]==|===\[\s.*?]===|====\[\s.*?]====|=====\[\s.*?]=====|======\[\s.*?]======)](?=\s|$)/,
```
RunDevelopment: How about this:
(Assuming there can be any number of `=`)

```suggestion
				pattern: /(^|\s)!\[(=*)\[\s.*?\]\2\](?=\s|$)/,
```

catb0t: I didn't know browser javascript regex had backreferences either, but thanks, i've fixed it

catb0t:
(Sorry for force-push, fixing components.js merge conflict)

components/prism-factor.js
```diff
@@ -0,0 +1,371 @@
...
+				// ! single-line exclamation point comments with whitespace after/around the !
+				pattern: /(^|\s)(?:! .*|!$)/,
+				lookbehind: true,
+				inside: comment_inside
+			},
+
+			/* from basis/multiline: */
+			{
+				// /* comment */, /* comment*/
+				pattern: /(^|\s)\/\*\s(?:.|\s)*?\*\/(?=\s|$)/,
```
RunDevelopment: The `(?:.|\s)` is no good. `.` is defined as _all characters except for line breaks_ and `\s` as _all white space characters_ (including line breaks). So `(?:.|\s)` is the same as `[\s\S]` but with the additional problem that `(?:.|\s)` might cause exponential backtracking if the regex engine doesn't optimize.

components/prism-factor.js
```diff
@@ -0,0 +1,370 @@
+(function (Prism) {
+
+	var comment_inside = {
+		function: {
+			pattern: /(\b)(?:TODO|FIX|FIXME|NOTE|BUG|XX+|HACK|\?\?+|!!+)(?=\b|:)/,
```
RunDevelopment: 1. `\b` is zero-width, so you don't need the lookbehind, because it will only ever capture the empty string.
2. `\?{2,}|!{2,}` might be more readable? This is just preference, so if you like the current version better, just ignore this nit.

components/prism-factor.js
```diff
@@ -0,0 +1,370 @@
...
+			lookbehind: true
+		}
+	};
+	var string_inside = {
+		number: /(?:\\[^\s']|%\w)/
```
RunDevelopment: Unnecessary non-capturing group.

```suggestion
		number: /\\[^\s']|%\w/
```

components/prism-factor.js
```diff
@@ -0,0 +1,370 @@
...
+		'comment': [
+			{
+				// ! single-line exclamation point comments with whitespace after/around the !
+				pattern: /(^|\s)(?:! .*|!$)/,
+				lookbehind: true,
+				inside: comment_inside
+			},
+
+			/* from basis/multiline: */
+			{
```
RunDevelopment: Shouldn't all the below comment patterns be greedy?

components/prism-factor.js
```diff
@@ -0,0 +1,370 @@
...
+					The values of MMMMMMMMMMMMM and EEEE map directly to the mantissa and exponent fields of the binary IEEE 754 representation."
+					<https://docs.factorcode.org/content/article-syntax-floats.html>
+				*/
+				pattern: /(^|\s)[+-]?0(?:b1\.[01]*|o1\.[0-7]*|d1\.\d*|x1\.[\dA-F]*)p\d+(?=\s|$)/i,
+				lookbehind: true
+			}
+		],
+
+		'regexp': {
+			pattern: /(^|\s)R\/\s+(?:\\\S|[^/])*?[^\\]\//,
```
RunDevelopment: The backslash rule seems kinda odd because `[^/]` can also match `\`, so this pattern will not match `R/ \\/`. How about this:
```suggestion
			pattern: /(^|\s)R\/\s+(?:\\\S|[^\\/])*\//,
```

components/prism-factor.js
```diff
@@ -0,0 +1,370 @@
...
+			lookbehind: true,
+			alias: 'number',
+			inside: {
+				variable: /(?:\\\S)/,
```
RunDevelopment:
```suggestion
				'variable': /\\\S/,
```

components/prism-factor.js
```diff
@@ -0,0 +1,370 @@
...
+				keyword: /(?!\\)[?*\[\]^$(){}.|]/
```
RunDevelopment: `(?!\\)` will always be true, so you can just remove it. (If you were worried about escaped versions of those (e.g `\?`), then don't be because the above `variable` pattern will already have consumed all of them.)

Also, is `+` missing?

```suggestion
				'keyword': /[?*\[\]^$(){}.|]/
```

components/prism-factor.js
```diff
@@ -0,0 +1,370 @@
...
+			lookbehind: true
+		},
+
+		'custom-string': {
+			pattern: /(^|\s)[A-Z0-9\-]+"\s(?:\\\S|[^"\\])*"/,
+			lookbehind: true,
+			greedy: true,
+			alias: 'string',
+			inside: {
+				number: /(?:\\\S|%\w|\/)/
```
RunDevelopment: Why `number`?

```suggestion
				'number': /\\\S|%\w|\//
```

components/prism-factor.js
```diff
@@ -0,0 +1,370 @@
...
+
+		'other-builtin-syntax': {
+			pattern: null,
+			lookbehind: true,
+			alias: 'operator'
+		},
+
+		'conventionally-named-word': {
+			// for certain word naming conventions like +lt+, sequence?, readers>>, >>setters, writers<<
+			pattern: /(^|\s)(?!")(?:set-\S+|change-\S+|>[^>\s]+|[^:>\s]+>|[^>\s]+>[^>\s]+|\+[^\+\s]+\+|[^?\s]+\?|\?[^?\s]+|[^>\s]+>>|>>[^>\s]+|[^<\s]+<<|\([^()\s]+\)|[^!\s]+!|[^\s]\S*\*|[^\s]\S*\.)(?=\s|$)/,
```
RunDevelopment: `[^\s]` == `\S`, so either make `[^\s]\S*` -> `\S+` or add the suffix to the set (e.g. `[^\s]\S*\*` -> `[^*\s]\S*\*`).
```suggestion
			pattern: /(^|\s)(?!")(?:set-\S+|change-\S+|>[^>\s]+|[^:>\s]+>|[^>\s]+>[^>\s]+|\+[^+\s]+\+|[^?\s]+\?|\?[^?\s]+|[^>\s]+>>|>>[^>\s]+|[^<\s]+<<|\([^()\s]+\)|[^!\s]+!|\S+\*|\S+\.)(?=\s|$)/,
```

components/prism-factor.js
```diff
@@ -0,0 +1,370 @@
...
+			lookbehind: true,
+			alias: 'keyword'
+		},
+
+		'colon-syntax': {
+			pattern: /(^|\s)(?:[A-Z]*#?):{1,2}\s+(?:;\S+|(?!;)\S+)(?=\s|$)/,
```
RunDevelopment:
```suggestion
			pattern: /(^|\s)[A-Z]*#?:{1,2}\s+(?:;\S+|(?!;)\S+)(?=\s|$)/,
```

components/prism-factor.js
catb0t: i missed including the suffix in the set, thanks

components/prism-factor.js
catb0t: thanks! the theme here is that JS RegExp supports a lot more features than expected -- i thought `{n,}` was a Perl RE extension

components/prism-factor.js
catb0t: In my editor, escapes in strings are highlighted like numbers, and I like the high contrast. It seems to be the best semantic fit in my opinion, since there is usually contrast between string and number text colours in Prism themes.

components/prism-factor.js
catb0t: It was supposed to be followed by an omitted `?` quantifier; bug fixed and test added

components/prism-factor.js
catb0t: Thanks!

components/prism-factor.js
catb0t: Being at the top of the file, I don't think it matters, and I can't think of an input that would be classified differently. I've made them greedy for correctness though.

components/prism-factor.js
RunDevelopment: Well, [JS Regex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) doesn't have a lot of features but it does have all a few. Though you can't use all of them because we still want to support older browsers (stuff like RE lookbehinds, Unicode properties, and named capturing groups are all too new).

components/prism-factor.js
RunDevelopment: That's a valid reason.

catb0t:
I forgot about `regexp` flags until now. That's the last change I wanted to get in, so I'm done adding commits unless you request further changes.

components/prism-factor.js
```diff
@@ -84,12 +84,16 @@
...
-			pattern: /(^|\s)R\/\s+(?:\\\S|[^\\/])*\//,
+			pattern: /(^|\s)R\/\s+(?:\\\S|[^\\/])*\/(?:[idmsr]*|[idmsr]+-[idmsr]+)(?=\s|$)/,
 			lookbehind: true,
 			alias: 'number',
 			inside: {
 				'variable': /\\\S/,
-				'keyword': /[+?*\[\]^$(){}.|]/
+				'keyword': /[+?*\[\]^$(){}.|]/,
+				'operator': {
+					pattern: /(\/)[idmsr]+(-[idmsr]+)?/,
```
RunDevelopment: This will also match: `R/ \/i/`.
```suggestion
					pattern: /(\/)[idmsr]+(-[idmsr]+)?$/,
```

catb0t: There's already a passing test against this, and I'm pretty sure the reason it works as-is, is as you said above, `variable` and `keyword` will have consumed the `\/` and not cause `operator` to match `/i` (with lookbehind) in that case.

I can add the end-of-string assertion anyway though.

RunDevelopment: Ah true. I didn't think of the `variable` pattern.

> I can add the end-of-string assertion anyway though.

Not necessary, the test is enough.

components/prism-factor.js
```diff
@@ -0,0 +1,403 @@
...
+		],
+
+		'special-using': {
+			pattern: /(^|\s)USING:(?:\s\S+)*(?=\s+;(?:\s|$))/,
+			lookbehind: true,
+			alias: 'function',
+			inside: {
+				// this is essentially a regex for vocab names, which i don't want to specify
+				// but the USING: gets picked up as a vocab name
+				string: {
```
RunDevelopment:
```suggestion
				'string': {
```

examples/prism-factor.html
```diff
@@ -0,0 +1,86 @@
+<h2>Comments</h2>
+<pre> <code>! a comment</code> </pre>
```
RunDevelopment: Please remove the extra spaces. The other one below as well.

```suggestion
<pre><code>! a comment</code></pre>
```

tests/languages/factor/other_builtin_parsing_words_feature.test
```diff
@@ -0,0 +1,29 @@
...
+	[ "normal-word", "code" ],
+	[ "other-builtin-syntax", ">>" ],
+	[ "other-builtin-syntax", "\\" ],
+	[ "other-builtin-syntax", "M\\" ]
+]
+
+
+--------------------------------------------------
+
+asdf
```
RunDevelopment: Ahhhhh. Yes. Documentation.

tests/languages/factor/strings_feature.test
```diff
@@ -0,0 +1,129 @@
...
+		"P\" as",
+		[ "number", "\\\\" ],
+		"\""
+	] ],
+	"\""
+]
+
+----------------------------------------------------------
+
+TODO:
```
RunDevelopment: I think you can remove the TODO.

RunDevelopment:
Thank you for contributing @catb0t!

catb0t:
Thank you for your incredible patience, and also universal knowledge of regex, @RunDevelopment!

Issue #2200, opened 2020-02-05
Bug or feature? Toolbar's `data-toolbar-reorder` deletes callbacks


RunDevelopment:
While writing #2199, I noticed that [this line](https://github.com/PrismJS/prism/blob/a7d67ca31be4c26c0a6e287769336c78e8e246cf/plugins/toolbar/prism-toolbar.js#L94) reorders the buttons and can also be used to disable buttons since all buttons which aren't in the list won't show up. 

This has been in there since #891, so @mAAdhaTTah: Is it a bug or a feature?

mAAdhaTTah:
I think Imma go with feature. I didn't think about that use case when I did it in the first place, but I think if the user expresses their intent as `[a, b]`, then it should render `a`, then `b`. Adding additional automatically seems to violate their intent. Which kinda suggests we don't need #2199 as the user can use `order` and keep the one(s) they want.

RunDevelopment:
Feature it is then. But we still need #2199 because `data-toolbar-reorder` is actually a global reordering/selection of buttons and will be applied to all elements because it is currently implemented as a `body` attribute (not inherited). So it wouldn't be possible to disable a single button which is [the issue](https://github.com/PrismJS/prism/issues/1381#issuecomment-582516029) #2199 is trying to solve albeit in a more general manner.

That being said, if `data-toolbar-reorder` were to be inherited, it would also solve the issue motivating #2199.

mAAdhaTTah:
I was actually going to suggest that. We do that sort of "inheriting" thing elsewhere, so that seems more in line with the pattern elsewhere.

I am also wondering if it makes sense to not have the toolbar appear on the inline `code` blocks (e.g. blocks without a parent `pre` element) at all, cuz that's usually used for small words or snippets that don't need the toolbar.

RunDevelopment:
Alright, then I'll do that.

> I am also wondering if it makes sense to not have the toolbar appear on the inline code blocks

Uhm. [It doesn't](https://github.com/PrismJS/prism/blob/2e0eff7670dd658aaa74458970e0030fec25db84/plugins/toolbar/prism-toolbar.js#L75).

mAAdhaTTah:
Ahh, I misinterpreted this comment from the original request:

> Copy to Clipboard - Would really prefer to disable it on single line code blocks.

But I get it now. So yeah, updating to inherit would be good.

RunDevelopment:
No problem. I'll close the issue as the initial question has been answered.

PR #2205, opened 2020-02-08
Toolbar: `data-toolbar-order` is now inherited

RunDevelopment:
What the title says. More details in #2200.


plugins/toolbar/prism-toolbar.js
```diff
@@ -81,22 +103,27 @@
...
 		wrapper.appendChild(pre);
 
 		// Setup the toolbar
 		var toolbar = document.createElement('div');
 		toolbar.classList.add('toolbar');
 
-		if (document.body.hasAttribute('data-toolbar-order')) {
-			callbacks = document.body.getAttribute('data-toolbar-order').split(',').map(function(key) {
+		// order callbacks
+		var elementCallbacks;
```
mAAdhaTTah: assign to `callbacks` here then kill the `else` block.

plugins/toolbar/prism-toolbar.js
```diff
@@ -63,6 +63,28 @@
...
+				order = order.trim();
+				if (order.length) {
+					return order.split(/\s*,\s*/g);
+				} else {
+					return [];
+				}
+			}
+			element = element.parentElement;
+		}
+		return undefined;
```
mAAdhaTTah: We don't need to explicitly return `undefined` here. Just kill this `return`.

RunDevelopment: Ah right. JS does that.

RunDevelopment:
Nice. What should we do with #2199? I still think that the `disable` feature might be easier to use that `order` in some cases. But their functionalities really overlap.

mAAdhaTTah:
Honestly, I think we should close it. If the same result can be produced with the current API, it would be better to push users towards that API than have multiple ways of doing the same thing. Plus the associated code size increase.

PR #2209, opened 2020-02-10
Shell-session: Added `console` alias

### LINKED ISSUE #2207: lang-console alias for lang-shell-session
**Motivation**
In Markdown (no idea on which flavors), the "Shell Session" equivalent is called "Console".

**Description**
Could "Console" be added as an alias for "Shell Session"?

RunDevelopment:
This resolves #2207.


mAAdhaTTah:
I'm not so sure I'm into this. `console` doesn't say "shell" to me, and if we were to add this, I could see `terminal` being a valid alias and it starts to feel like we're just adding a bunch of potential names here. Although it's not a ton of work, so maybe it's not a big deal? Thoughts?

RunDevelopment:
I agree. I haven't merged this yet because I don't really like the `console` alias either.
It's incredibly non-specific and doesn't imply bash/shell in any way. The people that do PowerShell for a living will most likely not think of bash if they hear "console". Technically, the output of every console application fits this label.

mAAdhaTTah:
Yeah... ok, I'm going to close this.

PR #2216, opened 2020-02-17
A test for examples

RunDevelopment:
This includes two things:

The examples checks are now their own test file, so the example tests are now part of `npm test` and not the gulp `premerge` task. I can't remember why I made it part of `permerge` in the first place. No comments in #2128 either.

Added a new check to validate the example files. It's rather strict.

---

The main intention of strict validation is to make my life as a reviewer easier. Many people (myself included) will add example files with unescaped "<" characters and other problems. 
Those problems are hard to spot for humans, so I made this test. It will also speed up the code review process because almost all people will figure out how to make the tests pass.


examples/prism-markup.html
```diff
@@ -72,6 +72,6 @@ <h2>Multi-line attribute values</h2>
 baz"></code></pre>
 
 <h2>XML tags with non-ASCII characters</h2>
-<pre><code>&lt;L&auml;ufer&gt;foo&lt;/L&auml;ufer&gt;
-&lt;tag l&auml;ufer="l&auml;ufer"&gt;bar&lt;/tag&gt;
-&lt;l&auml;ufer:tag&gt;baz&lt;/l&auml;ufer:tag&gt;</code></pre>
\ No newline at end of file
+<pre><code>&lt;Läufer&gt;foo&lt;/Läufer&gt;
+&lt;tag läufer="läufer"&gt;bar&lt;/tag&gt;
+&lt;läufer:tag&gt;baz&lt;/läufer:tag&gt;</code></pre>
```
RunDevelopment: UTF-8 should be supported in this day and age, so there's no need for the HTML entities.

Issue #2214, opened 2020-02-17
YAML anchors with hyphens are not correctly highlighted


toaster:
**Information**
- Language: YAML
- Plugins: none

**Description**

In a YAML document containing anchors with hyphens only the part of the anchor up to the first hyphen is highlighted.

**Code snippet**

<details>
<summary>The code being highlighted incorrectly.</summary>

```
x-foo-bar: &foo-bar 13
```

</details>

toaster:
@RunDevelopment you're definitely fast! :)

However, #2215 will break some other cases. Therefore I propose to use #2217 instead.

PR #2217, opened 2020-02-18
fix highlighting of YAMLs anchor & alias

toaster:
more sophisticated fix for #2214


components/prism-yaml.js
```diff
@@ -16,32 +16,32 @@ Prism.languages.yaml = {
...
 	},
 	'boolean': {
-		pattern: /([:\-,[{]\s*(?:![^\s]+)?[ \t]*)(?:true|false)[ \t]*(?=$|,|]|})/im,
+		pattern: /([:\-,[{]\s*(?:![^\s]+)?(?:[&*][^\s\[\]{},]+)?[ \t]*)(?:true|false)[ \t]*(?=$|,|]|})/im,
 		lookbehind: true,
 		alias: 'important'
 	},
 	'null': {
-		pattern: /([:\-,[{]\s*(?:![^\s]+)?[ \t]*)(?:null|~)[ \t]*(?=$|,|]|})/im,
+		pattern: /([:\-,[{]\s*(?:![^\s]+)?(?:[&*][^\s\[\]{},]+)?[ \t]*)(?:null|~)[ \t]*(?=$|,|]|})/im,
```
RunDevelopment: I don't really use YAML, so please help me here: The `(?:![^\s]+)?` is supposed to match [tags](http://blogs.perl.org/users/tinita/2018/01/introduction-to-yaml-schemas-and-tags.html), right?

So I have a few questions:

1. Is this valid: `foo: !!tag &anchor null` ? (not matched right now)
2. Is this valid: `foo: !!tag,&,an{}chor null` ? (matched)
3. Does this contain an anchor: `foo: !!tag&anchor null` ? (matched)

Nr. 2 and 3 are matched because the `[^\s]+` will consume the "anchor".

tests/languages/yaml/anchor_and_alias_feature.test
```diff
@@ -0,0 +1,59 @@
+x-number: &num-ber 13
+x-string: &stri_ng "good"
+x-null: &nu-ll null
```
RunDevelopment: Please also add a case with tags.

components/prism-yaml.js
toaster: 1. Yes, I'm working on it.
2. This is not an anchor (and not matched as such). If this is a valid tag, everything's fine. If not, fixing tag matching is out-of-scope for this PR.
3. The same as 2.

RunDevelopment: > If not, fixing tag matching is out-of-scope for this PR.

Fair enough. I looked it up in the spec, and yes, it's a bug. I'll make a PR for this issue after we merge this.

Regarding 1.:
Just make it this here. It'll work.

```suggestion
		pattern: /([:\-,[{]\s*(?:![^\s]+[ \t]*)?(?:[&*][^\s\[\]{},]+[ \t]*)?)(?:null|~)[ \t]*(?=$|,|]|})/im,
```

components/prism-yaml.js
```diff
@@ -16,32 +16,32 @@ Prism.languages.yaml = {
...
 	},
 	'boolean': {
-		pattern: /([:\-,[{]\s*(?:![^\s]+)?[ \t]*)(?:true|false)[ \t]*(?=$|,|]|})/im,
+		pattern: /([:\-,[{]\s*(?:\s[!&*][^\s]+)*[ \t]*)(?:true|false)[ \t]*(?=$|,|]|})/im,
 		lookbehind: true,
 		alias: 'important'
 	},
 	'null': {
-		pattern: /([:\-,[{]\s*(?:![^\s]+)?[ \t]*)(?:null|~)[ \t]*(?=$|,|]|})/im,
+		pattern: /([:\-,[{]\s*(?:\s[!&*][^\s]+)*[ \t]*)(?:null|~)[ \t]*(?=$|,|]|})/im,
```
RunDevelopment: Unfortunately, this won't match `foo:!!tag null` and there can be [at most 2](https://yaml.org/spec/1.2/spec.html#c-ns-properties(n,c)).

```suggestion
		pattern: /([:\-,[{]\s*(?:(?:![^\s]+|[*&][^\s[\]{},]+)[ \t]*){0,2})(?:null|~)[ \t]*(?=$|,|]|})/im,
```

toaster: Obviously there are a lot of cases that weren't tested before :(.

RunDevelopment: Yes, but here it's a case which was highlighted before and now isn't. We generally can't expect people to format their documents properly (that'd be too easy...), so our language definitions have to robust enough to be able to deal with situations where non-essential whitespaces are missing.

And yeah. We definitely need more tests for YAML... It's such a complex language and yet we have so few test cases for it.

toaster: And yes, there are at most 2 but we don't check if it's max 1 anchor or alias, we don't check if it is max 1 tag, we don't check if anchor _and_ alias are given (which probably is invalid).
There are a lot of semantic cases that aren't checked here. But this is neither the goal nor even possible with regexps.

RunDevelopment: Well, yes. The goal is just to match the language as close as possible within reasonable bounds (e.g. that regexes that are less than ~200 chars long and at least a little understandable).

That why I suggested the above pattern. It's not perfect but it's closer to the ground truth compared to the current pattern.

toaster: Oh, `foo:!!tag null` is invalid.

toaster: Okay, it's valid but it's a scalar:
```ruby
[10] pry(main)> YAML.parse("foo:!!tag null")
=> #<Psych::Nodes::Document:0x00007f9cb9c72c90
 @children=[#<Psych::Nodes::Scalar:0x00007f9cb9c72c18 @anchor=nil, @plain=true, @quoted=false, @style=1, @tag=nil, @value="foo:!!tag null">],
 @implicit=true,
 @implicit_end=true,
 @tag_directives=[],
 @version=[]>
[11] pry(main)> YAML.parse("foo: !!tag null")
=> #<Psych::Nodes::Document:0x00007f9cb9c92fe0
 @children=
  [#<Psych::Nodes::Mapping:0x00007f9cb9c92f90
    @anchor=nil,
    @children=
     [#<Psych::Nodes::Scalar:0x00007f9cb9c92f18 @anchor=nil, @plain=true, @quoted=false, @style=1, @tag=nil, @value="foo">,
      #<Psych::Nodes::Scalar:0x00007f9cb9c92ea0 @anchor=nil, @plain=false, @quoted=false, @style=1, @tag="tag:yaml.org,2002:tag", @value="null">],
    @implicit=true,
    @style=1,
    @tag=nil>],
 @implicit=true,
 @implicit_end=true,
 @tag_directives=[],
 @version=[]>
```

toaster: So it doesn't have to match here.

RunDevelopment: That's unexpected. In that case, your pattern is better.

RunDevelopment:
Please rebuild, so the CI check passes and we can merge this.

RunDevelopment:
Thank you for contributing @toaster!

PR #2218, opened 2020-02-18
added empty prism-modx js file and added modx to components json

frankie-tech:


frankie-tech:
my bad, meant to do this with my local copy, please ignore

PR #2221, opened 2020-02-20
Add support for LLVM IR

### LINKED ISSUE #2109: Language Request: LLVM IR
**Language**
LLVM IR is the internal language of the LLVM compiler toolchain. It's nice to have it highlighted while writing about compilers.

**Additional resources**
https://llvm.org/docs/LangRef.html

porglezomp:
LLVM IR has tons of keywords, has prefixed variables, and is relatively unstable, so it makes sense to consider everything outside of a small set of exceptions to be keywords.

Fixes #2109


RunDevelopment:
I will review this shortly. 

In the mean-time: The build doesn't pass because some of the generated files weren't updated. 
Please run `npx gulp` to rebuild Prism properly and update/create all minified files.

porglezomp:
Ah thanks, I wasn't sure if they should be checked in for conflict reasons. I went and looked at some other older language additions and didn't see those files being updated at the time so I wasn't sure.

components/prism-llvm.js
```diff
@@ -0,0 +1,12 @@
+(function(Prism) {
+	Prism.languages.llvm = {
+		'comment': /;.*/,
+		'string': /"[^"]*"/,
+		'variable': /[%@!#](?:(?:[-a-zA-Z$._]|\\[a-fA-F0-9]{2})(?:[-a-zA-Z$._0-9]|\\[a-fA-F0-9]{2})*|[0-9]+)/,
+		'label': /(?:[-a-zA-Z$._]|\\[a-fA-F0-9]{2})(?:[-a-zA-Z$._0-9]|\\[a-fA-F0-9]{2})*:/,
+		'type': /\b(?:void|label|token|metadata|float|double|half)\b/,
+		'keyword': /[a-z_][a-z_0-9]*/,
+		'number': /[+-]?[0-9]+(?:.[0-9]+)?(?:[eE][+-]?[0-9]+)?|0x[0-9A-Fa-f]+/,
```
RunDevelopment: Is `-0xABC` not a number?

Also, please add a test specifically for numeric literals like [this one](https://github.com/PrismJS/prism/blob/master/tests/languages/javascript/number_feature.test).

```suggestion
		'number': /[+-]?\b\d+(?:\.\d+)?(?:[eE][+-]?\d+)?\b|\b0x[\dA-Fa-f]+\b/,
```

components/prism-llvm.js
```diff
@@ -0,0 +1,12 @@
+(function(Prism) {
+	Prism.languages.llvm = {
+		'comment': /;.*/,
+		'string': /"[^"]*"/,
+		'variable': /[%@!#](?:(?:[-a-zA-Z$._]|\\[a-fA-F0-9]{2})(?:[-a-zA-Z$._0-9]|\\[a-fA-F0-9]{2})*|[0-9]+)/,
+		'label': /(?:[-a-zA-Z$._]|\\[a-fA-F0-9]{2})(?:[-a-zA-Z$._0-9]|\\[a-fA-F0-9]{2})*:/,
+		'type': /\b(?:void|label|token|metadata|float|double|half)\b/,
+		'keyword': /[a-z_][a-z_0-9]*/,
```
RunDevelopment: 1. We need to checks for word boundaries or numbers like `0xabc` will be partially matched as `keyword`.
2. Is it really a good idea to highlight every literal we haven't matched just as a `keyword`? We often use long keyword lists for this to avoid false positives. Could this be applied here as well? (I'm asking because I have no idea about LLVM.)

components/prism-llvm.js
```diff
@@ -0,0 +1,12 @@
+(function(Prism) {
+	Prism.languages.llvm = {
+		'comment': /;.*/,
+		'string': /"[^"]*"/,
```
RunDevelopment: Should the `c` prefix be also highlighted? 
E.g. `c"foo"`

And please make it greedy or strings like `"oh ; no"` will not be highlighted correctly.

components/prism-llvm.js
```diff
@@ -0,0 +1,12 @@
...
+		'variable': /[%@!#](?:(?:[-a-zA-Z$._]|\\[a-fA-F0-9]{2})(?:[-a-zA-Z$._0-9]|\\[a-fA-F0-9]{2})*|[0-9]+)/,
+		'label': /(?:[-a-zA-Z$._]|\\[a-fA-F0-9]{2})(?:[-a-zA-Z$._0-9]|\\[a-fA-F0-9]{2})*:/,
+		'type': /\b(?:void|label|token|metadata|float|double|half)\b/,
```
RunDevelopment: 1. Please sort this list to improve the readability.
2. Aren't there many types missing? I mean all of the [integer types](https://llvm.org/docs/LangRef.html#integer-type) are highlighted as `keyword`, a few [floating-point types](https://llvm.org/docs/LangRef.html#floating-point-types) are missing, and [this thing](https://llvm.org/docs/LangRef.html#x86-mmx-type) is also missing.
3. `type` isn't a colored by most themes, so please give it an alias `class-name`.

components/prism-llvm.js
porglezomp: LLVM has thousands of keywords and doesn't make fine-grained stability promises, adding and removing them especially in extensions. Also, this looks like it's what they're doing in the syntax highlighting they use in their examples.

RunDevelopment: Alright. That makes things simpler for us.
Thank you for sharing your insight.

```suggestion
		'keyword': /\b[a-z_][a-z_0-9]*\b/,
```

porglezomp:
Thanks for your detailed review, sorry for the delay in making those changes.

components/prism-llvm.js
porglezomp: Interestingly -0xABC isn't a number because hexadecimal is only used for bit-exact floating point literals.

tests/languages/llvm/edge-cases.test
```diff
@@ -0,0 +1,11 @@
+" ; "
```
RunDevelopment: This might as well be a `string` test with strings like `""`, `"\00\FF"`, `"/path/to"`, and `c"foo"`.

porglezomp: I'm gonna note here that the highlighting the LLVM LangRef keeps the `c` in the `c"bytes"` literals separately, which is why I left it this way.

RunDevelopment: That's fair.

RunDevelopment:
Thank you for contributing @porglezomp!

PR #2222, opened 2020-02-20
Added test to sort the language list

RunDevelopment:
When adding a new language, there's always the question of where to put it in the language list. 
The language list is generally sorted alphabetically, so I added a test which exactly that with some slight modification to the normal ASCII order with the goal to have a sensible language order.

I also changed some title which I felt where too long (e.g. "Augmented Backus–Naur form" -> "ABNF") or to make easier to find (e.g. "Gamemaker Language" -> "GML: Gamemaker Language").

This should also make code review easier because misplaced entries in the language list are quite common.


components.json
```diff
@@ -79,7 +79,7 @@
 			"owner": "dellagustin"
 		},
 		"abnf": {
-			"title": "Augmented Backus–Naur form",
+			"title": "ABNF",
```
mAAdhaTTah, line 82: Maybe this is better as an alias?

components.json
```diff
@@ -274,7 +274,7 @@
 			"owner": "JustinBeckwith"
 		},
 		"ebnf": {
-			"title": "Extended Backus–Naur form",
+			"title": "EBFN",
```
mAAdhaTTah, line 277: Same, alias?

components.json
RunDevelopment: Agreed.

RunDevelopment: Ahh, no. Looking at it again, I remembered the reason why I didn't add this as an alias in the first place: I can't. An alias title requires at least one alias (of the language id) or else the alias test will fail.

RunDevelopment:
I resolved the conflicts.

Regarding aliases: I just reverted my name changes.

RunDevelopment:
I'm such an idiot. Maybe I should also read the suggested changes?

Yeah. Let me just change the titles to ABFN, BFN, and EBNF. I think people will know what the abbreviations mean and for those that don't, there's Google.

components.json
mAAdhaTTah: Ok, so that makes sense but then why prefer `ABNF` over the long-form?

mAAdhaTTah: lol I need to read everything before commenting. I don't know anything about these languages so I'm not sure whether the abbreviations are more common. I'm ok with this if you are

RunDevelopment: I was thinking that if I searched for this language, I'd put "ABNF" into my search box. But for example, HighlightJS offers it under the long name. 

We don't really have rules for language names and the use of alias titles.

RunDevelopment: Alright, your last comment just reached me... 
I'll just go with the abbreviations for now.

PR #2227, opened 2020-02-22
gulp: Inline regex source improvement

RunDevelopment:
I noticed that regexes which contained `\r\n` did not have their source inline correctly. E.g. `/[\r\n]/.source` mapped to `"[\r\\n]"` instead of `"[\r\n]"`.

This fixes this minor issue.


RunDevelopment:
If it's ok, I'll just merge this. @mAAdhaTTah @Golmote 
It's just a minor fix after all.

PR #2233, opened 2020-02-25
Website: Fix diff highlight plugin page title

sidvishnoi:
https://prismjs.com/plugins/diff-highlight/ shows "Data-URI highlight" as page's `<title>`


RunDevelopment:
Thank you for fixing this @sidvishnoi!

PR #2236, opened 2020-02-27
Add C++20 keywords

shilch:
A list can be found at https://en.cppreference.com/w/cpp/keyword


RunDevelopment:
Thank you for submitting this pull request @shilch!

Before we can merge this, you have to rebuild Prism (using the command `npm ci && npx gulp`) and add the new keywords to [C++' keyword test](https://github.com/PrismJS/prism/blob/master/tests/languages/cpp/keyword_feature.test).

Issue #1132, opened 2017-05-03
Alignment bug with line numbers


mat-lentils:
Just wanted to leave a short note, that the current release version including the line numbers plugin has a small alignment bug. I've tested it in Chrome 58, FF Nightly and IE Edge, each of them displays the line numbers 2 Pixels below the base line. With some doctypes + IE the alignment gets even worse and is a whole line height below the base line.

The solution for my project is actually pretty straight forward: Add display: inline-block to the <code> tag.

Again, this was only tested in the browsers mentioned above.

Golmote:
Hi! This is a subtle one, but you're actually right.
Unfortunately, the fix you propose won't work well with the Coy theme.

PR #2237, opened 2020-02-29
Line highlight: Fixed top offset in combination with Line numbers

RunDevelopment:
This adds support for extra elements above the `<code>` element in combination with Line highlight and Line numbers.

<details>

image

</details>

---

However, there is one problem I wasn't able to fix: The slight error of the alignment of the line highlight. The cause for the issue is that the `<code>` element isn't perfectly aligned with the `<pre>` element (or any elements before it) which means that the line numbers are a few pixels too off. See #1132.
Because the line highlight is aligned by measuring the line numbers, it's off as well.
You can see this issue everywhere on the [Line numbers page](https://prismjs.com/plugins/line-numbers).

<details>

image

</details>

There is actually another fix for #1132 apart from the one proposed by the reporter of the issue: Set `display: block` for the `<code>` element. 
This is compatible with Coy but not with Funky. I also tried a few different things but none of them worked.

If possible, I'd like to include the fix of the alignment here as well (because it's __very__ visible after this update) but I have no idea how to fix this.


RunDevelopment:
@siavashs I'm sorry but I can't figure out how to fix the alignment issue, so this pull request might take a while.

plugins/line-highlight/prism-line-highlight.js
```diff
@@ -39,35 +39,72 @@
...
+		 * Returns the numeric value of the given pixel value.
+		 *
+		 * @param {string} px
+		 */
+		function pxToNumber(px) {
+			return +px.substr(0, px.length - 2);
+		}
+
+		return child.offsetTop + pxToNumber(childStyle.borderTopWidth) + pxToNumber(childStyle.paddingTop)
+			- pxToNumber(parentStyle.paddingTop);
```
mAAdhaTTah, line 62: NItty style suggestion: breaking one operation onto a separate line like this makes it a bit tough to read. All on one line (so I can scan left to right) or breaking them all (so I can scan top to bottom) would be easier to read.

mAAdhaTTah (approved):
This is basically gtg w/ a style nit + a refresh.

PR #2248, opened 2020-03-11
Add support for DAX

### LINKED ISSUE #2121: Support for DAX (Data Analysis Expressions)
**Language**
Data Analysis Expressions (DAX) is a library of functions and operators that can be combined to build formulas and expressions in Power BI Desktop, Azure Analysis Services, SQL Server Analysis Services, and Power Pivot in Excel.

**Additional resources**
[https://docs.microsoft.com/en-us/dax/](https://docs.microsoft.com/en-us/dax/)

peterbud:
Adding syntax highlighting for DAX (Data Analysis Expressions)

Pls note this is my first PR to this repo. Let me know if you would like to further optimize the keywords section.

Fixes #2121


components/prism-dax.js
```diff
@@ -0,0 +1,32 @@
...
+    },
+    'keyword': {
+        pattern: /\b(?:DEFINE|MEASURE|EVALUATE|ORDER BY|RETURN|VAR|START AT|ASC|DESC)\b/i
```
RunDevelopment: Can there by any number of spaces between keyword? E.g. `"ORDER\x20\x20BY"` (I can't do multiple spaces in inline code (thanks Markdown), so I used `\x20` instead.)

And what about tabs? Do they work?

tests/languages/dax/datafield_feature.test
```diff
@@ -0,0 +1,21 @@
+'table'[field]
+'table name'[field]
+'table'[field name]
+'table name'[field name]
+table[field]
+table[field name]
```
RunDevelopment: No tests for the `'[\w]+'` case.

examples/prism-dax.html
```diff
@@ -0,0 +1,42 @@
+<h2>Comments</h2>
+<pre><code>// This is a comment</code></pre>
+
+<h2>Simple example</h2>
+<pre><code>
+    Sales YTD :=
```
RunDevelopment: Is there any reason why all lines are indented?

examples/prism-dax.html
```diff
@@ -0,0 +1,42 @@
...
+<pre><code>
+    Sales YTD := 
+    CALCULATE (
+        [Sales Amount], 
+        DATESYTD( 'Date'[Date] )
+    )
+</code></pre>
+
+<h2>Full example</h2>
+<pre><code>
```
RunDevelopment: Please remove the leading newlines.

components/prism-dax.js
```diff
@@ -0,0 +1,32 @@
+Prism.languages.dax = {
+    'comment': {
+        pattern: /(^|[^\\])(?:\/\*[\s\S]*?\*\/|(?:--|\/\/).*)/,
+        lookbehind: true
+    },
+    'string': {
+        pattern: /(")(?:(?!\1)[^\\\r\n])*\1/,
```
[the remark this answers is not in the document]
peterbud: Thanks for the review! Unfortunately you are right: the DAX specification is not a great definition. I have been using DAX for a few years now, and I had the impression that string literals cannot be multi-line, and when I have now tested it turned out they actually can be. So I have modified the regex pattern and added test cases appropriately.

components/prism-dax.js
```diff
@@ -0,0 +1,32 @@
...
+        greedy: true
+    },
+    'measure': {
+        pattern: /([^'\w])\[\b.*?\b\]/,
+        lookbehind: true,
+        alias: "constant"
```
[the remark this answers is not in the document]
peterbud: Done

components/prism-dax.js
```diff
@@ -0,0 +1,32 @@
...
+    },
```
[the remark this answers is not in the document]
peterbud: Thanks for the advice, it is done as well.

components/prism-dax.js
peterbud: Same as above: never tried before, but now I have done some test and it is allowed to have arbitrary number of whitespace and even newline between `ORDER `and `BY`. I have updated the pattern and also added testcases for that.

examples/prism-dax.html
peterbud: OK, I have fixed these as well.

components/prism-dax.js
```diff
@@ -0,0 +1,28 @@
...
+    'keyword': /\b(?:DEFINE|MEASURE|EVALUATE|ORDER[\s]+BY|RETURN|VAR|START[\s]+AT|ASC|DESC)\b/i,
```
RunDevelopment:
```suggestion
    'keyword': /\b(?:DEFINE|MEASURE|EVALUATE|ORDER\s+BY|RETURN|VAR|START\s+AT|ASC|DESC)\b/i,
```

peterbud: Right! I'm still learning regex, apologies. Fixed

tests/languages/dax/datafield_feature.test
peterbud: Added test cases

components/prism-dax.js
```diff
@@ -0,0 +1,32 @@
...
+    'comment': {
+        pattern: /(^|[^\\])(?:\/\*[\s\S]*?\*\/|(?:--|\/\/).*)/,
+        lookbehind: true
+    },
+    'string': {
+        pattern: /(")(?:(?!\1)[^\\\r\n])*\1/,
+        greedy: true
+    },
+    'measure': {
+        pattern: /([^'\w])\[\b.*?\b\]/,
```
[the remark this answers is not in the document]
peterbud: Regex fixed. Also I followed your suggestion and swapped `data-field` and `measure`

components/prism-dax.js
```diff
@@ -0,0 +1,32 @@
...
+        lookbehind: true,
+        alias: "constant"
+    },
+    'data-field': {
+        pattern: /'[\w ]+'\[[^\]]+\]|[\w]+\[[^\]]+\]|'[\w]+'/,
```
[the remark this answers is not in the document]
peterbud: Turned out that table and field names can contain special character like `'` or `"`, but new line is not allowed, I have tested it. Re-wrote the regex from scratch added new test cases so pls check it once more.

components/prism-dax.js
```diff
@@ -0,0 +1,27 @@
+Prism.languages.dax = {
+    'comment': {
+        pattern: /(^|[^\\])(?:\/\*[\s\S]*?\*\/|(?:--|\/\/).*)/,
+        lookbehind: true
+    },
+    'data-field': {
+        pattern: /'[\w '"]+'(?:\[(?:[^\]])+\](?!\])){0,1}|\w+\[(?:[^\]])+\](?!\])/,
```
RunDevelopment: From my understanding, there can be basically any character in between the single quotes or square brackets, right? I'm kinda shocked that it allows single quotes inside single quotes cause it's just... weird.

Do field and table names allow other non-Latin characters (e.g. Japanese like こんにちは)?

Also, what's the reason behind the `(?!\])`?

(And please use `?` instead of `{0,1}`)

peterbud: I was also surprised, you can have table names with `'` or `"` on them. Also Unicode character can be in field or table names. I have highlighted them as an example below. What is the best way to change regex pattern for handling Unicode characters?
image

RunDevelopment: In your example, it seems like single-quoted table names behave like strings (you even have to do the `''` to escape one `'`) and the [spec reads like this as well](https://docs.microsoft.com/en-us/dax/dax-syntax-reference#tables).

For the column names, we can use `[ \w\xA0-\uFFFF]` to approximate all Unicode letters plus spaces. So for column names, we get `\[[ \w\xA0-\uFFFF]+\]`. (Should this be copied to `measure`?)

```suggestion
        pattern: /'(?:[^']|'')*'(?!')(?:\[[ \w\xA0-\uFFFF]+\])?|\w+\[[ \w\xA0-\uFFFF]+\]/,
```

peterbud: Thanks for this help. I have made the changes, updated test cases

components/prism-dax.js
```diff
@@ -0,0 +1,27 @@
...
+    'comment': {
+        pattern: /(^|[^\\])(?:\/\*[\s\S]*?\*\/|(?:--|\/\/).*)/,
+        lookbehind: true
+    },
+    'data-field': {
+        pattern: /'(?:[^']|'')*'(?!')(?:\[[ \w\xA0-\uFFFF]+\])?|\w+\[[ \w\xA0-\uFFFF]+\]/,
+        alias: 'symbol'
+    },
+    'measure': {
+        pattern: /[[ \w\xA0-\uFFFF]+\]/,
```
RunDevelopment: Almost slipped by. Maybe some tests for `measure` as well?

```suggestion
        pattern: /\[[ \w\xA0-\uFFFF]+\]/,
```

peterbud: Indeed that was missing. Added as well.

RunDevelopment:
Thank you for contributing @peterbud!

PR #2250, opened 2020-03-15
Add support for Power Query

peterbud:
Adding syntax highlighting for Power Query.
[Documentation](https://docs.microsoft.com/en-us/powerquery-m/)
[Specification](https://docs.microsoft.com/en-us/powerquery-m/power-query-m-language-specification)


components/prism-powerquery.js
```diff
@@ -0,0 +1,88 @@
...
+        /\bValue\.(?:Add|Compare|Divide|Equals|Metadata|Multiply|NativeQuery|NullableEquals|RemoveMetadata|ReplaceMetadata|Subtract|Type)\b/,
+        /\b(?:Byte|Currency|Decimal|Double|Guid|Int8|Int16|Int32|Int64|Percentage|Single)\.From\b/,
+        /\bJson.FromValue\b/
+    ],
+    'data-type':
+    {
+        pattern: /\b(?:any|anynonnull|binary|date|datetime|datetimezone|duration|function|list|logical|none|number|record|table|text|time|type)\b/,
+        alias: 'variable'
+    },
+    'number': /\b0x[\da-f]+\b|[+-]?\b\d+\.?\d*e?\d*|\B\.\d+\b/i,
```
RunDevelopment: The [spec](https://docs.microsoft.com/en-us/powerquery-m/power-query-m-language-specification) has a nice grammar for number on pages 20-21.

```suggestion
    'number': /\b0x[\da-f]+\b|[+-]?(?:\b\d+\.|\B\.)?\d+(?:e[+-]?\d+)?/i,
```

Could you also add the following numbers to the numbers test, please?
```
123
123e45
123e+45
123e-45
123E45
123E+45
123E-45
123.78e-45
.78e-45
```

components/prism-powerquery.js
```diff
@@ -0,0 +1,88 @@
...
+    'operator': /[-+*\/&?@^]|<(?:=>?|>)?|>(?:=?)|=(?:>?)/,
```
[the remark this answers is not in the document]
RunDevelopment: On page 106, it also lists `..` (for lists) and `...` as operators. What about them?

components/prism-powerquery.js
```diff
@@ -0,0 +1,88 @@
...
+
+Prism.languages.powerquery = {
+    'comment': {
+        pattern: /(^|[^\\])(?:\/\*[\s\S]*?\*\/|(?:\/\/).*)/,
+        lookbehind: true
+    },
+    'quoted-identifier': {
+        pattern: /#"(?:[^"\r\n]|"")*"(?!")/,
+        greedy: true,
+        alias: 'constant'
```
RunDevelopment: Aren't these just variable names? Why the `constant` alias?

components/prism-powerquery.js
```diff
@@ -0,0 +1,88 @@
...
+        /\bJoinSide\.(?:Left|Right)\b/,
+        /\bPrecision\.(?:Double|Decimal)\b/,
+        /\bRelativePosition\.(?:From(?:End|Start))\b/,
+        /\bTextEncoding\.(?:Ascii|BigEndianUnicode|Unicode|Utf8|Utf16|Windows)\b/,
+        /\bValue\.(?:As|Is|ReplaceType)\b/,
+        /\b(?:Any|Binary|Date|DateTime|DateTimeZone|Duration|Int8|Int16|Int32|Int64|Function|List|Logical|None|Number|Record|Table|Text|Time)\.Type\b/,
+        /\b(?:true|false|null)\b/
+    ],
+    'keyword': /\b(?:and|as|each|else|error|if|in|is|let|meta|not|nullable|optional|or|otherwise|section|shared|then|try|type)\b|#(?:binary|date|datetime|datetimezone|duration|infinity|nan|sections|shared|table|time)\b/,
+    'function': [
```
RunDevelopment: Can't we identify functions using the call syntax? 
Wouldn't the following also work:
```js
{
    pattern: /(^|[^#\w.])(?!\d)[\w.]+(?=\s*\()/,
    lookbehind: true
}
```
Or is there another language feature which also uses parenthesis I didn't consider?

components/prism-powerquery.js
```diff
@@ -0,0 +1,88 @@
...
+        /\bGroupKind\.(?:Global|Local)\b/,
+        /\bExtraValues\.(?:List|Ignore|Error)\b/,
+        /\bJoinAlgorithm\.(?:Dynamic|PairwiseHash|SortMerge|LeftHash|RightHash|LeftIndex|RightIndex)\b/,
+        /\bJoinSide\.(?:Left|Right)\b/,
+        /\bPrecision\.(?:Double|Decimal)\b/,
+        /\bRelativePosition\.(?:From(?:End|Start))\b/,
+        /\bTextEncoding\.(?:Ascii|BigEndianUnicode|Unicode|Utf8|Utf16|Windows)\b/,
+        /\bValue\.(?:As|Is|ReplaceType)\b/,
+        /\b(?:Any|Binary|Date|DateTime|DateTimeZone|Duration|Int8|Int16|Int32|Int64|Function|List|Logical|None|Number|Record|Table|Text|Time)\.Type\b/,
+        /\b(?:true|false|null)\b/
```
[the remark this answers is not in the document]
peterbud: OK, moved to a separate feature. Added also tests

components/prism-powerquery.js
```diff
@@ -0,0 +1,88 @@
...
+        /\bRoundingMode\.(?:AwayFromZero|Down|ToEven|TowardZero|Up)\b/,
+        /\bMissingField\.(?:Error|Ignore|UseNull)\b/,
+        /\bQuoteStyle\.(?:Csv|None)\b/,
+        /\bJoinKind\.(?:Inner|LeftOuter|RightOuter|FullOuter|LeftAnti|RightAnti)\b/,
+        /\bGroupKind\.(?:Global|Local)\b/,
+        /\bExtraValues\.(?:List|Ignore|Error)\b/,
+        /\bJoinAlgorithm\.(?:Dynamic|PairwiseHash|SortMerge|LeftHash|RightHash|LeftIndex|RightIndex)\b/,
+        /\bJoinSide\.(?:Left|Right)\b/,
+        /\bPrecision\.(?:Double|Decimal)\b/,
+        /\bRelativePosition\.(?:From(?:End|Start))\b/,
```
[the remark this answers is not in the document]
peterbud: Good catch!

components/prism-powerquery.js
```diff
@@ -0,0 +1,88 @@
...
+        /\bText\.(?:At|(?:After|Before)Delimiter|BetweenDelimiters|Clean|Combine|Contains|End(?:sWith)?|From(?:Binary)?|InferNumberType|Insert|Length|Lower|Middle|NewGuid|Pad(?:End|Start)|PositionOf(?:Any)?|Proper|Range|Remove(?:Range)?|Repeat|Replace(?:Range)?|Reverse|Select|Split(?:Any)?|Start(?:sWith)?|To(?:Binary|List)|Trim(?:End|Start)?|Upper)\b/,
+        /\bTime\.(?:EndOfHour|From(?:Text)?|Hour|Minute|Second|StartOfHour|To(?:Record|Text))\b/,
+        /\bType\.(?:AddTableKey|ClosedRecord|Facets|For(?:Function|Record)|Function(?:Parameters|RequiredParameters|Return)|Is(?:Nullable|OpenRecord)?|ListItem|NonNullable|OpenRecord|RecordFields|Replace(?:Facets|TableKeys)|Table(?:Column|Keys|Row|Schema)|Union)\b/,
+        /\bUri\.(?:BuildQueryString|Combine|EscapeDataString|Parts)\b/,
+        /\bValue\.(?:Add|Compare|Divide|Equals|Metadata|Multiply|NativeQuery|NullableEquals|RemoveMetadata|ReplaceMetadata|Subtract|Type)\b/,
+        /\b(?:Byte|Currency|Decimal|Double|Guid|Int8|Int16|Int32|Int64|Percentage|Single)\.From\b/,
+        /\bJson.FromValue\b/
+    ],
+    'data-type':
+    {
```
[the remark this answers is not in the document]
peterbud: Done

components/prism-powerquery.js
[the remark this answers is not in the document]
peterbud: Interestingly, I know `..` but I'm not sure where the tripe dot operator `...` is used in real life. Nevertheless I have added them to the pattern, and added test cases for them as well

components/prism-powerquery.js
peterbud: OK changed to `variable`

components/prism-powerquery.js
peterbud: That is much better, I made this change, but still there is one issue: This pattern above is now recognizing the double dot `..` operator. Example:
```
AA = .12,
BB = {1..100},
```
It recognizes 1.**.100** as a number (aka it includes one dot from the double dot operator in the number token.

I have tried to fix this, but I'm not so good with regex. As I see I cannot use `(?<!\.)` - could you maybe suggest a good solution pls?

peterbud: And also noticed that this pattern recognizes variable names as numbers if the have a digit in their names, like `MyVariable2`

RunDevelopment: My bad. This pattern should work.

```js
{
    pattern: /\b0x[\da-f]+\b|(?:[+-]?(?:\b\d+\.)?\b\d+|[+-]\.\d+|(^|[^.])\B\.\d+)(?:e[+-]?\d+)?\b/,
    lookbehind: true
}
```

(Thank god, rewriting the `(?<!\.)` using a Prism lookbehind wasn't tedious at all... \s
Why did JS only add support for lookbehinds two years ago...)

peterbud: Super, thanks!

components/prism-powerquery.js
peterbud: That simplifies the pattern a lot :-0. Works perfectly, thanks

tests/languages/powerquery/number_feature.test
```diff
@@ -0,0 +1,43 @@
...
+123
+123e45
+123e+45
+123e-45
+123E45
+123E+45
+123E-45
+123.78e-45
+.78e-45
+
```
RunDevelopment: Could you please add an example for a range like `123..456`. This broke the pattern before, so I think we should include a test case.

peterbud: OK, I have added it here to check that it properly recognizes the `..` operator

peterbud: Should I simplify the testfile for the functions, maybe?

RunDevelopment: Yeah, that'd be nice. Maybe also add a case for user-defined functions.

After that, I think this PR is good to go.

peterbud: OK done. Function tests are now very simple, hope this is OK.

RunDevelopment:
Thank you for contributing @peterbud!

PR #2256, opened 2020-03-21
Bash: Added support for escaped quotes

### LINKED ISSUE #2255: [bash] issue while escaping quote
- Prism version: 1.19
- Environment: Firefox 74.0, XUbuntu 19.10.

Prism looks "lost" with Bash code having this syntax :
```
printf -v aval "%d" \'a                   # 'a' ascii value
printf -v zval "%d" \'z                   # 'z' ascii value
```
It looks like is considers the first quote in comment to be the end of a string, the string start being the escaped quote. What is most disturbing is that all lines following these lines are wrongly colorized, as we can see on:
[https://exercism.io/my/solutions/7a25d82ea512403280f7fd168040eefb?iteration_idx=1](https://exercism.io/my/solutions/7a25d82ea512403280f7fd168040eefb?iteration_idx=1), where colors are all wrong after line 56.

Workaround 1: The following works with prism (no single quote in comment) :
```
printf -v aval "%d" \'a                  # "a" ascii value
printf -v zval "%d" \'z                  # "z" ascii value
```
Workaround 2: using double quotes instead of escaping with backslash :
```
printf -v aval "%d" "'a"                # 'a' ascii value
printf -v zval "%d" "'z"                # 'z' ascii value
```

RunDevelopment:
This fixes #2255.

image


tests/languages/shell-session/command_string_feature.test
```diff
@@ -13,6 +13,8 @@ foo
 bar
 STRING_END
 
+$ echo \'a # '
```
mAAdhaTTah: Why doesn't this addition result in additional tokens?

RunDevelopment: Good catch! 
(Could have sworn I added the tokens...)

PR #2258, opened 2020-03-21
Remove self & document from IIFE arguments

### LINKED ISSUE #2257: ReferenceError: self is not defined - prism-keep-markup
**Information:**
- Prism version: "prismjs": "^1.19.0"
- Plugins: `.babelrc` `{ ... ["line-numbers", "show-language", "keep-markup"]`
- Environment: [Next.js]

**Does the latest version of Prism from the [download page](https://prismjs.com/download.html) also have this issue?**
Yes

**Description**
After adding `"keep-markup"` to the plugins list, the app fails:
image

**Example**
Reproduction available below
_PS you can stop the app from crashing by disabling `"keep-markup"` in .babelrc_
https://codesandbox.io/s/nextjs-prismjs-slate-u5rmz

mAAdhaTTah:
If self or document don't exist, they'll error instead of getting to the
checks.

Fixes #2257.


ScottAgirs:
Hey guys, will there be a patch published or how does this work? While on localhost it's possible to manually delete the buggy parts from node_modules, when deploying npm install is still broken...

An idea of when this could be live or what to do to have this running would be much appreciated 🙌 🙏

RunDevelopment:
We will make a release soon but it might take some time. In the meantime, you can let npm checkout the commit instead.

```
npm i github:PrismJS/prism#3c043338154fc2a660fc3f6035a8589899f4a644
```

PR #2265, opened 2020-03-24
Added treeview plugin

### LINKED ISSUE #2264: language request: treeview
**Language**

format code blocks which is the result of the `tree` command shout be included in the core.

example of treeview:

```
root_folder/
|-- a first folder/
|   |-- holidays.mov
|   |-- javascript-file.js
|   `-- some_picture.jpg
|-- documents/
```

**Additional resources**

https://www.npmjs.com/package/tree-cli
https://github.com/Golmote/prism-treeview/ from @Golmote
https://codepen.io/anon/pen/GWxyNj

RunDevelopment:
This is a port of @Golmote's [prism-treeview](https://github.com/Golmote/prism-treeview/) plugin with some minor changes.

This resolves #2264.


plugins/treeview/prism-treeview.js
```diff
@@ -0,0 +1,60 @@
...
+
+				if (/(^|[^\\])[=*|]\s*$/.test(env.content)) {
+					env.content = env.content.slice(0, -1);
+				}
+
+				var parts = env.content.toLowerCase().replace(/\s+/, '').split('.');
+				while (parts.length > 1) {
+					parts.shift();
+					// Ex. 'foo.min.js' would become '<span class="token keyword ext-min-js ext-js">foo.min.js</span>'
+					env.classes.push('ext-' + parts.join('-'));
```
muescha: maybe add also a `ext-js` here with only the last extension if someone use points in normal filename like: `this.is.my.file.html.js`

muescha:
make it sense to split the css into 2 parts:

- one with the colors `plugins/treeview/prism-treeview.css`
- one with only icons `plugins/treeview/prism-treeview-icons.css`

this way it is possible to:
- see it without icons (and can have the css payload low)
- exchange with a different css to for example the github file icons 
(like with this chrome plugin https://github.com/lvarayut/github-file-icons -> [CSS](https://github.com/lvarayut/github-file-icons/blob/master/vendors/file-icons-js/css/style.css) 
which uses [file-icons-js](https://github.com/websemantics/file-icons-js/))

plugins/treeview/prism-treeview.js
Golmote: @muescha This is already the case I believe. In the example you gave, the token would have the classes
```
ext-is-my-file-html-js ext-my-file-html-js ext-html-js ext-js
```

Golmote:
@muescha I do not think a Prism plugin can have multiple CSS files currently. Whoever wants to change the icons will need to replace the whole file, or override just the needed parts.
But the codeblock colors will come from the selected theme anyway, the only remaining color is the color of the borders used to draw the filetree lines, which can be overriden too.

mAAdhaTTah:
Yeah, multiple CSS files would make the babel plugin more complicated as well, which assumes max one CSS file per plugin.

plugins/treeview/prism-treeview.js
RunDevelopment: It behaves as described @Golmote. I think it's ok to add few more classes than necessary in these edge cases.

RunDevelopment:
Splitting the CSS is sorta possible with #2162. 
Assuming that users build their theme on the download page, we can just cut out the predefined icons (and maybe even insert other ones).

RunDevelopment:
Regarding the babel plugin: We could use something like a `no-default-icon-set` class to mark code blocks with a custom one. This won't do anything for CSS bundle size, but it'll work.
(The class will be inherited, so you can define it in the `<body>` and have it affect the whole page.)

plugins/treeview/prism-treeview.js
muescha: ah yes 🤦‍♂ its a loop :) 

great :)

muescha:
make it sense also include beginning slashes as directory?

```treeview
/my-my-site
└── my-config.js
└── /src
└── /plugins
```

RunDevelopment:
Directories are usually denoted by a __trailing__ slash:

```
foo/       # directory
`-- bar    # file
```

That being said, the plugin will remove the trailing slash (and even similar file markers).

PR #2271, opened 2020-03-26
Add Support for Shortcodes (WordPress)

mccomaschris:
To use this language, use the class "language-shortcode"


RunDevelopment:
@mccomaschris 
What is the difference between this language and BBCode? I read some articles but they seem pretty much the same to me.

mccomaschris:
> @mccomaschris
> What is the difference between this language and BBCode? I read some articles but they seem pretty much the same to me.

Crap, I had totally missed that one (and not used BBCode, or even thought of it, in over a decade. I'll go ahead and close this for now. Thanks!

RunDevelopment:
One question: Should we give BBCode an alias "shortcode", so you can still use `language-shortcode`?

mccomaschris:
I think it could be beneficial, I tried that language class. I know of a couple very popular WordPress plugins using language-php for their shortcodes in their documentation, they like me may not just put BBCode and WordPress Shortcodes together as the same language. I'll submit a new PR to add "shortcode" as an alias to BBCode.

PR #2273, opened 2020-03-26
Add shortcode alias to BBCode

mccomaschris:
This just adds the shortcode alias to BBCode, letting people use the class "language-shortcode"


RunDevelopment:
Thank you for contributing @mccomaschris!

PR #2276, opened 2020-03-28
Add Eta, a JS templating engine

nebrelbug:
Website: https://eta-docs.now.sh

Eta is a JS template engine very similar to EJS (hence why many of the examples and tests were only slightly modified from EJS').


nebrelbug:
I didn't modify any components other than those for Eta, but some of the files minified differently... would clicking "Revert Changes" (under the Files tab on GitHub) for each of those files be good?

RunDevelopment:
Seems like you're using a different uglify version. Try using `npm ci` instead of `npm i` when installing Prism's dependencies.

nebrelbug:
@RunDevelopment aha -- that worked! Thanks! Just pushed my changes.

I also realized I used Yarn to install the first time, so maybe that was part of the problem :man_shrugging:

RunDevelopment:
I looked at your language definition and from a syntax-highlight standpoint, I think that EJS is enough. The differences between ETA and EJS are very minor and not noticeable enough to justify another language IMO.

The only differences (from a syntax point of view) are:

1. ETA doesn't do `#` comments.
2. ETA supports additional modifiers (the `(?:\s*(?:=|~))?` part).
3. ETA supports `%>` inside of strings and comments. (Only some strings thou and not in single-line comments.)

I think it will be good enough to just give EJS an `eta` alias as the languages are very similar.

---

Also, as you seem to be the main developer of ETA (yes, I do my research ;) ), I think that I should point out that [the pattern](https://github.com/eta-dev/eta/blob/b51fdaaef4aa9d8a8a036f6e84aa4da47c44839d/src/parse.ts#L50) used to detect templates is not ideal. Here a few points to improve it:

<details>

1. Support for single-line comments:
    This can be done like this `\/\/.*(?!.)`. The lookahead is very important because if you don't have it, this part of the pattern will cause exponential backtracking in your pattern.
2. Better multiline comments:
    This is just a guess but I don't think it's intentional that your pattern can match the string `<% /* %> */`. If it is intentional, just ignore this comment.
    This can be fixed by using this pattern for multiline comments: `\/\*(?:[^*]|\*(?!\/))*\*\/`.
3. Strings:
    Normal JS strings can actually span multiple lines by using a backslash before the line break. E.g.
    ```js
    "foo\
    bar" // same for single-quoted strings
    ```
    Nobody uses this but it's valid syntax.
4. Template strings:
    Because of the interpolation expressions of template strings are normal JS expressions, the syntax of template strings is actually context-free and cannot be described (or matched) by a regular expression. E.g. This is a valid template string `` `a${`b`}c` `` (yes, you can nest these as deep as you want to.) Prism also has huge problems with template strings.
    The best thing you can do here with regular expressions is to support a finite number of nested levels but the regular expression will be hugely complex even for just two levels of nested template strings.
5. Exponential backtracking:
    I really find it amazing that your pattern doesn't backtrack like crazy. 
    Because the quantified element of `(?:[^]*?(?:...)?)*?` can match the empty string, you force the underlying JS engine to optimize your pattern which saves you from exponential backtracking. I'm really impressed that this yields such an optimized pattern.
    That being said, I personally wouldn't do this because it relies on the JS engine being somewhat smart, so there's no guarantee that it will not backtrack like crazy, but if you only need to support major (speak well-optimized) browsers, I guess it's ok.

</details>

nebrelbug:
@RunDevelopment thanks so much for the incredibly helpful feedback!

I just finished [rewriting the parser](https://github.com/eta-dev/eta/blob/v1.0.0-beta.3/src/parse.ts) based off your suggestions :smiley:. The new parser throws when there are unclosed strings or comments, and template literals should work better as well. I also got rid of the backtracking RegExp :)

As far as single-line comments, they seem to be working as far as I can tell by experimenting on Eta's [playground page](https://cdn.statically.io/gh/eta-dev/eta/master/browser-tests/demo.html). Strings with backslashes before the line break actually worked before because of some tricky stuff in the RegExp, and I just verified they still work.

As far as adding an `eta` alias to EJS, that probably makes sense as long as most users don't have delimiters inside quotes or comments. Perhaps I should close this PR and open up a new one with the alias?

Anyways, thanks again for the suggestions! If you have any additional feedback on the changes I made, I'd love it :smile:

RunDevelopment:
Glad I could help.

> Perhaps I should close this PR and open up a new one with the alias?

Yes, that would be the best course of action.

I'll close this now.

nebrelbug:
Just created a PR: #2282

PR #2281, opened 2020-04-03
Diff: Add nest token

chenxsan:
When using `diff` lang, 

```diff
  {
    "name": "webpack-demo",
    "version": "1.0.0",
    "description": "",
    "scripts": {
-      "test": "echo \"Error: no test specified\" && exit 1"
+      "test": "echo \"Error: no test specified\" && exit 1",
+      "build": "webpack"
    },
    "keywords": [],
    "author": "",
    "license": "ISC",
    "devDependencies": {
      "webpack": "^4.20.2",
      "webpack-cli": "^3.1.2"
    },
    "dependencies": {
      "lodash": "^4.17.5"
    }
  }
```
we sometimes do not want to copy `+` `-` along with other texts. We can make it if we can select `+` `-` [with css class](https://github.com/webpack/webpack.js.org/issues/3578#issue-569463262):

```css
.target-diff-prefix { user-select: none;}
```

But we can't make it without [diff highlight plugin](https://prismjs.com/plugins/diff-highlight/) which is unfortunately not supported in [refractor](https://github.com/wooorm/refractor#plugins) we use.

This pull request would bring the support without plugin.


components/prism-diff.js
```diff
@@ -39,11 +39,24 @@
...
 		}
 
-		Prism.languages.diff[name] = {
-			// pattern: /^(?:[_].*(?:\r\n?|\n|(?![\s\S])))+/m
-			pattern: RegExp('^(?:[' + prefix + '].*(?:\r\n?|\n|(?![\\s\\S])))+', 'm'),
-			alias: alias
-		};
+		var pattern = RegExp('^(?:[' + prefix + '].*(?:\r\n?|\n|(?![\\s\\S])))+', 'm');
+
+		if (['-', '<', '+', '>'].indexOf(prefix) !== -1) {
```
RunDevelopment, line 44: I don't see why you want to add the prefix token for a select few. 

The motivation is that you don't want to copy the prefix, right? Then why is it ok to copy e.g. the space prefix?

chenxsan: Because whitespace won't stop the code from running but `+` `-` will, which means we have to remove`+` `-` etc. to make the code work. Also, Code Formatter like prettier could fix the whitespace problem easily.

components/prism-diff.js
```diff
@@ -39,11 +39,24 @@
...
+			var inside = {
+				prefix: RegExp('^[' + prefix + ']', 'm')
+			};
```
RunDevelopment, line 47: This inside won't work if the next character after the prefix is the same character as the prefix.
Example:

```
++
--
```

```suggestion
			var inside = {
				'line': {
				    pattern: /^(.).+(?:\r\n?|\n)?/,
				    lookbehind: true
				},
				'prefix': /[\s\S]/
			};
```

components/prism-diff.js
RunDevelopment: I don't know what language you're using but python code will definitely stop running with an extra space.

Also, what about the `!` prefix?

components/prism-diff.js
chenxsan: Sorry, I don't quite understand the suggested change here, it wraps every single character with `prefix` now which is not really what I want:

image

RunDevelopment: Please also look at the `line` token. The trick is that the line token will capture every line (plus line ends) but leaves out the prefix. So the only thing left is the prefix.

components/prism-diff.js
chenxsan: I use HTML/CSS/JavaScript mostly. I would rather cover use cases I already known. When people using other languages want to target those prefixes too, they could improve it later. Handling something I'm not familiar with is pretty dangerous.

components/prism-diff.js
chenxsan: Thank u! It's my fault to put `prefix` key before `line`.

components/prism-diff.js
RunDevelopment: > they could improve it later.

Most people will probably just use another syntax highlighter instead of submitting a PR. (Thank you for doing the latter)

> Handling something I'm not familiar with is pretty dangerous.

Agreed but in this case, it's pretty safe. 
Worst-case: The highlighting is a bit off.
Worst-worst-case: Some extra characters get copied.

From my point of view, handling all prefixes equally is a better solution. It's simpler, works for all, and is [less surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) (why only these prefixes?).

components/prism-diff.js
```diff
@@ -39,11 +39,28 @@
...
-			// pattern: /^(?:[_].*(?:\r\n?|\n|(?![\s\S])))+/m
-			pattern: RegExp('^(?:[' + prefix + '].*(?:\r\n?|\n|(?![\\s\\S])))+', 'm'),
-			alias: alias
-		};
+		var pattern = RegExp('^(?:[' + prefix + '].*(?:\r\n?|\n|(?![\\s\\S])))+', 'm');
+
+		if (['-', '<', '+', '>'].indexOf(prefix) !== -1) {
+			var inside = {
+				line: {
+					pattern: /^(.).+(?:\r\n?|\n)?/,
```
RunDevelopment, line 47: Sorry, I made a mistake here. This pattern has problems with empty lines (no chars after the prefix).
Here's the fix:

```suggestion
					pattern: /(.)(?=[\s\S]).*(?:\r\n?|\n)?/,
```

components/prism-diff.js
```diff
@@ -39,11 +39,28 @@
...
+					lookbehind: true
+				},
+				prefix: /[\s\S]/
```
RunDevelopment, line 50: Please add single quotes around the `prefix` and `line` properties.

We do this to differentiate Prism grammar tokens from normal object properties. This is just a small convention to make language definitions easier to read.

```suggestion
				'prefix': /[\s\S]/
```

components/prism-diff.js
chenxsan: I agree with you.

components/prism-diff.js
```diff
@@ -39,11 +39,28 @@
...
 		}
 
-		Prism.languages.diff[name] = {
-			// pattern: /^(?:[_].*(?:\r\n?|\n|(?![\s\S])))+/m
-			pattern: RegExp('^(?:[' + prefix + '].*(?:\r\n?|\n|(?![\\s\\S])))+', 'm'),
-			alias: alias
-		};
+		var pattern = RegExp('^(?:[' + prefix + '].*(?:\r\n?|\n|(?![\\s\\S])))+', 'm');
+
+		if (['-', '<', '+', '>', '!'].indexOf(prefix) !== -1) {
```
RunDevelopment, line 44: Why is the `if` still here?

chenxsan: Updated. Misunderstood the `unchanged` one.

components/prism-diff.js
```diff
@@ -40,10 +40,17 @@
...
-			// pattern: /^(?:[_].*(?:\r\n?|\n|(?![\s\S])))+/m
 			pattern: RegExp('^(?:[' + prefix + '].*(?:\r\n?|\n|(?![\\s\\S])))+', 'm'),
-			alias: alias
+			alias: alias,
+			inside: {
+				'line': {
+					pattern: /(.)(?=[\s\S]).*(?:\r\n?|\n)?/,
+					lookbehind: true
+				},
+				'prefix': /[\s\S]/
```
RunDevelopment, line 50: One last thing:
Diff highlight also adds an alias according the prefix type. E.g. `<span class="token prefix inserted">+</span>`

The code for extracting the category is on line 36.

chenxsan: I've added `alias` to `'prefix'`.

chenxsan:
@RunDevelopment Thanks for your kind help!

karlhorky:
Really cool, thanks for this @chenxsan and @RunDevelopment !

PR #2282, opened 2020-04-03
Add 'eta' alias to EJS

nebrelbug:
[Eta](https://eta.js.org) is fairly similar to EJS. If it ever gets popular, I might throw together a new pull request with proper support for strings and comments inside delimiters :)


components.json
```diff
@@ -291,7 +291,8 @@
 		"ejs": {
 			"title": "EJS",
 			"require": ["javascript", "markup-templating"],
-			"owner": "RunDevelopment"
+			"owner": "RunDevelopment",
+			"alias": "eta"
```
RunDevelopment, line 295: You should also give this an `aliasTitle`, similar to how `svg` is an alias of `markup` with the title `SVG`.

nebrelbug: Done :)

plugins/show-language/prism-show-language.js
```diff
@@ -57,6 +57,7 @@
 		"dockerfile": "Docker",
 		"ebnf": "Extended Backus–Naur form",
 		"ejs": "EJS",
+		"eta": "EJS",
```
RunDevelopment, line 60: This isn't good. It should be `ETA`.
That's a bug in Prism's build. I'll investigate.

RunDevelopment: Will be fixed by #2283. Let's merge this PR!

RunDevelopment:
Thank you for contributing @nebrelbug!

nebrelbug:
@RunDevelopment thank you!

PR #2290, opened 2020-04-07
Command Line refactoring

### LINKED ISSUE #2289: command-line plugin does not display the prompt (spans outside of their container)
**Information:**
- Prism version: 1.20.0
- Plugins: line-numbers, autolinker, show-language, keep-markup, command-line
- Environment: Webpack Encore (a Symfony project)

**Description**
The plugin is activated (detects output lines) but it does not display the user and host. It shows on page load (< 1 sec) and then gets removed from the DOM.
I added the `keep-markup` plugin too but this time, the `<span>` elements are placed outside the `<span class="command-line-prompt"></span>` which seems to be the container according to the [example page](https://prismjs.com/plugins/command-line/) I inspected.

I use Webpack Encore, with the Babel plugin as recommended. Before calling Prism's highlight function I add the attributes on the `<pre>` (the HTML content is generated from MarkDown stored in the database). Here is my full script:
```js
// Slide content container
let content = document.getElementById('slide_content');

// Add classes for plugins
content.querySelectorAll('pre').forEach(function (pre) {

    // Display command line output
    let code = pre.querySelector(':scope > code');
    if (
        code.classList.contains('language-shell')
        || code.classList.contains('language-shell-session')
        || code.classList.contains('language-bash')
        || code.classList.contains('language-git')
    ) {
        pre.classList.add('command-line');
        pre.dataset.user = 'aymdev';
        pre.dataset.host = 'local';
        pre.dataset.filterOutput = '[output]';

    } else {

        // Or add line numbers
        pre.classList.add('line-numbers');
    }
});

// Plugins assets
require('prismjs/plugins/command-line/prism-command-line.css');

// Prism JS (configuration with Babel plugin)
import Prism from 'prismjs';
Prism.highlightAllUnder(content);
```
Removing all other plugins had no effect. No idea if this could be related to #1456 

**Temporary workaround:**
To fix this I added this at the bottom of my script:
```js
window.addEventListener('DOMContentLoaded', function () {
    content.querySelectorAll('pre.command-line > code').forEach(function (codeElement) {
        let prompt = codeElement.querySelector('span.command-line-prompt');

        // Starts at 1 as the first element is the prompt
        for (let i = 1; i < codeElement.children.length; i++) {
            // Move spans inside the prompt
            if (codeElement.children[i].getAttribute('data-user') || codeElement.children[i].attributes.length === 0) {
                prompt.insertBefore(codeElement.children[i], null);
                i--;
            } else {
                // Stop as spans to move are placed before highlighted tokens
                return;
            }
        }
    });
});
```

RunDevelopment:
This fixes #2289.

The actual fix is quite simple:

```diff
-if (env.element.querySelector('.command-line-prompt')) { // Abort if prompt already exists.
-	commandLine.complete = true;
-	return;
-}
+// The element might be highlighted multiple times, so we just remove the previous prompt
+var existingPrompt = env.element.querySelector('.command-line-prompt');
+if (existingPrompt) {
+	existingPrompt.remove();
+}
```

But I also took the time to refactor Command Line a little.


mAAdhaTTah:
It would be easier to review this if the stylistic changes were in a separate PR. Between the refactor, the actual change, & the indentation, I'm not really sure what I'm looking at here.

PR #2298, opened 2020-04-09
Bash: Add PHP dependency manager `composer` as a command

Ayesh:
Hi there,
`composer` is PHP's dependency manager, and I would like to propose that we add it as a command to bash language support. I see that `npm` is already in the list, which has equal grounds as `composer`. 

Thank you.


RunDevelopment:
Thank you for contributing @Ayesh!

Ayesh:
Thank you @RunDevelopment. I'm sorry I forgot to update the tests.

Ayesh:
Awesome, this is merged, thank you 🙏.

PR #2308, opened 2020-04-11
Minor improvement for default theme

jorgepvenegas:
While implementing the default theme on my website, I found this little issue with the background color on tokens. Here is the way it looks before applying the change proposed in this PR for JavaScript syntax:
before
Notice that behind `=` and `||` there is a slight difference in the background color. This is how it looks after applying this change:
after

Although I believe this doesn't depend on a specific syntax, here is example with Rust:

Before (look at the `=`):
rust-before
After:
rust-after

The change is minimal, just remove a `background` CSS rule from `theme/prism.css`. Just to give more context, I went through all the other template files (`theme/prism-*.css`) and realized they aren't specifying this rule.

### DIFF
--- a/themes/prism.css
+++ b/themes/prism.css
@@ -106,7 +106,6 @@ pre[class*="language-"] {
 .language-css .token.string,
 .style .token.string {
 	color: #9a6e3a;
-	background: hsla(0, 0%, 100%, .5);
 }
 
 .token.atrule,


RunDevelopment:
That's not a bug, it's a feature.

I know that this sentence is usually used ironically but for real: it's a feature. I agree that it looks somewhat... special and you [aren't the first one to report this as a bug](https://github.com/PrismJS/prism/issues/1691) but the white background is intentional.

jorgepvenegas:
@RunDevelopment interesting.

Issue #1347, opened 2018-03-11
Token operator should not have a background


paladox:
Like the other tokens, token operator should not have a background, make things look a little wierd.

<img width="677" alt="screen shot 2018-03-11 at 22 48 44" src="https://user-images.githubusercontent.com/5727000/37259452-6056a228-257e-11e8-9d74-82019cbe8c43.png">

See ^^

Golmote:
Since you decided to change the background color of the theme, you should probably change or remove the background color of the operators, indeed.

This is ok as part of the original theme, tho.

paladox:
I didn't change the colour of it though :)

Golmote:
The background color of the code block, I meant. This greenish background isn't the original background of the default theme is it?

paladox:
Ah i see, yeh.

Golmote:
So I consider you did some customization, and so you might also need to customize more to make it look less weird.

paladox:
Yeh, we are using the highlight function so it dosen't implement the language-<lang> class. (well not really we are as i am trying to implement this in gerrit https://gerrit-review.googlesource.com/c/gerrit/+/160034/ but need #1325 to be resolved :))

paladox:
@Golmote im wondering if we could remove it offically from the token or is it needed?

Golmote:
IMO no we should not. It looks good on the theme and the author did it like this, I won't discuss it.
This is theme customization, and theme customization should be in your own CSS.

PR #2309, opened 2020-04-11
Default Theme: Add a comment that declares the background color of `operator` tokens as intentional

RunDevelopment:
Since people seem to somewhat [regularly](https://github.com/PrismJS/prism/issues/1347) [report](https://github.com/PrismJS/prism/issues/1691) [and](https://github.com/PrismJS/prism/pull/1692) [fix](https://github.com/PrismJS/prism/pull/2308) the background color of `operator` tokens in the default theme, I added a little comment to clarify that it's indeed intentional.

NGL, when I first saw Prism, I also thought it's a bug.

### DIFF
--- a/themes/prism.css
+++ b/themes/prism.css
@@ -106,6 +106,7 @@ pre[class*="language-"] {
 .language-css .token.string,
 .style .token.string {
 	color: #9a6e3a;
+	/* This background color was intended by the author of this theme. */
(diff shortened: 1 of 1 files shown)


ccorcos:
Hey, is there's a reason for this intentional background statement?

RunDevelopment:
The styles are entirely up to the author of the theme. They probably like the light emphasis the background color puts on operators.

PR #2326, opened 2020-04-23
Autoloader: Fixed data-dependencies and extensions

### LINKED ISSUE #1976: Autoloader endlessly loads extensions blocking the whole page
**Information:**
- Prism version: 1.16
- Environment: Browser & everything that Autoloader

**Description**
When using Autoloader on an element which has an extension set as its language, the Autoloader will endlessly try to load this language and block the whole page in the process.

I know that using an extension as a language is a mistake itself but it is something which shouldn't cause the Autoloader to just crash the page.

**Example**
```
<code class="language-js-extra"></code>
```

### LINKED ISSUE #1994: Autoloader doesn't load `data-dependencies` of elements with already loaded languages
**Information:**
- Prism version: 1.16
- Environment: Browser

**Description**
Autoloader will ignore the `data-dependencies` of elements if the language of the element is loaded already.

Suppose we have already loaded the JavaScript language. If we then want to highlight this element:

```html
<pre data-dependencies="js-extras"><code class="language-javascript">
    console.log("foo");
</code></pre>
```

Then Autoloader will do nothing. JS Extras will not be loaded.

**Example**
An example of this can be seen in #1936 on the examples page. Open the examples page, select JavaScript, select JS Extras, and see that JS Extras wasn't loaded.

RunDevelopment:
This fixes #1994. This fixes #1976. 

It does so by using a new method to check whether a language is already loaded. It now checks whether the language and `data-dependencies` have been loaded instead of checking whether the current grammar is defined.


RunDevelopment:
I should also point out that this fixes the bug that blocks #1936.

plugins/autoloader/prism-autoloader.js
```diff
@@ -253,42 +253,55 @@
...
-			+ (config.use_minified ? '.min' : '') + '.js'
+	function getDependencies(element) {
+		var deps = (element.getAttribute('data-dependencies') || '').trim();
+		if (!deps) {
+			var parent = element.parentElement;
+			if (parent && parent.tagName.toLowerCase() === 'pre') {
+				deps = (parent.getAttribute('data-dependencies') || '').trim();
+			}
+		}
+		return deps ? deps.split(/\s*,\s*/g) : [];
```
mAAdhaTTah, line 269: We have a couple APIs that look up the tree for attributes like this. Is it worthwhile to make this "inherited" like that?

plugins/autoloader/prism-autoloader.js
```diff
@@ -253,42 +253,55 @@
...
+			// the given language is already loaded
+			return true;
 		}
 
-		loadLanguages(deps ? deps.split(/\s*,\s*/g) : [], function () {
-			loadLanguage(lang, function () {
-				Prism.highlightElement(elt);
-			});
-		});
+		// has already been loaded
```
mAAdhaTTah, line 291: This comment seems out of place.

plugins/autoloader/prism-autoloader.js
```diff
@@ -253,42 +253,55 @@
...
 	 *
 	 * @param {string} lang
-	 * @param {HTMLElement} elt
+	 * @returns {boolean}
 	 */
-	function registerElement(lang, elt) {
-		if (lang in lang_aliases) {
-			lang = lang_aliases[lang];
+	function isLoaded(lang) {
+		if (lang.indexOf('!') >= 0) {
```
mAAdhaTTah, line 279: We seem to be doing some `force` checks with a variable in `load`. Is this code path actually hit or is this a few feature or what's the story here?

mAAdhaTTah: Nevermind, I needed to reread the whole file.

plugins/autoloader/prism-autoloader.js
RunDevelopment: What do you mean by "look up the tree"? It only checks the `code` and the `pre` elements. We have a few that look for classes until they reach the root of the documents but this isn't one of them.

mAAdhaTTah: Yeah, I was wondering if it would be worth doing that, but if not, no worries.

plugins/autoloader/prism-autoloader.js
```diff
@@ -423,10 +439,20 @@
...
+		if (!element || !language || language === ignored_language) {
+			return;
+		}
+
+		var deps = getDependencies(element);
+		deps.push(language);
+
+		if (!deps.every(isLoaded)) {
+			// the language or some dependencies aren't loaded
+			loadLanguages(dependencies, function () {
```
RunDevelopment, line 453: Guess who meant to write `deps` and not `dependencies`? Fixed in #2449.

PR #2331, opened 2020-04-26
Alias objectivec as objc

rchee:
Objectivec also know as objc. Maybe a good choice to alias it as objc.


RunDevelopment:
Thank you for contributing!

PR #2337, opened 2020-04-28
fix(line-numbers): lineNumberWrapper might be null when code is ""

lynette-li:
When code is "" or undefined or null,the following error occurs when resizing the window because the element of ".line-numbers-rows" is null.

image


plugins/line-numbers/prism-line-numbers.js
```diff
@@ -42,7 +42,9 @@
 			codeLines.forEach(function (line, lineNumber) {
 				lineNumberSizer.textContent = line || '\n';
 				var lineSize = lineNumberSizer.getBoundingClientRect().height;
-				lineNumbersWrapper.children[lineNumber].style.height = lineSize + 'px';
+				if(lineNumbersWrapper){
```
RunDevelopment, line 45: Since the code block doesn't have line numbers at all, I think we can just return early instead of treating the symptoms here. So before we get here, we can do something like this:

```diff
 			var codeElement = element.querySelector('code');
 			var lineNumbersWrapper = element.querySelector('.line-numbers-rows');
+			if (!codeElement || !lineNumbersWrapper) {
+				return;
+			}
 			var lineNumberSizer = element.querySelector('.line-numbers-sizer');
 			var codeLines = codeElement.textContent.split(NEW_LINE_EXP);
```

lynette-li: I think it's good!Thanks!

RunDevelopment:
Thank you for contributing @lynette-li!

PR #2342, opened 2020-04-29
C++: Add enum class to class names

jsmolka:
Prism falsy indentifies the `class` in `enum class` as a class name and therefore applies the wrong color.

image

This commit fixes it. The order, before `enum` itself, is important.

image


components/prism-cpp.js
```diff
@@ -1,6 +1,6 @@
 Prism.languages.cpp = Prism.languages.extend('c', {
 	'class-name': {
-		pattern: /(\b(?:class|enum|struct)\s+)\w+/,
+		pattern: /(\b(?:class|enum class|enum|struct)\s+)\w+/,
```
RunDevelopment, line 3: Rather than this solution, I think it's better to disallow keyword class names.
```suggestion
		pattern: /(\b(?:class|enum|struct)\s+)(?!class|enum|struct)\w+/,
```

RunDevelopment:
Thank you for this fix @jsmolka!

To make the build pass, you also have to commit the minified files. Install Prism's dependencies using `npm ci`, run `npx gulp`, and commit all changes. The build should now pass.

Please also add the previous error to [this](https://github.com/PrismJS/prism/blob/master/tests/languages/cpp/class-name_feature.test) test.

jsmolka:
@RunDevelopment I have already tried altering the test file in the following way

```
class Foo
class Foo_bar
struct foo
enum bar
enum class FooBar

[
	["keyword", "class"], ["class-name", "Foo"],
	["keyword", "class"], ["class-name", "Foo_bar"],
	["keyword", "struct"], ["class-name", "foo"],
	["keyword", "enum"], ["class-name", "bar"],
	["keyword", "enum class"], ["class-name", "FooBar"]
]

Checks for class names.
```

After running gulp I got

```
: expected [ Array(11) ] to deeply equal [ Array(10) ]
      + expected - actual

           "bar"
         ]
         [
           "keyword"
      -    "enum"
      +    "enum class"
         ]
         [
      -    "keyword"
      -    "class"
      -  ]
      -  [
           "class-name"
           "FooBar"
         ]
       ]
```

RunDevelopment:
Yes, the [test suite](https://prismjs.com/test-suite.html) because the expected JSON wasn't updated. Use `npm run test:languages -- --pretty` to get correct formatted JSON.

jsmolka:
The test still fails :/

RunDevelopment:
The correct JSON is probably this:

```json
[
	["keyword", "class"], ["class-name", "Foo"],
	["keyword", "class"], ["class-name", "Foo_bar"],
	["keyword", "struct"], ["class-name", "foo"],
	["keyword", "enum"], ["class-name", "bar"],
	["keyword", "enum"], ["keyword", "class"], ["class-name", "FooBar"]
]
```

jsmolka:
That fixed it, thanks!

RunDevelopment:
Thank you for contributing @jsmolka!

PR #2355, opened 2020-05-02
Update for dart keyword.

CaiJingLong:
Add some keyword.
The document of [dart keyword in website](https://dart.dev/guides/language/language-tour#keywords).


components/prism-dart.js
```diff
@@ -11,7 +11,7 @@ Prism.languages.dart = Prism.languages.extend('clike', {
 	],
 	'keyword': [
 		/\b(?:async|sync|yield)\*/,
-		/\b(?:abstract|assert|async|await|break|case|catch|class|const|continue|default|deferred|do|dynamic|else|enum|export|external|extends|factory|final|finally|for|get|if|implements|import|in|library|new|null|operator|part|rethrow|return|set|static|super|switch|this|throw|try|typedef|var|void|while|with|yield)\b/
+		/\b(?:abstract|as|assert|async|await|break|case|catch|class|const|continue|covariant|default|deferred|do|dynamic|else|enum|export|extends|extension|external|factory|false|final|finally|for|Function|get|hide|if|implements|import|in|interface|is|library|mixin|new|null|on|operator|part|rethrow|return|set|show|static|super|switch|sync|this|throw|true|try|typedef|var|void|while|with|yield)\b/
```
RunDevelopment, line 14: `false` and `true` are already part of a different token called `boolean`. You don't see this token here because it is inherited from [`clike`](https://github.com/PrismJS/prism/blob/c93244768ec7031caa6359655289c1e756a90194/components/prism-clike.js#L25).

Also, please update [Dart's keyword test](https://github.com/PrismJS/prism/blob/master/tests/languages/dart/keyword_feature.test), so your new keywords are tested.

RunDevelopment:
Thank you for making this PR @CaiJingLong!

To make the build pass, you have to install Prism's dependencies with `npm ci` and then run `npx gulp`.

components/prism-dart.js
CaiJingLong: OK, I'll check it, and add new commit.

RunDevelopment:
Thank you for contributing!

PR #2356, opened 2020-05-02
Added `npm run build` command

RunDevelopment:
I added a `build` command to our `package.json`. This name is commonly used and should make it easier for people to find it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "prism.js",
   "style": "themes/prism.css",
   "scripts": {
+    "build": "npx gulp",
     "test:aliases": "mocha tests/aliases-test.js",
     "test:core": "mocha tests/core/**/*.js",
     "test:dependencies": "mocha tests/dependencies-test.js",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "prism.js",
   "style": "themes/prism.css",
   "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
   "main": "prism.js",
   "style": "themes/prism.css",
   "scripts": {
+    "build": "npx gulp",
```
mAAdhaTTah, line 8: You don't need `npx` here cuz running an npm script will preprend your node_modules/bin scripts to your PATH.

```suggestion
    "build": "gulp",
```

RunDevelopment: npm will add your __global__ `node_modules/bin` to PATH but I want the __local__ version of gulp. I.e. on my PC, just using `gulp` will give the following.

```
C:\Users\micha\Git\prism>gulp
Der Befehl "gulp" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.
```

(The text says: "The command "gulp" is either mistyped or could not be found.")

mAAdhaTTah: When you run npm scripts that's what it does. It prepends the local to your path, just for that run. Try it locally.

RunDevelopment: Ahh right. That's what you meant. I thought that with PATH you meant the global PATH variable and not the one local to each program. My bad.

Tested it and it worked.

PR #2358, opened 2020-05-03
pug: Improved class and id in tag detection

dev-itsheng:
pug: Improved class(named `"attr-class"`) and id(named `"attr-id"`) in tag detection, and added corresponding unit test.

In order to distinguish the id, class and pure tag declared by the tag in the pug, two regular expressions judgment are added to the corresponding code, and analogous unit testings.

Similarly, the next PR is to add judgments on id, class, pesudo (pesudo-class and pesudo-element), ... in Stylus, it is comming soon.


components/prism-pug.js
```diff
@@ -112,6 +112,14 @@
 			pattern: /(^[\t ]*)(?!-)[\w\-#.]*[\w\-](?:(?:&[^(]+)?\([^)]+\))*\/?:?/m,
 			lookbehind: true,
 			inside: {
+				'attr-id': {
+					pattern: /(^[\t ]*(?!-)[\w\-.]*)#[\w\-]+/,
```
RunDevelopment, line 116: The lookbehind seems really complicated. From my understanding, [ids and classes](https://pugjs.org/language/attributes.html) aren't matches right now, right? So couldn't we just match them _after_ everything else? Wouldn't inserting the following after `'punctuation': /:/` also work:

```js
'attr-id': /#[\w\-]+/,
'attr-class': /\.[\w\-]+/
```

Or am I missing something?

RunDevelopment:
Thank you for making this PR @dev-itsheng!

To make the build pass, you have to install Prism's dependencies using `npm ci` and then run the `npx gulp` command to rebuild all generated files (all `.min.js` files and some more).

Please also take a look at my comment, and please stop lewding Chino.

dev-itsheng:
Yes, your solution is better than mine.

At first, I find that `/#[\w\-]+/` will match `#hash` of `a(href="example.html#hash")`, so I add `^[\t ]*(?!-)` to match the white-space front the tag, `[\w\-.]*` to match the beginning of the tag or attr-class. It seems to solve the problem.

If insert it after `'punctuation'`, the problem will not be occurred.

I will make a new PR by correct method soon. Thanks you for this reply.

RunDevelopment:
Glad, I can help.

Also, you don't need to make a new PR, you can just commit and push updates in this one.

dev-itsheng:
I rewrite code, run `npm i`, `npx gulp`, try to commit and push it.

dev-itsheng:
My previous avatar is just for fun, but I changed it immediately.

components/prism-pug.js
```diff
@@ -130,7 +130,9 @@
 						}
 					}
 				],
-				'punctuation': /:/
+				'punctuation': /:/,
+				'attr-id': /#[\w\-]+/,
```
RunDevelopment, line 134: One last thing: I think we should name these tokens like the corresponding CSS selector [tokens](https://github.com/PrismJS/prism/blob/c93244768ec7031caa6359655289c1e756a90194/components/prism-css-extras.js#L11), so just `id` and `class`. This should make it easier for someone to write a theme.

dev-itsheng:
In fact, I has already considered it.

In my opinion, prism supported many languages, it's possible that another language has similar `id` or `class` definition, to avoid it, I finally use `attr-` prefix.

On the other hand, I reviewed some theme CSS files, they use `.token.boolean` selector rather than `[class*="language-..."] .token.boolean`, I think that `id` in pug is different from `id` in CSS and CSS extension language.

Final, I choose `attr-id` and `attr-class`.

RunDevelopment:
Sound reasoning. Then let's keep it as is.

RunDevelopment:

dev-itsheng:
By the way, my GitHub experience is very poor, if I want to create a new PR to solve the another problem, should I delete my fork and create a new fork from this project? or continue to commit and push in my repo?

The next PR will fix the Stylus bug:

image

RunDevelopment:
You could delete your fork (I see some people do this, don't be like them) but you can also just use the features git has for this:

I won't go into the details because git can be complicated at first but this is what you have to do:
1. Open a terminal in your `prism` folder.
2. Since we added a new commit to the master branch of PrismJS/prism, you have to get these changes in your fork. So first, we have to teach git about the PrismJS/prism repo. Run `git remote add upstream https://github.com/PrismJS/prism.git` in your terminal. `upstream` will now refer to the PrismJS/prism repo.
3. Now git knows that the PrismJS/prism repo exists, but it didn't actually download anything yet. Run `git fetch upstream`.
4. Now we merge the changes from `upstream` (PrismJS/prism) into your fork. First, switch to the master branch of your fork with `git checkout master`, and then merge the changes from `upstream` (PrismJS/prism) with `git merge upstream/master`.
5. Push to your fork with `git push`.

Your fork is now up to date.

For a new PR, make a branch from the master branch of your fork. The IDE you use probably has some nice way of doing this. Make your commits into that new branch. To make a PR, publish your branch to your fork. After that is done, GitHub itself will offer you to make a PR when you visit either PrismJS/prism or your fork.

I hope this helps.
For more details, just google stuff, I guess? There are a lot of resources for git.

dev-itsheng:
Thanks for your help.

In fact, I has already try to use Google to search like 'GitHub new PR after PR' and other keywords, but don't get any useful information, and GitHub always gives me a suggest of deleting this fork, like this:

image

So I'm afraid of previous commit and push will have an influence on next PR, but after reading your answer(my English is poor, too, hope my understand is correct), my worry is unnecessary. I can continue to commit and push code in this fork(after up to date, if my commit is latest, I couldn't do something about merge).

RunDevelopment:
Yeah, I can image that you won't find anything useful there. From my experience, when I started with git, I only found one guy who actually explained how you do this fork updating (but only that). Git isn't easy to learn but once you get it, it's super useful because literally everything you might want to do with git, you can do. Git is super powerful.

GitHub is really just a server of git repos with a nice web page. It doesn't do a whole lot when you actually use git.

My recommendation: Take your time to really learn git (not just GitHub). It doesn't have to be all at once. 

As a start, try to understand why you had to do everything I described in my previous comment. [This article](https://thepilcrow.net/explaining-basic-concepts-git-and-github/) (GitHub focused) should explain it. Focus on the _concepts_ not the commands.

PR #2369, opened 2020-05-06
PureBasic

HeX0R101:
New Language PureBasic added.


tests/languages/purebasic/comment_feature.test
```diff
@@ -0,0 +1,13 @@
+;
+;gugug du Affe
```
RunDevelopment, line 2: Maybe a slightly less aggressive example?

components/prism-purebasic.js
```diff
@@ -0,0 +1,85 @@
...
+  'tag': /(?:\#[a-zA-Z0-9_]+)/,
+  'asm': {
+    pattern: /(^\s*)(?:\!.*)/im,
+    lookbehind: true,
+    inside: Prism.languages.pbfasm,
+    alias: 'tag'
+  }
+});
+
+Prism.languages.pbfasm = Prism.languages['purebasic'];
```
RunDevelopment, line 82: This seems strange. What's the intention here?

tests/languages/purebasic/comment_feature.test
HeX0R101: Well... yes, this seems to be a forgotten corpse from all my tests, although I wouldn't name this aggressive?!
We can change it to something like
;This is a comment.

btw.:
The test procedure for comments drove me crazy!
I wanted to add a test for comments BEHIND a legal code line.
This seems to be impossible, and the error messages are also quite useless (at least for me)

components/prism-purebasic.js
HeX0R101: As stated on the head of the prism-purebasic.js file already:
This whole code is not mine, I just made it work in prism.
Not sure which part you find strange, this is for the inline ASM, code example:

```
Procedure.i XorTwoBlocks2(*buffer1, *buffer2, length)
  ; move all the required data to source reg, destination reg and counter reg
  !mov esi, [p.p_buffer1]           ; read 32-bit integer from p.p_buffer1 and move to esi
  !mov edi, [p.p_buffer2]           ; read 32-bit integer from p.p_buffer2 and move to edi
  !mov ecx, [p.v_length]            ; read 32-bit integer from p.v_length and move to ecx
  
  !@@:                              ; anonymous label, can be reached by @b (back) or @f (forward)
    !mov al, byte [edi + ecx - 1]   ; read byte from destination
    !xor byte [esi + ecx - 1], al   ; xor source with destination (i.e. xor bytes from both blocks)
    !dec ecx                        ; decrease counter
  !jne @b                           ; jumb back to first anonymous label behind
  ProcedureReturn 0
EndProcedure
```

HeX0R101:
> 
> 
> Thank you for making this PR @HeX0R101!
> 
> I left you two comments.
> Also, there are a lot of small things I want to change before merging this. Is it ok if I just commit these changes directly into your branch?

Sure!

components/prism-purebasic.js
RunDevelopment: So PureBasic and its assembly language are the same language. I thought they were distinct languages, so an alias didn't make sense for me.

tests/languages/purebasic/comment_feature.test
RunDevelopment: > The test procedure for comments drove me crazy!

What happened here? I don't see how this can cause an error? What code did you try to test?
(Just a guess: Did you get a JSON parse/syntax error?)

HeX0R101: I'm quite sure I misunderstood something, but from my understanding, this here should be fine:
```
a$ = 2 ;Test

[
	["comment", ";Test"]
]

Checks for comments.
```
The comment part (to test) is ;Test, so what is wrong here?

Here the (cryptic) error message, which drove me crazy:

```
: expected [ Array(4) ] to deeply equal [ [ 'comment', ';Test' ] ]
      + expected - actual

       [
      -  "a$ "
         [
      -    "operator"
      -    "="
      -  ]
      -  [
      -    "number"
      -    "2"
      -  ]
      -  [
           "comment"
           ";Test"
         ]
       ]
```

RunDevelopment: That's the diff of the expected token stream (your JSON) and the actual token stream Prism output when highlighting `a$ = 2 ;Test`. (But yes, the diff isn't helpful at all here. It can be useful when there are only minor differences but I usually ignore it.)

Your JSON has to be the token stream of the entire code snippet. That includes whatever tokens `a$ = 2` will be.

I get that this can be confusing. I'll add it to the [doc](https://prismjs.com/test-suite.html).

components/prism-purebasic.js
```diff
@@ -0,0 +1,73 @@
...
+	'keyword': /\b(?:declarecdll|declaredll|compilerselect|compilercase|compilerdefault|compilerendselect|compilererror|enableexplicit|disableexplicit|not|and|or|xor|calldebugger|debuglevel|enabledebugger|disabledebugger|restore|read|includepath|includebinary|threaded|runtime|with|endwith|structureunion|endstructureunion|align|newlist|newmap|interface|endinterface|extends|enumeration|endenumeration|swap|foreach|continue|fakereturn|goto|gosub|return|break|module|endmodule|declaremodule|enddeclaremodule|declare|declarec|prototype|prototypec|enableasm|disableasm|dim|redim|data|datasection|enddatasection|to|procedurereturn|debug|default|case|select|endselect|as|import|endimport|importc|compilerif|compilerelse|compilerendif|compilerelseif|end|structure|endstructure|while|wend|for|next|step|if|else|elseif|endif|repeat|until|procedure|proceduredll|procedurec|procedurecdll|endprocedure|protected|shared|static|global|define|includefile|xincludefile|macro|endmacro)\b/i,
+	'function': /\b\w+(?:\.\w+)?\s*(?=\()/,
+	'number': /(?:\$[\da-f]+|\b-?\d*\.?\d+(?:e[+-]?\d+)?)\b/i,
+	'operator': /@\*\w+|(?:@|\?)\w+|\*\w+|-[>-]?|\+\+?|!=?|<<?=?|>>?=?|==?|&&?|\|?\||[~^%?*/@]/,
+	'important': /\*\w+\b/,
```
RunDevelopment, line 16: Do you know what this pattern is supposed to capture?
I'm asking because it doesn't work. The `operator` pattern above will already capture all `\*\w+`.

HeX0R101: Seems to be unneccessary, indeed.
Just throw it out!

RunDevelopment:
Thank you for contributing @HeX0R101!

Issue #2008, opened 2019-08-01
CSS extras combined with markup templating retains placeholders


wooorm:
**Information:**
- Prism version: 1.17.1 (probably since markup templating was introduced a year ago)
- Environment: Browser, Node

**Does the latest version of Prism from the [download page](https://prismjs.com/download.html) also have this issue?**

Yes

**Description**

Markup templating injects placeholders: stuff like `___PHP0___`.
If this is injected inside CSS, CSS Extras picks up on the number and creates a number token for it.
Because the placeholder is no longer a single text token, the placeholder can no longer be tokenised and remains in the resulting text instead of being replaced.

**Example**

Say we want to highlight this code as php:

```php
<img style="width:<?php echo (80 / count($images)) ?>%"/>
```

Wrapped in HTML like so:

```html
<!DOCTYPE html>
<html>
  <head>
    <link href="./prism.css" rel="stylesheet" />
  </head>
  <body>
    <pre><code class="language-php">&lt;img style="width:&lt;?php echo (80 / count($images)) ?>%"/></code></pre>
    <script src="./prism.js"></script>
  </body>
</html>
```

...where `prism.css` refers to any theme, and `prism.js` refers to a Prism build including php, markup-templating, and css-extras.

Yields:

![result](https://user-images.githubusercontent.com/944406/62303640-cd9be080-b47c-11e9-93e4-8450d7cb123c.png)

**More info**

The problematic line is here:

https://github.com/PrismJS/prism/blob/3af5d744fd7359086c4a1fc5ac4b66fe4001c935/components/prism-css-extras.js#L75

It was changed in https://github.com/PrismJS/prism/pull/1450 8 months ago. But it may have been around for at least 6 years (checking blames up to get to the initial commit). It’s more likely that this was introduced through markup templating in https://github.com/PrismJS/prism/pull/1367

RunDevelopment:
Thank you for reporting!

wooorm:
Thanks @RunDevelopment for solving this!

PR #2371, opened 2020-05-07
Added a test for identifier support across all languages

RunDevelopment:
Motivated by [this comment](https://github.com/PrismJS/prism/pull/2368#issuecomment-624784685), I added a test that goes through all languages and checks that identifiers aren't broken.

_What does "identifiers aren't broken" mean?_
It means that any identifier (`/[_a-zA-Z][_a-zA-Z0-9]*/`) will be tokenized as either one token or not at all. I.e. the identifier `foo123` would be broken if the language tokenized the `123` part as a number. The test will see how the languages handle identifiers like this and others. It will also check for numbers.

_Why do we need this?_
As pointed out in the comment, Markup templating (MT) assumes that its placeholders (which are identifiers) aren't broken up. If they are, MT will stop working. In the past, it caused [this issue](#2008).

_How is this implemented?_
The test is quite simple. It has a list (actually 3) of identifiers and just tests that those identifiers aren't broken for any given language. Because some languages don't have identifiers, you can selectively disable the test for a certain class or all classes of identifiers.
The error message of this test includes an explanation of what broken identifiers are and how to fix them. Instructions on disabling are also included.

(The problem with the current implementation of this test is that I only do a `Prism.tokenize` on every identifier. I don't test `inside` grammars because these are usually very specific to the parent pattern, so there are almost only false positives.)

The actual changes to the languages are just boundary assertions. (I didn't just blindly throw some `\b` in there tho. I went and looked up the spec/doc of every language I didn't know.)
In some cases, I even had to change some test cases because they were wrong. Markdown changed the most because I didn't know that `foo_italic_` won't make anything italic at the time. That's fixed now. For languages that had a faulty number pattern, I didn't create any new test files because we now have this test.


components/prism-css-extras.js
```diff
@@ -101,6 +101,7 @@
 				}
 			}
 		],
+		// it's import that there is no boundary assertion after the hex digits
```
mAAdhaTTah, line 104:
```suggestion
		// it's important that there is no boundary assertion after the hex digits
```

PR #2372, opened 2020-05-07
Stylus: Put the `comment` token behind the `string` token and set the `greedy` property to `true`

dev-itsheng:
After I solved the bug of comment which break the string and url, I also realized the comment can break the `@namespace` rule which named `atrule-declaration` token in Stylus tokenize. the grammar like [this](https://developer.mozilla.org/zh-CN/docs/Web/CSS/%40namespace):

```text
@namespace url(http://www.w3.org/1999/xhtml)
@namespace svg url(http://www.w3.org/2000/svg)
```

So at first I try to add `greedy: true` in `'atrule-declaration'` token, but I realized the same problem probably be happened in other tokens, such as `variable-declaration`, `statement`, `property-declaration` and `selector` (In fact, I test them and the problems happened). 

We can certainly add `greedy: true` in every token, but I think there is a better solution, as the title says, is putting the token behind them.

After that, I remove the `greedy: true` in `property-declaration` token.

Some extra change: 

To match the comment in `selector` token, I added `comment` property in `property.inside` object.

To match the following comment, the `greedy: true` in `comment` should be set:

```text
/*
comment
*/
```

Finally, about test:

To improve the changed code correctness, the tests of all scenes (all scenes which I can think about, there could be something that I ignore) are added.


dev-itsheng:
I forgot to change the language in MDN, the English site is [this](https://developer.mozilla.org/en-US/docs/Web/CSS/@namespace).

RunDevelopment:
I have nothing to add. Really good fix!

PR #2385, opened 2020-05-16
Ren'Py: Added `rpy` alias

proudust:
Add rpy alias to Ren'py language because Ren'Py script uses the .rpy filename extension.


RunDevelopment:
Thank you for contributing @proudust!

PR #2387, opened 2020-05-16
JS: Improved `get`/`set` and parameter detection

### LINKED ISSUE #2386: set is not a JavaScript keyword
**Information**
- Language: JavaScript
- Plugins: none

**Description**
The JavaScript highlighter incorrectly highlights `set` as a keyword but as per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Keywords it isn't one. This causes weird highlighting when writing such code:

Syntax highlighting fail example

**Code snippet**

<details>

```
func(x, y)
set(x, y)
```

</details>

RunDevelopment:
[`get`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) and [`set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set) now have to be followed by `(?=\s*[\[$\w\xA0-\uFFFF])`.

While writing the tests for this, I noticed that `parameter` doesn't handle computed properties, so I fixed that as well. Example:

```js
const obj = {
	set [expr](val) { } // "val" wasn't detected.
};
```

---

This resolves #2386.


madx:
Yay! Looks great!

madx:
Hurray! :champagne:

PR #2390, opened 2020-05-17
Add new aliases

valtlai:
* Adds aliases:
    * `atom` and `rss` for `markup`
    * `webmanifest` for `json` (see [Web App Manifest](https://w3c.github.io/manifest/)).
* Reorders the aliases in `components.json` to match those in `components/prism-markup.js`.

It seems that the name `atom` collides with something, since it doesn’t get added to `plugins/show-language/prism-show-language.js` (or its minified version) when running gulp. I couldn’t find the reason for this.


RunDevelopment:
Thank you for making this PR @valtlai!

Could you please clarify what "Atom" is? I only know [this Atom](https://en.wikipedia.org/wiki/Atom_(text_editor)) and [this atom](https://en.wikipedia.org/wiki/Atom).

> It seems that the name `atom` collides with something, since it doesn’t get added to `plugins/show-language/prism-show-language.js` (or its minified version) when running gulp. I couldn’t find the reason for this.

Most language ids are just the name of the language in all-lowercase, so Show Language will try to guess the title by capitalizing the id. If that guess is correct, then the id-title pair won't be stored in `prism-show-language.js`. 
If I remember correctly, we guess about 30% correctly, so this decreases the file size quite a bit.

Also, don't worry. If any aliases/ids collide, the tests won't pass.

valtlai:
Hi, @RunDevelopment!

> Could you please clarify what "Atom" is? I only know [this Atom](https://en.wikipedia.org/wiki/Atom_(text_editor)) and [this atom](https://en.wikipedia.org/wiki/Atom).

It’s [Atom feed](https://validator.w3.org/feed/docs/atom.html), like RSS but less known.

> Most language ids are just the name of the language in all-lowercase, so Show Language will try to guess the title by capitalizing the id. If that guess is correct, then the id-title pair won't be stored in `prism-show-language.js`.

Oh, that’s clever!

RunDevelopment:
Thanks for the clarification.

One last thing before we merge this: Could you please revert the reordering? The ordering in `prism-markup.js` is a pure implementation detail and has nothing to do with the order in `components.json`. Also, I like the older order because it's roughly sorted from most-used to least-used.
I think it's ok if we just append new aliases. (If we created some rule for their ordering then we'd also have to write a test that enforces that rule.)

RunDevelopment:

PR #2392, opened 2020-05-18
minor changes to better handle comments in proc

cedporter:
Check the tests. There were instances where comments in special proc weren't being captured as comments. Promoting/adding the comment object fixes this issue.


RunDevelopment:

PR #2406, opened 2020-05-27
Support multiline graphql descriptions (Fix #2405)

### LINKED ISSUE #2405: Support multiline GraphQL descriptions
**Information:**
- Prism version: latest from the download page
- Plugins: none
- Environment: Browser

**Description**
GraphQL descriptions, specifically multi-line, as described in [section 3.3 of the GraphQL spec](https://spec.graphql.org/June2018/#sec-Descriptions) are not highlighted correctly.

Also, since the spec calls out that descriptions can be formatted in CommonMark, would it make sense to format the description contents as markdown?

**Example**
https://jsfiddle.net/AlexMabry/5swkjcob/7/

AlexMabry:
I took a stab at this, hoping to save you some time.


RunDevelopment:
Thank you @AlexMabry for this PR!

To make the CI pass, you have to rebuild Prism using the command `npx gulp`.

Before we merge this, we also have to talk about whether we want Markdown highlighting inside multiline strings. You added this because they are frequently used for this purpose (I assume) but 1) single-line strings may also be used for [descriptions](http://spec.graphql.org/draft/#Description) and 2) multi-line strings may not contain Markdown if used as values.

Here is my suggestion on how we could do this:
```js
'description': {
	pattern: /(?:"""(?:[^"]|(?!""")")*"""|"(?:\\.|[^\\"\r\n])*")(?=\s*[a-z_])/i,
	greedy: true,
	alias: 'string',
	inside: {
		// TODO: Exclude leading and trailing "
		'language-markdown': {
			pattern: /[\s\S]+/,
			inside: Prism.languages.markdown
		}
	}
},
'string': {
	pattern: /"""(?:[^"]|(?!""")")*"""|"(?:\\.|[^\\"\r\n])*"/,
	greedy: true
},
```
The basic idea is that a description is always followed by either a keyword or a name (see [grammar](http://spec.graphql.org/draft/#sec-Document-Syntax)), so the `(?=\s*[a-z_])` lookahead _should_ be enough to differentiate description strings and (value) strings. 

Thoughts?
(I don't use GraphQL, so I don't know whether there are some cases for which this doesn't work.)

(I also had to slightly change the multi-line string pattern because of the lookahead (see [lazy trap](https://www.rexegg.com/regex-quantifiers.html#lazytrap)).)

mAAdhaTTah:
Also, if Markdown is optional, I think it needs to be added to the language conditionally.

RunDevelopment:
@mAAdhaTTah I think it's totally fine to have a `language-markup` wrapper around markdown without highlighting. The token stream is more stable that way compared to if we were to only conditionally add the `language-markup` or even `description` token.

AlexMabry:
@RunDevelopment Thanks for the suggestions.  I agree with you that descriptions and strings should be treated differently.

After making your change, I also made the additional changes:
- `components/`
   - Changed the alias for descriptions from 'string' to 'comment' which I think is more appropriate 
- `examples/`
  - Updated the string and description examples  
- `tests/languages/graphql/` 
  - Added a multi-line string test
  - Added some basic description tests
- `tests/languages/markdown+graphql/`
  - Added a description test that checks markdown in descriptions

It looks likes CI is still failing, so I will look into that

AlexMabry:
I just made the changes that you recommended.

RunDevelopment:
Thank you very much for contributing @AlexMabry!

PR #2409, opened 2020-05-27
Add ⍥

abrudz:
Introduced into Dyalog APL 18.0 which is just about to be released.


RunDevelopment:
Thank you for contributing @abrudz!

PR #2414, opened 2020-06-02
Protobuf: Added support for RPC

RunDevelopment:
This adds support for the [Protobuf RPC syntax](https://developers.google.com/protocol-buffers/docs/proto3#services).

I also made some minor improvements such as a lookahead for keywords to prevent false positives with field names and some pattern improvements.


tests/languages/protobuf/rcp_feature.test
```diff
@@ -0,0 +1,68 @@
+rpc LotsOfReplies(HelloRequest) returns (stream HelloResponse);
```
mAAdhaTTah, line 1: Nit: filename should be `rpc_`?

RunDevelopment: Thanks. I always mix that up.

RunDevelopment:
Thanks for the reviews @mAAdhaTTah!

mAAdhaTTah:
@RunDevelopment np. Going to take a closer look through the templating one later.

PR #2416, opened 2020-06-07
ApacheConf: alias apacheconf to conf

r33drichards:
a little bit of backstory on why I'd like this alias. 

emacs highlights config files in `org-mode` with the conf keyword

2020-06-06-222455_950x312_scrot

using [orgajs](https://github.com/orgapp/orgajs) this syntax in `org-mode` exports the code block to have the class `language-conf` which doesn't have a supported language definition. `conf-mode` in emacs  is what provides the syntax highlighting for these files, and its similar enough to the apache conf syntax that I think that it is reasonable to provide this alias as well. 

I followed the instructions here https://prismjs.com/extending.html#creating-a-new-language-definition for this Pull Request and if you want me to create an issue for this lmk.


RunDevelopment:
I'm not familiar with the emacs ecosystem but the `conf` language name/id is so general that you could think that it's any configuration format. From what you presented, the id doesn't seem to be common either but is just used by this one library.

From what I see, this is an issue with [orgajs](https://github.com/orgapp/orgajs) and not Prism. I think the easiest solution here is to add an option to [orgajs](https://github.com/orgapp/orgajs) to allow users to change the `language-xxxx` class. (`language-conf` is undescriptive anyway.)

RunDevelopment:
Feel free to reopen this if I am mistaken or misunderstood something but as is, I can't merge this.

PR #2417, opened 2020-06-09
New Language: Keepalived configure file

dev-itsheng:
[Keepalived](https://www.keepalived.org/) is a load balancing tools, it has a configure grammar which like Nginx but not same:

1. Semicolons are not used to delimit directives.
2. Its keywords (In fact, I prefer to say it 'property' like CSS.) are different from Nginx.

So I collect all of properties from [its document](https://www.keepalived.org/doc/configuration_synopsis.html) and write some token like `ip`, `path` which has `alias` pattern of commonly token name.

But the rule could not exactly right, because its document has also bug. In fact, I have made some PR to its GitHub repo with the wrong I found, There are still some errors that I did not find. Once found, I will update this language syntax and send PR to the that repo.

By the way, I think the nginx language could be rewrite like this. It is not C-Like grammar and could make lot of bugs in display. Maybe in next PR I will do it.


components/prism-keepalived.js
```diff
@@ -0,0 +1,32 @@
+Prism.languages.keepalived = {
+	'comment': {
+		pattern: /(^|[^"{\\])#.*/,
+		lookbehind: true
+	},
+	'property': {
+		pattern: /([\s\t]*)(?:\bglobal_defs|notification_email|notification_email_from|smtp_server|smtp_connect_timeout|router_id|virtual_server|fwmark|delay_loop|lb_algo|lb_kind|persistence_timeout|persistence_granularity|virtualhost|protocol|sorry_server|real_server|weight|TCP_CHECK|MISC_CHECK|misc_path|HTTP_GET|SSL_GET|url|path|digest|connect_port|connect_timeout|retry|delay_before_retry|vrrp_instance|state|interface|mcast_src_ip|lvs_sync_daemon_inteface|virtual_router_id|priority|advert_int|smtp_alert|authentication|auth_type|auth_pass|virtual_ipaddress|virtual_ipaddress_excluded|notify_master|notify_backup|notify_fault|vrrp_sync_group|vrrp_script|interval|script)\b/,
```
RunDevelopment, line 7: 1. The lookbehind is unnecessary.
2. I think this pattern is supposed to be of the form `\b(?:foo|...|bar)\b`, isn't it?
3. Please sort the list.

components/prism-keepalived.js
```diff
@@ -0,0 +1,32 @@
+Prism.languages.keepalived = {
+	'comment': {
+		pattern: /(^|[^"{\\])#.*/,
```
RunDevelopment, line 3: Could you please explain the `[^"{\\]` class? Why are these characters not allowed?

components/prism-keepalived.js
```diff
@@ -0,0 +1,32 @@
...
+	'email': {
+		pattern: /(?:[\w-])+@(?:[\w-])+(?:(?:\.[\w-]{2,3}){1,2})/,
+		alias: 'string',
+	},
+	'path': {
+		pattern: /(?:\/(?:[^\/\s]+\/)*[^\/\s]*)|(?:[a-zA-Z]:\\(?:\w+\\)*(?:\w+\.\w+)?)/,
+		alias: 'string',
+	},
+	'ip': {
+		pattern: /(?:(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)\.){3}(?:225[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)/,
```
RunDevelopment, line 24: Some boundaries to prevent false positives/partially matched incorrect IP addresses.

```suggestion
		pattern: /\b(?:(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)\.){3}(?:225[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)\b/,
```

components/prism-keepalived.js
RunDevelopment, line 3: > Comments start with '#' or '!' to the end of the  line  and  can  start
       anywhere in a line.
[Source](https://www.keepalived.org/manpage.html)

components/prism-keepalived.js
```diff
@@ -0,0 +1,32 @@
...
+	},
+	'property': {
+		pattern: /([\s\t]*)(?:\bglobal_defs|notification_email|notification_email_from|smtp_server|smtp_connect_timeout|router_id|virtual_server|fwmark|delay_loop|lb_algo|lb_kind|persistence_timeout|persistence_granularity|virtualhost|protocol|sorry_server|real_server|weight|TCP_CHECK|MISC_CHECK|misc_path|HTTP_GET|SSL_GET|url|path|digest|connect_port|connect_timeout|retry|delay_before_retry|vrrp_instance|state|interface|mcast_src_ip|lvs_sync_daemon_inteface|virtual_router_id|priority|advert_int|smtp_alert|authentication|auth_type|auth_pass|virtual_ipaddress|virtual_ipaddress_excluded|notify_master|notify_backup|notify_fault|vrrp_sync_group|vrrp_script|interval|script)\b/,
+		lookbehind: true
+	},
+	'constant': {
+		pattern: /(^[ \t]*)(?:rr|wrr|lc|wlc|sh|dh|lblc|NAT|DR|TUN|TCP|UDP|MASTER|BACKUP|PASS|AH)\b/,
+		lookbehind: true
+	},
+	'string': /("|')(?:\\(?:\r\n|[\s\S])|(?!\1)[^\\\r\n])*\1/,
```
RunDevelopment, line 14: This should probably be greedy. `"foo # bar"` is a valid string, right?

components/prism-keepalived.js
```diff
@@ -0,0 +1,32 @@
...
+	'email': {
+		pattern: /(?:[\w-])+@(?:[\w-])+(?:(?:\.[\w-]{2,3}){1,2})/,
```
RunDevelopment, line 16: Unnecessary non-capturing groups make your pattern seem more complex than it really is. Try to remove them if not necessary.

```suggestion
		pattern: /[\w-]+@[\w-]+(?:\.[\w-]{2,3}){1,2}/,
```

components/prism-keepalived.js
```diff
@@ -0,0 +1,32 @@
...
+	'comment': {
+		pattern: /(^|[^"{\\])#.*/,
+		lookbehind: true
+	},
+	'property': {
+		pattern: /([\s\t]*)(?:\bglobal_defs|notification_email|notification_email_from|smtp_server|smtp_connect_timeout|router_id|virtual_server|fwmark|delay_loop|lb_algo|lb_kind|persistence_timeout|persistence_granularity|virtualhost|protocol|sorry_server|real_server|weight|TCP_CHECK|MISC_CHECK|misc_path|HTTP_GET|SSL_GET|url|path|digest|connect_port|connect_timeout|retry|delay_before_retry|vrrp_instance|state|interface|mcast_src_ip|lvs_sync_daemon_inteface|virtual_router_id|priority|advert_int|smtp_alert|authentication|auth_type|auth_pass|virtual_ipaddress|virtual_ipaddress_excluded|notify_master|notify_backup|notify_fault|vrrp_sync_group|vrrp_script|interval|script)\b/,
+		lookbehind: true
+	},
+	'constant': {
+		pattern: /(^[ \t]*)(?:rr|wrr|lc|wlc|sh|dh|lblc|NAT|DR|TUN|TCP|UDP|MASTER|BACKUP|PASS|AH)\b/,
```
RunDevelopment, line 11: Please sort the list.

components/prism-keepalived.js
```diff
@@ -0,0 +1,32 @@
...
+		lookbehind: true
+	},
+	'string': /("|')(?:\\(?:\r\n|[\s\S])|(?!\1)[^\\\r\n])*\1/,
+	'email': {
+		pattern: /(?:[\w-])+@(?:[\w-])+(?:(?:\.[\w-]{2,3}){1,2})/,
+		alias: 'string',
+	},
+	'path': {
+		pattern: /(?:\/(?:[^\/\s]+\/)*[^\/\s]*)|(?:[a-zA-Z]:\\(?:\w+\\)*(?:\w+\.\w+)?)/,
```
RunDevelopment, line 20: This is supposed to capture unix-like and windows-like file/directory names, right? Unix-like paths look kinda fine but as for windows: Why not allow any characters like in unix-like paths like this: `[a-zA-Z]:\\(?:[^\\\s]+\\)*[^\\\s]*`?

components/prism-keepalived.js
dev-itsheng: > The lookbehind is unnecessary.

The `comment` token is copied from nginx.

If I remove `(^|[^"{\\])` from `/(^|[^"{\\])#.*/`, it can match `"it is a # string"`, so I put the `string` token before it, can pass the test and above example.

> I think this pattern is supposed to be of the form \b(?:foo|...|bar)\b, isn't it?

In fact, when I use nginx, I found that `\b(?:index|types|...)\b` can match `index.html` and `/etc/nginx/mime.types`, this result is not what we except, so I try to find a solution to avoid it. This is just a temporary practice, I have not thought of a better solution.

> Please sort the list.

The order of the list is specified in the document. In the [old document](https://www.keepalived.org/doc/configuration_synopsis.html), there is a table describing all the attributes. I also just know that the new document is updated in another interface and this list will be updated. Do you think it is better to use the order of appearance in the document (now), or alphabetical sorting?

components/prism-keepalived.js
dev-itsheng: I don't know too and remove it now. See the above comment.

components/prism-keepalived.js
dev-itsheng: Yes, I will add it.

It looks like the `number` token, `(^|[^\w.-])` in `/(^|[^\w.-])-?\d*\.?\d+/`.

Can we change `number` token in this or another language to this?

components/prism-keepalived.js
dev-itsheng: Yes, it could be changed to `/(#|!).*/`.

components/prism-keepalived.js
dev-itsheng: I think that adjusting the order can solve this problem, as modified in Stylus.

components/prism-keepalived.js
dev-itsheng: Yes, I remove them.

components/prism-keepalived.js
dev-itsheng: Similar to the reason above, I think it is more convenient to put the related values together.

components/prism-keepalived.js
RunDevelopment: >  I found that \b(?:index|types|...)\b can match index.html and /etc/nginx/mime.types

You already have a pattern that matches all paths. Can't we just match paths before this pattern?

> Do you think it is better to use the order of appearance in the document (now), or alphabetical sorting?

Alphabetical sorting. It's easy to see if certain words are included and it's obvious where you have to add new ones (if new ones were added).

components/prism-keepalived.js
dev-itsheng: Yes. Keppalived can't support Windows not, but:

1. It probably support in future.
2. This pattern can also be used in nginx.

The windows-like regex is copied from my previous code, I think your opinion is right. I change it now.

components/prism-keepalived.js
RunDevelopment: Not really. If you just adjust the order, the comment `# foo "bar"` won't be highlighted correctly. They have to be greedy.

components/prism-keepalived.js
RunDevelopment: That's an acceptable reason here (but the other list is too long for that).

components/prism-keepalived.js
dev-itsheng: The list is so long that could I use a string array to build it? like [factor](https://github.com/PrismJS/prism/blob/master/components/prism-factor.js#L381-L399).

RunDevelopment: In the case of factor, I think it's justified because the combinations there contain a lot of special regex characters that have to be escaped. That's not the case here, so I don't see the point in this, so I think that the overhead of a function is a little too much.

If you just want to do something like `pattern: RegExp('\\b' + ['word1', 'word2'].join('|') + '\\b')`, then sure! Go for it!

dev-itsheng:
I accepted all the suggestions and fixed them. 

In addition, I also did:

1. Add support for IPv6 and subnet mask (in `ip` token).
2. According to the latest documentation of keepalived, add supoprt for variable, conditional-configuration, boolean and operator.
3. Make `property` token to two alphabetical sorting arrays, because the too long regex will make a error.
4. According to the latest documentation of keepalived, update property, constant.
5. Add more complete and detailed unit tests.

components/prism-keepalived.js
```diff
@@ -0,0 +1,367 @@
...
+	'ip': {
+		pattern: /\b(?:(?:(?:(?:[0-9A-Fa-f]{1,4}:){7}[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){6}:[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){5}:(?:[0-9A-Fa-f]{1,4}:)?[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){4}:(?:[0-9A-Fa-f]{1,4}:){0,2}[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){3}:(?:[0-9A-Fa-f]{1,4}:){0,3}[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){2}:(?:[0-9A-Fa-f]{1,4}:){0,4}[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){6}(?:(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)\.){3}(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)|(?:(?:[0-9A-Fa-f]{1,4}:){0,5}:(?:(?:25[0-5]|1\d{2}|2[0-4]\d|[1-9]\d|\d)\.){3}(?:25[0-5]|1\d{2}|2[0-4]\d|[1-9]\d|\d))|(?:::(?:[0-9A-Fa-f]{1,4}:){0,5}(?:(?:25[0-5]|1\d{2}|2[0-4]\d|[1-9]\d|\d)\.){3}(?:25[0-5]|1\d{2}|2[0-4]\d|[1-9]\d|\d))|(?:[0-9A-Fa-f]{1,4}::(?:[0-9A-Fa-f]{1,4}:){0,5}[0-9A-Fa-f]{1,4})|(?:::(?:[0-9A-Fa-f]{1,4}:){0,6}[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){1,7}:)))(?:\/\d{1,3})?)|(?:(?:(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)\.){3}(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)(?:\/\d{1,2})?)\b/,
+		alias: 'number'
+	},
+
+	// support *nix / Windows, directory / file
+	'path': {
+		pattern: /(\s+)(?:\/(?:[^\/\s]+\/)*[^\/\s]*)|(?:[a-zA-Z]:\\(?:[^\\\s]+\\)*[^\\\s]*)/,
```
RunDevelopment, line 16: What's the reason behind the `(\s+)` lookbehind?

Also, unnecessary non-capturing groups:
```suggestion
		pattern: /(\s+)\/(?:[^\/\s]+\/)*[^\/\s]*|[a-zA-Z]:\\(?:[^\\\s]+\\)*[^\\\s]*/,
```

components/prism-keepalived.js
```diff
@@ -0,0 +1,367 @@
...
+	'string': /("|')(?:\\(?:\r\n|[\s\S])|(?!\1)[^\\\r\n])*\1/,
+	'comment': {
+		pattern: /(?:#|!).*/,
+		greedy: true
+	},
+
+	// support IPv4, IPv6, subnet mask
+	'ip': {
+		pattern: /\b(?:(?:(?:(?:[0-9A-Fa-f]{1,4}:){7}[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){6}:[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){5}:(?:[0-9A-Fa-f]{1,4}:)?[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){4}:(?:[0-9A-Fa-f]{1,4}:){0,2}[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){3}:(?:[0-9A-Fa-f]{1,4}:){0,3}[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){2}:(?:[0-9A-Fa-f]{1,4}:){0,4}[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){6}(?:(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)\.){3}(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)|(?:(?:[0-9A-Fa-f]{1,4}:){0,5}:(?:(?:25[0-5]|1\d{2}|2[0-4]\d|[1-9]\d|\d)\.){3}(?:25[0-5]|1\d{2}|2[0-4]\d|[1-9]\d|\d))|(?:::(?:[0-9A-Fa-f]{1,4}:){0,5}(?:(?:25[0-5]|1\d{2}|2[0-4]\d|[1-9]\d|\d)\.){3}(?:25[0-5]|1\d{2}|2[0-4]\d|[1-9]\d|\d))|(?:[0-9A-Fa-f]{1,4}::(?:[0-9A-Fa-f]{1,4}:){0,5}[0-9A-Fa-f]{1,4})|(?:::(?:[0-9A-Fa-f]{1,4}:){0,6}[0-9A-Fa-f]{1,4})|(?:(?:[0-9A-Fa-f]{1,4}:){1,7}:)))(?:\/\d{1,3})?)|(?:(?:(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)\.){3}(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)(?:\/\d{1,2})?)\b/,
```
RunDevelopment, line 10: We need to talk about this. This pattern is really complicated. I simplified it somewhat but it's still long.
Is there any way we can make this simpler?

```suggestion
		pattern: RegExp(/\b(?:(?:(?:[\da-f]{1,4}:){7}[\da-f]{1,4}|(?:[\da-f]{1,4}:){6}:[\da-f]{1,4}|(?:[\da-f]{1,4}:){5}:(?:[\da-f]{1,4}:)?[\da-f]{1,4}|(?:[\da-f]{1,4}:){4}:(?:[\da-f]{1,4}:){0,2}[\da-f]{1,4}|(?:[\da-f]{1,4}:){3}:(?:[\da-f]{1,4}:){0,3}[\da-f]{1,4}|(?:[\da-f]{1,4}:){2}:(?:[\da-f]{1,4}:){0,4}[\da-f]{1,4}|(?:[\da-f]{1,4}:){6}<ipv4>|(?:[\da-f]{1,4}:){0,5}:<ipv4>|::(?:[\da-f]{1,4}:){0,5}<ipv4>|[\da-f]{1,4}::(?:[\da-f]{1,4}:){0,5}[\da-f]{1,4}|::(?:[\da-f]{1,4}:){0,6}[\da-f]{1,4}|(?:[\da-f]{1,4}:){1,7}:)(?:\/\d{1,3})?|<ipv4>(?:\/\d{1,2})?)\b/.source.replace(/<ipv4>/g, function () { return /(?:(?:<255>\.){3}<255>)/.source; }).replace(/<255>/g, function () { return /(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]\d|\d)/.source; }), 'i'),
```

RunDevelopment:
I quickly updated and finished the PR.

RunDevelopment (approved):

RunDevelopment:

PR #2418, opened 2020-06-09
jsstacktrace: add language

sbrl:
I was getting really frustrated this afternoon with not being able to read long Javascript stack traces easily, so I implemented a simple grammar for Prism :-)


RunDevelopment:
Seems interesting. JS stack traces are 100% implementation-defined, so this might not be 100% perfect but why not?

Before we can merge this:

1. Please clean up the file. There are a lot of commented-out code snippets.
1. This whole thing get's easier if you wrap every stack frame in a token and highlight its parts. This is easier and you don't have to worry about accidentally highlighting parts of the error message. (See [Java stack trace](https://github.com/PrismJS/prism/blob/2a2e79ed5d32af6f0d0d90ea7aad002665051083/components/prism-javastacktrace.js#L42))
1. Please only use single quotes.
1. Please separate words with dashes (`-`) in token names. (~`notmycode`~ `not-my-code`).
1. Please add an example page.

Also, maybe add support for deno as well? Right now it's very centered on node.

<details>
<summary>Examples stack traces from deno</summary>

```
Error: Some error
    at throwsA (<unknown>:1:23)
    at <unknown>:1:13
    at evaluate ($deno$/repl.ts:64:34)
    at Object.replLoop ($deno$/repl.ts:153:13)
```

```
Uncaught NotFound: No such file or directory (os error 2)
    at DenoError (deno/js/errors.ts:22:5)
    at maybeError (deno/js/errors.ts:41:12)
    at handleAsyncMsgFromRust (deno/js/dispatch.ts:27:17)
```

</details>

sbrl:
Thanks for the feedback! I'll review those comments and make some improvements.

sbrl:
Done, @RunDevelopment - all suggestions applied, if you could take another look?

sbrl:
Ah, thanks for all those comments @RunDevelopment! I've made some additional changes. Some notes:

 - I've used your new patterns, but I made some tweaks:
     - `<` and `>` are needed in functions because a function can be anonymous - e.g. `foo.<anonymous>`
     - In your modified filename pattern, you accidentally dropped support for absolute paths on Windows. While I don't have a Windows machine myself (nor would I recommend it to anyone else), I think that supporting it is still a good idea.

Many great suggestions, thanks again!

components/prism-jsstacktrace.js
```diff
@@ -0,0 +1,53 @@
...
+			},
+			
+			'filename': {
+				pattern: /(\bat\s+|\()(?:[a-zA-Z]:)?[^():]+(?=:)/,
+				lookbehind: true,
+				alias: 'url'
+			},
+			
+			'function': {
+				pattern: /(at\s+(?:new\s+)?)[_$a-zA-Z\xA0-\uFFFF][.$\w\xA0-\uFFFF<>]*/,
```
RunDevelopment, line 22: Doesn't work with:

```
at new <anonymous> (_http_common.js:159:16)
```

```suggestion
				pattern: /(at\s+(?:new\s+)?)[_$a-zA-Z\xA0-\uFFFF<][.$\w\xA0-\uFFFF<>]*/,
```

components/prism-jsstacktrace.js
```diff
@@ -0,0 +1,53 @@
...
+			'keyword': /\b(?:at|new)\b/,
+			
+			'alias': {
+				pattern: /\[(?:as\s+)?[_$a-zA-Z\xA0-\uFFFF][$\w\xA0-\uFFFF]*\]/,
+				alias: 'variable'
+			},
+			
+			
+			// 'filename': {
+			// 	pattern: /(\()[^):]+(?=[:)])/,
```
RunDevelopment, line 40: Please remove the commented-out code.

components/prism-jsstacktrace.js
```diff
@@ -0,0 +1,53 @@
...
+			// 	lookbehind: true,
+			// 	alias: 'url'
+			// },
+			
+			'line-number': {
+				pattern: /:[0-9]+(?::[0-9]+)?\b/,
+				lookbehind: true,
+				alias: 'number'
+			},
```
RunDevelopment, line 49: Suggestion: Highlight `:` as punctuation.

Also, the unnecessary `lookbehind: true` is why the tests failed.

```suggestion
			'line-number': {
				pattern: /:[0-9]+(?::[0-9]+)?\b/,
				alias: 'number',
				inside: {
					'punctuation': /:/
				}
			},
```

components.json
```diff
@@ -579,6 +579,10 @@
 			"require": "json",
 			"owner": "RunDevelopment"
 		},
+		"jsstacktrace": {
+			"title": "Javascript stack trace",
```
RunDevelopment, line 583: 1. Please call it "JS stack trace". Except for JavaScript itself, we always abbreviate it as JS for all other components.
2. Please move this under "JS Extras" and above "JS Templates". The list sorted roughly alphabetically.

sbrl:
Applied those changes, @RunDevelopment. Is this good to go now?

sbrl:
Phew! Glad to finally have this approved. Thanks for the help in the review process!

I've rebuilt it and pushed a new commit .

RunDevelopment:
Thank you for contributing @sbrl!

PR #2426, opened 2020-06-20
C++: Added support for default comparison operator

RunDevelopment:
1. Added support for [default comparison operator](https://en.cppreference.com/w/cpp/language/default_comparisons).
2. Concept declarations are now highlighted as class names.
3. Removed an `inside: null` that is never set to any value. (Seems like I forgot to remove this in a previous PR.)


components/prism-cpp.js
```diff
@@ -17,19 +17,16 @@
 			// This will capture the class name before destructors like:
 			//   Foo::~Foo() {}
 			/\b[A-Z_]\w*(?=\s*::\s*~\w+\s*\()/i,
-			{
-				// This also intends to capture the class name of method implementations but here the class has template
-				// parameters, so it can't be a namespace (until C++ adds generic namespaces).
-				pattern: /\w+(?=\s*<(?:[^<>]|<(?:[^<>]|<[^<>]*>)*>)*>\s*::\s*\w+\s*\()/,
-				inside: null // see below
```
mAAdhaTTah, line 24: Do we know what this referred to before?

RunDevelopment: Nothing. I think it used to refer to something during the development of #2348, that changed, and I forgot to remove it.

PR #2434, opened 2020-06-26
Added general de/activation mechanism for plugins

### LINKED ISSUE #2433: Class to escape keep markup on certain code elements
I love Keep Markup, but it's interacting badly with a dynamic code snippet. It would be nice to be able to escape it on a case by case basis, in the same way one can escape whitespace normalization.

Perhaps a `drop-markup` or `dont-keep-markup` class?

RunDevelopment:
The whole thing is implemented as a small `isActive` function `prism-core`. 

This adds about 150bytes to Core and about 50bytes to Keep Markup but saves about 130bytes in LN and 40bytes in NW. So we gain a few bytes overall but we also gain functionality.

---

This resolves #2433.


LeaVerou:
LGTM. I may have done the implementation of `isActive` differently, e.g. in one of the following ways:

1. Use ``element.closest(`.${className}, .no-${className}`)`` and then examine what class the result has
2. Use ``element.closest(`.${className}`)`` and ``element.closest(`.no-${className}`)`` and then if both exist use `element1.contains(element2)`

But that's just optimization :)

RunDevelopment:
I never heard of `closest` but with it (and some newer JS syntax) you can do the whole function in a one-liner:

```js
function isActive(element, className, defaultActivation) {
    return element?.closest(`${className}, no-${className}`)?.classList?.contains(className) ?? defaultActivation;
}
```

Unfortunately, my favorite browser [doesn't support it](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Browser_compatibility).

LeaVerou:
I would avoid unsupported syntax, though I'm more lenient with unsupported functions, since they can always be polyfilled. And we're discussing dropping IE11 anyway, right?

Also if I'm reading the one liner above correctly, it looks like it would fail (would return true instead of false) if an element had no-classname and defaultActivation was true.

RunDevelopment:
> And we're discussing dropping IE11 anyway, right?

Yes, but we haven't yet.

Hmmm. Maybe it's a good idea to have a tracking issue of all things that should be done after we drop IE11?

> Also if I'm reading the one liner above correctly, it looks like it would fail (would return true instead of false) if an element had no-classname and defaultActivation was true.

Nope, `(false ?? true) === false`. `defaultActivation` will only come into play if and any of the chained values is nullish.

LeaVerou:
> Hmmm. Maybe it's a good idea to have a tracking issue of all things that should be done after we drop IE11?

Absolutely.

> Nope,

Oh right.

PR #2435, opened 2020-06-26
Line numbers: Style inline code

RunDevelopment:
I noticed that some inline code snippets [aren't highlighted](https://prismjs.com/plugins/line-numbers/#examples).


RunDevelopment:
I'm just gonna merge this because of how small the change is.

PR #2438, opened 2020-06-28
Tests: Fixed all type errors

RunDevelopment:
This fixes all type errors in our test files. I just added and changed comments, so no functional changes.

The most "controversial" part of this might be the `types.d.ts` file I added. Some types cannot be expressed in JSDoc, so I just added a small declaration file and imported the types I need from there.
(Btw. the feature not supported in JSDoc is recursive type definitions.)

### DIFF
--- a/tests/dependencies-test.js
+++ b/tests/dependencies-test.js
@@ -255,7 +255,7 @@ describe('components.json', function () {
 			// and an alias, so we force the lazy alias resolver to check all aliases
 			allIds.push('js');
 
-			getLoader(components, allIds).getIds();
+			getLoader(/** @type {any} */(components), allIds).getIds();
 		} catch (error) {
 			assert.fail(error.toString());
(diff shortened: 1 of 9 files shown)


mAAdhaTTah:
How'd you confirm all the type errors were resolved? Do we want to integrate that step to maintain this going forward? Not 100% sure we _should_ but if we wanted to, how would we?

RunDevelopment:
> How'd you confirm all the type errors were resolved?

Just inside the IDE.

> Do we want to integrate that step to maintain this going forward? [...] if we wanted to, how would we?

We just need a `tsconfig.json` configured to not emit any code and then run `tsc`.

RunDevelopment:
I'll close this now. 

We plan to integrate TS via JSDoc in v2, so we will have to fix all type errors + add TS at some point anyway.

PR #2440, opened 2020-06-28
C/C++: Improved macros and expressions

RunDevelopment:
This can be seen as an extension of #2320.

Macro expressions after the directive are now highlighted. I also added some highlighting for macro-specific syntaxes like `\` at line ends and the concatenation operator.


blotfi:
when could I download the prismjs that include this #2440 enhancement?

RunDevelopment:
@blotfi Right now!

blotfi:
just to be more precise, it must be valid in **C language** (and therefore in C++ language) #2440 enhancement
thanks

blotfi:
> @blotfi Right now!

you mean if i download
https://prismjs.com/download.html#themes=prism&languages=clike+c+cpp+json&plugins=line-numbers+file-highlight+keep-markup

I will find the modification (for C and CPP) ?

RunDevelopment:
> you mean if i download [...] I will find the modification (for C and CPP) ?

Yes.

> it must be valid in C language

Macro expressions will be highlighted as C code in C and as C++ code in C++. You can see this with `#define` macros that contain C++ keywords.

C:
image

C++:
image

Is that what you meant?

blotfi:
I am testing now, 2 min
I see you put:

			'directive-hash': /^#/,
			'punctuation': /##|\\(?=[\r\n])/,
			'expression': {
				pattern: /\S[\s\S]*/,
				inside: Prism.languages.c

blotfi:
Works great
Thanks

I will open another issue for `#include <xc.h>`
it works well with file 
`<pre data-src="code/TP0/cmde.c"></pre>`
but not if it is included in the html

```
<pre data-start="0" ><code class="language-cpp line-numbers">
#include <xc.h>
return;
#ifdef NON
void list()
{
}

```


image

Issue #1665, opened 2018-12-18
Support Vue SFC syntax


hirokiosame:
Currently, it doesn't seem like Prism (or any other syntax highlighting library) supports the [Vue Single File Components](https://vuejs.org/v2/guide/single-file-components.html) syntax.

I'd like to request support for this syntax.

jwkicklighter:
Would love support for this

CesarApodaca:
I really love this library and would be great to have Vue SFC support

mAAdhaTTah:
I actually think this may be supportable without a custom syntax. If you load JavaScript & CSS then markup, you should be able to render Vue SFCs in the `markup` language and have everything highlighted properly.

RunDevelopment:
@mAAdhaTTah That won't work. Ignoring the templating syntax, there still is the issue that the `style` and `script` tags do not have to contain JS and CSS ([an example with Stylus](https://gist.github.com/chrisvfritz/1c9f2daea9bc078dcb47e9a82e5f7587)). Markup might be a good starting point but there is no avoiding a new language.

saulhardman:
> I actually think this may be supportable without a custom syntax. If you load JavaScript & CSS then markup, you should be able to render Vue SFCs in the `markup` language and have everything highlighted properly.

`html` also works nicely in my experience – thanks for the suggestion 👍

littlee:
Which version will support this?
v1.22.0 is still not cool enough when parsing `v-bind` attributes

![image](https://user-images.githubusercontent.com/7534668/98215605-f6b1f780-1f82-11eb-913a-9fd4777671b3.png)

hatboysam:
Driving by with a +1 for this issue. I'm very happy with Prism on https://codeapprove.com but this is one of the main sticking points.

BrettGregson:
+1 for Vue syntax highlighting

linlinyang:
typescript in vue3 work failed:
```vue
<template>{{ num }}</template>

<script lang="ts" setup>
const num: number = 123;
</script>
```
`num: number` syntax cannot highlight

LeaVerou:
> Which version will support this? v1.22.0 is still not cool enough when parsing `v-bind` attributes

Most likely v2, since v1 is feature frozen.

PR #2442, opened 2020-06-29
Markup: Added support type="text/plain" scripts

RunDevelopment:
This adds support for properly unhighlighted `<script type="text/plain">` regions:

image

---

The main reason I implemented this isn't `text/plain`. It's Vue (#1665).

Vue supports [embedded `script` and `style` tags](https://vue-loader.vuejs.org/guide/pre-processors.html#using-pre-processors) with languages other than Javascript and CSS. It differentiates the languages based on a `lang` attribute. 

The problem here is that the `lang` attribute can be anywhere in the list of attributes. So I changed the `addInlined` function to support arbitrary attributes as additional conditions. Example:

```js
// this will only highlight script tags with a `lang="ts"` or `lang='ts'` or `lang=ts` attribute as TS
addInlined('script', 'typescript', [{ name: /lang/, value: /ts/ }]);
```

It supports multiple attributes and will generate a regex that only matches tags that fit all criteria.

---

After this and #1991 are through, I can implement Vue.


mAAdhaTTah:
> The main reason I implemented this isn't text/plain. It's Vue (#1665).

I wonder if maybe this code should be moved into a "Vue" language? Bot hasn't spoken yet, but this seems like a lot of code to add to the core language for a non-default use case.

RunDevelopment:
Bot has spoken and it doesn't look good.

Maybe we could make `addInline` its own "language"? But JS and CSS need it, so I don't know if it'll help.

components/prism-markup.js
```diff
@@ -108,15 +117,112 @@ Object.defineProperty(Prism.languages.markup.tag, 'addInlined', {
...
-		Prism.languages.insertBefore('markup', 'cdata', def);
+		var beforeArray = before ? (Array.isArray(before) ? before : [before]) : [];
+		/** @type {Array} */
+		var existingTokens = Prism.languages.markup[tagName];
+		if (!existingTokens) {
+			var def = {};
+			def[tagName] = [token];
+			Prism.languages.insertBefore('markup', 'cdata', def);
+		} else {
+			var index = existingTokens.findIndex(function (t) {
```
mAAdhaTTah, line 217: `findIndex` isn't supported in IE11 😢

mAAdhaTTah:
> Maybe we could make `addInline` its own "language"? But JS and CSS need it, so I don't know if it'll help.

Yeah, the issue isn't the function as a whole; it's the extra logic required to support this use case. I don't know if I have a solution off the top tho.

RunDevelopment:
The new function is basically a generalization of the old one. I could factor out the code that generates the attribute loop (= the new generalization) and replace it with a function. That could work. I'll look for a solution tomorrow.

RunDevelopment:
Hmmm. Markup still gains about 200B but I don't think that I can scrap off some more.

components/prism-markup.js
```diff
@@ -108,23 +120,62 @@ Object.defineProperty(Prism.languages.markup.tag, 'addInlined', {
...
+	});
+
+	Prism.languages.html = Prism.languages.markup;
+	Prism.languages.mathml = Prism.languages.markup;
+	Prism.languages.svg = Prism.languages.markup;
+
+	Prism.languages.xml = Prism.languages.extend('markup', {});
+	Prism.languages.ssml = Prism.languages.xml;
+	Prism.languages.atom = Prism.languages.xml;
+	Prism.languages.rss = Prism.languages.xml;
```
mAAdhaTTah, line 179: I think we can flatten these two assignment series down to one line for each. See if that saves anything.

components.json
```diff
@@ -711,6 +711,11 @@
 			"alias": "md",
 			"owner": "Golmote"
 		},
+		"markup-attributes": {
+			"title": "Markup attributes",
+			"require": "markup",
+			"owner": "RunDevelopment"
+		},
```
mAAdhaTTah, line 718: If what we're doing now is designed primarily for Vue, would it be worth just making this a full-fledged Vue language for Vue SFCs? If we need this more generally for multiple languages, we can extract it then, but `markup-attributes` isn't terribly clear as to what the languages actually does.

components/prism-markup.js
RunDevelopment: 30 bytes or about 15% down.

components.json
RunDevelopment: I think that the modification it makes to markup justifies it being a "language". It would be kinda weird if Vue added support for plain text scripts in HTML. Honestly, the whole thing should be split even further: 1 "language" with just the attributes function and 1 language that adds support for plain text scripts.

Also, it's not like the whole thing is useful for just Vue and HTML. [ASP.net could also profit from it](https://github.com/PrismJS/prism/blob/b40f8f4b471ed5cb28fe0e9b33e5c4333e3a2bf4/components/prism-aspnet.js#L43) but I'll need to make some more adjustments for this.

> but markup-attributes isn't terribly clear as to what the languages actually does.

True. I can't think of a better name thou...

Issue #2441, opened 2020-06-29
#include <xc.h>  should look like that in Cpp and not be seen as html markup, also comparaison and << >> bit manipulation


blotfi:
**Language**
C and C++

**Additional resources**
the issue is for includes compilation directives `#include <xc.h>`
it works well with file 
`<pre data-src="code/TP0/cmde.c"></pre>`
but not if it is included in the html

```
<pre data-start="0" ><code class="language-cpp line-numbers">
#include <xc.h>
return;
#ifdef NON
void list()
{
}

```

if not possible to add it, how can I not use markup
or I am obliged to change all my < into `&lt; ` manually? (boring)
thanks

RunDevelopment:
You can use the [Unescaped Markup plugin](https://prismjs.com/plugins/unescaped-markup/). 

However, if you use Unescaped Markup's comment style, be sure to not use the [`-->` operator](https://stackoverflow.com/questions/1642028/what-is-the-operator-in-c) in your code. 😄

blotfi:
yes I tired it and it didn't work

and now after multiple tries I know why it was not working for me

https://prismjs.com/plugins/unescaped-markup/
if you don't ask for 
`Markup + HTML + XML + SVG + MathML + SSML + Atom + RSS 2.15KB`
it will not work

So you should clearly mention this on the page: https://prismjs.com/plugins/unescaped-markup/
Or if you check this plugins, automatically it checks the 1st syntax option 

so for my C/C++/JSON the minimum I need is;
https://prismjs.com/download.html#themes=prism&languages=markup+clike+c+cpp+json&plugins=line-numbers+file-highlight+keep-markup+unescaped-markup

RunDevelopment:
I had no idea that the plugin required Markup... It doesn't even use it, it's just a requirement... I'll remove the requirement.

RunDevelopment:
The requirement is removed now. You can remove Markup from your build. 
Sorry for the inconvenience.

blotfi:
I confirm, it works now without the markup.
At least, we found this bug!
You can _now_ close this issue

PR #2444, opened 2020-06-29
Unescaped Markup: Removed markup requirement

RunDevelopment:
See [this comment](https://github.com/PrismJS/prism/issues/2441#issuecomment-651254572).


RunDevelopment:
I'm just gonna go ahead and merge this. The change is very simple and just removes a dependency that, according to `components.json`, shouldn't even be there.

PR #2451, opened 2020-07-01
Handle edge case of string macro functions

cedporter:
The macro string functions don't necessarily need any string markup to be considered a string. This new construction handles identifying the content within a string function as a string.


components/prism-sas.js
```diff
@@ -237,7 +237,22 @@
 			lookbehind: true,
 			inside: args
 		},
-
+		'macro-string-functions': {
+			pattern: /((?:^|\s|=))%(?:NRBQUOTE|NRQUOTE|NRSTR|BQUOTE|QUOTE|STR)\(.*?(?:[^%]\))/i,
+			lookbehind: true,
+			inside: {
+				function: {
```
RunDevelopment, line 244: Please wrap this and `string` and `punctuation` into single quotes to keep things consistent.

components/prism-sas.js
```diff
@@ -237,7 +237,22 @@
...
+					pattern: /%(?:NRBQUOTE|NRQUOTE|NRSTR|BQUOTE|QUOTE|STR)/i,
+					alias: 'keyword'
+				},
+				string: {
+					pattern: /(\()(?:[\s\S]*?)(?=(?!%)\))/,
```
RunDevelopment, line 249: 1. `(?!%)\)` We know that the next character is `)`, so `(?!%)` will always be true.

The pattern is now `(\()[\s\S]*?(?=\))`. This has two other problems: 1) It can potentially produce empty tokens, so let's make it `(\()[\s\S]+?(?=\))`. 2) `[\s\S]+?(?=\))` is slower than `[^)]+`.

```suggestion
					pattern: /(\()[^)]+/,
```

components/prism-sas.js
```diff
@@ -237,7 +237,22 @@
...
+					lookbehind: true,
+					alias: 'string'
```
RunDevelopment, line 251: What? A `string` alias `string`?

components/prism-sas.js
cedporter: Addressed as well as quoting function.

components/prism-sas.js
cedporter: Yeah, wasn't thinking here.

RunDevelopment:

PR #2455, opened 2020-07-03
fix: typos "Prims"

AndersDJohnson:
Noticed some typos "Prims" in the source.


RunDevelopment:
Thank you @AndersDJohnson!

I make this typo way too often and it seems like spell check doesn't catch it... So thank you for fixing that.

PR #2460, opened 2020-07-05
Component module require fix

neel1996:
The require function looks for the component modules relatively within the current directory `E.g: require("./prismjs/components/prism-java");`. This is not working with `create-react-app` as the dependency cannot be resolved properly. 

The fix modifies the require function to check for the module globally.

### DIFF
--- a/components/index.js
+++ b/components/index.js
@@ -36,13 +36,13 @@ function loadLanguages(languages) {
 			return;
 		}
 
-		const pathToLanguage = './prism-' + lang;
+		const pathToLanguage = '/prism-' + lang;
 
 		// remove from require cache and from Prism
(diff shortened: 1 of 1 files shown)


RunDevelopment:
Could you please elaborate on why relative requires don't work?

I can't believe that we're the only project that uses relative paths to refer to files. Isn't this a problem with `create-react-app`?

mAAdhaTTah:
You shouldn't be using `loadLanguages` with `create-react-app` in the first. Unfortunately, because CRA doesn't provide a way to modify its `.babelrc` file, so the only way we could support CRA is with a babel macro. Filed a ticket in the babel plugin for that: https://github.com/mAAdhaTTah/babel-plugin-prismjs/issues/111

PR #2463, opened 2020-07-06
Macro string functions

cedporter:
Don't actually need to mark the contents of these macro string functions as strings. Do need to ensure that unmatched quotes within them are not erroneously identified as the start and end of strings. Also improved macro variable identification.


components/prism-sas.js
```diff
@@ -245,19 +255,14 @@
 					pattern: /%(?:NRBQUOTE|NRQUOTE|NRSTR|BQUOTE|QUOTE|STR)/i,
 					alias: 'keyword'
 				},
-				'string': {
-					pattern: /(\()[^)]+/,
-					lookbehind: true
+				'macro-keyword': macroKeyword,
+				'macro-variable': macroVariable,
+				'escaped-char': {
+					pattern: /%(?:'|"|\(|\)|<|>|=|¬|\^|~|;|,|#)/i,
```
RunDevelopment, line 261:
```suggestion
					pattern: /%['"()<>=¬^~;,#]/,
```

components/prism-sas.js
```diff
@@ -8,6 +8,16 @@
 		alias: 'number'
 	};
 
+	var macroVariable = {
+		pattern: /&[a-z_][a-z_1-9]*/i
```
RunDevelopment, line 12: `0` is not allowed?

cedporter: It is allowed. Thanks for catching that.

RunDevelopment:
Thank you for contributing!

PR #2470, opened 2020-07-10
Autoloader: Invert dependencies and aliases maps

RunDevelopment:
This inverts Autoloader's dependencies and aliases map. 

The idea behind this is that in a dependency map, most elements will depend on only a few elements. The best example of this is `clike`. Previously, the string `"clike"` occurred as many times as there were components that depended on C-like. Now, the string appears once. Doing this for all dependencies and aliases saves about 0.5kB (or about 10%) for the minified file's size.

---

Please tell me whether we actually want this. A 10% decrease isn't much and this does add more complexity with the double inversion.


gulpfile.js/index.js
```diff
@@ -168,9 +170,54 @@ async function languagePlugins() {
...
+		}
+		inlineElement(inv);
+		return inv;
+	}
+
 	const jsonLanguagesMap = formattedStringify(nonNullLanguageMap);
-	const jsonDependenciesMap = formattedStringify(dependenciesMap);
-	const jsonAliasMap = formattedStringify(aliasMap);
+	const jsonDependenciesMap = `invertRecord(${formattedStringify(invertRecord(dependenciesMap))})`;
+	const jsonAliasMap = `invertRecord(${formattedStringify(invertRecord(aliasMap))})`;
```
mAAdhaTTah, line 220: Is this right? This looks weird. Is the intention to prefix the string with `invertRecord`?

RunDevelopment: Autoloader has an `invertRecord` function and the prefix is there to call that function. We have to invert the structure twice to get the original back.

RunDevelopment:
Bot has spoken.

PR #2471, opened 2020-07-10
EditorConfig support

### LINKED ISSUE #2462: EditorConfig syntax highlight support
**Language**
.editorconfig

**Additional resources**
Main Page: https://editorconfig.org/
Specification: https://editorconfig-specification.readthedocs.io/en/latest/

osipxd:
Tried to add EditorConfig support. Please check if I've done it right.  
image

Closes #2462


components/prism-editorconfig.js
```diff
@@ -0,0 +1,20 @@
...
+	'comment': /[;#].*$/m,
+	'section': {
+		pattern: /^[ \t]*\[.+]/m,
+		alias: 'keyword',
+		inside: {
+			'regex': /\\\\[\[\]{},!?.*]/, // Escape special characters with '\\'
+			'punctuation': /[\[\]{},!?]|\.\.|\*{1,2}/
+		}
+	},
+	'property': /^[ \t]*[^\s=]+?(?=[ \t]*=)/m,
```
RunDevelopment, line 12: The `[^\s=]+?` doesn't have to be lazy because we know that the character after it will be either a space, tab, or equals sign because of the lookahead.

```suggestion
	'property': /^[ \t]*[^\s=]+(?=[ \t]*=)/m,
```

components/prism-editorconfig.js
```diff
@@ -0,0 +1,20 @@
+Prism.languages.editorconfig = {
+	// https://editorconfig-specification.readthedocs.io/en/latest/
+	'comment': /[;#].*$/m,
+	'section': {
+		pattern: /^[ \t]*\[.+]/m,
```
RunDevelopment, line 5:
```suggestion
		pattern: /^[ \t]*\[.+\]/m,
```

components/prism-editorconfig.js
```diff
@@ -0,0 +1,20 @@
+Prism.languages.editorconfig = {
+	// https://editorconfig-specification.readthedocs.io/en/latest/
+	'comment': /[;#].*$/m,
```
RunDevelopment, line 3: You don't need the `$`. If the `m` flag is present, `$` will be the same as `(?!.)`. `.*` is greedy, so we know that the next character (if any) will not match `.`, so the `$` == `(?!.)` is redundant.

```suggestion
	'comment': /[;#].*/,
```

components/prism-editorconfig.js
```diff
@@ -0,0 +1,20 @@
...
+	'section': {
+		pattern: /^[ \t]*\[.+]/m,
+		alias: 'keyword',
+		inside: {
+			'regex': /\\\\[\[\]{},!?.*]/, // Escape special characters with '\\'
+			'punctuation': /[\[\]{},!?]|\.\.|\*{1,2}/
```
RunDevelopment, line 9: I think it would great if the `*`, `**`, `?`, `!`, and `..` were visually distinct from the rest.

```suggestion
			'operator': /[!?]|\.\.|\*{1,2}/,
			'punctuation': /[\[\]{},]/
```

components/prism-editorconfig.js
```diff
@@ -0,0 +1,20 @@
...
+		}
+	},
+	'property': /^[ \t]*[^\s=]+?(?=[ \t]*=)/m,
+	'value': {
+		pattern: /=.*/,
+		alias: 'string',
+		inside: {
+			'punctuation': /^[=]/
```
RunDevelopment, line 17:
```suggestion
			'punctuation': /^=/
```

components/prism-editorconfig.js
osipxd: My IDE highlights it as a redundant escape.
image
Should I just ignore this?

components/prism-editorconfig.js
osipxd: Yes, it looks much better

components/prism-editorconfig.js
osipxd: Fixed all comments except this one

RunDevelopment: Technically, it is redundant. I don't like unescaped `]`a for various reasons, but this shouldn't hold this pull back.

RunDevelopment:

PR #2481, opened 2020-07-22
Added support for .gitignore

### LINKED ISSUE #2475: Add gitignore support
**Language**
.gitignore
I think it also covers .hgignore

**Additional resources**
https://git-scm.com/docs/gitignore
https://www.mercurial-scm.org/wiki/.hgignore

osipxd:
Closes #2475

I've called the main language `ignore` because it can be used in many ignore configs. Like `.hgignore` or `.npmignore`.  
Implementation inspired with [.ignore](https://plugins.jetbrains.com/plugin/7495--ignore) plugin.


components/prism-ignore.js
```diff
@@ -0,0 +1,34 @@
+(function (Prism) {
+	Prism.languages.ignore = {
+		// https://git-scm.com/docs/gitignore
+		'header': {
+			pattern: /(^|[^\\])###.*/,
```
RunDevelopment, line 5: Since the whole format is line-based, we can simply require a header to start at the start of a line. That way, we don't have to check for any backslashes and we won't have problems with comments like this:

```
# comment ### foo
```

Please use the same trick for the `section` and `comment` patterns as well.

```suggestion
			pattern: /^###.*/m,
```

components/prism-ignore.js
```diff
@@ -0,0 +1,34 @@
...
+		},
+		'section': {
+			pattern: /(^|[^\\])##.*/,
+			lookbehind: true,
+			alias: 'comment'
+		},
+		'comment': {
+			pattern: /(^|[^\\])#.*/,
+			lookbehind: true,
+			alias: 'namespace'
```
RunDevelopment, line 17: What? Please let the themes style comments as comments.

components/prism-ignore.js
```diff
@@ -0,0 +1,34 @@
...
+		},
+		'entry': {
+			pattern: /\S(?:.*(?:(?:\\ )|\S))?/,
+			alias: 'string',
+			inside: {
+				'regex': {
+					pattern: /^!|(^|[^\\])[\[\]]/,
+					lookbehind: true
+				},
```
RunDevelopment, line 26: I think the `!` should be highlighted as an operator.

Also, instead of highlighting the `[]` delimiters of regexes, wouldn't it be more useful to highlight the whole regex?

```suggestion
			inside: {
				'operator': /^!/,
				'regex': {
					pattern: /(^|[^\\])\[[^\[\]]*\]/,
					lookbehind: true
				},
```

components/prism-ignore.js
```diff
@@ -0,0 +1,34 @@
...
+		// https://git-scm.com/docs/gitignore
+		'header': {
+			pattern: /(^|[^\\])###.*/,
+			lookbehind: true,
+			alias: ['comment', 'bold']
+		},
+		'section': {
+			pattern: /(^|[^\\])##.*/,
+			lookbehind: true,
+			alias: 'comment'
```
RunDevelopment, line 12: Maybe use `important` instead?

components/prism-ignore.js
osipxd: Yes, that looks better

osipxd:
> What are headers and sections? The spec doesn't mention those.

It isn't from specification it is just additional highlight to better structurize gitignore file in case if it is large.
I'll remove `section` and `header` if you think it will be better.
 
> Also, it might be nice to have an alias for [`.npmignore` files](https://docs.npmjs.com/using-npm/developers.html#:~:text=.npmignore%20files%20follow%20the%20same%20pattern%20rules%20as%20.gitignore%20files) _edit:_ and `.hgignore` files.

Ok, I'll add it

components/prism-ignore.js
osipxd: It's just a hack to make comments semi-transparent.
I wait for your decision about sections and headers and then I'll fix it

components/prism-ignore.js
osipxd: I wait for your decision about sections and headers and then I'll fix it

RunDevelopment:
> It isn't from specification it is just additional highlight to better structurize gitignore file in case if it is large.

It tried searching for this practice and I wasn't able to find anything. It doesn't seem to be common, so let's just go with the spec for now and remove `section` and `header`.

components/prism-ignore.js
osipxd: Removed alias

components/prism-ignore.js
osipxd: Removed section and header

RunDevelopment:

PR #2482, opened 2020-07-22
EditorConfig trim spaces before key and section title

### LINKED ISSUE #2472: Leading whitespaces shouldn’t be relevant in EditorConfig
**Information**

- Language: EditorConfig
- Plugins: none

**Does the problem still occur in the latest version of Prism?**

[Yes](https://prismjs.com/test.html#language=editorconfig).

**Description**

Contrary to the [specification](https://editorconfig-specification.readthedocs.io/en/latest/#file-format), all leading whitespaces on each line are not considered irrelevant.

**Code snippet**

<details>

Prism renders

```editor-config
    [Section1]
    key1=value1
```

in this way

```html
<span class="token section keyword">    <span class="token punctuation">[</span>Section1<span class="token punctuation">]</span></span>
<span class="token property">    key1</span><span class="token value string"><span class="token punctuation">=</span>value1</span>
```

</details>

osipxd:
Fixes #2472


components/prism-editorconfig.js
```diff
@@ -2,15 +2,19 @@ Prism.languages.editorconfig = {
 	// https://editorconfig-specification.readthedocs.io/en/latest/
 	'comment': /[;#].*/,
 	'section': {
-		pattern: /^[ \t]*\[.+]/m,
+		pattern: /(^|[ \t]*)\[.+]/,
```
RunDevelopment, line 5: I think that's not what you meant. Since the lookbehind can match the empty string in the `[ \t]*` alternative, the lookbehind will trivially accept any string. This means that it will even match things like this:

```
foo[what?]
```

```suggestion
		pattern: /(^[ \t]*)\[.+\]/m,
```

components/prism-editorconfig.js
```diff
@@ -2,15 +2,19 @@ Prism.languages.editorconfig = {
...
 		alias: 'keyword',
 		inside: {
 			'regex': /\\\\[\[\]{},!?.*]/, // Escape special characters with '\\'
 			'operator': /[!?]|\.\.|\*{1,2}/,
 			'punctuation': /[\[\]{},]/
 		}
 	},
-	'property': /^[ \t]*[^\s=]+(?=[ \t]*=)/m,
+	'property': {
+		pattern: /(^|[ \t]*)[^\s=]+(?=[ \t]*=)/,
```
RunDevelopment, line 15: Same here.
```suggestion
		pattern: /(^[ \t]*)[^\s=]+(?=[ \t]*=)/s,
```

components/prism-editorconfig.js
osipxd: Yes, you're right. Fixed it.

components/prism-editorconfig.js
osipxd: Done

RunDevelopment:
Thank you for contributing!

PR #2484, opened 2020-07-24
Fixed Treeview page

### LINKED ISSUE #2483: The treeview plugin page is broken
Hello.

On the treeview plugin page:
https://prismjs.com/plugins/treeview/

seems like some assets files missed and in header and footer the page show 404 error.

Please check it.

RunDevelopment:
This fixes #2483.

Seems like I forgot to include this page into #2180.

### DIFF
--- a/plugins/treeview/index.html
+++ b/plugins/treeview/index.html
@@ -3,7 +3,7 @@
 <head>
 
 	<meta charset="utf-8" />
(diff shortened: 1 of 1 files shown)


RunDevelopment:
I will just merge this because the fix is both simple and urgent.

PR #2485, opened 2020-07-26
Batch: Fix escaped double quote

proudust:
Fix string token not taking escaped double quotes into account.


components/prism-batch.js
```diff
@@ -7,7 +7,7 @@
 			'punctuation': /:/
 		}
 	};
-	var string = /"[^"]*"/;
+	var string = /"(?:[\\"]"|[^"])*"/;
```
RunDevelopment, line 10: Wait. So you can escape a `"` with either `\"` or `""`?

proudust: See [this stackoverflow post](https://stackoverflow.com/questions/7760545/escape-double-quotes-in-parameter). I also confirmed it, but confirmed the same behavior.

RunDevelopment: I see. Then let's quickly fix the pattern and then merge this.

The problem is that right now, it might cause [exponential backtracking](https://www.rexegg.com/regex-explosive-quantifiers.html#:~:text=Backtracking%20is%20a%20wonderful%20feature,have%20taken%20a%20different%20path.). The problem is that we can match strings like `"a""b"` as either one match `["a""b"]` or two matches `["a"]["b"]`. 

TBH, it doesn't actually cause exponential backtracking right now. In this pattern it's not in a loop (`*`) and in the below pattern, there is no suffix. So while it's technically not a problem right now, I still want to fix it. People (myself included) tend to just copy and paste small patterns like this making it really easy to accidentally create a pattern with exponential runtime.
A good example of where this might happen is the below pattern. If you add anything at all after the last `*` (e.g. a lookahead to exclude some false positives), the string part of the pattern is going to cause exponential backtracking.

Well, enough explaining. How do we fix it? It's quite simple. We only have to make sure that can't match another string directly after a string, so we do this:
```suggestion
	var string = /"(?:[\\"]"|[^"])*"(?!")/;
```

components/prism-batch.js
```diff
@@ -76,7 +76,7 @@
 			},
 			{
 				// Other commands
-				pattern: /((?:^|[&(])[ \t]*@?)\w+\b(?:[^^&)\r\n]|\^(?:\r\n|[\s\S]))*/im,
+				pattern: /((?:^|[&(])[ \t]*@?)\w+\b(?:"(?:[\\"]"|[^"])*"|[^^&)\r\n]|\^(?:\r\n|[\s\S]))*/im,
```
RunDevelopment, line 79: For similar reasons as mentioned above, this might cause exponential backtracking. This is how to fix it.

```suggestion
				pattern: /((?:^|[&(])[ \t]*@?)\w+\b(?:"(?:[\\"]"|[^"])*"(?!")|[^"^&)\r\n]|\^(?:\r\n|[\s\S]))*/im,
```

RunDevelopment:
Thank you for contributing @proudust!

PR #2489, opened 2020-07-28
Kotlin: backticks in function name

osipxd:
According to [this](https://kotlinlang.org/docs/reference/java-interop.html#escaping-for-java-identifiers-that-are-keywords-in-kotlin) and [this](https://kotlinlang.org/docs/reference/coding-conventions.html#names-for-test-methods)


components/prism-kotlin.js
```diff
@@ -1,15 +1,22 @@
 (function (Prism) {
+
+	var functionName = /(?:`[\w !"#$%^&()*+,=?@{|}~\-]+`|\w+)/.source;
```
RunDevelopment, line 3: According to [Kotlin's spec](https://kotlinlang.org/spec/syntax-and-grammar.html#identifiers), it should be this:

```suggestion
	var functionName = /(?:`[^\r\n`]+`|\w+)/.source;
```

With this, the pattern is short enough to be inlined, I think.

osipxd: Fixed

RunDevelopment (approved):

RunDevelopment:

PR #2494, opened 2020-08-04
Add support for NaniScript

Elringus:
NaniScript is a scenario scripting language used by Naninovel visual novel extension for Unity: https://naninovel.com


Elringus:
Sorry for the ongoing commits; it's ready now.

components/prism-naniscript.js
```diff
@@ -0,0 +1,189 @@
...
+		if (token.type === 'generic-text') {
+			var content = getContents(token, []).join('');
+			if (!isBracketsBalanced(content)) {
+				token.type = 'bad-line';
+				token.content = content;
+				env.tokens[index] = token;
+			}
+		}
+		return false;
+	});
```
RunDevelopment, line 151:
```suggestion
	env.tokens.forEach(function (token, index) {
		if (token.type === 'generic-text') {
			var content = getContents(token, []).join('');
			if (!isBracketsBalanced(content)) {
				token.type = 'bad-line';
				token.content = content;
				env.tokens[index] = token;
			}
		}
	});
```

components/prism-naniscript.js
```diff
@@ -0,0 +1,189 @@
...
+		lookbehind: true,
+	},
+	// > ...
+	// Define is a control line starting with '>' followed by a word, a space and a text.
+	'define': {
+		pattern: /^>.+$/m,
+		alias: 'tag',
+		inside: {
+			value: {
+				pattern: /(^>[a-zA-Z0-9_]+)[\t ]+[^{}\r\n]+/,
```
RunDevelopment, line 21: We generally don't want tokens to start or end with spaces. There are cases where that's hard to do but here it's easy.

```suggestion
				pattern: /(^>[a-zA-Z0-9_]+[\t ]+)[^{}\r\n]+/,
```

components/prism-naniscript.js
```diff
@@ -0,0 +1,189 @@
...
+					'quoted-string': {
+						pattern: /"(?:[^"\\]|\\.)*"/
+					},
+					'command-param-id': {
+						alias: 'property',
+						pattern: /([^"\\]?)[a-zA-Z0-9_]+:/,
+						lookbehind: true,
+					},
+					'command-param-name': {
+						pattern: /^(\[)[\t ]*[a-zA-Z0-9_]+/,
```
RunDevelopment, line 109:
```suggestion
						pattern: /^(\[[\t ]*)[a-zA-Z0-9_]+/,
```

components/prism-naniscript.js
```diff
@@ -0,0 +1,189 @@
...
+							pattern: expressionDef(false),
+							alias: 'selector',
+						},
+						{
+							pattern: /([\t ]+).+?(?=[\t ]|$)/,
+							alias: 'operator',
+							lookbehind: true
+						},
+						{
+							pattern: /.+/,
```
RunDevelopment, line 125: This also captures all spaces between parameters.
```suggestion
							pattern: /\S(?:.*\S)?/,
```

components/prism-naniscript.js
```diff
@@ -0,0 +1,189 @@
...
+				pattern: /\[[\t ]*[a-zA-Z0-9_]+[^\r\n\[]*\]/m,
+				greedy: true,
+				alias: 'function',
+				inside: {
+					'quoted-string': {
+						pattern: /"(?:[^"\\]|\\.)*"/
+					},
+					'command-param-id': {
+						alias: 'property',
+						pattern: /([^"\\]?)[a-zA-Z0-9_]+:/,
```
RunDevelopment, line 105: What's the intention behind the lookbehind?

components/prism-naniscript.js
```diff
@@ -0,0 +1,189 @@
...
+		pattern: /^([\t ]*)@[a-zA-Z0-9_]+(?=[\t ]+[\S]{1,}|[\t ]+['"][\S]|[\t ]+$|$).*/m,
+		lookbehind: true,
+		alias: 'function',
+		inside: {
+			expression: {
+				pattern: expressionDef(),
+				greedy: true,
+				alias: 'selector'
+			},
+			'whitespaces': /[\t ]+$/,
```
RunDevelopment, line 47: Why? I don't see any value in highlighting whitespaces.

Elringus:
Thank you for the review! @vladdancer is actually the main contributor here, hopefully he'll be able to provide some feedback on the suggestions. A PR to the fork would be great, thanks!

RunDevelopment:
Quick question: How does escaping in strings work? The [doc](https://naninovel.com/guide/naninovel-scripts.html#comment-lines) gives this example:

```
"Saying \\"Stop the car\\" was a mistake."
```

But I guess that this isn't how it is supposed to be tokenized:

image
(Screenshot of the DOM of the website)

So how does it work? It's not mentioned anywhere and the double escaping seems weird to me.

Elringus:
That's a special case here. The value of `if` operators is treated as expression by default (to save the user from typing curly braces every time) and expressions are evaluated with their own parser after being parsed by the main one, hence the need for double escape: `\\"` is replaced with `\"` on the initial parse and then the expression evaluator receive a properly escaped quote.

RunDevelopment:
I see. Since doing this correctly will be quite difficult, let's just leave it as is.

components/prism-naniscript.js
```diff
@@ -0,0 +1,189 @@
+var expressionDef = function (isMultiline) {
+	if (isMultiline === void 0) { isMultiline = true; }
+	return isMultiline
+		? /\{[^\r\n\[\]{]*\}(?=[^\r\n\[\]\}]*)/m
```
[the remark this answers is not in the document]
vladdancer: I think it is leaved from some code refactor, even before writing tests.
I've tried to validate string for unclosed brackets using regex, but then moved to validating them by function.
Thank you for heads up.

components/prism-naniscript.js
```diff
@@ -0,0 +1,189 @@
...
+			},
+			key: {
+				pattern: /(^>)[a-zA-Z0-9_]+/,
+				lookbehind: true,
+			}
+		}
+	},
+	// # ...
+	'label': {
+		pattern: /^([\t ]*)[#]{1}[\t ]*[a-zA-Z0-9]+[\t ]?$/m,
```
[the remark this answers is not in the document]
vladdancer: Thank for the tip. I'm not a regex guru. Yeah, '#' plus as many tabs,spaces

components/prism-naniscript.js
```diff
@@ -0,0 +1,189 @@
...
+			'escaped-char': {
+				pattern: /\\[{}\[\]"]{1}/m
+			},
+			expression: {
+				pattern: expressionDef(),
+				greedy: true,
+				alias: 'selector'
+			},
+			'inline-command': {
+				pattern: /\[[\t ]*[a-zA-Z0-9_]+[^\r\n\[]*\]/m,
```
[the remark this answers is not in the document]
vladdancer: You a right. Fixed.

vladdancer:
@RunDevelopment thank you for a such great review! 
I fixed remain issues and a couple of small improvements, added PR in Elringus repo.

components/prism-naniscript.js
```diff
@@ -0,0 +1,172 @@
+(function (Prism) {
+
+	function expressionDef(isMultiline) {
+		return RegExp(/\{[^\r\n\[\]{}]*\}/, isMultiline ? 'm': '');
+	};
```
RunDevelopment, line 5: The m flag doesn't affect the pattern at all, so let's just make this a variable instead of a function.

```suggestion
	var expressionDef = /\{[^\r\n\[\]{}]*\}/;
```

tests/languages/naniscript/label_feature.test
```diff
@@ -0,0 +1,24 @@
+#
+# Section1
+#Section2
+# Section4
+#			SectionWithMultipleTabsBefore
+## Section3
+# Section with multiple words
+	# Section with
+	# Section with tab
```
RunDevelopment, line 9: Are these valid sections or not? The JSON says no but the name of those "sections" makes it seems like they should be valid.

components/prism-naniscript.js
```diff
@@ -0,0 +1,172 @@
...
+				},
+				'key': {
+					pattern: /(^>)\w+/,
+					lookbehind: true,
+				}
+			}
+		},
+		// # ...
+		'label': {
+			pattern: /^([\t ]*)#[\t ]*\w+[\t ]?$/m,
```
RunDevelopment, line 60: This still has to be changed.

```suggestion
			pattern: /^([\t ]*)#[\t ]*\w+[\t ]*$/m,
```

tests/languages/naniscript/label_feature.test
Elringus: Those are not valid. Not sure which JSON you're referring to.

RunDevelopment: > Those are not valid.

Ok. It's all good then.

> Not sure which JSON you're referring to.

The JSON of the expected token stream below.

components/prism-naniscript.js
```diff
@@ -0,0 +1,170 @@
...
+			alias: 'regex'
+		},
+		'command': {
+			pattern: /^([\t ]*)@\w+(?=[\t ]|$).*/m,
+			lookbehind: true,
+			alias: 'function',
+			inside: {
+				'command-name': /^@\w+/,
+				'expression': {
+					pattern: expressionDef(true),
```
RunDevelopment, line 69: It's not a function anymore.

Elringus: Fixed, thanks!

RunDevelopment:
Thank you for contributing @Elringus and @vladdancer!

Elringus:
Thank you for the help and patience with us!

vladdancer:
@RunDevelopment Thanks one more time for such great review, it gives me more details about regular expressions!

PR #2505, opened 2020-08-05
Added support for TypoScript

dkern:
Added support for TypoScript/TSConfig highlighting.


components.json
```diff
@@ -1137,6 +1137,15 @@
 			"alias": "ts",
 			"owner": "vkbansal"
 		},
+		"typoscript": {
+			"title": "TypoScript",
+			"require": "javascript",
+			"alias": "tsconfig",
+			"aliasTitles": {
+				"tsconfig": "TSConfig"
+			},
```
RunDevelopment, line 1146: Given the popularity of TypeScript, I'm really not a fan of this alias. It's very confusing because if I hear "tsconfig", I'll think of [this](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html).

components/prism-typoscript.js
```diff
@@ -0,0 +1,31 @@
+Prism.languages.typoscript = Prism.languages.extend('javascript', {
+    'comment': /(?:\s|^)[#][^#^\r\n]{2,}?(?:\r?\n|$)|(?:\s|^)\/\/.*(?:\r?\n|$)/g,
```
RunDevelopment, line 2: The preceding space shouldn't be part of the token.

Also, you don't need a `g` flag in any patterns. If Prism needs it, it will be added.

```suggestion
	'comment': {
		pattern: /(^|\s)(?:#[^#^\r\n]{2,}(?:\r?\n|$)|\/\/.*)/,
		lookbehind: true
	},
```

components/prism-typoscript.js
```diff
@@ -0,0 +1,31 @@
...
+    }, {
+        pattern: /pagets__.*(?:\s|=|.)/g
+    }],
```
RunDevelopment, line 8: 1. Any grammar token `{ pattern: /foo/ }` can be replaced with `/foo/`. Please also do this with all other objects of this form.
1. Please sort long keyword lists. This makes it easy to identify missing or duplicate keywords.
1. Regarding `/pagets__.*(?:\s|=|.)/`
    1. Shouldn't it start with `\b`?
    1. `.*` is greedy. Does that mean it's allowed to match e.g. spaces?

```suggestion
	'tag': [
		/\b(?:_CSS_DEFAULT_STYLE|_DEFAULT_PI_VARS|_LOCAL_LANG|_offset|absRefPrefix|accessibility|accessKey|ACT|ACTIFSUB|ACTIFSUBR|ACTRO|addAttributes|addChr10BetweenParagraphs|addClasses|addClassesCol|addClassesImage|addExtUrlsAndShortCuts|addItems|additionalHeaders|additionalParams|addParams|addQueryString|adjustItemsH|adjustSubItemsH|adminPanelStyles|admPanel|after|afterImg|afterImgLink|afterImgTagParams|afterROImg|afterWrap|age|alertPopups|align|all|allImageCaptions|allow|allowCaching|allowedAttribs|allowedClasses|allowedCols|allowedNewTables|allowTags|allowTVlisting|allSaveFunctions|allStdWrap|allWrap|alt_print|alternateBgColors|alternativeSortingField|alternativeTempPath|altImgResource|altLabels|altTarget|altText|altUrl|altUrl_noDefaultParams|altWrap|always|alwaysActivePIDlist|alwaysLink|alwaysShowClickMenuInTopFrame|andWhere|angle|antiAlias|append|applicationContext|applyTotalH|applyTotalW|archive|archiveTypoLink|arrayReturnMode|arrowACT|arrowImgParams|arrowNO|as|ascii|ATagAfterWrap|ATagBeforeWrap|ATagParams|ATagTitle|atLeast|atMost|attribute|auth|autoLevels|autonumber|backColor|backend_layout|BackendLayouts|background|badMess|baseURL|BE|be_groups|be_users|before|beforeImg|beforeImgLink|beforeImgTagParams|beforeLastTag|beforeROImg|beforeWrap|begin|beLoginLinkIPList|beLoginLinkIPList_login|beLoginLinkIPList_logout|bgCol|bgImg|blankStrEqFalse|blur|bm|bodyTag|bodyTagAdd|bodyTagCObject|bodyTagMargins|bodytext|border|borderClass|borderCol|borderPadding|borderSpace|bordersWithin|borderThick|borderWidth|bottomBackColor|bottomContent|bottomHeight|bottomImg|bottomImg_mask|BOX|br|browse|browser|brTag|bullet|bulletlist|bullets|bytes|cache|cache_clearAtMidnight|cache_period|caption|caption_stdWrap|captionAlign|captionHeader|captionSplit|CARRAY|case|casesensitiveComp|categories|cellpadding|cellspacing|centerImgACT|centerImgCUR|centerImgNO|centerLeftImgACT|centerLeftImgCUR|centerLeftImgNO|centerRightImgACT|centerRightImgCUR|centerRightImgNO|char|charcoal|charMapConfig|check|CHECK|class|classesAnchor|classesCharacter|classesImage|classesParagraph|classicPageEditMode|classStdWrap|clear|clearCache|clearCache_disable|clearCache_pageGrandParent|clearCache_pageSiblingChildren|clearCacheCmd|clearCacheLevels|clearCacheOfPages|CLEARGIF|clickMenuTimeOut|clickTitleMode|clipboardNumberPads|cMargins|cObj|cObject|cObjNum|colCount|collapse|collections|color|color1|color2|color3|color4|colors|colour|colPos|colPos_list|colRelations|cols|colSpace|colspan|columns|COLUMNS|columnSpacing|columnStdWrap|COMMENT|comment_auto|commentWrap|compensateFieldWidth|compressCss|compressJs|compX|compY|concatenateCss|concatenateJs|conf|config|CONFIG|constant|constants|CONSTANTS|content|content_from_pid_allowOutsideDomain|contextMenu|copy|copyLevels|count_HMENU_MENUOBJ|count_menuItems|count_MENUOBJ|create|createFoldersInEB|crop|csConv|CSS_inlineStyle|CTABLE|CType|CUR|CURIFSUB|CURIFSUBRO|current|CURRO|curUid|cut|cWidth|data|dataArray|dataKey|dataProcessing|dataWrap|date|date_stdWrap|datePrefix|dayofmonth|dayofweek|DB|db_list|debug|debugData|debugFunc|debugItemConf|debugRenumberedObject|default|defaultAlign|defaultCmd|defaultFileUploads|defaultHeaderType|defaultOutput|defaults|defaultType|delete|denyTags|depth|DESC|description|device|dimensions|directImageLink|directionLeft|directionUp|directory|directReturn|disableAdvanced|disableAllHeaderCode|disableAltText|disableBigButtons|disableBodyTag|disableCacheSelector|disableCharsetHeader|disabled|disableDelete|disableDocSelector|disableHideAtCopy|disableIconLinkToContextmenu|disableItems|disableNewContentElementWizard|disableNoMatchingValueElement|disablePageExternalUrl|disablePrefixComment|disablePrependAtCopy|disableSearchBox|disableSingleTableView|disableTabInTextarea|displayActiveOnLoad|displayContent|displayFieldIcons|displayIcons|displayMessages|displayQueries|displayRecord|displayTimes|distributeX|distributeY|div|DIV|doctype|doctypeSwitch|DOCUMENT_BODY|doktype|doNotLinkIt|doNotShowLink|doNotStripHTML|dontCheckPid|dontFollowMouse|dontHideOnMouseUp|dontLinkIfSubmenu|dontShowPalettesOnFocusInAB|dontWrapInTable|doubleBrTag|dWorkArea|dynCSS|edge|edit|edit_access|edit_docModuleUplaod|edit_docModuleUpload|edit_pageheader|edit_RTE|editFieldsAtATime|editFormsOnPage|editIcons|editIconsStdWrap|editNoPopup|editPanel|EFFECT|element|elements|else|email|emailMeAtLogin|emailMess|emboss|emptyTitleHandling|enable|encapsLines|encapsLinesStdWrap|encapsTagList|end|entryLevel|equalH|equals|everybody|excludeDoktypes|excludeUidList|expAll|expand|explode|ext|external|externalBlocks|extTarget|face|fallbackRendering|false|FE|fe_adminLib|fe_groups|fe_users|feadmin|field|fieldDelimiter|fieldName|fieldOrder|fieldRequired|fields|fieldWrap|file|file1|file2|file3|file4|file5|filelink|filelist|fileTarget|firstLabel|firstLabelGeneral|fixAttrib|flip|flop|fluidtemplate|folder|folders|folderTree|foldoutMenu|foldSpeed|foldTimer|fontColor|fontFile|fontOffset|fontSize|fontSizeMultiplicator|fontTag|forceDisplayFieldIcons|forceDisplayIcons|forceTemplateParsing|forceTypeValue|FORM|format|frame|FRAME|frameReloadIfNotInFrameset|frameSet|FRAMESET|freezeMouseover|ftu|function|Functions|gamma|gapBgCol|gapLineCol|gapLineThickness|gapWidth|get|getBorder|getLeft|getRight|global|globalNesting|globalString|globalVar|gmenu_foldout|GMENU_FOLDOUT|gmenu_layers|GMENU_LAYERS|goodMess|GP|gray|group|groupBy|groupid|header|header_layout|headerComment|headerData|headerSpace|headTag|height|helpText|hidden|hiddenFields|hide|hideButCreateMap|hideMenuTimer|hideMenuWhenNotOver|hidePStyleItems|hideRecords|hideSubmoduleIcons|highColor|history|hostname|hour|html|HTML|HTMLparser|HTMLparser_tags|htmlSpecialChars|htmlTag_dir|htmlTag_langKey|htmlTag_setParams|http|icon|icon_image_ext_list|icon_link|iconCObject|iconTitle|id|IENV|if|ifBlank|ifEmpty|IFSUB|IFSUBRO|image|image_compression|image_effects|image_frames|imageColumnStdWrap|imageLinkWrap|imagePath|images|imageStdWrap|imageStdWrapNoWidth|imageTextSplit|imageWrapIfAny|imgList|imgMap|imgMapExtras|imgMax|IMGMENUITEM|imgNameNotRandom|imgNamePrefix|imgObjNum|imgParams|imgPath|imgResource|imgStart|imgTagStdWrap|imgText|IMGTEXT|inBranch|inc|INCLUDE_TYPOSCRIPT|includeCSS|includeCSSLibs|includeJS|includeJSFooter|includeJSFooterlibs|includeJSLibs|includeLibrary|includeLibs|includeNotInMenu|includeRecordsWithoutDefaultTranslation|includeSpacer|incT3Lib_htmlmail|index|index_descrLgd|index_enable|index_externals|index_metatags|info|inlineStyle2TempFile|innerStdWrap|innerStdWrap_all|innerWrap|innerWrap2|input|inputLevels|insertClassesFromRTE|insertData|insertDmailerBoundaries|intensity|intTarget|intval|invert|IP|IProcFunc|isFalse|isGreaterThan|isInList|isLessThan|isPositive|isTrue|itemArrayProcFunc|itemH|items|itemsProcFunc|iterations|join|jsmenu|JSMENU|JSMENUITEM|JSwindow|JSWindow|JSwindow_params|jumpurl|jumpUrl|jumpurl_enable|jumpurl_mailto_disable|jumpUrl_transferSession|keep|keepEntries|keepNonMatchedTags|key|keyword3|label|LABEL|labelStdWrap|labelWrap|lang|language|language_alt|languageField|lastRowStdWrap|layer_menu_id|layerStyle|layout|layoutKey|layoutRootPath|layoutRootPaths|left|leftIcons|leftImgACT|leftImgCUR|leftImgNO|leftjoin|leftOffset|levels|leveltitle|leveluid|lib|limit|line|lineColor|lineThickness|linkParams|linkPrefix|linkTitleToSelf|linkVars|linkWrap|list|listNum|listOnlyInSingleTableView|LIT|lll|lm|locale_all|localNesting|locationData|lockFilePath|lockPosition|lockPosition_addSelf|lockPosition_adjust|lockToIP|login|loginUser|longdescURL|lowColor|lower|LR|mailform|mailto|main|mainScript|makelinks|markerWrap|marks|mask|max|maxAge|maxChars|maxGalleryWidth|maxGalleryWidthInText|maxH|maxHeight|maxItems|maxW|maxWidth|maxWInText|mayNotCreateEditShortcuts|media|mediaQuery|menu|menu_type|menuBackColor|menuHeight|menuName|menuOffset|menuWidth|message_page_is_being_generated|message_preview|meta|metaCharset|method|min|minH|minItems|minute|minW|mod|mode|module|moduleMenuCollapsable|month|move_wizard|MP_defaults|MP_disableTypolinkClosestMPvalue|MP_mapRootPoints|multimedia|MULTIMEDIA|name|navFrameResizable|navFrameWidth|negate|nesting|netprintApplicationLink|neverHideAtCopy|new|NEW|new_wizard|newPageWiz|newRecordFromTable|newWindow|newWizards|next|nicetext|niceText|NO|no_cache|no_search|noAttrib|noCache|noCaption|noCols|noCreateRecordsLink|noLink|noMatchingValue_label|noMenuMode|nonCachedSubst|none|nonTypoTagStdWrap|nonTypoTagUserFunc|nonWrappedTag|noOrderBy|noPageTitle|noResultObj|noRows|noRowsStdWrap|noScaleUp|noStretchAndMarginCells|noThumbsInEB|noThumbsInRTEimageSelect|notification_email_charset|notification_email_encoding|notification_email_urlmode|noTrimWrap|noValueInsert|numRows|obj|offset|offsetWrap|onlineWorkspaceInfo|onlyCurrentPid|opacity|options|order|orderBy|OTABLE|outerWrap|outline|outputLevels|override|overrideAttribs|overrideId|overridePageModule|overrideWithExtension|page|PAGE_TARGET|PAGE_TSCONFIG_ID|PAGE_TSCONFIG_IDLIST|PAGE_TSCONFIG_STR|pageFrameObj|pageGenScript|pages|pages_language_overlay|pageTitleFirst|pageTree|parameter|params|parseFunc|parseFunc_RTE|parser|parseTarget|partialRootPath|partialRootPaths|password|paste|path|permissions|perms|pid|pid_list|pidInList|PIDinRootline|PIDupinRootline|pixelDensity|pixelSpaceFontSizeRef|plaintextLib|plainTextStdWrap|plugin|postCObject|postform|postform_newThread|postLineBlanks|postLineChar|postLineLen|postUserFunc|postUserFuncInt|preBlanks|preCObject|prefix|prefixComment|prefixLocalAnchors|prefixRelPathWith|preIfEmptyListNum|preLineBlanks|preLineChar|preLineLen|prepend|preRenderRegisters|preserveEntities|preUserFunc|prev|preview|previewBorder|prevnextToSection|printheader|prioriCalc|proc|properties|protect|protectLvar|publish|publish_levels|QEisDefault|quality|radio|RADIO|radioWrap|range|rawUrlEncode|recipient|RECORDS|recursive|recursiveDelete|redirect|redirectToURL|reduceColors|references|register|relation|relativeToParentLayer|relativeToTriggerItem|relPathPrefix|remap|remapTag|REMOTE_ADDR|removeBadHTML|removeDefaultJS|removeEmptyEntries|removeIfEquals|removeIfFalse|removeItems|removeObjectsOfDummy|removePrependedNumbers|removeTags|removeWrapping|renderCharset|rendering|renderMethod|renderObj|renderWrap|replace|replacement|REQ|required|reset|resources|resultObj|returnKey|returnLast|returnUrl|rightImgACT|rightImgCUR|rightImgNO|rightjoin|rm|rmTagIfNoAttrib|RO|RO_chBgColor|rootline|rotate|rowCount|rows|rowSpace|rowStdWrap|RTE|RTE_compliant|RTEfullScreenWidth|rules|sample|saveClipboard|saveDocNew|search|SEARCHRESULT|secondRow|section|sectionIndex|select|select_key|selectFields|sendCacheHeaders|separateRows|separator|set|setContentToCurrent|setCurrent|setfixed|setFixedHeight|setFixedWidth|setJS_mouseOver|setJS_openPic|setOnly|setup|shadow|SHARED|sharpen|shear|short|shortcut|shortcut_onEditId_dontSetPageTree|shortcut_onEditId_keepExistingExpanded|shortcutFrame|shortcutGroups|shortcutIcon|show|showAccessRestrictedPages|showActive|showClipControlPanelsDespiteOfCMlayers|showFirst|showHiddenPages|showHiddenRecords|showHistory|showPageIdWithTitle|showTagFreeClasses|simulateDate|simulateUserGroup|singleCaption|singleNoCaption|singlePid|singleStdWrap|site_author|site_reserved|sitemap|sitetitle|siteUrl|size|small|smallRetina|solarize|sorting|source|sourceCollection|space|spaceAfter|spaceBefore|spaceBelowAbove|spaceLeft|spaceRight|spacing|spamProtectEmailAddresses|spamProtectEmailAddresses_atSubst|spamProtectEmailAddresses_lastDotSubst|SPC|special|split|splitCaption|splitChar|splitRendering|src|srcsetCandidate|startInTaskCenter|stayFolded|stdheader|stdWrap|stdWrap2|strftime|stripHtml|styles|stylesheet|submenuObjSuffixes|subMenuOffset|submit|subparts|subst_elementUid|substMarksSeparately|substring|swirl|sword|sword_noMixedCase|SWORD_PARAMS|sword_standAlone|sys_dmail|sys_domain|sys_filemounts|sys_language_mode|sys_language_overlay|sys_language_softMergeIfNotBlank|sys_language_uid|sys_note|sys_template|system|tabel|table|tableCellColor|tableParams|tables|tableStdWrap|tableStyle|tableWidth|tags|target|TCAdefaults|TCEFORM|TCEMAIN|TDparams|temp|template|templateContent|templateFile|templateName|templateRootPath|templateRootPaths|text|textarea|textMargin|textMargin_outOfText|textMaxLength|textmedia|textObjNum|textpic|textPos|textStyle|thickness|this|thumbnailsByDefault|tile|time_stdWrap|tipafriendLib|title|titleInLink|titleInLinkAndImg|titleLen|titleTagFunction|titleText|tm|TMENU_LAYERS|TMENUITEM|token|top|topOffset|totalWidth|transparentBackground|transparentColor|treatIdAsReference|treeLevel|trim|true|tsdebug|tsdebug_tree|TSFE|tt_content|twice|type|typeNum|types|typolink|typolinkCheckRootline|uid|uidInList|uniqueGlobal|uniqueLocal|unset|unsetEmpty|updated|uploadFieldsInTopOfEB|uploads|upper|url|us|useCacheHash|useLargestItemX|useLargestItemY|user|user_task|useragent|USERDEF1|USERDEF1RO|USERDEF2|USERDEF2RO|userdefined|useRegExp|userFunc|userfunction|usergroup|userid|USERNAME_substToken|userProc|USR|USRRO|value|valueArray|version|view|wave|web_func|web_info|web_layout|web_list|web_t|where|width|wiz|wordSpacing|workArea|workOnSubpart|wrap|wrap1|wrap2|wrap3|wrapAfterTags|wrapAlign|wrapFieldName|wrapItemAndSub|wrapNonWrappedLines|wraps|xhtml_cleanig|xhtml_strict|xhtml_trans|xmlprologue|xPosOffset|XY|ypMenu|yPosOffset)\b/,
		/pagets__.*(?:\s|=|.)/,
	],
```

components/prism-typoscript.js
```diff
@@ -0,0 +1,31 @@
...
+    'function': [{
+        pattern: /@import/g,
+        greed: true,
+    }, {
+        pattern: /{[$#].*}|{(?:register|field|cObj):.*}|(?:TSFE|file):.*\n?\s?/g
```
RunDevelopment, line 13: 1. The `.*` in `{[$#].*}` and `{(?:register|field|cObj):.*}` is greedy. Does that mean that it is allowed to match `{` and `}`?
2. Why the `\n?\s?`?

components/prism-typoscript.js
```diff
@@ -0,0 +1,31 @@
...
+        inside: {
+            'function': {
+                pattern: /{[$#].*}|{(?:register|field|cObj):.*}|(?:TSFE|file):.*\n?\s?/g
+            },
+            'keyword': {
+                pattern: /\b(?:_GIFBUILDER|FLUIDTEMPLATE|CARRAY|CASE|CLEARGIF|COA|COA_INT|CONSTANTS|CONTENT|EDITPANEL|EFFECT|FILE|FORM|FRAME|FRAMESET|GIFBUILDER|global|globalString|globalVar|GMENU|GMENU_FOLDOUT|GMENU_LAYERS|GP|HMENU|HRULER|HTML|IENV|IMAGE|IMG_RESOURCE|IMGMENU|IMGMENUITEM|IMGTEXT|JSMENU|JSMENUITEM|LOAD_REGISTER|PAGE|RECORDS|RESTORE_REGISTER|TEMPLATE|TEXT|TMENU|TMENU_LAYERS|TMENUITEM|USER|USER_INT)\b/g
+            }
+        }
+    }, {
+        pattern: /\n?\s+["'].*["']/g,
```
RunDevelopment, line 27: Why does this start with `\n?\s+`

components/prism-typoscript.js
```diff
@@ -0,0 +1,31 @@
+Prism.languages.typoscript = Prism.languages.extend('javascript', {
```
RunDevelopment, line 1: I forget this: Why do you extend JS? TypoScript and JS don't seem to have much in common.

components.json
dkern: Well, you refer to a `json` file named `tsconfig`. But `tsconfig` is an own filetype. The syntax is the same as for `TypoScript`, but `TSConfig` controlls different parts of an installation. Those types, `TypoScript` and `TSConfig` are common. I would still suggest to add `TSConfig` as alias, because of the filetypes and the fact, that the label of Prism is correct when i highlight such code, because saying `TypoScript` there could be missleading where the settings should be located in the installation.

grafik

components/prism-typoscript.js
dkern: I've done a complete rewrite here, i will add this asap.

components/prism-typoscript.js
dkern: About 1.: Sure, it's just more readably (imho), if there are more groups. But jey, sure.
About 2: I've managed to remove the long list completely. I add this update asap.
About 3: This was removed too in the update, so nevermind. ;)

components/prism-typoscript.js
dkern: I've reworked this too, will add it asap.

components/prism-typoscript.js
dkern: I've reworked this too, will add it asap.

components/prism-typoscript.js
dkern: Good question. I think this was my starting point i just dont question myself if this has a reason anymore. I remove the dependeny completly. This will be added with my update too.

components/prism-typoscript.js
dkern: So about 1. again: Is this an issue for you, or is it okay to use blocks like this below. I found it just way more readable.

```JS
'baz': [
    {
        pattern: /foo/,
    },
    {
        pattern: /bar/,
        lookbehind: true,
        greedy: true
    }
],
```

dkern:
>     1. None of the regex literals need a `g` flag. If Prism needs it, it will add it.
Will be fixed! ;)

>     2. We use tabs for indentation.
Will check this too.

>     3. Please add some [tests](https://prismjs.com/test-suite.html).
I take a short look at it, but to be honest, i already spend way to much time on this language creation. 😉

dkern:
Two further questions:

1. Are comments allowed, explaining some of the regex in the non-minified file?

2. Is there a way to use internal reverence, or is it possible to declare a variable somewhere to store something? I need the keywords matching regex on two places. To save data, i could use the same regex for both. Is there any way or should i just insert them two times?

components.json
RunDevelopment: So `tsconfig` is a file extension. Then it's ok, I guess.

components/prism-typoscript.js
RunDevelopment: > So about 1. again:

That's acceptable. I mainly don't want those two:

```js
'foo': {
    pattern: /foo/
},
'bar': [ // all items of this form
    {
        pattern: /bar/
    },
    {
        pattern: /baz/
    }
]
```

instead of simply:

```js
'foo': /foo/,
'bar': [
    /bar/,
    /baz/
]
```

I get that a random regex literal in an array object looks a little strange, so it's ok to use the object-form there.

RunDevelopment:
> Are comments allowed

Of course! 

> Is there a way to use internal reference, or is it possible to declare a variable somewhere to store something? I need the keywords matching regex on two places. To save data, i could use the same regex for both. Is there any way or should i just insert them two times?

[Here](https://github.com/PrismJS/prism/blob/388ad996c4b576205de4d4feda69202bd26c1345/components/prism-java.js#L47) is how the Java language definition does it. Please note that the IIFE is necessary to prevent namespace pollution.

dkern:
Thank you for your reply. I'm now completely finished with my work, but the default tests fails, because of one regex group. But i have no idea how to fix this. I need a bit of help here.

The Problem is my comment matching. General a comment always starts with `//`. But if there is a leading `:` or `\\` it is not a comment, or whenever there is `=` with optional space directly in front, it's not a comment too. I have the regex working fine, but the prism tests fails, because there is a matching group. But i can't make the group non-matching with `?:`, because of the lookbehind.

My regex:
```REGEX
/(^|(?<!\\|:|=\s*))\/\/.*/
```

Here is a working example of it. Everything is fine, but i would not pass the test:
https://regex101.com/r/RYh0Og/1/

Would be nice if someone has a hint how to do this. Thanks!

RunDevelopment:
The problem in `/(^|(?<!\\|:|=\s*))\/\/.*/` is the ES2018 lookbehind `(?<!\\|:|=\s*)`. [ES2018 lookbehinds](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Browser_compatibility) are fairly new, so we can't use them. That's why we have these pseudo-lookbehind groups via a capturing group. 

But the problem with Prism lookbehind groups is that we can't easily negate them. This is not trivial, so here's your pattern with the ES2018 lookbehind: `/(^|[^\\:=\s]|(?:^|[^=\s])\s+)\/\/.*/`

However, your tests still don't pass because with the second comment here not being highlighted:

```
none match text // a comment
// a comment
```

Why is the second comment not highlighted? The problem is that after the first comment, the next character is a `\n`. `\s` matches `\n`, so we need to find a character that matches `[^\s=]` before it. But we cannot find such a character because the regex engine started matching at `\n` and cannot go back (without ES2018 lookbehinds) to the `t` of the previous comment. Thus the second comment cannot be matched.

How do we fix that? 
We can't... We can only change the pattern. If we use the pseudo-lookbehind of `(?<!\\|:|=[ \t]*)` instead, it will work. This means that code like the following to be a false positive but I hope that's acceptable.

```
=
// false positive
```

Final pattern: `/(^|[^\\:= \t]|(?:^|[^= \t])[ \t]+)\/\/.*/`

dkern:
Thank you @RunDevelopment ! I've tested your pattern and it works perfectly! Such things are way above my regex-knowledge. ;)

So, a second before your comment, I've pushed my changes. I've completely redone my language definition and managed to removed the whole _tag_ list, so it is now more a _syntax based_ highlighting. I'm pretty pleased with the outcome. The highlighting is much better then in my first submission. Overall, i have noting left to do. Everything is covered so far.

I think i have managed to get all your things in the review covered. I've added some tests too and hope that everything is ok now.

components/prism-typoscript.js
```diff
@@ -0,0 +1,79 @@
+(function(Prism) {
+
+	var keywords = /\b(?:ACT|ACTIFSUB|CARRAY|CASE|CLEARGIF|COA|COA_INT|CONSTANTS|CONTENT|CUR|EDITPANEL|EFFECT|EXT|FILE|FLUIDTEMPLATE|FORM|FRAME|FRAMESET|GIFBUILDER|GMENU|GMENU_FOLDOUT|GMENU_LAYERS|GP|HMENU|HRULER|HTML|IENV|IFSUB|IMAGE|IMGMENU|IMGMENUITEM|IMGTEXT|IMG_RESOURCE|INCLUDE_TYPOSCRIPT|JSMENU|JSMENUITEM|LLL|LOAD_REGISTER|NO|PAGE|RECORDS|RESTORE_REGISTER|TEMPLATE|TEXT|TMENU|TMENUITEM|TMENU_LAYERS|USER|USER_INT|_GIFBUILDER|global|globalString|globalVar)\b/;
+
+	Prism.languages.typoscript = {
+		'comment': [
+			{
+				// multiline comments /* */
+				pattern: /[^\/]\/\*[^.][\s\S]*?\*\//,
```
RunDevelopment, line 9: 1. You don't have to worry about single-line comments breaking your multiline comments. 
1. Are empty comments not possible? E.g. `/**/`

components/prism-typoscript.js
```diff
@@ -0,0 +1,79 @@
...
+		'keyword': keywords,
+		'number': {
+			// special highlighting for indexes of arrays in tags
+			pattern: /[0-9]+\s*[.{=]/,
+			inside: {
+				'operator': /[.{=]/,
+			}
+		},
+		'tag': {
+			pattern: /\s*\.?[A-z0-9-_]+\.?\s*/,
```
RunDevelopment, line 68: 1. `A-z`: That's probably not right, is it? This will match all characters from `A` (`\x41`) to `z` (`\x7a`) including characters like these: `@_[]`.
2. Please remove the leading and trailing `\s*`.

```suggestion
			pattern: /\.?[\w-]+\.?/,
```

components/prism-typoscript.js
```diff
@@ -0,0 +1,79 @@
...
+				greedy: true
+			}
+		],
+		'function': [
+			{
+				// old include style
+				pattern: /<INCLUDE_TYPOSCRIPT:\s*source\s*=\s*['"].*['"]\s*>/,
+				inside: {
+					'string': {
+						pattern: /['"].*['"]/,
```
RunDevelopment, line 31: You probably mean `"[^"\r\n]*"|'[^'\r\n]'`, right? (Same for all the occurrences.)

docs/Prism.html
```diff
@@ -1210,13 +1210,7 @@ <h4 class="name" id=".tokenize"><span class="type-signature">(static) </span>tok
...
-        console.log(`Found numeric literal: ${token.content}`);
-    }
-});</code></pre>
+    <pre class="prettyprint"><code>let code = `var foo = 0;`;
let tokens = Prism.tokenize(code, Prism.languages.javascript);
tokens.forEach(token => {
    if (token instanceof Prism.Token &amp;&amp; token.type === 'number') {
        console.log(`Found numeric literal: ${token.content}`);
    }
});</code></pre>
```
RunDevelopment, line 1213: How? This `^M` means `\r`. How did the line ends of these code blocks specifically get converted to `\r`?

Please revert these changes.

components/prism-typoscript.js
dkern: I've simply copied this part from `clike`. I tought i don't have to do this again, as the comments are the same. ;)

docs/Prism.html
dkern: That seems to happen when i run the build scripts. I never touched that file by myself. I try to resolve that file.

components/prism-typoscript.js
dkern: I think there are no multiline versions of this. So yes, i can use yours too. Will update the pattern.

components/prism-typoscript.js
dkern: I only found `\` as other character used in string, so i went with `/\.?[\w-\\]+\.?/`.

components/prism-typoscript.js
dkern: I can do it like this, yes. But i added a star. ;)

dkern:
I think I've resolved everything so far. Hopefully we're fine now. 😉

components/prism-typoscript.js
RunDevelopment: > But i added a star. ;)

Yeah... My bad.

components/prism-typoscript.js
```diff
@@ -0,0 +1,79 @@
...
+		'keyword': keywords,
+		'number': {
+			// special highlighting for indexes of arrays in tags
+			pattern: /[0-9]+\s*[.{=]/,
+			inside: {
+				'operator': /[.{=]/,
+			}
+		},
+		'tag': {
+            pattern: /\.?[\w-\\]+\.?/,
```
RunDevelopment, line 68: 1. Indentation.
2. Maybe we should highlight `\` as punctuation too in that case?

components/prism-typoscript.js
RunDevelopment: Copied? The [`comment` pattern of `clike`](https://github.com/PrismJS/prism/blob/388ad996c4b576205de4d4feda69202bd26c1345/components/prism-clike.js#L4) looks very different.

components/prism-typoscript.js
dkern: 1. fixed
2. No, the only thing that has `\` in it are Class names. But the whole classname is a tag, it's not separated by this. TypoScript will basically be split into an php array by `.` when parsed. So something like this:

```TYPOSCRIPT
Foo.Dkern\Classes\Test {
    pid = 1
  }
}
```

Will become this:

```PHP
$arr['Foo']['Dkern\Classes\Test']['pid'] = 1;
```

So i would not make `\` an punctuation, as this could be missleading.

RunDevelopment: Ahh, I assumed that `\` was punctuation since PHP namespaces are like this.

RunDevelopment:
@dkern The last remaining issue is the `comment` pattern. The easiest solution might be to copy it from `clike`.

components/prism-typoscript.js
dkern: Hm. Right. I have no idea where I've copied it from. Maybe an older version I had on my machine?! Don't know. But I copied it. ;) If you want a change please just say what I should add. I don't care.

dkern:
> @dkern The last remaining issue is the `comment` pattern. The easiest solution might be to copy it from `clike`.
Done.

RunDevelopment:
Thank you for contributing @dkern!

kronos361 (approved):
Excellent job

PR #2509, opened 2020-08-06
Shell session & bash: All command after start of Heredoc string

### LINKED ISSUE #2503: Fix hello-document support in shell-session language
**Information:**
- Prism version: 1.20.0
- Plugins: none
- Environment: Browser

Does the latest version of Prism from the [download page](https://prismjs.com/download.html) also have this issue?

yes

**Description**

First, thanks so much for Prism.

Putting together the regular expressions to pull this off must be painstaking!

Consider the following code block using the `shell-session` language.

```
$ cp /etc/ipsec.secrets /etc/ipsec.secrets.backup

$ cat << "EOF" > /etc/ipsec.secrets
: RSA vpn-server-a.key
EOF

$ cat << "EOF" > /etc/ipsec.secrets
: RSA vpn-server-a.key
# : RSA vpn-server-b.key
EOF
```

Issues:

- Everything after `cat` is part of the output token. I believe the above code block should render three command tokens with no output.

- Lines that start with `#` (or `$`) inside hello-documents become new commands. I believe commands inside of hello-documents should be ignored as they are part of a parent command.

See following screenshot.

here-document

RunDevelopment:
This resolves #2503.

I made it so that the "rest" command is highlighted as `punctuation` to be visually distinct from the rest of the string.

image

image


kadirselcuk:
Rest komutları görsel paylaşmanız güzel olmuş

PR #2514, opened 2020-08-10
Set up CI with Azure Pipelines

kadirselcuk:
[skip ci]


mAAdhaTTah:
We are not using and do not intend to use Azure Pipelines. Was this sent to the right repo?

PR #2515, opened 2020-08-10
Tests: Added an option to accept the actual token stream

### LINKED ISSUE #2401: Automatically generate expected JSON for .test files
**Motivation**
Writing test files, while important, has always been a pain. I made #1801 because I got tired of having to format the token stream __by hand__.
I'm even lazier now and want the test runner to insert the token stream for me.

Jokes aside, the basic idea is that you only need to add some code and a description and the test runner will insert the JSON itself. This makes it a lot easier to create test files because you don't even have to copy-paste the token stream anymore, you just accept the highlighting as correct. 
Updating the test files after small changes (e.g. adding a few new keywords) will also be easier.

I think that this may lead to test files with errors since the developer no longer has to look at the token stream (though, it is questionable that devs look at the token stream when copy-pasting it).

**Description**
Add a CLI option `--insert` to `npm run test:languages` that will automatically insert the actual token stream in `.test` files where the actual and the accept token stream aren't equal.

RunDevelopment:
Running `npm run test:languages -- --accept` will now insert the actual token stream of a test case to be the expect JSON. This will only be done if no expected JSON is specified.

### Example

Test case:

```
var foo = 0;
```

Run `npm run test:languages -- --accept`.

Test case:

```
var foo = 0;

[
	["keyword", "var"],
	" foo ",
	["operator", "="],
	["number", "5"],
	["punctuation", ";"]
]
```

### Motivation

This is super useful for creating many test cases at once. 
When I create and change languages, I always have test drive open constantly check that every works. I then just copy the code from test drive I know work into test cases and copy their JSON from console output. This command makes my usual workflow a lot quicker.

It also makes it easier for first-time contributors to create test cases.

---

This resolves #2401.

### DIFF
--- a/test-suite.html
+++ b/test-suite.html
@@ -94,7 +94,10 @@
 	<p>Your file is built up of two or three sections, separated by ten or more dashes <code>-</code>, starting at the begin of the line:</p>
 	<ol>
 		<li>Your language snippet. The code you want to tokenize using Prism. (<strong>required</strong>)</li>
-		<li>The simplified token stream you expect. Needs to be valid JSON. (<strong>required</strong> or <em>optional</em>)</li>
+		<li>
+			The simplified token stream you expect. Needs to be valid JSON. (<em>optional</em>) <br>
+			If there no token stream defined, the test case will fail unless the <code>--accept</code> flag is present when running the test command (e.g. <code>npm run test:languages -- --accept</code>). If the flag is present and there is no expected token stream, the runner will insert the actual token stream into the test case file, changing it.
+		</li>
 		<li>A comment explaining the test case. (<em>optional</em>)</li>
 	</ol>
 	<p>The easiest way would be to look at an existing test file:</p>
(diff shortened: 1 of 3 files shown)


test-suite.html
```diff
@@ -93,8 +93,8 @@ <h2 id="writing-tests-writing-your-test">Writing your test</h2>
 
 	<p>Your file is built up of two or three sections, separated by ten or more dashes <code>-</code>, starting at the begin of the line:</p>
 	<ol>
-		<li>Your language snippet. The code you want to compile using Prism. (<strong>required</strong>)</li>
-		<li>The simplified token stream you expect. Needs to be valid JSON. (<strong>required</strong>)</li>
+		<li>Your language snippet. The code you want to tokenize using Prism. (<strong>required</strong>)</li>
+		<li>The simplified token stream you expect. Needs to be valid JSON. (<strong>required</strong> or <em>optional</em>)</li>
```
mAAdhaTTah, line 97: This should probs just say "optional" (since it can't really be both), and explain the behavior when it's not present.

PR #2518, opened 2020-08-13
MongoDB syntax

airs0urce:
I applied changes as we discussed here: https://github.com/PrismJS/prism/pull/2502

And also I have one question. I see with default theme tokens "builtin" and "string" have same color which visually looks like parsing has been broken:

Screen Shot 2020-08-13 at 5 38 19 PM

In some themes "builtin" token has different color, not like "string" token, which looks good:
Screen Shot 2020-08-13 at 5 40 59 PM

Question is - is there any recommendations by making "builtin" looks different from string with default theme?
Sure, I could create another theme and use it, but if the pull request will be approved, for people who decides to use this syntax it will be working bad by default like in the first screenshot.


airs0urce:
Another way - I could name that token not "builtin", but something that has different color, but then meaning will be lost, which is bad too

RunDevelopment:
You could use an alias. If you define `class-name` as an alias, the builtin classes will be highlighted in the same color as function in the default theme and the color in Tomorrow Night will remain the same.

components/prism-mongodb.js
```diff
@@ -0,0 +1,100 @@
...
+		'RegExp',
+		'ISODate',
+		'UUID',
+	];
+
+	operators = operators.map(function(operator) { 
+		return operator.replace('$', '\\$');
+	});
+
+	var operatorsRegex = '(?:' + operators.join('(?:\\b|:)|') + ')\\b';
```
RunDevelopment, line 65: 1. You don't need the `(?:\\b|:)`.
1. Please rename the variable. It's not a regex. Maybe `operatorsSource`?

```suggestion
	var operatorsRegex = '(?:' + operators.join('|') + ')\\b';
```

components/prism-mongodb.js
```diff
@@ -0,0 +1,100 @@
...
+			inside: {
+				'keyword': RegExp('^([\'"])?' + operatorsRegex + '(?:\\1)?$')
+			}
+		}
+	});
+
+	Prism.languages.mongodb.string.inside = {
+		url: {
+			// url pattern
+			pattern: /https?:\/\/(?:www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b[-a-zA-Z0-9()@:%_\+.~#?&//=]*/,
```
RunDevelopment, line 82: Same as before. I think URLs and IP addresses should be highlighted only if the whole string is one to prevent false positives.

Also, the pattern can be simplified.

```suggestion
			pattern: /https?:\/\/[-\w@:%.+~#=]{1,256}\.[a-z0-9()]{1,6}\b[-\w()@:%+.~#?&/=]*/i,  
```

components/prism-mongodb.js
```diff
@@ -0,0 +1,103 @@
...
+
+	Prism.languages.insertBefore('mongodb', 'constant', {
+		'builtin': {
+			pattern: RegExp('\\b(?:' + builtinFunctions.join('|') + ')\\b'),
+			alias: 'keyword'
+		}
+	});
+
+}(Prism));
+
```
RunDevelopment, line 100: Please remove the unnecessary empty lines.

airs0urce:
@RunDevelopment 
Thank you, I set "keyword" as alias:
alias: "keyword"

Didn't use "function" as generic JS function and mongodb function should be highlighted different.
Looks like everything is ready now.

components/prism-mongodb.js
```diff
@@ -0,0 +1,103 @@
...
+
```
RunDevelopment, line 2: Please remove the empty lines.

airs0urce:
@RunDevelopment yes, agree about everything, will make the changes. 
But not sure about "url" and "ip". The goal of adding them is to be able to easy catch those entities inside long string, for example I have logs written in MongoDB, I have long stack traces there and it will help to parse logs by eyes more quick:

Screen Shot 2020-08-13 at 8 22 58 PM

If we highlight only full string containing URL or IP it will break the original idea of adding highlight because it's easy to catch URL or IP if it's fully inside string. So, if you think that false positives possible then I can just remove url and ip from the definition. Or I can change as you suggested, but right now not sure if it will be useful if it's done this way.

RunDevelopment:
> I have logs written in MongoDB, I have long stack traces there and it will help to parse logs by eyes more quick

If that was the explicit intention, then it's ok to keep it as is.

components/prism-mongodb.js
RunDevelopment: Resolved [here](https://github.com/PrismJS/prism/pull/2518#issuecomment-673480969).

airs0urce:
@RunDevelopment Yes, it was explicit. many developers use MongoDB for logging purpose, so it should help to parse strings by eyes.

components/prism-mongodb.js
airs0urce: Already added this change as part of commits

components/prism-mongodb.js
airs0urce: Already added this change as part of commits

components/prism-mongodb.js
airs0urce: Already added this change as part of commits

RunDevelopment:
Thank you for contributing @airs0urce!

Issue #2519, opened 2020-08-13
When creating a new language, keywords do not work if they are not in English


Diversus23:
**Information:**
- Prism version: 1.21.0
- Plugins: none
- Environment: Chrome, npm 6.14.6,

**Description**
I made a new language  BSL (1C:Enterprise) https://github.com/Diversus23/prism 
In Russia, the popular development environment 1C:Enterprise 8. It is possible to use two language in the development of Russian and English. For example:

```
While a = 1 Do
...
EndDo;

```
Analog in Russian:

```
Пока а = 1 Цикл
...
КонецЦикла;
```

I use flag /u for RegExp (for Unicode characters), but but it doesn't work for Russian language (for English ok)... How do I fix this?

Pattern:

```
	'keyword': [
		{
			// RU
			pattern: /\b(?:пока|для|новый|прервать|...|конеццикла)\b/uig,
			lookbehind: true		
		},
		{
			// Eng
			pattern: /\b(?:while|for|new|break|try|...|enddo)\b/ig,
			lookbehind: true
		}
	],
```

Screenshots:

![2020-08-13_15-48-27](https://user-images.githubusercontent.com/19926084/90136360-bcb6c800-dd7c-11ea-8571-fe4c7ca1ddd7.png)
![2020-08-13_15-49-08](https://user-images.githubusercontent.com/19926084/90136362-bde7f500-dd7c-11ea-9e96-a16c96693948.png)
![2020-08-13_15-49-43](https://user-images.githubusercontent.com/19926084/90136363-be808b80-dd7c-11ea-8068-dd328f39ef0c.png)

**Example**
An example which reproduces the bug. This can be a link to a website, a [JSFiddle](https://jsfiddle.net/), an attached example project, etc.

RunDevelopment:
The problem is `\b`.

`\b` means word boundary where a word is defined as a `\w` character. But `\w` only includes the letters A-Z (uppercase and lowercase), digits 0-9 (ASCII only), and the underscore (`_`). That's it; It doesn't include any non-Latin characters. This doesn't change when adding the `u` flag.

So from the perspective of `\b`, all Russian characters are non-word characters, meaning that there is no word boundary between `(` and `п` because neither are word characters.

__Workaround 1__

[...]

PR #2520, opened 2020-08-13
Add language BSL (1C:Enterprise)

Diversus23:
The language BSL (1C:Enterprise) added.

There is one problem, since the constructions use /u (Cyrillic language constructions are used) the "_npm test_" command is down.


Diversus23:
My fork: https://github.com/Diversus23/prism

image

RunDevelopment:
First, please let me apologize. I assumed that you made a language definition for private use in which case the `u`
 flag and ES2018 lookbehinds might not have been an issue for you. But for Prism, where we have to support older browsers as well, we can't use either. I'm sorry for giving you the wrong information.

To make the tests pass, we have to get rid of the `u` flag and the ES2018 lookbehinds.

Getting rid of the ES2018 lookbehinds is easy: you just have to use workaround 2.

Getting rid of the `u` flag is a lot harder... We basically have to create a character class that behaves like `[\p{L}\p{N}_]` without the use of Unicode properties. In #2519, I said to use Babel but that's not an option here because Babel will transpile `\p{L}` into a pattern that is 4kB long. We need `\p{L}` multiple times, so let's not do this.

Instead, let's redefine the problem. Since we only really need to support [Cyrillic script](https://en.wikipedia.org/wiki/Cyrillic_script_in_Unicode), we can cut down on the numbers of characters. Instead of `[\p{L}\p{N}_]`, let's use `[\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f]`. It will match all Latin and Cyrillic word characters, is reasonably short, and doesn't need the `u` flag.

Diversus23:
1. I have a question about tests. I run the test "npm run test: languages -- --language=bsl --pretty" and the test tells me that everything is OK. But then in the global test "npm test" my tests do not pass. Why don't I also run global test blocks in a local test? It is a long time to wait for the entire global test to be checked. Is it possible to run npm test only for my language, but completely?
2. Consider making sure that regular expressions support national languages natively in prismjs. There are languages that don't work based on English syntax https://en.wikipedia.org/wiki/Non-English-based_programming_languages and there are many of them. Or still deal with the flag /u. Prismjs will not be able to support them normally without dancing with a tambourine :)
3. Replaced [\p{L}\p{N}_] = > [\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f] but the tests don't pass
image
4. There is a problem with numbers. Pattern 'number': /(?:\b\d+\.?\d*|\B\.\d+)(?:E[+-]?\d+)?/i does not work normally for the same reason because there is \b
image

components/prism-bsl.js
```diff
@@ -0,0 +1,57 @@
...
+		// date & time
+		{
+			pattern: /\'(?:\\?.)*?\'/,
+			lookbehind: true			
+		}
+	],
+	'keyword': [
+		{
+			// RU
+			pattern: /(?:(?<=[\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f])(?![\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f])|(?<![\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f])(?=[\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f]))(?:пока|для|новый|прервать|попытка|исключение|вызватьисключение|иначе|конецпопытки|неопределено|функция|перем|возврат|конецфункции|если|иначеесли|процедура|конецпроцедуры|тогда|знач|экспорт|конецесли|из|каждого|истина|ложь|по|цикл|конеццикла|истина|ложь)(?:(?<=[\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f])(?![\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f])|(?<![\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f])(?=[\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f]))/i
```
RunDevelopment, line 22: This will get rid of the ES2018 lookbehinds.

Also, you might want to replace the `\b`s in the English keyword list and the Russian and English operators. Otherwise, mixed-Latin-Cyrillic identifiers could be an issue (e.g. in `whileготов`, `while` will be highlighted as a keyword).

```suggestion
			pattern: /(^|[^\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f])(?:пока|для|новый|прервать|попытка|исключение|вызватьисключение|иначе|конецпопытки|неопределено|функция|перем|возврат|конецфункции|если|иначеесли|процедура|конецпроцедуры|тогда|знач|экспорт|конецесли|из|каждого|истина|ложь|по|цикл|конеццикла|истина|ложь)(?![\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f])/i,
			lookbehind: true
```

components/prism-bsl.js
```diff
@@ -0,0 +1,57 @@
...
+	'string': [
+		// strings
+		{
+			pattern: /"(?:[^"]|"")*"(?!")/,
+			greedy: true
+		},
+		// date & time
+		{
+			pattern: /\'(?:\\?.)*?\'/,
+			lookbehind: true
```
RunDevelopment, line 16: Unnecessary lookbehind.

components/prism-bsl.js
```diff
@@ -0,0 +1,57 @@
...
+	'comment': /\/\/.*/,
+	'string': [
+		// strings
+		{
+			pattern: /"(?:[^"]|"")*"(?!")/,
+			greedy: true
+		},
+		// date & time
+		{
+			pattern: /\'(?:\\?.)*?\'/,
```
RunDevelopment, line 15: This pattern will backtrack exponentially for the string `'\\\\\\\\\\`.

```suggestion
			pattern: /'(?:[^'\r\n\\]|\\.)*'/,
```

components/prism-bsl.js
```diff
@@ -0,0 +1,57 @@
...
+		{
+			pattern: /\b(?:and|or|not)\b/i
+		}
+		
+	],
+	'punctuation': /\(\.|\.\)|[()\[\]:;,.]/,	
+	'directive': [
+		// Теги препроцессора вида &Клиент and &Сервер
+		{
+			pattern: /^\s*&.*/gm,
```
RunDevelopment, line 46: 1. You don't need the `g` flag.
1. Please exclude the leading spaces like this:

(Same for the other pattern as well.)

```suggestion
			pattern: /^(\s*)&.*/m,
			lookbehind: true,
```

components/prism-bsl.js
```diff
@@ -0,0 +1,57 @@
...
+		},		
+		// EN
+		{
+			pattern: /\b(?:and|or|not)\b/i
+		}
+		
+	],
+	'punctuation': /\(\.|\.\)|[()\[\]:;,.]/,	
+	'directive': [
+		// Теги препроцессора вида &Клиент and &Сервер
```
RunDevelopment, line 44: Please write you comments in English or add a translation.

---

I once worked on code from a guy who wrote all his comments in Russian. I don't speak Russian so I had to google-translate all of his comments. The problem with that is that Google Translate has no idea about context and the (I assume) technical language used didn't make this any easier. In the end, I couldn't even understand a third of his comments...

Admittedly, your comments [translate really well](https://translate.google.de/?hl=en&tab=rT#view=home&op=translate&sl=ru&tl=en&text=%D0%A2%D0%B5%D0%B3%D0%B8%20%D0%BF%D1%80%D0%B5%D0%BF%D1%80%D0%BE%D1%86%D0%B5%D1%81%D1%81%D0%BE%D1%80%D0%B0%20%D0%B2%D0%B8%D0%B4%D0%B0%20%23%D0%9E%D0%B1%D0%BB%D0%B0%D1%81%D1%82%D1%8C%20%D0%B8%20%23%D0%9A%D0%BE%D0%BD%D0%B5%D1%86%D0%9E%D0%B1%D0%BB%D0%B0%D1%81%D1%82%D0%B8) but please at least add a translation.

components/prism-bsl.js
```diff
@@ -0,0 +1,57 @@
...
+		},
+		{
+			// EN
+			pattern: /\b(?:while|for|new|break|try|except|raise|else|endtry|undefined|function|var|return|endfunction|null|if|elseif|procedure|endprocedure|then|val|export|endif|in|each|true|false|to|do|enddo|true|false)\b/i
+		}
+	],
+	'number': /(?:\b\d+\.?\d*|\B\.\d+)(?:E[+-]?\d+)?/i,
```
RunDevelopment, line 29: It's tricky to translate the `\b` and `\B` here. The problem is that we only have __one__ and only one Prism lookbehind group.

To do this, I will rewrite the pattern a little and use lookaheads to "pick" the right alternative like this: `(?:\b(?=\d)|\B(?=\.)(?:\d+\.?\d*|\.\d+)(?:E[+-]?\d+)?`. This can be converted into a single Prism lookbehind group and after a little optimization, we get this:

```suggestion
	'number': {
		pattern: /(^(?=\d)|[^\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f])(?:\d+\.?\d*|\.\d+)(?:E[+-]?\d+)?/i,
		lookbehind: true
	},
```

components/prism-bsl.js
```diff
@@ -0,0 +1,57 @@
...
+	'operator': [	
+		/<[=]?|>[=]?|[+\-*\/]=?|[\%=]/i,
```
RunDevelopment, line 31: This can be simplified.

```suggestion
		/[<>+\-*/]=?|[%=]/,
```

components/prism-bsl.js
Diversus23: Ok. Sorry.

Diversus23:
Thank you for your help. I will be happy if my changes for the new BSL language are added to the main repository. In Russia, this language is very popular and has a huge community.

components/prism-bsl.js
RunDevelopment: No need to apologize.

Btw. I really like that you made all comments in Russian and English.

docs/Prism.html
```diff
@@ -1210,13 +1210,7 @@ <h4 class="name" id=".tokenize"><span class="type-signature">(static) </span>tok
...
-        console.log(`Found numeric literal: ${token.content}`);
-    }
-});</code></pre>
+    <pre class="prettyprint"><code>let code = `var foo = 0;`;
let tokens = Prism.tokenize(code, Prism.languages.javascript);
tokens.forEach(token => {
    if (token instanceof Prism.Token &amp;&amp; token.type === 'number') {
        console.log(`Found numeric literal: ${token.content}`);
    }
});</code></pre>
```
RunDevelopment, line 1213: I have [seen this before](https://github.com/PrismJS/prism/pull/2505#discussion_r468180842). I hoped this was an one-of-a-kind occurrence but it seems to be systematic.

May I ask what OS you're using?

components/prism-bsl.js
```diff
@@ -0,0 +1,71 @@
...
+	],
+	'keyword': [
+		{
+			// RU
+			pattern: /(^|[^\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f])(?:пока|для|новый|прервать|попытка|исключение|вызватьисключение|иначе|конецпопытки|неопределено|функция|перем|возврат|конецфункции|если|иначеесли|процедура|конецпроцедуры|тогда|знач|экспорт|конецесли|из|каждого|истина|ложь|по|цикл|конеццикла|выполнить)(?![\w\u0400-\u0484\u0487-\u052f\u1c80-\u1c88\u1d2b\u1d78\u2de0-\u2dff\ua640-\ua69f\ufe2e\ufe2f])/i,
+			lookbehind: true
+		},
+		{
+			// EN
+			pattern: /\b(?:while|for|new|break|try|except|raise|else|endtry|undefined|function|var|return|endfunction|null|if|elseif|procedure|endprocedure|then|val|export|endif|in|each|true|false|to|do|enddo|execute)\b/i
```
RunDevelopment, line 27: As I pointed out [here](https://github.com/PrismJS/prism/pull/2520#discussion_r471029326), you might want to replace `\b` or is there a reason against this?

RunDevelopment:
Thanks for the changes @Diversus23!

Apart from my two comments, I think this is ready to be merged. 

Please resolve the merge conflicts. This should be as simple as merging our master branch run running `npm run build`.

docs/Prism.html
Diversus23: > May I ask what OS you're using?
Windows 10

components/prism-bsl.js
Diversus23: I wanted to explicitly separate EN and RU constructs. I think this will be correct and easier to follow in the future.
And since \b is used everywhere, for EN, I do the same.

docs/Prism.html
RunDevelopment: > Windows 10

Same. So it's not that.

I don't want this to hold back your pull any longer but one last question: What line ends are used in this section of the file specifically? It should be all line feeds and look like this:

image

Diversus23: CR LF
image

Diversus23: For Docs/prism.html
image

RunDevelopment: Thank you for the information!

So it actually generates the `CR`s... I'll look into that some more but only after we merge this. Please revert the changes to the `docs/**` files and rebuild to make the CI pass.

RunDevelopment:
There are still 4 files that changed:

```
components.js                                    | 2 +-
docs/Prism.languages.html                        | 8 +++++++-
plugins/autoloader/prism-autoloader.min.js       | 2 +-
plugins/show-language/prism-show-language.min.js | 2 +-
```

You also have to revert `docs/Prism.languages.html`. 
The rest are generated minified files. In theory, you only have to run `npm run build`. Does that not work?

RunDevelopment:
Seems like you accidentally committed the `docs/**` files again. Everything else is fine.

RunDevelopment:
You can make JSDoc generate the correct line ends by converting the line ends of `components/prism-core.js` to LF and running `npm run build` again. This may be a little easier than using git to revert some files. 

Btw. I filed a [bug report](https://github.com/jsdoc/jsdoc/issues/1837) about the `CR`s in the `docs/**` files. I'll soon make a temporary hack to fix the line ends until the bug is fixed upstream.

Diversus23:
I only ran `npm run build`. Changes to docs/** were created automatically. And I probably should not have put the modified docs/** files. To be honest, I don't often work with repositories with the team and I don't know what to do next to solve the problem.

RunDevelopment:
> I only ran npm run build. Changes to docs/** were created automatically.

Sorry, that's a problem on our part. I will fix this soon.

Until then. to make the CI pass, do the following:

1. Change the line ends of `components/prism-core.js` to LF. Most editors will have an option to do that for you.
1. Run `npm run build`. The `docs/**` files will change (this is what we want!).
1. Commit only the changed `docs/**` files. Do not commit the changes to `components/prism-core.js` or any other changed files outside of the `docs` folder.

Again, I'm really sorry that this is so difficult. This isn't something anybody should have to deal with. This will be fixed soon.

RunDevelopment:
Thank you for contributing @Diversus23!

Diversus23:
Thank you @RunDevelopment!
I am always inspired by people who do their job well and are real professionals. Good luck to you!

PR #2523, opened 2020-08-17
JSDoc: Fixed line ends

RunDevelopment:
In [two](https://github.com/PrismJS/prism/pull/2505#discussion_r468180842) [recent](https://github.com/PrismJS/prism/pull/2520#discussion_r471158872) pull requests, we had [this issue](https://github.com/jsdoc/jsdoc/issues/1837).

This is a temporary fix to adjust the wrong line ends until the issue is fixed upstream.

I would like to merge this soon because it can really hinder contributors because is prevents the CI from passing (which is a requirement to merge any pull request).

@mAAdhaTTah

### DIFF
--- a/gulpfile.js/docs.js
+++ b/gulpfile.js/docs.js
@@ -45,7 +45,16 @@ function docsRemoveExcessFiles() {
 	return del(jsDoc.junk);
 }
 
-const docs = series(docsClean, docsCreate, docsRemoveExcessFiles, docsAddFavicon);
+function docsFixLineEnds(cb) {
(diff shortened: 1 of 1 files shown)


RunDevelopment:
It __does__ depend on the line ends of the input file. If the input file uses LF, everything will be fine but under Windows, you can configure git to checkout code and convert the line ends to CRLF (which is the default I think?).

Also, it doesn't generate CRLF from LF. It generates CR from CRLF which is the problem.

RunDevelopment:
@mAAdhaTTah Is this good to merge?

mAAdhaTTah:
Yeah go for it.

PR #2526, opened 2020-08-26
Added PureScript language definition

sriharshachilakapati:
Added PureScript language definition with test cases and examples.


RunDevelopment:
Thank you for making this language definition @sriharshachilakapati!

I will review it once the issue with the many generated files is resolved. This _should_ be quite simple. My guess is that you ran `npm i` instead of `npm ci`. The output of uglify is very sensitive to the specific uglify version used. 
Deleting the local `node_modules` folder (and reverting changes to `package.json` and `package-lock.json`, if any), and then running `npm ci`, followed by `npm run build` should fix the issue.

sriharshachilakapati:
I was using pnpm instead of npm and it didn't respect that lock file. My bad, will correct that and push soon.

components/prism-purescript.js
```diff
@@ -0,0 +1,38 @@
+Prism.languages.purescript = {
+	'comment': {
+		pattern: /(^|[^-!#$%*+=?&@|~.:<>^\\\/])(?:--[^-!#$%*+=?&@|~.:<>^\\\/].*|{-[\s\S]*?-})/m,
+		lookbehind: true
+	},
+	'char': /'(?:[^\\']|\\(?:[abfnrtv\\"'&]|\^[A-Z@[\]^_]|NUL|SOH|STX|ETX|EOT|ENQ|ACK|BEL|BS|HT|LF|VT|FF|CR|SO|SI|DLE|DC1|DC2|DC3|DC4|NAK|SYN|ETB|CAN|EM|SUB|ESC|FS|GS|RS|US|SP|DEL|\d+|o[0-7]+|x[0-9a-fA-F]+))'/,
+	'string': {
+		pattern: /"(?:[^\\"]|\\(?:[abfnrtv\\"'&]|\^[A-Z@[\]^_]|NUL|SOH|STX|ETX|EOT|ENQ|ACK|BEL|BS|HT|LF|VT|FF|CR|SO|SI|DLE|DC1|DC2|DC3|DC4|NAK|SYN|ETB|CAN|EM|SUB|ESC|FS|GS|RS|US|SP|DEL|\d+|o[0-7]+|x[0-9a-fA-F]+)|\\\s+\\)*"/,
```
RunDevelopment, line 8: This pattern has an exponential runtime for certain input strings for two reasons:

1. The inner loop can match `\SOH` in two ways: 1) it can be matched whole (via the `SOH` alternative) and 2) it can be matched as `\SO` (via the `SO` alternative) followed by an `H` character (using `[^\\"]`). This means that a string of the form `/"(?:\\SOH)+/` will cause exponential backtracking. ([Try it out](https://regexr.com/5avbu).)
2. The `\d+|o[0-7]+|x[0-9a-fA-F]+` will all cause exponential backtracking for a similar reason.

I think the easiest fix here, is to not validate escape sequences.

Note: The `char` pattern is safe because it doesn't repeat characters.

```suggestion
		pattern: /"(?:[^\\"]|\\(?\S|\s+\\)*"/,
```

components/prism-purescript.js
```diff
@@ -0,0 +1,38 @@
+Prism.languages.purescript = {
+	'comment': {
+		pattern: /(^|[^-!#$%*+=?&@|~.:<>^\\\/])(?:--[^-!#$%*+=?&@|~.:<>^\\\/].*|{-[\s\S]*?-})/m,
+		lookbehind: true
+	},
+	'char': /'(?:[^\\']|\\(?:[abfnrtv\\"'&]|\^[A-Z@[\]^_]|NUL|SOH|STX|ETX|EOT|ENQ|ACK|BEL|BS|HT|LF|VT|FF|CR|SO|SI|DLE|DC1|DC2|DC3|DC4|NAK|SYN|ETB|CAN|EM|SUB|ESC|FS|GS|RS|US|SP|DEL|\d+|o[0-7]+|x[0-9a-fA-F]+))'/,
```
RunDevelopment, line 6: Please give this the `string` alias. Otherwise, it won't get a pretty color in most themes.

components/prism-purescript.js
```diff
@@ -0,0 +1,38 @@
...
+	'string': {
+		pattern: /"(?:[^\\"]|\\(?:[abfnrtv\\"'&]|\^[A-Z@[\]^_]|NUL|SOH|STX|ETX|EOT|ENQ|ACK|BEL|BS|HT|LF|VT|FF|CR|SO|SI|DLE|DC1|DC2|DC3|DC4|NAK|SYN|ETB|CAN|EM|SUB|ESC|FS|GS|RS|US|SP|DEL|\d+|o[0-7]+|x[0-9a-fA-F]+)|\\\s+\\)*"/,
+		greedy: true
+	},
+	'keyword': /\b(?:ado|case|class|data|derive|do|else|if|in|infixl|infixr|instance|let|module|newtype|of|primitive|then|type|where)\b/,
+	'import_statement': {
```
RunDevelopment, line 12: Snake-case, please.

```suggestion
	'import-statement': {
```

components/prism-purescript.js
```diff
@@ -0,0 +1,38 @@
...
+		// The imported or hidden names are not included in this import
+		// statement. This is because we want to highlight those exactly like
+		// we do for the names in the program.
+		pattern: /((?:\r?\n|\r|^)\s*)import\s+(?:[A-Z][\w']*)(?:\.[A-Z][\w']*)*(?:\s+as\s+(?:[A-Z][_a-zA-Z0-9']*)(?:\.[A-Z][\w']*)*)?(?:\s+hiding\b)?/m,
```
RunDevelopment, line 16: 1. `[_a-zA-Z0-9']` == `[\w']`
1. `/((?:\r?\n|\r|^)\s*).../m` == `/(^\s*).../m` (only true because of the Prism lookbehind)

```suggestion
		pattern: /(^\s*)import\s+[A-Z][\w']*(?:\.[A-Z][\w']*)*(?:\s+as\s+[A-Z][\w']*(?:\.[A-Z][\w']*)*)?(?:\s+hiding\b)?/m,
```

components/prism-purescript.js
```diff
@@ -0,0 +1,38 @@
...
+		lookbehind: true,
+		inside: {
+			'keyword': /\b(?:import|as|hiding)\b/
+		}
+	},
+	// These are builtin variables only. Constructors are highlighted later as a constant.
+	'builtin': /\b(?:when|unless|liftA1|apply|bind|discard|join|ifM|identity|whenM|unlessM|liftM1|ap|compose|otherwise|top|bottom|recip|eq|notEq|degree|div|mod|lcm|gcd|flip|const|map|void|flap|conj|disj|not|mempty|compare|min|max|comparing|clamp|between|sub|negate|append|add|zero|mul|one|show|unit|absurd)\b/,
```
RunDevelopment, line 23: Sorted.

```suggestion
	'builtin': /\b(?:absurd|add|ap|append|apply|between|bind|bottom|clamp|compare|comparing|compose|conj|const|degree|discard|disj|div|eq|flap|flip|gcd|identity|ifM|join|lcm|liftA1|liftM1|map|max|mempty|min|mod|mul|negate|not|notEq|one|otherwise|recip|show|sub|top|unit|unless|unlessM|void|when|whenM|zero)\b/,
```

components/prism-purescript.js
```diff
@@ -0,0 +1,38 @@
...
+	// decimal integers and floating point numbers | octal integers | hexadecimal integers
+	'number': /\b(?:\d+(?:\.\d+)?(?:e[+-]?\d+)?|0o[0-7]+|0x[0-9a-f]+)\b/i,
```
RunDevelopment, line 25: Is "0XFF" a valid number too?

components.json
```diff
@@ -913,6 +913,11 @@
 			"alias": "pbfasm",
 			"owner": "HeX0R101"
 		},
+		"purescript": {
+			"title": "PureScript",
+			"alias": "purs",
+			"owner": "Sri Harsha Chilakapati"
```
RunDevelopment, line 919: We use this to link to your GitHub profile, so please use your GitHub username.

```suggestion
			"owner": "sriharshachilakapati"
```

tests/languages/purescript/string_feature.test
```diff
@@ -0,0 +1,19 @@
+""
+"fo\"o"
+"foo \
+  \ bar"
+"foo -- comment \
```
RunDevelopment, line 5: It's not really a comment, is it?

```suggestion
"foo -- comment lookalike \
```

RunDevelopment (changes requested):
Again, thank you for the language definition @sriharshachilakapati!

I left you a few comments with suggestions and improvements. Please take a look!

sriharshachilakapati:
The syntax of PureScript follows closely to that of Haskell, and hence I derived this from Haskell's one. Will reply to the comments one by one.

components/prism-purescript.js
```diff
@@ -0,0 +1,38 @@
+Prism.languages.purescript = {
+	'comment': {
+		pattern: /(^|[^-!#$%*+=?&@|~.:<>^\\\/])(?:--[^-!#$%*+=?&@|~.:<>^\\\/].*|{-[\s\S]*?-})/m,
```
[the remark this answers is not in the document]
sriharshachilakapati: Empty comments indeed are possible in PureScript.

```purescript
something --
{--}
```

Both the comments in the above snippet are valid.

RunDevelopment:
>  I derived this from Haskell

It's ok to use another language as the base but we usually use the [`extend`](https://prismjs.com/docs/Prism.languages.html#.extend) function to do so. The function basically copies a language and lets you make some changes to it. With `extend`, PureScript can be implemented like this:

```js
Prism.languages.purescript = Prism.languages.extend('haskell', {
	'keyword': /\b(?:ado|case|class|data|derive|do|else|if|in|infixl|infixr|instance|let|module|newtype|of|primitive|then|type|where)\b/,
	'import_statement': {
		// The imported or hidden names are not included in this import
		// statement. This is because we want to highlight those exactly like
		// we do for the names in the program.
		pattern: /((?:\r?\n|\r|^)\s*)import\s+(?:[A-Z][\w']*)(?:\.[A-Z][\w']*)*(?:\s+as\s+(?:[A-Z][_a-zA-Z0-9']*)(?:\.[A-Z][\w']*)*)?(?:\s+hiding\b)?/m,
		lookbehind: true,
		inside: {
			'keyword': /\b(?:import|as|hiding)\b/
		}
	},
	// These are builtin variables only. Constructors are highlighted later as a constant.
	'builtin': /\b(?:when|unless|liftA1|apply|bind|discard|join|ifM|identity|whenM|unlessM|liftM1|ap|compose|otherwise|top|bottom|recip|eq|notEq|degree|div|mod|lcm|gcd|flip|const|map|void|flap|conj|disj|not|mempty|compare|min|max|comparing|clamp|between|sub|negate|append|add|zero|mul|one|show|unit|absurd)\b/,
});
```

The main benefits of this approach are that you have to write less code and that improvements from the extended language will trickle down (e.g. I will make a pull to fix the exponential backtracking in Haskell).

Please use the above code. It's equivalent to your current approach.

I will close all comments on patterns that are identical to Haskell's ones.

components/prism-purescript.js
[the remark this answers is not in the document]
RunDevelopment: I'll fix this issue in Haskell.

components/prism-purescript.js
RunDevelopment: This still has to be changed but will only work after #2535, so be sure to update your fork.

sriharshachilakapati:
@RunDevelopment I've updated the definition to use extend, but now the test cases are failing.

```
1) Testing language 'purescript'
       – should pass test case 'builtin_feature':
     Error: Cannot extend 'haskell' because it is not defined in Prism.languages.
      at Object.extendTest (tests/helper/checks.js:27:10)
      at Object.object.<computed> [as extend] (tests/helper/checks.js:7:10)
      at eval (eval at <anonymous> (tests/helper/prism-loader.js:66:18), <anonymous>:3:46)
      at languageFunction (tests/helper/prism-loader.js:67:57)
      at /Users/sriharshachilakapati/Projects/prism/tests/helper/prism-loader.js:69:4
      at handleId (dependencies.js:269:13)
      at loadComponentsInOrder (dependencies.js:291:4)
      at Object.load (dependencies.js:437:12)
      at Object.loadLanguages (tests/helper/prism-loader.js:56:57)
      at Object.createInstance (tests/helper/prism-loader.js:38:18)
      at Object.runTestCase (tests/helper/test-case.js:58:29)
      at Context.<anonymous> (tests/run.js:32:16)
      at processImmediate (internal/timers.js:458:21)
```

Where do I specify that dependency in test suite?

RunDevelopment:
My bad. You have add `"require": "haskell"` to PureScript in `components.json` (and rebuild).

RunDevelopment (changes requested):
A rebuild is required after making any changes to `components.json` or any language definition.

Also, my unresolved review comments still stand.

tests/languages/purescript/string_feature.test
sriharshachilakapati: This one is a string, but results in an error because it is not closed.

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @sriharshachilakapati!

sriharshachilakapati:
@RunDevelopment Thanks for your patience in guiding me through the PR. By the way, may I know if there are any plans for a release anytime soon?

RunDevelopment:
> any plans for a release anytime soon?

No. We release rather infrequently, so maybe in a few weeks? 
However, all merged features are available right now on the [download page](https://prismjs.com/download.html#themes=prism&languages=haskell+purescript).

sriharshachilakapati:
I'm asking cause I came here from another project. If released, I can directly update the package.json file there and it should work for me.

mAAdhaTTah:
@sriharshachilakapati I can't commit to anything in particular, but I will spend some time going through PRs and looking at what we should include in our next release.

@RunDevelopment We get this question a lot. We may want to consider either a regular release cycle and/or automating the release process.

RunDevelopment:
@mAAdhaTTah I was thinking about a regular release cycle too (maybe monthly?) and automating sounds good. How would we go about automating? (I always hand it off to you after making the changelog, so how involved is the process of making a new release?)

mAAdhaTTah:
It's not super involved, I think you really just need publish perms. But if we can set up a bot to do this stuff for us, it would be a lot easier.

RunDevelopment:
Perms?

We could set up a bot with GH actions. They finally support [manual triggers](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). 
I'm not sure _who_ can use these triggers thou. Would be a problem if anybody could make a new release.

mAAdhaTTah:
perms => permissions.

We could definitely have the bot only act in response to certain users (org members, probably). I'll create an issue to flesh this out a bit.

georgefst:
I can see I've narrowly missed the boat here, but I've always felt that highlighting some arbitrary set of "built-in" functions from isn't very true to the spirit of these languages, since there isn't actually anything special about them.

There was a discussion about this in [the Haskell Vscode plugin](https://github.com/JustusAdam/language-haskell/issues/11), and it was decided to remove it.

PR #2532, opened 2020-08-27
Regex: Changed how languages embed regexes

RunDevelopment:
Because of how Regex modified JS, [whether JSX got regex highlighting](https://github.com/wooorm/refractor/issues/34) depended on the loading order of JSX and Regex.

This pull fixes this issue by redoing the dependencies around Regex. JS and Vala will now have Regex as an optional dependency. This means that all languages that extend JS will automatically get regex highlighting (if available).


RunDevelopment:
Regarding conflicts: Most of the conflicts we have are in generated files. These conflicts are trivial to resolve (merge, run `npm run build`, commit everything), so should I add a little npm script that does this automatically? Might be useful for first-time contributors.

mAAdhaTTah:
That would only work if the bot has access to commit to the PR, so we'd need to make sure it's resilient. Maybe we need to consider switching to GitHub actions? That would make some of this stuff easier.

RunDevelopment:
That sounds a little bit too automated for my taste. I was thinking about a little script I could point to when people have to resolve merge conflicts.

> That would make some of this stuff easier.

Yeah, for stuff like automated releases that'd be easier. Should I make a quick PR to move the current Travis pipeline?

(Also, one annoying side of GH actions is that workflows are hard to change. You can't easily make a new branch and change them. GH will reject your push. Pretty annoying but can be worked around.)

PR #2533, opened 2020-08-27
JS Extras: Highlight import and export bindings

RunDevelopment:
This wraps the binding declarations of import/export statements.

---

@karlhorky
This should also work for you, right?

image


karlhorky:
LGTM, thanks!

karlhorky (approved):

mAAdhaTTah:
`maybe-class-name` seems a bit odd to me. What's the idea here?

karlhorky:
I think that's not new - looks like it was added in https://github.com/PrismJS/prism/pull/1743

RunDevelopment:
Yes, it's been there for a while. The idea is that class names are usually written in pascal case, so we can use this naming convention. However, it's not perfect, so I didn't want to name it "class-name".

mAAdhaTTah:
Thanks for the clarification!

karlhorky:
@RunDevelopment @mAAdhaTTah could we have this also published as a new version of `prismjs`?

Version `1.21.0` (current version as of 8 Oct 2020) doesn't include this change: https://unpkg.com/browse/prismjs@1.21.0/components/prism-js-extras.js

RunDevelopment:
Well, it's been 2 months since the last release.

@mAAdhaTTah I'll prepare the changelog.

mAAdhaTTah:
@karlhorky @RunDevelopment v1.22.0 has been published.

RunDevelopment:
@mAAdhaTTah Thanks! 

On that note: Should we include all changes in the GH release notes? I always removed/didn't include them because I don't want to maintain two copies of the same thing. How do we want to handle this going forward? Always/never include the changes or just have a link to our changelog in the release notes?

mAAdhaTTah:
@RunDevelopment That makes sense, just do it in the Changelog.

karlhorky:
Cool, thanks for that!

PR #2542, opened 2020-09-06
Added support for Birb

Calamity210:
Added support for a custom language(Birb)


RunDevelopment:
Thank you for making this language definition @Calamity210!

However, I don't know if we can merge this.
My main problem is that there is no language specification or detailed documentation for Birb. Without such a document, I have no way of knowing that your language definition is actually correct. 

Right now, the [documentation for your language](https://gc.spidev.codes/docs/birb/) is too little for others to work with. Specifically for this PR, the doc is inconsistent with the Prism language definition. E.g. at no point does the doc mention triple-quote string literals and the doc explicitly states that [single-quote string literals can be multiline](https://gc.spidev.codes/docs/birb/core/String.html#string-type) but that isn't supported by the language definition.

The lack of documentation also means that the language definition will be practically unmaintainable since we won't be able to review any patches or enhancements to the Birb language definition either.

Thoughts? @Calamity210 @mAAdhaTTah

Calamity210:
For sure makes sense to me, the triple quotes was a typo. Birb doesnt support it. Mind if I request a review from you once our docs are improved? @RunDevelopment

RunDevelopment:
Sure! 

(Maybe add a comment with a link to the docs to your language definition once you're finished.)

Calamity210:
@RunDevelopment We are still working on an api documentation, but the tour documents the complete syntax. https://birbolang.web.app/docs/ this document contains more or less everything about Birbs syntax itself as it stands right now, what exactly are you looking for with the docs?

RunDevelopment:
> what exactly are you looking for with the docs?

Prism highlights concepts (functions, classes, operators, ...), so when reading through language documentation, I'm always looking for all the concepts the language has to offer. I then go through each concept and look for its syntax (most docs do this via examples but some even give you the BNF grammar of single concepts or even the whole language).
To summaries: I look for everything the language can do and how that is expressed in code.

I think the doc is ok in that regard. One of the nice sides of a minimal language is that you don't need a lot of documentation to describe it.

Ready for review?

---

Also, I'm a little curious as to how Birb infers the type of a variable with `var`. Example:

```
var a = 100; // clearly int
var b = 0.5; // clearly double
var c = 1e2; // int or double?
var d = 1e100; // too big for int, so is it a double?
```

Calamity210:
> Ready for review?
Yeah, lets do it!


> Also, I'm a little curious as to how Birb infers the type of a variable with `var`. Example:
> 
> ```
> var a = 100; // clearly int
> var b = 0.5; // clearly double
> var c = 1e2; // int or double?
> var d = 1e100; // too big for int, so is it a double?
> ```

`var` is still being worked on to better handle edge cases, but in the case you provided above, both c and d would be doubles. Exponential numbers in birb are doubles by default.

tests/languages/birb/string_feature.test
```diff
@@ -0,0 +1,13 @@
+"" ''
+"fo\"o" 'fo\'o'
```
RunDevelopment, line 2: Please add a multiline string example.

This also leads to another point: C-like strings aren't multiline. You need to make your own pattern like this: `/("|')(?:\\[\s\S]|(?!\1)[^\\])*\1/`

tests/languages/java/function_featrue.test
```diff
@@ -1,31 +1,31 @@
-void foo(int a) {}
```
RunDevelopment, line 1: Please revert these changes to an unrelated test file.

tests/languages/birb/operator_feature.test
```diff
@@ -0,0 +1,27 @@
+++ --
+* / % ~/
++ - ! ~
+<< >> ?
```
RunDevelopment, line 4: I suggest to also make `:` an operator to keep the highlighting of `cond ? if_case : else_case` consistent.

components/prism-birb.js
```diff
@@ -0,0 +1,25 @@
+Prism.languages.birb = Prism.languages.extend('clike', {
+	'class-name': [
+		/\b[A-Z](?:\w*[a-zA-Z]\w*)?\b/,
+    	/\b[A-Z]\w*(?=\s+\w+\s*[;,=())])/
+    ],
+	'keyword': /\b(?:assert|bool|break|case|class|const|default|double|else|enum|final|follows|for|grab|if|int|nest|next|new|noSeeb|return|static|switch|throw|var|void|while)\b/,
```
RunDevelopment, line 6: `bool`, `double`, and `int` are keywords? They're not in the [keyword list](https://birbolang.web.app/docs/#keywords).

tests/languages/birb/keyword_feature.test
```diff
@@ -0,0 +1,43 @@
+assert
+bool break case
+class;
+const default double
+else enum
+final
+follows;
+for grab
+if int List
```
RunDevelopment, line 9: `List` and a few others aren't keywords. What are they doing here?

RunDevelopment (changes requested):
Here's the review.

Apart from my comment, please use tabs for indentation. There are a few lines where tabs and spaces are mixed.

Calamity210:
> Here's the review.
> 
> Apart from my comment, please use tabs for indentation. There are a few lines where tabs and spaces are mixed.

Had I mixed them, i had a bit of a brainfog the day I worked on this, ill correct this along with the comments in a bit. Thanks!

Calamity210:
@RunDevelopment I've taken a look at the comments and corrected them.  I removed the pattern I had for functions as it seemed like I was misunderstanding something. Everything else should be fine afaict.

components/prism-birb.js
```diff
@@ -0,0 +1,25 @@
...
+	'variable': /[a-z_]\w*/,
```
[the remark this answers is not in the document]
Calamity210: > Please move this to the end of the object. It's misleading otherwise.

I have moved the items above lower to according to the positions of the overwritten tokens, but im not quite sure if this is what you suggested. Please let me know if it doesn't look right

components/prism-birb.js
```diff
@@ -0,0 +1,23 @@
...
+		// matches variable and function return types (parameters as well).
+		/\b[A-Z]\w*(?=\s+\w+\s*[;,=()])/
+	],
+	'string': {
+		pattern: /r?("|')(?:\\.|(?!\1)[^\\])*\1/,
+		greedy: true
+	},
+	'keyword': /\b(?:assert|break|case|class|const|default|else|enum|final|follows|for|grab|if|nest|next|new|noSeeb|return|static|switch|throw|var|void|while)\b/,
+	'variable': /\b[a-z_]\??\w*\b/,
+	'operator': /\+\+|--|&&|\|\||<<=?|>>=?|~(?:\/=?)?|[+\-*\/%&^|=!<>:]=?|\?/,
```
RunDevelopment, line 14: `:=` is an operator?

components/prism-birb.js
```diff
@@ -0,0 +1,23 @@
+Prism.languages.birb = Prism.languages.extend('clike', {
+	'class-name': [
+		/\b[A-Z](?:[\d_]*[a-zA-Z]\w*)?\b/,
+
+		// matches variable and function return types (parameters as well).
+		/\b[A-Z]\w*(?=\s+\w+\s*[;,=()])/
+	],
+	'string': {
+		pattern: /r?("|')(?:\\.|(?!\1)[^\\])*\1/,
```
RunDevelopment, line 9: What is the `r?` for?

components/prism-birb.js
```diff
@@ -0,0 +1,23 @@
...
+		greedy: true
+	},
+	'keyword': /\b(?:assert|break|case|class|const|default|else|enum|final|follows|for|grab|if|nest|next|new|noSeeb|return|static|switch|throw|var|void|while)\b/,
+	'variable': /\b[a-z_]\??\w*\b/,
```
RunDevelopment, line 13: What is the `\??` for?

Also, please move it after `operator` since `operator` is also a token of C-like.

RunDevelopment (changes requested):

components/prism-birb.js
Calamity210: I am adding "raw" strings to Birb sometime today, any string starting with `r` as in ```
r`foo\n`
``` would be a raw string (treats \ as a literal and not an escape character)

components/prism-birb.js
Calamity210: it is not an operator, this is to match smth like `int?`, datatypes ending in `?` are nullable. Though, I have misplaced it

components/prism-birb.js
Calamity210: Nope, misplacement. Sry

components/prism-birb.js
Calamity210: I've removed it as we are still testing/implementing it and haven't included it in the docs yet

RunDevelopment (approved):

RunDevelopment:
Looks good!

To resolve the merge conflicts, run `npm run build` after merging Prism's master branch and commit everything as is.

Calamity210:
> Looks good!
> 
> To resolve the merge conflicts, run `npm run build` after merging Prism's master branch and commit everything as is.

Yup give me one min

Calamity210:
Resolved @RunDevelopment

RunDevelopment:
Thank you for contributing @Calamity210!

PR #2549, opened 2020-09-11
Autoloader: Fixed file detection regexes

RunDevelopment:
Guess who forgot to make a group optional and then blindly copy-pasted it. Yep, me.

This fixes the bug introduced in #1928. The Autoloader plugin will now also recognize non-minified files.


mAAdhaTTah (approved):

8thHalfHour:
So, is it now possible to include the link to the autoloader CDN and not have to also set the `languages_path` to the components folder? In other words, is this still required after linking to the script:

```
<script>
          Prism.plugins.autoloader.languages_path = 'https://cdnjs.cloudflare.com/ajax/libs/prism/1.22.0/components/';
</script>
```

RunDevelopment:
@8thHalfHour Yes, this will now work. You don't have to manually set `languages_path` anymore.

Issue #2546, opened 2020-09-09
Provide minified theme css


mhatvan:
**Motivation**
It is common to have min.css files in distributed npm packages to profit from file size savings.

**Description**
Currently I am using `import 'prismjs/themes/prism-okaidia.css';` but the css file is not minified.

RunDevelopment:
Related to #956.

Issue #956, opened 2016-05-16
Provide option to download minified CSS


CITguy:
The download page includes the option to get a full development vs minified source, but this option does not extend to the CSS (see gif below). 

![prismjs-download-minified-no-css](https://cloud.githubusercontent.com/assets/545605/15304218/a8ffe9a2-1b81-11e6-8d6b-12b881722254.gif)

Is it possible to add minification to the CSS source?

LeaVerou:
The reason is that most people modify the CSS at least to a degree, but don't modify the JS. Also the CSS is much smaller, so it's not really worth it to minify.

hc-codersatlas:
Useful to provide this for users using prism via cdn, e.g. jsdelivr and unpkg? We override the css after including the CDN file.

PR #2553, opened 2020-09-14
Added Minified CSS using gulp-clean-css API

ashwinjayan:
Minified css automatically generated using gulp-clean-css added to gulpfile.js folder


RunDevelopment:
Thank you for making this PR @ashwinjayan!
(You forgot to add `gulp-clean-css` as a dev dependency. Please commit the changes to `package.json` and `package-lock.json`.)

@ashwinjayan @mAAdhaTTah @LeaVerou @Golmote 
Is this something we want/need for Prism? I see that it makes sense to offer minified themes and plugin CSS but I never saw this as a priority because the file size of minified themes is only about 25% to 30% less than the unminified versions.

Note: If we added the minified files, we should also add support for this to the download page (shouldn't be too hard). However, changes to the download page shouldn't be part of this PR, IMO.

Thoughts?

---

Related issues: #2546, #956

ashwinjayan:
Hi @RunDevelopment, I missed to save the dependency. Have added it to dev-dependency and committed.

gulpfile.js/paths.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
 	componentsFile: 'components.json',
 	componentsFileJS: 'components.js',
 	components: ['components/**/*.js', '!components/index.js', '!components/**/*.min.js'],
+	themes: 'themes/*.css',
```
RunDevelopment, line 7: Do you see the `.min.min.css` files as well?

```suggestion
	themes: ['themes/*.css', '!themes/*.min.css'],
```

ashwinjayan: Have made this changes and committed the code again.

gulpfile.js/index.js
```diff
@@ -209,13 +210,17 @@ async function languagePlugins() {
...
+const theme = minifyCSS;
 
 module.exports = {
 	watch: watchComponentsAndPlugins,
-	default: series(parallel(components, plugins, componentsJsonToJs, build), docs),
+	default: series(parallel(components, theme, plugins, componentsJsonToJs, build), docs),
 	linkify,
 	changes
-};
+};
```
mAAdhaTTah, line 226: Nit, but this should keep the newline.

mAAdhaTTah:
@RunDevelopment How hard would it be to add to the download page? @ashwinjayan Is this something you'd be willing to tackle?

I do think it's a little curious we provide the JS minified and not the CSS so I'm open to adding this, but we should probs include the download page change as part of this.

RunDevelopment:
> How hard would it be to add to the download page?

Pretty easy from what I see. If plugin CSS was also minified, then [this](https://github.com/PrismJS/prism/blob/0604793c64906c4307456d3cceb1b823d1ca9d9b/assets/download.js#L184) is the only line that has to be changed.

RunDevelopment:
I'll close this now in favor of #3073.

PR #2561, opened 2020-09-17
YAML: Improved key pattern

### LINKED ISSUE #2560: YAML: inconsistent mapping key highlighting
**Information:**
- Prism version: Latest
- Plugins: None
- Environment: Browser

**Description**

Prism doesn't handle all strings equally when they are mapping keys, in particular when there are spaces/dashes.

**Example**

```yaml
hello: world
hello the-world: 23
```

RunDevelopment:
I just spend over an hour reading through YAML's spec and simplifying the plain key pattern, so it could be expressed as a regex... Good times.

Anyway, this changes the key pattern to be very close to the spec. It also adds proper support for string keys.

This fixes #2560.


eepp:
YAML can be quite complex.

Thanks a lot for taking the time!

PR #2566, opened 2020-10-01
feature: support for PHP 7.4 (typehint, return types)

TomPavelec:
fix: `Foo::bar` Foo matched as class name
fix: `new self` self matched as keyword instead of class name
fix: `use function` is matched as package


components/prism-php.js
```diff
@@ -42,14 +49,19 @@
...
-			pattern: /(\\|namespace\s+|use\s+)[\w\\]+/,
+			pattern: /(\\|namespace\s+|use\s+(function\s+)*)[a-zA-Z_][\w\\]+/,
 			lookbehind: true,
 			inside: {
 				punctuation: /\\/
 			}
+		},
+		'type': {
+			pattern: /(bool|boolean|int|integer|float|string|object|void|array|mixed)(?!_)/,
+			greedy: !0
```
RunDevelopment, line 64: Same for all the others.

```suggestion
			greedy: true
```

components/prism-php.js
```diff
@@ -42,14 +49,19 @@
 		}
 	});
 
-	Prism.languages.insertBefore('php', 'keyword', {
+	// Must be defined before class-name (because primitive types must be matched before class names in function typehint)
+	Prism.languages.insertBefore('php', 'class-name', {
 		'variable': /\$+(?:\w+\b|(?={))/i,
 		'package': {
-			pattern: /(\\|namespace\s+|use\s+)[\w\\]+/,
+			pattern: /(\\|namespace\s+|use\s+(function\s+)*)[a-zA-Z_][\w\\]+/,
```
RunDevelopment, line 56: Keywords are case insensitive, so please add the `i` flag.

> `(function\s+)*` 

Star? Shouldn't that be a `?`?

> `[a-zA-Z_][\w\\]+`

Why the `[a-zA-Z_]`? This prevents the pattern from matching text like `use \bar` in full.

components/prism-php.js
```diff
@@ -114,6 +126,25 @@
 	// The different types of PHP strings "replace" the C-like standard string
 	delete Prism.languages.php['string'];
 
+	Prism.languages.insertBefore('php', 'class-name', {
+		'class-static-call': {
+			pattern: /[a-zA-Z_][\w\\]+(?=\:\:)/,
```
RunDevelopment, line 131: 1. We want to support single-character class names.
1. There may be any number of spaces before `::`.
1. The `\b` will prevent false positives.

```suggestion
			pattern: /\b[a-zA-Z_][\w\\]*(?=\s*::)/,
```

components/prism-php.js
```diff
@@ -114,6 +126,25 @@
...
+			greedy: !0,
+			alias: 'class-name'
+		},
+		'class-type-hint': {
+			pattern: /[a-zA-Z_][\w\\]+(?=\s*\$)/,
```
RunDevelopment, line 136:
```suggestion
			pattern: /\b[a-zA-Z_][\w\\]*(?=\s*\$)/,
```

components/prism-php.js
```diff
@@ -114,6 +126,25 @@
...
+			greedy: !0,
+			alias: 'class-name'
+		},
+		'class-return-type': {
+			pattern: /(\)\s*\:\s*\?*\s*)[a-zA-Z_][\w\\]+/,
```
RunDevelopment, line 141:
```suggestion
			pattern: /(\)\s*:\s*\?*\s*)[a-zA-Z_][\w\\]*/,
```

components/prism-php.js
```diff
@@ -42,14 +49,19 @@
...
 		'package': {
-			pattern: /(\\|namespace\s+|use\s+)[\w\\]+/,
+			pattern: /(\\|namespace\s+|use\s+(function\s+)*)[a-zA-Z_][\w\\]+/,
 			lookbehind: true,
 			inside: {
 				punctuation: /\\/
 			}
+		},
+		'type': {
+			pattern: /(bool|boolean|int|integer|float|string|object|void|array|mixed)(?!_)/,
```
RunDevelopment, line 63: 1. Please use non-capturing groups (one of the reasons the tests fail).
1. Please use boundary assertions.

```suggestion
			pattern: /\b(?:bool|boolean|int|integer|float|string|object|void|array|mixed)\b/,
```

components/prism-php.js
```diff
@@ -12,7 +12,7 @@
  */
 (function (Prism) {
 	Prism.languages.php = Prism.languages.extend('clike', {
-		'keyword': /\b(?:__halt_compiler|abstract|and|array|as|break|callable|case|catch|class|clone|const|continue|declare|default|die|do|echo|else|elseif|empty|enddeclare|endfor|endforeach|endif|endswitch|endwhile|eval|exit|extends|final|finally|for|foreach|function|global|goto|if|implements|include|include_once|instanceof|insteadof|interface|isset|list|namespace|new|or|parent|print|private|protected|public|require|require_once|return|static|switch|throw|trait|try|unset|use|var|while|xor|yield)\b/i,
+		'keyword': /\b(?:__halt_compiler|abstract|and|array|as|break|callable|case|catch|class|clone|const|continue|declare|default|die|do|echo|else|elseif|empty|enddeclare|endfor|endforeach|endif|endswitch|endwhile|eval|exit|extends|final|finally|for|foreach|function|global|goto|if|implements|include|include_once|instanceof|insteadof|interface|isset|list|namespace|new|or|parent|print|private|protected|public|require|require_once|return|self|static|switch|throw|trait|try|unset|use|var|while|xor|yield)\b/i,
```
RunDevelopment, line 15: Is `self` a keyword?

tests/languages/php/class-name_feature.test
```diff
@@ -0,0 +1,34 @@
...
+(Foo $variable)
+): Foo
+): ?Foo
+
+----------------------------------------------------
+
+[
+	["class-name", ["Foo"]],
+	["punctuation", ":"],
+	["punctuation", ":"],
```
RunDevelopment, line 11: These two should be one token. Please overwrite the `punctuation` pattern with `/::|[{}[\];(),.:]/`. To do this, add the following lines after the `class-name` token:

```diff
 Prism.languages.php = Prism.languages.extend('clike', {
 	....
 	'class-name': {
 		...
-	}
+	},
+	'punctuation': /::|[{}[\];(),.:]/
 });
```

RunDevelopment (changes requested):
Thank you for improving the PHP language definition @TomPavelec!

There are a few things that have to be improved but it looks ok overall.

components/prism-php.js
TomPavelec: copy pasted from minified code -> fixed in next commit

components/prism-php.js
TomPavelec: Yes star changed to `?`
My point was to not match names beginning with number (that is syntax error) -> fixed in next commit

components/prism-php.js
TomPavelec: Same as namespace, do not  match class names beginning with number -> fixed in next commit

components/prism-php.js
TomPavelec: fixed

components/prism-php.js
TomPavelec: fixed

components/prism-php.js
TomPavelec: fixed

tests/languages/php/class-name_feature.test
TomPavelec: double colon `::` is actually operator (scope / static access) -> moved to operators

components/prism-php.js
TomPavelec: Yes it is a keyword in context like `new self();` or `self::$var`.
But in it is a type (return type) in `function foo(): self`

Both contexts fixed in next commit

RunDevelopment (changes requested):
Wow. Better number, operator, and punctuation patterns and even more PHP features (execution operators). I really want to get this merged now but before that, we still have fix some issues.

components/prism-php.js
```diff
@@ -114,6 +146,61 @@
...
+		},
+		'class-name static-context': {
+			pattern: /\b[a-zA-Z_][\w]*(?!\\)\b(?=\s*::)/i,
+			greedy: true,
+		},
+		'class-name class-name-fully-qualified static-context': {
+			pattern: /(?:\\*\b[a-zA-Z_][\w]*)+\b(?!\\)(?=\s*::)/i,
+			greedy: true,
+			inside: {
+				punctuation: /\\/
```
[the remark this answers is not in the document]
TomPavelec: fixed

components/prism-php.js
```diff
@@ -114,6 +146,61 @@
...
+			},
+			lookbehind: true
+		},
+		'class-name static-context': {
+			pattern: /\b[a-zA-Z_][\w]*(?!\\)\b(?=\s*::)/i,
+			greedy: true,
+		},
+		'class-name class-name-fully-qualified static-context': {
+			pattern: /(?:\\*\b[a-zA-Z_][\w]*)+\b(?!\\)(?=\s*::)/i,
```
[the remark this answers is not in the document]
TomPavelec: fixed

components/prism-php.js
```diff
@@ -114,6 +146,61 @@
...
+			greedy: true,
+			inside: {
+				punctuation: /\\/
+			}
+		},
+		'class-name type-hint': {
+			pattern: /([(,?]\s*)\b[a-zA-Z_][\w]*(?!\\)\b(?=\s*\$)/i,
```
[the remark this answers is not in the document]
TomPavelec: fixed

components/prism-php.js
```diff
@@ -114,6 +146,61 @@
...
+			greedy: true,
+			lookbehind: true
+		},
+		'class-name class-name-fully-qualified type-hint': {
+			pattern: /([(,?]\s*)(?:\\*\b[a-zA-Z_][\w]*)+\b(?!\\)(?=\s*\$)/i,
```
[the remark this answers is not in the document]
TomPavelec: fixed

components/prism-php.js
```diff
@@ -114,6 +146,61 @@
...
+		'class-name class-name-fully-qualified': {
+			pattern: /(\b(?:extends|implements|instanceof|new(?!\s+self|\s+static))\s+|\bcatch\s+\()(?:\\*\b[a-zA-Z_][\w]*)+\b(?!\\)/i,
+			greedy: true,
+			inside: {
+				punctuation: /\\/
+			},
+			lookbehind: true
+		},
+		'class-name static-context': {
+			pattern: /\b[a-zA-Z_][\w]*(?!\\)\b(?=\s*::)/i,
```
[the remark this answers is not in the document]
TomPavelec: fixed

components/prism-php.js
```diff
@@ -114,6 +146,61 @@
...
+	Prism.languages.insertBefore('php', 'keyword', {
+		'keyword static-context': {
+			pattern: /\b(?:parent|self|static)(?=\s*::)/i,
+			greedy: true
+		}
+	});
+
+	Prism.languages.insertBefore('php', 'boolean', {
+		'class-name class-name-fully-qualified': {
+			pattern: /(\b(?:extends|implements|instanceof|new(?!\s+self|\s+static))\s+|\bcatch\s+\()(?:\\*\b[a-zA-Z_][\w]*)+\b(?!\\)/i,
```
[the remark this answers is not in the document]
TomPavelec: fixed

components/prism-php.js
```diff
@@ -42,14 +49,34 @@
...
 		'package': {
-			pattern: /(\\|namespace\s+|use\s+)[\w\\]+/,
+			pattern: /(namespace\s+|use\s+(?:function\s+)?)(?:\\*\b[a-zA-Z_][\w]*)+\b(?!\\)/i,
 			lookbehind: true,
 			inside: {
 				punctuation: /\\/
 			}
+		},
+		'type type-casting': {
+			pattern: /(\(\s*)\b(?:bool|boolean|int|integer|float|string|object|array(?!\s*\())\b(?=\s*\))/i,
```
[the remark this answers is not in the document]
TomPavelec: fixed

components/prism-php.js
```diff
@@ -42,14 +49,34 @@
 		}
 	});
 
-	Prism.languages.insertBefore('php', 'keyword', {
+	// Must be defined before class-name (because primitive types must be matched before class names in function typehint)
+	Prism.languages.insertBefore('php', 'class-name', {
 		'variable': /\$+(?:\w+\b|(?={))/i,
 		'package': {
-			pattern: /(\\|namespace\s+|use\s+)[\w\\]+/,
+			pattern: /(namespace\s+|use\s+(?:function\s+)?)(?:\\*\b[a-zA-Z_][\w]*)+\b(?!\\)/i,
```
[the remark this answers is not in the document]
TomPavelec: fixed

components/prism-php.js
```diff
@@ -114,6 +146,61 @@
 	// The different types of PHP strings "replace" the C-like standard string
 	delete Prism.languages.php['string'];
 
+	Prism.languages.insertBefore('php', 'keyword', {
+		'keyword static-context': {
```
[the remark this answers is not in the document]
TomPavelec: I refactored it to this system with aliases e.g. `alias: ['class-name-fully-qualified', 'static-context'],` for `class-name`.

But now I have "problem" in tests.
In test drive the `Foo\Bar\Baz::Q` have classes `class-name class-name-fully-qualified static-context`, that's right.
But in test it is tokenized only as `class-name` and aliases are not included.
And that's the "problem" I can't test different scopes. I can only test that it is `class-name`.

Is it real problem for you? Should i test only top lvl  name?

I have an idea: (and already functional prototype)
- rename test from `class-name_feature.test` to something like `class-name_feature.test_with_alias`
- in `tests/run.js` check that extension
- in `TestCase.runTestCase` add new parameter `withAlias`
- in `TestCase.simpleTokenize` if `withAlias` is true add alias to token type

tests/pattern-tests.js
```diff
@@ -291,7 +291,7 @@ function testPatterns(Prism) {
 	});
 
 	it('- should have nice names and aliases', function () {
-		const niceName = /^[a-z][a-z\d]*(?:[-_][a-z\d]+)*$/;
+		const niceName = /^[a-z][-_a-z\d]*(?: [a-z][-_a-z\d]*)*$/;
```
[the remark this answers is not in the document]
TomPavelec: Yes, I agree. Returned back.

components/prism-php.js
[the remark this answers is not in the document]
TomPavelec: @RunDevelopment wdyt? https://github.com/PrismJS/prism/pull/2566/commits/a7f885c69d0d361fc1cefb022c2953e60835ff04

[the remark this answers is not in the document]
RunDevelopment: @TomPavelec Sorry for the delay. 

> Is it real problem for you? Should i test only top lvl name?

Honestly, it's not really a problem. It's nice that we can differentiate different usages of class names/types/etc. but only a small number of themes (if any) will actually use this. Therefore it's not strictly necessary to test these aliases.

I think that your prototype is nice and usable but we already have [this format](https://github.com/PrismJS/prism/blob/master/tests/languages/markup/entity_feature.js) that can serve a similar purpose.

components/prism-php.js
```diff
@@ -1,118 +1,204 @@
...
+				lookbehind: true
+			},
+			{
+				pattern: /(\)\s*:\s*\?*\s*)\b(?:bool|boolean|int|integer|float|string|object|void|array(?!\s*\()|mixed|self|static)\b/i,
+				alias: 'return-type',
+				greedy: true,
+				lookbehind: true
+			},
+			{
+				pattern: /\b(?:bool|boolean|int|integer|float|string|object|void|array(?!\s*\()|mixed)\b/i,
```
RunDevelopment, line 56: Including `array` here is a bit of an issue because `array` is a keyword. This causes inconsistent highlighting for snippets like:

```php
function functionName (array $argumentName){}
```

Question: Should `array` be a keyword here or is the non-keyword highlighting fine?

components/prism-php.js
```diff
@@ -1,118 +1,204 @@
...
-				punctuation: /\\/
+				'punctuation': /\\/
 			}
-		}
-	});
-
-	// Must be defined after the function pattern
-	Prism.languages.insertBefore('php', 'operator', {
+		},
+		'type': [
```
RunDevelopment, line 36: Question: What is the difference between class name and type? I get that these are builtin types but should we make this distinction?

components/prism-php.js
```diff
@@ -1,118 +1,204 @@
...
-	Prism.languages.insertBefore('php', 'keyword', {
+		},
+		'comment': [
+			{
+				pattern: /(^|[^\\])(?:\/\*[\s\S]*?\*\/|\/\/.*)/,
+				lookbehind: true
+			},
+			{
+				pattern: /(^|[^\\])#.*/,
+				alias: 'shell-comment',
```
RunDevelopment, line 24: Is this alias necessary? If not, then we can combine both patterns into `/(^|[^\\])(?:\/\*[\s\S]*?\*\/|\/\/.*|#.*)/`.

Also, please make both `comment` patterns greedy.

components/prism-php.js
```diff
@@ -1,118 +1,204 @@
...
+				lookbehind: true
+			}
+		],
 		'variable': /\$+(?:\w+\b|(?={))/i,
 		'package': {
-			pattern: /(\\|namespace\s+|use\s+)[\w\\]+/,
+			pattern: /(namespace\s+|use\s+(?:function\s+)?)(?:\\?\b[a-zA-Z_]\w*)+\b(?!\\)/i,
```
RunDevelopment, line 30: Because of the i Flag, we can simplify `[a-zA-Z_]` to `[a-z_]`.

Same for almost all of the other occurrences of ``[a-zA-Z_]``.

```suggestion
			pattern: /(namespace\s+|use\s+(?:function\s+)?)(?:\\?\b[a-z_]\w*)+\b(?!\\)/i,
```

RunDevelopment (changes requested):
@TomPavelec Again, sorry for the delay.

I reviewed it again and I think that it's almost ready to merge. We still have to decide what to do with your alias test implementation and I also left you some comments.

The conflicts resulting from #2574 also have to be resolved.

If you want to, then I'll make you a PR resolving some of my comments.

components/prism-php.js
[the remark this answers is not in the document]
TomPavelec: I think that if there is already system which can test aliases i can use it instead of my solution.
I just didn't know that it exists.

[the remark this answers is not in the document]
RunDevelopment: The format is described in [the test suite documentation](https://prismjs.com/test-suite.html#writing-specific-tests) but it's rarely used (only 5 out of ~1700 test files use this format).

tests/languages/php/class-name_feature.test
```diff
@@ -0,0 +1,124 @@
...
+\Package\Foo::bar();
+
+(Foo $variable)
+(\Foo $variable)
+(\Package\Foo $variable)
+
+): Foo
+): \Foo
+): \Package\Foo
+): ?Foo
```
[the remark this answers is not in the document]
TomPavelec: fixed

components/prism-php.js
TomPavelec: fixed

components/prism-php.js
TomPavelec: I removed `shell-comment` alias (it's not necessary).
But when pattern it is greedy, then `// ?>` is matched as comment (instead of `//` comment and `?>` delimiter)

components/prism-php.js
TomPavelec: Actualy `array` is keyword in all contexts, I moved types under keyword.
So `array` is always keyword and alias is added as context (type hint, return type, etc.)

components/prism-php.js
TomPavelec: Class name is always matched as `class-name` and aliases (type hint, return type, type declaration) are added for context

components/prism-php.js
RunDevelopment: That's true. Then let's keep it non-greedy and hope for the best!

components/prism-php.js
```diff
@@ -1,118 +1,210 @@
...
-	});
-
-	Prism.languages.insertBefore('php', 'keyword', {
+		},
+		'comment': [
+			/\/\*\*\//,
+			{
+				pattern: /\/\*\*[\s\S]*?\*\//,
+				alias: 'doc-comment'
+			},
```
RunDevelopment, line 19: What's the idea here? Is PHPDoc causing problems?

TomPavelec: PHP doc comments `/** */` (beginning with two asterisks) are special type of comment, so I think they deserve own alias

RunDevelopment: PHPDoc will automatically add a [`doc-comment` token](https://github.com/PrismJS/prism/blob/master/tests/languages/phpdoc%2Bphp/phpdoc_inclusion.test), so I don't think this is necessary.

TomPavelec: Oh, i see. OK

TomPavelec:
What should I do with failing Latte tests?
`<a n:href="Post:show $post->id">link</a>` attribute value `Post:show $post->id` is not valid php syntax

RunDevelopment:
[Links](https://doc.nette.org/en/3.0/creating-links) seem to enjoy special syntax in Nette Latte. However, this is a problem with Latte and not PHP.

I think it's ok to just accept the Latte error for now. Please just set the current actual token stream as the expected token stream in the test file. I'll open an issue for this after we merge this PR.

RunDevelopment (approved):

RunDevelopment:
Your changes are a major improvement to Prism's PHP language definition!

components/prism-php.js
dg: btw `boolean` and `integer` are not valid typehints in PHP (they are understood as class names), only `bool` and `int`. In PHP 8.0 is valid typehint also `false` (but not `true`).

TomPavelec: You are right, i will fix it in PHP8 features PR https://github.com/PrismJS/prism/pull/2591

components/prism-php.js
dg: here should be `\??` instead of `\?*`.

TomPavelec: yes, i'll fix it in https://github.com/PrismJS/prism/pull/2591

PR #2575, opened 2020-10-04
Test page: Added share option

RunDevelopment:
It can be quite useful to share a link to the test page with a specific language but it wasn't possible to share the highlighted text as well.

This PR adds a "Share" link to the test page. The link points to the test page with the current language + text in the URL hash. The link is automatically updated after changes to the current language or text.

### Example

image

The `href` of the "Share" link is `#language=markup&text=%3Cb%3Efoo%3C%2Fb%3E`.


mAAdhaTTah:
I like this idea; would be helpful for people to file bugs too. Can we update the bug template to include suggesting a link to this page?

RunDevelopment:
In this case, I should probably replace the link with a button that copies the link into the user's clipboard.

RunDevelopment:
I added a little popup around the link.

ffoo

mAAdhaTTah (approved):

PR #2582, opened 2020-10-10
Core: Moved DFS to util namespace

RunDevelopment:
This moves `Prism.languages.DFS` to `Prism.util.DFS`. 

The reason for this is that the DFS function doesn't depend on the values in `Prism.languages` like `insertBefore` and `extend` do. `DFS` is a function that implements a general DFS algorithm that doesn't depend on Prism' languages, so I believe that it shouldn't be part of the `Prism.languages` namespace.


mAAdhaTTah:
I want to agree with this, but I wonder if this is considered a public API.

RunDevelopment:
It isn't documented anywhere, so it's internal IMO. Functions in the `util` namespace [are known to be internal](https://github.com/PrismJS/prism/blob/e01ecd00b0ebdae88ceed0095941ee162072a0bc/components/prism-core.js#L54), so I think it fits.

But we can also wait until 2.0.

mAAdhaTTah:
> Functions in the util namespace are known to be internal

It does, but it's not currently in the `util` namespace, it's in the `language` namespace, which is public. It's also useful, so making it private may not really be necessary.

RunDevelopment:
> It's also useful

To who? I don't think that anybody outside of Prism uses this function. And even we only 3 components that use it...

> making it private

Not private but internal. Plugins will still use the function.

mAAdhaTTah:
> I don't think that anybody outside of Prism uses this function.

Prism is used in a lot of contexts. I don't know that I would assume no one has taken advantage of it.

> Not private but internal. Plugins will still use the function.

From the POV of an external consumer, these are the same.

I think I'd like to hold this for v2.0.

RunDevelopment:
Implemented in v2.

PR #2593, opened 2020-10-15
Set type="button" attribute for copy to clipboard plugin.

alexsegura:
Hi. 

When using Prism + copy to clipboard inside a HTML form, the form is submitted when copying. 
It's because by default the button is seen as a submit button. 
Adding the `type="button"` attribute prevents this.


mAAdhaTTah:
Makes sense. Thanks for the contribution!

alexsegura:
🙌

@mAAdhaTTah is a new release planned anytime soon?

RunDevelopment:
@mAAdhaTTah I think a new release after #2584 might be a good idea. I'm waiting for a confirmation from @yetingli that all vulnerabilities have been fixed but I'm reasonably confident that there won't be many others because of #2590.

mAAdhaTTah:
@RunDevelopment I'm on board with that. We did a release recently, so maybe next one should be a patch?

RunDevelopment:
@mAAdhaTTah Yeah. All commits since v1.22.0 were enhancements and bug fixes, so I agree that it should be a patch.

Issue #2594, opened 2020-10-16
Generics with React TSX Language Leave `plain-text` Token Open


karlhorky:
**Information:**
- Prism version: Test Drive version
- Plugins: Test Drive plugins
- Environment: Test Drive page

**Description**

The generic `Foo<string>` in the following code breaks syntax highlighting of the second function `Add2` with the **React TSX** language:

```tsx
function Add1(a, b) { return <div>a + b</div> }

type Bar = Foo<string>;

function Add2(a, b) { return <div>a + b</div> }
```

**Example**

Screenshots from Test Drive:

![Screen Shot 2020-10-16 at 18 26 59](https://user-images.githubusercontent.com/1935696/96284448-df17cc80-0fdd-11eb-960b-ebdf98eda882.png)

![Screen Shot 2020-10-16 at 18 28 12](https://user-images.githubusercontent.com/1935696/96284450-dfb06300-0fdd-11eb-9b28-110f757da65e.png)

https://prismjs.com/test.html#language=tsx

karlhorky:
Another example, this time with more common, real-world code (highlighting on `handleChange` function is broken):

```tsx
import React, { MouseEvent, ChangeEvent, FormEvent } from 'react';

function handleSubmit(event: FormEvent<HTMLFormElement>) {
  event.preventDefault();
}

function handleChange(event: ChangeEvent<HTMLInputElement>) {
  console.log(event.target.value);
}

function handleClick(event: MouseEvent) {
  console.log(event.button);
}

export default function Form() {
  return (
    <form onSubmit={handleSubmit}>
      <input onChange={handleChange} placeholder="Name" />
      <button onClick={handleClick}></button>
    </form>
  );
}
```

RunDevelopment:
I looked into it and I don't see an easy fix for this issue.

I'm sorry to say this but this might be open for a while.

karlhorky:
Yeah, I was thinking that it may not be an easy one to fix!

Wonder if this fact could help: there is no matching closing tag for generics but JSX requires a matching closing tag.

karlhorky:
Or, another idea: require a non-letter character `[^a-zA-Z]` in front of a JSX tag - otherwise assume it's a generic.

It's rare to have text up against the left side of a starting tag in JSX - and it can and most often is formatted differently than this.

RunDevelopment:
> Wonder if this fact could help: there is no matching closing tag for generics but JSX requires a matching closing tag.

[...]

PR #2596, opened 2020-10-17
TSX: Temporary workaround for the collisions of JSX tags and TS generics

RunDevelopment:
This implements [the workaround](https://github.com/PrismJS/prism/issues/2594#issuecomment-710666928) proposed by @karlhorky.

(I also had to change the JSX tag pattern to get rid of the backreferences.)

This also fixes TSX's global namespace pollution.


karlhorky:
Nice, thanks! Will the version with this be `1.22.1`?

RunDevelopment:
This will be part of 1.23.0 but before that, we will merge a few more PRs.

karlhorky:
Sounds good, cheers :)

badsyntax:
Just want to say this change has fixed things for me. See https://github.com/sergioramos/remark-prism/issues/85

Thanks for the fix!

PR #2597, opened 2020-10-17
Added test for polynomial backtracking

RunDevelopment:
This was branched from #2590. Blocked by #2590.

---

Very similar to #2590, I added a test that checks for polynomial backtracking in Prism's regexes. It only detects one kind of polynomial backtracking but even that yields hundreds of potential vulnerabilities. The problem this time around is that I don't know how to fix them all.

__The problem__

Let's look at one example from JavaScript to illustrate the problem: `/[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*=>)/i`
Do you see it? Here is the output of the test:

```
Prism.languages.javascript.parameter[1].pattern: Polynomial backtracking. By repeating any character that matches /[\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]/i, an attack string can be created.

    [$\w\xA0-\uFFFF]*(?=\s*
    ^~~~~~~~~~~~~~~~~   ^~~

Full pattern:
/[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*=>)/i
                   ^~~~~~~~~~~~~~~~~   ^~~
```

The problem is that `[$\w\xA0-\uFFFF]` and `\s` are __not disjoint__, so we can just take any character that the two have in common and create an attack string. The test tells us that the available characters are `[\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]`.

Using this knowledge, we can create the attack string `"\xA0".repeat(n)` for any `n`. The above regex will take O(n^3) time to reject strings of that form. On my laptop, it takes over 1 second for `/[_$a-z\xA0-\uFFFF][$\w\xA0-\uFFFF]*(?=\s*=>)/i.test("\xa0".repeat(2000))` to return `false`.

Now that you understand the example, __how do we fix this__?
Well, we "just" have to make `[$\w\xA0-\uFFFF]` and `\s` disjoint. The correct way to do it is to subtract `\s` from `[$\w\xA0-\uFFFF]` but how do we do that... The other is to subtract `[$\w\xA0-\uFFFF]` from `\s`.

This question goes out to my fellow maintainers. How do we fix this? 
There are a lot of other examples like this too. Just take a look at the TravisCI logs. There are literally hundreds of these cases and most are non-trivial to fix.

@Golmote @mAAdhaTTah @LeaVerou 

__How does the analysis work?__

<details>

It's not necessary to understand the analysis in order to know that all the problems it finds are vulnerabilities that should be fixed. If you don't care about the implementation details, skip this section.

I go through all quantifiers that have an unbounded maximum (e.g. `+`, `{2,}`). A quantifier's element is then converted into a character set such that the character set is guaranteed accepted by the quantifier's element (e.g. `a+` -> `[a]`, `(a|bb|c)*` -> `[ac]`, `(\w\s*|a\b|foo)*` -> `[]`). The basic idea is that we now know all characters that the quantifier will trivially accept.

To get polynomial backtracking, we just have to find a compatible quantifier and a path to get there. This is done by examining the regex AST nodes that can be reached from the initial quantifier and then recursively continuing the search until we cannot proceed or a compatible quantifier was found.

Examples:
1. The easiest example is `a*a*`. The initial quantifier is the left `a*` with the character set `[a]`. We then find the right `a*` reachable from the left `a*`. The right `a*` is compatible with `[a]`, so we just found a match.
    The attack string is `"a"*n+$`. (`$` is some suffix that causes the full regex to reject all inputs.)
2. `[bh\d]+[a-z][a-fA-F]+`: The initial quantifier is `[bh\d]+` with the character set `[0-9bh]`. We then find `[a-z]` reachable from `[bh\d]+`. To proceed, we have to narrow down the available character set from `[0-9bh]` to `[bh]`. From `[a-z]` reachable is `[a-fA-F]+`. We narrow down the available character set from `[bh]` to `[b]` to match the character set of `[a-fA-F]+`. Since we still have available characters, `[bh\d]+` and `[a-fA-F]+` are compatible.
    The attack string is `"b"*n+$`.

This is a very limited analysis but it's still quite good.

</details>


LeaVerou:
For single character matches, you can subtract character classes by using a negative lookahead before the character class. E.g. for a word character that's not also a number, you can do `/(?!\d)[\w]/`.

However, I am a bit wary of introducing a performance overhead on every single Prism usage, by default, to prevent this, which is only a concern when highlighting unsafe content.

RunDevelopment:
I usually try to avoid negative lookaheads to fix backtracking issues for 3 reasons:

1. The performance overhead. 
   Most browsers are unable to optimize away the lookahead `(?!a)b` (for any character sets `a` and `b`) even though it's a trivial optimization.
1. Tooling.
    Lookarounds are not easy to handle for tooling. The checks for polynomial and exponential backtracking both have trouble with them and are usually forced to ignore the branches that contain them. If you try to fix the backtracking with lookaheads and make a mistake, _you'll never find out_.
1. Simplicity.
    Having a bunch of seemingly random lookahead doesn't make it easier to understand a regex.

While I could do something about points 1 and 2 (tooling can be improved and we could transform our regexes to do these trivial optimizations as part of the build process), point 3 still remains.

> which is only a concern when highlighting unsafe content.

I'd rather have Prism be safe by default or at least as save as we can manage.

LeaVerou:
The alternative I suppose is to manually compute the subtraction result as a character class, but is that really better simplicity-wise?

> I'd rather have Prism be safe by default or at least as safe as we can manage.

At any cost?

RunDevelopment:
> The alternative I suppose is to manually compute the subtraction result as a character class, but is that really better simplicity-wise?

In some cases, it is but here... no. 

Should I write a little babel plugin that applies the lookaheads, so we don't have to worry as much about the performance overhead? At least the minified files should be fast. (see side note below)

> > I'd rather have Prism be safe by default or at least as safe as we can manage.
> 
> At any cost?

No. If I wanted to guarantee O(n) runtime for Prism on all input, I'd have to reimplement everything using something like an LR parser. Yeah... let's not for the time being.

<details>
<summary>Side note</summary>

Even the fixed version of the above example will take O(n^2) time for some strings: 
```js
/(?!\s)[_$a-z\xA0-\uFFFF](?:(?!\s)[$\w\xA0-\uFFFF])*(?=\s*=>)/i
    .test("a".repeat(5_000)) // takes about 35 milliseconds on my computer
```

To make it O(n) for all input, we'd have to add a lookbehind as well:
```js
/(?<!(?!\s)[_$a-z\d\xA0-\uFFFF])(?!\s)[_$a-z\xA0-\uFFFF](?:(?!\s)[$\w\xA0-\uFFFF])*(?=\s*=>)/i
    .test("a".repeat(5_000 * 1000)) // takes about 70 milliseconds on my computer
```

After applying all the `(?!\s)` lookaheads, the regex will run about 4 times faster and looks like this:
```js
/(?<![^\s\x00-\x08\x0e-\x1f!"#\x25-\x2f\x3a-\x40[\\\]`\x7b-\x9f^])[^\s\d\x00-\x08\x0e-\x1f!"#\x25-\x2f\x3a-\x5e`\x7b-\x9f][^\s\x00-\x08\x0e-\x1f!"#\x25-\x2f\x3a-\x40[\\\]`\x7b-\x9f^]*(?=\s*=>)/i
    .test("a".repeat(5_000 * 1000)) // takes about 16 milliseconds on my computer
```

/Side note

</details>

<details>
<summary>Side side note</summary>

While playing around, I found that the lookahead versions also have another problem: The stack.

For long strings (>1MB), both Chrome and Firefox throw an error:

```js
/(?<!(?!\s)[_$a-z\d\xA0-\uFFFF])(?!\s)[_$a-z\xA0-\uFFFF](?:(?!\s)[$\w\xA0-\uFFFF])*(?=\s*=>)/i.test("a".repeat(10e6))
// Chrome 86
> Uncaught RangeError: Maximum call stack size exceeded
// FireFox 80
> Uncaught InternalError: too much recursion
```

The version with applied lookaheads doesn't have this problem. The regex will happily test strings >100MB. Well, it will until we hit the hard limits of the browsers (about 500MB in Chrome and about 1GB in Firefox).

/Side side note

</details>

While it would really nice to have guaranteed O(n) runtime for all of our regexes, it is also impossible in general (e.g. `/(.+)\1/`) and not easy where it is possible. 
However, I do think that we should aim for guaranteed O(n^2). This will ensure that Prism is able to highlight malicious inputs <10kB in at most a few seconds. From what I've seen in the wild, this will probably cover >95% of our users. The added tests will point out (to the best of their ability) possible ways to go higher than O(n^2). If those issues are fixed, we'll be reasonably close to guaranteeing O(n^2) runtime for all regexes.

_The cost?_ Fixing 200-400 of Prism's regexes. 
Truth be told, I already fixed more than half of them (many systematic issues that are easy to fix). The problems are with regexes like the above where it's not easy to subtract the character sets. 

@LeaVerou You're right. I'll use looaheads when there's no other way and be done with it. I was hoping that there was some better way that I overlook but unfortunately it seems like there isn't. Talking about it help me clear this up. Thank you.

<details>
<summary>Side side side note</summary>

If we ever wanted to try to get O(n) for all regexes (where it's possible), we can use the current test for polynomial backtracking.

Basically, it's possible to exploit the linear backtracking of regex `R` if the regex `/[\s\S]*<R>/` is vulnerable to polynomial backtracking.

/Side side side note

</details>

LeaVerou:
Thank you for the thoughtful analysis @RunDevelopment!
I'm not sure an ad hoc Babel plugin would be the way to go, not to mention it's more work for you. 
Perhaps we could use a regexp library like XRegExp to do this client-side? It only needs doing once, so the benefits of precomputing it are small, whereas making the minified version functionally different is not ideal.

RunDevelopment:
> making the minified version functionally different is not ideal.

True, but we already do that to some extend (#2296). I'm all for using tricks to optimize the minified files but I'm also keenly aware that this might go wrong someday. We could run all language tests against the minified files to prevent that thou.

> Perhaps we could use a regexp library like XRegExp to do this client-side?

How do we use XRegExp here? It's an amazing library but it doesn't optimize regexes.

LeaVerou:
> How do we use XRegExp here? It's an amazing library but it doesn't optimize regexes.

It allows building regexes from smaller regexes, may make it easier to subtract groups, and once we use it, it may also help simplify our grammars.
(I haven't personally used XRegExp so I may be off here about what it does; I just respect Steven Levithan's regexp expertise so I kind of assume it does everything :grin: but I'm fine with any other library too)

RunDevelopment:
> We just need a helper to subtract regexes, right? That must exist.

Are you sure about that? 
The closest thing I know is [regexp-tree and its optimizer API](https://github.com/DmitrySoshnikov/regexp-tree#using-optimizer-api) and even that doesn't optimize the lookaheads away. (However, their [transpiler API](https://github.com/DmitrySoshnikov/regexp-tree#using-compat-transpiler-api) could allow us features like named capturing groups (not at runtime), so it might be worth looking into even though it seems like their optimizer has a few problems with [correctness](https://github.com/DmitrySoshnikov/regexp-tree/issues/81).) 

There is surprisingly little tooling for regexes available in the JS ecosystem.

RunDevelopment:
I finally got around to fix all remaining cases of polynomial backtracking, so far. There are still relatively easy ways to improve the detection but I will do that after extracting the detection logic into its own project/package.

The performance problem remains. As I outlined before, regexes that prevent polynomial backtracking using lookarounds are typically slower than equivalent regexes that do not use lookarounds. Unfortunately, I'm still not aware of any tool that can rewrite JS regexes to obtain equivalent regexes without lookarounds. However, I recently created something similar, so creating such a tool might be a nice next project.

mAAdhaTTah:
May be worth refreshing this so the bot can poke its head in?

LeaVerou:
Wow, that's a cool bot! Is it new? 🎉

mAAdhaTTah:
@LeaVerou Yup! Added in #2608 & fixed in #2638 if you wanna check out the source. It's pretty easy to modify if we want it to support other things.

mAAdhaTTah:
So I'm supportive of this from a bundle size perspective. I have no idea what the performance impact is though or whether that's large enough to be concerned about / blocking.

RunDevelopment:
> I have no idea what the performance impact is though or whether that's large enough to be concerned about / blocking.

I'm currently moving the detection logic into its own project and improving it. I'm also implementing a simpler fixer that will try to fix common patterns causing polynomial backtracking. I'll release it in a day or two.

I'll do performance tests after I'm done.

RunDevelopment:
Perf test result:

<details>

```
Found 9 cases with 27 files in total.
Estimated duration: 4m 30s

c

  | PrismJS@master                            0.09ms ±  2%   80smp 1.07x
  | RunDevelopment@test-poly-backtracking     0.08ms ±  0%   85smp
  | PrismJS@master                            0.02ms ±  1%   86smp 1.02x
  | RunDevelopment@test-poly-backtracking     0.02ms ±  0%   89smp
  https://raw.githubusercontent.com/git/git/master/remote.c (66 kB)
  | PrismJS@master                            3.08ms ±  1%   84smp
  | RunDevelopment@test-poly-backtracking     3.13ms ±  1%   81smp 1.02x
  https://raw.githubusercontent.com/git/git/master/remote.h (11 kB)
  | PrismJS@master                            0.39ms ±  1%   85smp 1.03x
  | RunDevelopment@test-poly-backtracking     0.38ms ±  1%   89smp

css

  ../../assets/style.css (7 kB)
  | PrismJS@master                            1.02ms ±  1%   87smp 1.11x
  | RunDevelopment@test-poly-backtracking     0.92ms ±  1%   88smp

css!+css-extras (css)

  ../../assets/style.css (7 kB)
  | PrismJS@master                            1.57ms ±  1%   87smp 1.04x
  | RunDevelopment@test-poly-backtracking     1.51ms ±  1%   86smp

javascript

  ../../assets/utopia.js (11 kB)
  | PrismJS@master                            1.45ms ±  1%   87smp
  | RunDevelopment@test-poly-backtracking     1.55ms ±  1%   87smp 1.06x
  ../../components.json (27 kB)
  | PrismJS@master                            3.94ms ±  0%   79smp
  | RunDevelopment@test-poly-backtracking     4.05ms ±  1%   85smp 1.03x
  ../../package-lock.json (206 kB)
  | PrismJS@master                           20.50ms ±  1%   58smp
  | RunDevelopment@test-poly-backtracking    21.02ms ±  1%   56smp 1.03x
  | PrismJS@master                            3.80ms ±  1%   82smp
  | RunDevelopment@test-poly-backtracking     3.97ms ±  1%   83smp 1.04x
  | PrismJS@master                            3.49ms ±  1%   86smp
  | RunDevelopment@test-poly-backtracking     3.63ms ±  1%   81smp 1.04x
  | PrismJS@master                           36.11ms ±  1%   45smp
  | RunDevelopment@test-poly-backtracking    40.46ms ±  1%   40smp 1.12x
  | PrismJS@master                           21.33ms ±  1%   57smp
  | RunDevelopment@test-poly-backtracking    23.00ms ±  2%   52smp 1.08x

json

  ../../components.json (27 kB)
  | PrismJS@master                            3.21ms ±  1%   85smp
  | RunDevelopment@test-poly-backtracking     3.30ms ±  1%   85smp 1.03x
  ../../package-lock.json (206 kB)
  | PrismJS@master                           15.10ms ±  1%   64smp
  | RunDevelopment@test-poly-backtracking    15.92ms ±  1%   61smp 1.05x

markup

  ../../download.html (4 kB)
  | PrismJS@master                            0.43ms ±  1%   87smp
  | RunDevelopment@test-poly-backtracking     0.44ms ±  0%   90smp 1.02x
  ../../index.html (19 kB)
  | PrismJS@master                            2.24ms ±  2%   84smp
  | RunDevelopment@test-poly-backtracking     2.29ms ±  0%   84smp 1.02x
  https://github.com/PrismJS/prism (170 kB)
  | PrismJS@master                           20.90ms ±  1%   58smp
  | RunDevelopment@test-poly-backtracking    21.33ms ±  1%   57smp 1.02x

  ../../download.html (4 kB)
  | PrismJS@master                            0.68ms ±  1%   90smp
  | RunDevelopment@test-poly-backtracking     0.69ms ±  1%   88smp 1.02x
  ../../index.html (19 kB)
  | PrismJS@master                            2.60ms ±  0%   85smp
  | RunDevelopment@test-poly-backtracking     2.66ms ±  1%   83smp 1.02x
  https://github.com/PrismJS/prism (170 kB)
  | PrismJS@master                           21.69ms ±  1%   55smp
  | RunDevelopment@test-poly-backtracking    22.34ms ±  1%   55smp 1.03x

ruby

  https://raw.githubusercontent.com/rails/rails/master/actionview/lib/action_view/base.rb (11 kB)
  | PrismJS@master                            0.65ms ±  1%   88smp
  | RunDevelopment@test-poly-backtracking     0.65ms ±  1%   89smp 1.01x
  | PrismJS@master                            0.87ms ±  0%   90smp
  | RunDevelopment@test-poly-backtracking     0.88ms ±  1%   91smp 1.01x
  https://raw.githubusercontent.com/rails/rails/master/actionview/lib/action_view/template.rb (13 kB)
  | PrismJS@master                            0.85ms ±  1%   90smp
  | RunDevelopment@test-poly-backtracking     0.86ms ±  1%   91smp 1.01x

rust

  https://raw.githubusercontent.com/rust-lang/regex/master/src/compile.rs (42 kB)
  | PrismJS@master                            5.30ms ±  1%   83smp
  | RunDevelopment@test-poly-backtracking     5.36ms ±  1%   81smp 1.01x
  | PrismJS@master                            0.49ms ±  1%   90smp
  | RunDevelopment@test-poly-backtracking     0.49ms ±  0%   90smp 1.00x
  | PrismJS@master                            1.16ms ±  1%   90smp
  | RunDevelopment@test-poly-backtracking     1.17ms ±  1%   88smp 1.01x

summary
                                         best  worst  relative
  PrismJS@master                           22      5     1.00x
  RunDevelopment@test-poly-backtracking     5     22     1.01x
```

</details>

It's slightly slower (going by the number of best/worst), but not by significant amount.

RunDevelopment:
@joshgoebel You might be interested in my [scslre](https://github.com/RunDevelopment/scslre) project. It finds polynomial and exponential backtracking (within its limitations). It can also find [super-linear time complexity caused by the matching algorithm itself](https://rundevelopment.github.io/scslre/interfaces/movereport.html), but those reports are really hard (or impossible) to fix and [disabled in Prism](https://github.com/PrismJS/prism/pull/2597/files#diff-bb6f85478391410e40b7b3a072a8a190b40d1a0cc1bbd27f5889b40672d17228R523).

mAAdhaTTah:
Both bundle size + perf impacts are acceptable to me. I'll give the entire PR a look over either tonight or this week but it would be nice to get this in the next release.

tests/pattern-tests.js
```diff
@@ -504,4 +512,118 @@ function testPatterns(Prism) {
...
+						break;
+					}
+					case 'Move': {
+						rangeOffset = 1;
+						rangeStr = patternStr.substring(1, polyCase.quant.end + 1);
+						rangeHighlight = highlight([polyCase.quant]);
+						break;
+					}
+					default:
+						throw Error();
```
mAAdhaTTah, line 559: And maybe include a message? `"Should never happen :p"`

```suggestion
						throw new Error();
```

tests/pattern-tests.js
```diff
@@ -6,14 +6,22 @@ const { BFS, parseRegex } = require('./helper/util');
...
 const { JS, Words, NFA } = require('refa');
+const scslre = require('scslre');
 
 /**
- * A set of all safe (non-exponentially-backtracking) RegExp literals (string).
+ * A set of all safe (non-exponentially backtracking) RegExp literals (string).
  *
  * @type {Set<string>}
  */
 const safeRegexes = new Set();
```
mAAdhaTTah, line 16: Maaaaaybe?
```suggestion
const expoSafeRegexes = new Set();
```

mAAdhaTTah:
Couple of nitty suggestions but this looks great!

joshgoebel:
> You might be interested in my scslre project. It finds polynomial and exponential backtracking (within its limitations).

Is this then just all the work/effort you did in Prism factored out into a stand-alone library?

RunDevelopment:
> Is this then just all the work/effort you did in Prism factored out into a stand-alone library?

Not all of it, just the polynomial backtracking detection originally introduced by this PR.

mAAdhaTTah (approved):
lgtm!

joshgoebel:
@RunDevelopment I added this to our todo list. :)

PR #2601, opened 2020-10-23
Remove Julia constants that are not exported by default

knuesel:
This PR removed a few unexported constants for Julia. These constants are defined in the `Base.MathConstants` module, but they are unexported. So it should not be assumed that these names (without a module prefix) represent constants in Julia code.

To illustrate, I ended up here because I was wondering why the `e` looks funny on https://docs.julialang.org/en/v1/base/arrays/#Base.hvcat . It turns out PrismJS treats `e` as a Julia constant. This is wrong: by default, someone typing `e` in Julia will get the error `e not defined`).


RunDevelopment:
Thank for the pull request @knuesel!

The change seems reasonable. I guess you left `pi` and `im` because nobody uses them as variables?

The make the build pass, you also have to run `npm run build` and change [Julia's constant test](https://github.com/PrismJS/prism/blob/master/tests/languages/julia/constant_feature.test). More information on tests can be found [here](https://prismjs.com/test-suite.html).

knuesel:
@RunDevelopment sorry for missing that, should be fixed now.

I left `im`, `pi`, `π` and `ℯ` (which is different from `e`) because these ones are exported, so they are constants in a default setup.

RunDevelopment (approved):

PR #2608, opened 2020-10-25
Add Dangerfile and provide bundle size info

### LINKED ISSUE #1787: Add buildsize bot
I'd love to be able to see the impact on the output size of various PRs. E.g. #1776 has some minor improvements that suggest it could be good for the build size but would be nice to have a bot confirm that for us, especially as build size is something we're generally concerned about overall.

mAAdhaTTah:
Fixes #1787.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,0 +1,114 @@
+const { markdown } = require('danger');
+const fs = require('fs').promises;
+const gzipSize = require('gzip-size');
+const git = require('simple-git/promise')(__dirname).silent(true);
+
+// https://stackoverflow.com/questions/15900485/correct-way-to-convert-size-in-bytes-to-kb-mb-gb-in-javascript
+const formatBytes = (bytes, decimals = 2) => {
+	if (bytes === 0) return '0 Bytes';
+
+	const k = 1000;
+	const dm = decimals < 0 ? 0 : decimals;
+	const sizes = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+
+	const i = Math.floor(Math.log(Math.abs(bytes)) / Math.log(k));
+
+	return parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + ' ' + sizes[i];
+}
+
+const maybePlus = (from, to) => from < to ? "+" : "";
+
+const absDiff = (from, to) => {
+	if (from === to) {
+		return formatBytes(0);
+	}
+
+	return `${maybePlus(from, to)}${formatBytes(to - from)}`;
+}
+
+const percDiff =  (from, to) => {
+	if (from === to) {
+		return '0%';
+	}
+
+	return `${maybePlus(from, to)}${
+		(100 * (to - from) / (from || to)).toFixed(1)
+	}%`;
+}
+
+const getSummary = (rows, totalMasterFileSize, totalFileSize) => {
+	const numFiles = rows.length;
+	const maybeS = rows.length > 0 ? 's' : '';
+	const byteDiff = absDiff(totalMasterFileSize, totalFileSize);
+	const percentDiff = percDiff(totalMasterFileSize, totalFileSize);
+
+	return `A total of ${numFiles} file${maybeS} have changed, with a combined diff of ${byteDiff} (${percentDiff}).`;
+}
+
+const getChangedMinifiedFiles = async () => {
+	const result = await git.diff(['--name-only', '--no-renames', 'master...']);
+
+	return result
+		? result.split(/\r?\n/g).filter(file => file.endsWith('.min.js'))
+		: [];
+};
+
+const run = async () => {
+	// Check if master exists & check it out if not.
+	const result = await git.branch(['--list', 'master']);
+	if (result.all.length === 0) {
+		await git.branch(['master', 'origin/master']);
+	}
+
+	const minified = await getChangedMinifiedFiles();
+
+	if (minified.length === 0) {
+		markdown(`## No JS Changes`);
+		return;
+	}
+
+	const rows = [];
+	let totalFileSize = 0;
+	let totalMasterFileSize = 0;
+
+	for (const file of minified) {
+		const [fileContents, fileMasterContents] = await Promise.all([
+			fs.readFile(file, 'utf-8').catch(() => ''),
+			git.show([`master:${file}`]).catch(() => ''),
+		]);
+
+		const [fileSize, fileMasterSize] = await Promise.all([
+			gzipSize(fileContents),
+			gzipSize(fileMasterContents),
+		]);
+
+		totalFileSize += fileSize;
+		totalMasterFileSize +=fileMasterSize
+
+		rows.push([
+			file,
+			formatBytes(fileMasterSize),
+			formatBytes(fileSize),
+			absDiff(fileMasterSize, fileSize),
+			percDiff(fileMasterSize, fileSize),
+		]);
+	}
+
+	markdown(`## JS File Size Changes (gzipped)
+
+${getSummary(rows, totalMasterFileSize, totalFileSize)}
+
+<details>
+
+| file | master | pull | size diff | % diff |
+| --- | --- | --- | --- | --- |
+${rows.map(row => `| ${row.join(' | ')} |`).join('\n')}
+
+</details>
+`);
+}
+
+run().catch(err => {
+	console.error(err);
+	process.exit(1);
+});
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -1,0 +1,20 @@
+name: Danger
+
+on: 
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  run:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          fetch-depth: 0
+      - run: npm ci
+      - name: Danger
+        run: npx danger ci
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.npmignore
+++ b/.npmignore
@@ -13,6 +13,7 @@ tests/
 *.html
 bower.json
 composer.json
+dangerfile.js
 gulpfile.js
 .editorconfig
 .gitattributes
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,193 @@
       "integrity": "sha512-oJtNJCMFdIMwXGmx+KxuaD7i3b8uS7TTFYW/FNG2BT8m+fmGHoiPYoH0Pe3gya07WuFmM5FCDIr1x0irkD/hyA==",
       "dev": true
     },
+    "@babel/polyfill": {
+      "version": "7.12.1",
+      "resolved": "https://registry.npmjs.org/@babel/polyfill/-/polyfill-7.12.1.tgz",
+      "integrity": "sha512-X0pi0V6gxLi6lFZpGmeNa4zxtwEmCs42isWLNjZZDE0Y8yVfgu0T2OAHlzBbdYlqbW/YXVvoBHpATEM+goCj8g==",
+      "dev": true,
+      "requires": {
+        "core-js": "^2.6.5",
+        "regenerator-runtime": "^0.13.4"
+      }
+    },
+    "@octokit/auth-token": {
+      "version": "2.4.2",
+      "resolved": "https://registry.npmjs.org/@octokit/auth-token/-/auth-token-2.4.2.tgz",
+      "integrity": "sha512-jE/lE/IKIz2v1+/P0u4fJqv0kYwXOTujKemJMFr6FeopsxlIK3+wKDCJGnysg81XID5TgZQbIfuJ5J0lnTiuyQ==",
+      "dev": true,
+      "requires": {
+        "@octokit/types": "^5.0.0"
+      }
+    },
+    "@octokit/endpoint": {
+      "version": "6.0.8",
+      "resolved": "https://registry.npmjs.org/@octokit/endpoint/-/endpoint-6.0.8.tgz",
+      "integrity": "sha512-MuRrgv+bM4Q+e9uEvxAB/Kf+Sj0O2JAOBA131uo1o6lgdq1iS8ejKwtqHgdfY91V3rN9R/hdGKFiQYMzVzVBEQ==",
+      "dev": true,
+      "requires": {
+        "@octokit/types": "^5.0.0",
+        "is-plain-object": "^5.0.0",
+        "universal-user-agent": "^6.0.0"
+      },
+      "dependencies": {
+        "is-plain-object": {
+          "version": "5.0.0",
+          "resolved": "https://registry.npmjs.org/is-plain-object/-/is-plain-object-5.0.0.tgz",
+          "integrity": "sha512-VRSzKkbMm5jMDoKLbltAkFQ5Qr7VDiTFGXxYFXXowVj387GeGNOCsOH6Msy00SGZ3Fp84b1Naa1psqgcCIEP5Q==",
+          "dev": true
+        },
+        "universal-user-agent": {
+          "version": "6.0.0",
+          "resolved": "https://registry.npmjs.org/universal-user-agent/-/universal-user-agent-6.0.0.tgz",
+          "integrity": "sha512-isyNax3wXoKaulPDZWHQqbmIx1k2tb9fb3GGDBRxCscfYV2Ch7WxPArBsFEG8s/safwXTT7H4QGhaIkTp9447w==",
+          "dev": true
+        }
+      }
+    },
+    "@octokit/plugin-paginate-rest": {
+      "version": "1.1.2",
+      "resolved": "https://registry.npmjs.org/@octokit/plugin-paginate-rest/-/plugin-paginate-rest-1.1.2.tgz",
+      "integrity": "sha512-jbsSoi5Q1pj63sC16XIUboklNw+8tL9VOnJsWycWYR78TKss5PVpIPb1TUUcMQ+bBh7cY579cVAWmf5qG+dw+Q==",
+      "dev": true,
+      "requires": {
+        "@octokit/types": "^2.0.1"
+      },
+      "dependencies": {
+        "@octokit/types": {
+          "version": "2.16.2",
+          "resolved": "https://registry.npmjs.org/@octokit/types/-/types-2.16.2.tgz",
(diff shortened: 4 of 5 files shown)


mAAdhaTTah:
This is fairly rudimentary but basically does the thing. I can add additional comparisons to the table, if desired, adding comparisons to the source files and/or adding comparisons to non-gzipped versions.

The reason I didn't use a lib like `bundlesize` is they enforce a max per file or glob checked and I don't think the purposes here is to enforce a max so much as provide additional information for evaluating adding few features.

Open to comments/suggestions, but this is ready for review!

dangerfile.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+const percDiff =  (first, second) =>
+	`${(100 * Math.abs( (first - second) / ( (first+second)/2 ) )).toFixed(2)}%`
+
+const run = async () => {
+	const minified = danger.git.modified_files.filter(file => file.includes('.min.js'))
+
+	if (minified.length === 0) {
+		return markdown(`# No JS Changes`);
```
RunDevelopment, line 27: This is very misleading.

Also, if there is nothing to report, do have to make a message at all?

```suggestion
		markdown(`# No JS Changes`);
		return;
```

RunDevelopment (changes requested):
Looking good but there are still some things to improve.

Could you please add a summary and the absolute change? It's nice that know the relative change for each file, but the absolute changes (per file and in total) are also interesting.

dangerfile.js
```diff
@@ -0,0 +1,67 @@
+import {message, danger, markdown} from "danger"
+
+const fs = require('fs').promises;
```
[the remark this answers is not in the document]
mAAdhaTTah: I think I did this cuz I have some weird issues w importing from something other than `danger` but that might have been a relic of some earlier explorations.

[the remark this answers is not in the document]
mAAdhaTTah: Yeah, ok, so I still get it (see the updated comment). I guess I could change the `danger` import to CJS if we want consistency.

dangerfile.js
```diff
@@ -0,0 +1,71 @@
...
+	const minified = danger.git.modified_files.filter(file => file.includes('.min.js'))
+
+	if (minified.length === 0) {
+		return markdown(`# No JS Changes`);
+	}
+
+	// Check if master exists & check it out if not.
+	const result = await git.branch(['--list', 'master']);
+	if (result.all.length === 0) {
+		await git.branch(['master', 'origin/master']);
```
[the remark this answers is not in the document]
mAAdhaTTah: Oh good question, will need to test that.

dangerfile.js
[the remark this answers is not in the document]
mAAdhaTTah: Gonna remove the commit, but here's the stack trace:

```
Cannot use import statement outside a module
dangerfile.js:2
import fs from 'fs/promises';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at wrapSafe (internal/modules/cjs/loader.js:931:16)
    at Module._compile (internal/modules/cjs/loader.js:979:27)
    at Object.requireFromString [as default] (/home/runner/work/prism/prism/node_modules/require-from-string/index.js:28:4)
    at Object.<anonymous> (/home/runner/work/prism/prism/node_modules/danger/distribution/runner/runners/inline.js:144:63)
    at step (/home/runner/work/prism/prism/node_modules/danger/distribution/runner/runners/inline.js:32:23)
    at Object.next (/home/runner/work/prism/prism/node_modules/danger/distribution/runner/runners/inline.js:13:53)
    at /home/runner/work/prism/prism/node_modules/danger/distribution/runner/runners/inline.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (/home/runner/work/prism/prism/node_modules/danger/distribution/runner/runners/inline.js:3:12)
    at exports.runDangerfileEnvironment (/home/runner/work/prism/prism/node_modules/danger/distribution/runner/runners/inline.js:105:136)
```

dangerfile.js
```diff
@@ -0,0 +1,71 @@
...
+const gzipSize = require('gzip-size');
+const git = require('simple-git/promise')(__dirname);
+
+// https://stackoverflow.com/questions/15900485/correct-way-to-convert-size-in-bytes-to-kb-mb-gb-in-javascript
+function formatBytes(bytes, decimals = 2) {
+	if (bytes === 0) return '0 Bytes';
+
+	const k = 1024;
+	const dm = decimals < 0 ? 0 : decimals;
+	const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
```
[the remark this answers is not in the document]
mAAdhaTTah: I guess I have mixed feelings. While `KiB` is technically more accurate, I don't know how often I see `KiB`. If we want technical accuracy, maybe we should switch `k = 1000`?

dangerfile.js
mAAdhaTTah: We don't _have_ to, but it does make it explicit that there is no change (and that Danger ran successfully and found nothing worth reporting).

mAAdhaTTah: Not sure I understand the "misleading" bit tho.

dangerfile.js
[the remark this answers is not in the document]
RunDevelopment: OOF. Seems like Danger does some import witchcraft under the hood. Yeah, let's not touch this and just use the mixed import style.

dangerfile.js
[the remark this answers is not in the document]
RunDevelopment: Switching to k=1000 is also ok.

dangerfile.js
```diff
@@ -0,0 +1,71 @@
...
+	let exitCode = 0;
+	try {
+		await run();
+	} catch (err) {
+		console.error(err);
+		exitCode = 1;
+	}
+
+	process.exit(exitCode);
+})();
```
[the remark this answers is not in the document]
mAAdhaTTah: I borrowed this code from like... Jest or React or something that used DangerJS for bundle size checking. I think it had cleanup logic though, now that I think about it, so we can go with the simpler version.

dangerfile.js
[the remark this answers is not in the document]
mAAdhaTTah: I changed it to `require` for everything for consistency, which does work.

dangerfile.js
```diff
@@ -0,0 +1,71 @@
...
+	if (result.all.length === 0) {
+		await git.branch(['master', 'origin/master']);
+	}
+
+	const rows = [];
+
+	for (const file of minified) {
+		const [fileContents, fileMasterContents] = await Promise.all([
+			fs.readFile(file, 'utf-8'),
+			git.show([`master:${file}`]),
```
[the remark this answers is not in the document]
mAAdhaTTah: Same, didn't test that.

dangerfile.js
RunDevelopment: > it does make it explicit that there is no change

That's a good point. Let's keep it then.

> Not sure I understand the "misleading" bit tho.

Yeah, I kinda changed the subject in the second sentence... With `return markdown(...)`, I expect that the returned value of `run` will be used which isn't the case. The return type of the `run` function is `void`. Please don't save a line by returning the result of `markdown(...)`.

RunDevelopment:
Hmmm. Did the file size change? I'm pretty sure `prism-core.min.js` was at 2.93KB (both master and pull) when I first saw the message.

mAAdhaTTah:
@RunDevelopment Yes, we went from 1024 -> 1000 for the calculation to make KB accurate.

RunDevelopment:
Ahh, yeah. I'm stupid. Maybe I could also remember what I say and agree to. That'd be great.

mAAdhaTTah:
lol don't be so hard yourself.

dangerfile.js
mAAdhaTTah: Fair enough. Fixed that.

mAAdhaTTah:
Ok, so need to do some testing w/ deleted files & remote forks, but before I dive into that, how does the output look?

dangerfile.js
```diff
@@ -0,0 +1,93 @@
...
+	return `${maybePlus(from, to)}${formatBytes(to - from)}`;
+}
+
+const percDiff =  (from, to) => {
+	if (from === to) {
+		return '0%';
+	}
+
+	return `${maybePlus(from, to)}${
+		(100 * (to - from) / ((from + to) / 2 )).toFixed(2)
```
RunDevelopment, line 35: I think that we should round the percentages to whole numbers. Sub-percent changes don't matter anyway.
```suggestion
		(100 * (to - from) / ((from + to) / 2 )).toFixed()
```

RunDevelopment:
CSS files rarely change, so I don't think that's necessary. Unminified files shouldn't be included either (otherwise I will remove all comments to keep those file sizes down 😄). 
Your initial approach (all `.min.js` files) seemed fine to me.

Regarding the output, I think it's fine. 
A summary would be nice if more than 1 minified file changes. Though it might not be a good idea to just add up the gzipped differences. GZip will likely optimize between files in a real bundle (I'm thinking one file with all languages). I don't know what the best solution for this is.

dangerfile.js
mAAdhaTTah: I disagree with this a bit. I think we're likely to see changes be in the 1% range where I'd like to know the % difference with a bit more granularity, so I'd like to keep the 2 decimal places.

mAAdhaTTah:
> A summary would be nice if more than 1 minified file changes.

Do you have something in mind for this? What sort of details would you want in a summary?

> Though it might not be a good idea to just add up the gzipped differences. GZip will likely optimize between files in a real bundle (I'm thinking one file with all languages). I don't know what the best solution for this is.

Yeah, that makes sense. The goal isn't strictly to provide exact figure for all these use cases but to give us info on the tradeoffs we're making. Relatedly, when we get the benchmark tests in (I'm going to take a run through the open PRs over the coming week), we can add those here for some additional info.

dangerfile.js
RunDevelopment: Ok, but let's turn it down to only 1 decimal digit. Most of our files are 1 to 2KB gzipped, so 0.01% represents 0.1 to 0.2 bytes.

RunDevelopment:
> > A summary would be nice if more than 1 minified file changes.
> 
> Do you have something in mind for this? What sort of details would you want in a summary?

Nothing grand. Just the number of relevant files changed and the total number of bytes saved/gained.

> The goal isn't strictly to provide exact figure for all these use cases but to give us info on the tradeoffs we're making.

In that case, we might as well just add up the gzipped differences for now and replace it with something better later.

> Relatedly, when we get the benchmark tests in [...], we can add those here for some additional info.

Yeah no. The benchmark in its current form takes about 2 to 5 minutes (depends on the configuration) to run after all files have been downloaded. I don't want us to wait that long.

But we could make the benchmark start via a manual trigger. Idk, make a 🚀 reaction to the comment or something.

---

Something I hope to implement with this is what I said in #1758. If the PR changes the language definitions, the bot will post a link to the test page, so you can just try out the changes.

I also really like the [example with test files here](https://danger.systems/js/guides/the_dangerfile.html#working-with-files) since we have the same situation. I'll probably make a few PRs after this one is merged.

dangerfile.js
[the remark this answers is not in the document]
RunDevelopment: I made a little test PR (#2610) and `git remote -v` returned
```
origin	https://github.com/PrismJS/prism (fetch)
origin	https://github.com/PrismJS/prism (push)
```

So this will work even if the PR source is from a fork.

mAAdhaTTah:
> The benchmark in its current form takes about 2 to 5 minutes (depends on the configuration) to run after all files have been downloaded.

FWIW, given the amount of time that tends to elapse between reviews, that's may not be a huge deal, esp if we split between two different runs. But we don't have to decide now.

RunDevelopment:
> esp if we split between two different runs

That's a good idea. I thought you wanted to display the results of the benchmark in the Danger comment because I want this comment to appear/update as quickly as possible.

> But we don't have to decide now.

That.

Let's try to get this in now and add more features in small PRs later. There are only two things that still have to be done:

1. Testing added/delete files. (You can probably just rename any minified file for this.)
1. A small summary.

mAAdhaTTah:
Summary is added.

mAAdhaTTah:
Ok, I think we're looking pretty good but let me know what you think!

dangerfile.js
```diff
@@ -0,0 +1,118 @@
...
+			gzipSize(fileMasterContents),
+		]);
+
+		totalFileSize += fileSize;
+		totalMasterFileSize +=fileMasterSize
+
+		rows.push([
+			file,
+			formatBytes(fileSize),
+			formatBytes(fileMasterSize),
```
RunDevelopment, line 100: Wrong order.
```suggestion
			formatBytes(fileMasterSize),
			formatBytes(fileSize),
```

RunDevelopment:
I'm pretty sure the renaming-solution you implemented won't work. 
Git's renaming detection is pretty good. It will even detect renames in files that *marginally* changed. This means that we can't assume that the content of the original and renaming file are the same. This is relevant because the diff in the summary will be wrong if we make this assumption.

In your test case, git wasn't able to detect the rename because you tested with minified files. Git's diff algorithm is line-based, so if you rename + change a one-line file, git won't be able to detect the rename. Please rename a file without changing it. Your current test case is just a deleted and an added file as far as git is concerned.

The problem with renames is that we don't know the name of the original file when using Danger. Danger only adds the renamed file in `modified_files`. You can't get the original file name. In #2616, I used git's API directly just so that I could disable git's renaming detection.

mAAdhaTTah:
You're right in the abstract, but I was going for a best-guess solution that will produce a somewhat-valid output instead of erroring. I don't think we're going to end up renaming minified files that often, so I don't think we need to maintain a bunch of logic to make sure it's perfect. We could display a warning saying "hey whoops this was renamed, we don't know" but I think it'll be pretty clear that it'll look kinda like a new file but the diff is a rename.

RunDevelopment:
> maintain a bunch of logic

So you agree to use the correct solution ;)

I mean, using git directly sounds scary but it's just 2 lines to get all changed files without renames.

```js
const result = await git.diff(['--name-only', '--no-renames', 'master...']);
return result ? result.split(/\r?\n/g) : [];
```

These two lines mean that we don't have to worry about renames all. Renamed files are just one deleted file and one added file. Getting the file content is simple:

```js
const [fileContents, fileMasterContents] = await Promise.all([
	fs.readFile(file, 'utf-8').catch(() => null),
	git.show([`master:${file}`]).catch(() => null),
]);
```

That's a lot simpler than the current error-handling solution, isn't it?

mAAdhaTTah:
Fair enough. Implemented that. Kept it `catch(() => '')` and added `silent` to `simplegit` cuz the `git show` was erroring to the console.

mAAdhaTTah:
Alright, I think that gets us there.

dangerfile.js
```diff
@@ -0,0 +1,120 @@
...
+const getSummary = (rows, totalMasterFileSize, totalFileSize) => {
+	const numFiles = rows.length;
+	const maybeS = rows.length > 0 ? 's' : '';
+	const byteDiff = absDiff(totalMasterFileSize, totalFileSize);
+	const percentDiff = percDiff(totalMasterFileSize, totalFileSize);
+
+	return `A total of ${numFiles} file${maybeS} have changed, with a combined diff of ${byteDiff} (${percentDiff}).`;
+}
+
+const handleRenamedError = (file, type) => async err => {
```
RunDevelopment, line 48: Unused function.

dangerfile.js
```diff
@@ -0,0 +1,120 @@
...
+	return `${maybePlus(from, to)}${formatBytes(to - from)}`;
+}
+
+const percDiff =  (from, to) => {
+	if (from === to) {
+		return '0%';
+	}
+
+	return `${maybePlus(from, to)}${
+		(100 * (to - from) / (from || to)).toFixed(1)
```
RunDevelopment, line 35: > from || to

How about we use a [proper formula](https://en.wikipedia.org/wiki/Relative_change_and_difference#Formulae)?

```suggestion
		(100 * (to - from) / Math.max(from, to)).toFixed(1)
```

dangerfile.js
```diff
@@ -0,0 +1,120 @@
...
+
+	// Otherwise, rethrow, cuz we haven't handled any other errors.
+	throw err;
+}
+
+const getChangedMinifiedFiles = async () => {
+	const result = await git.diff(['--name-only', '--no-renames', 'master...']);
+
+	return result
+		? result.split(/\r?\n/g).filter(file => file.includes('.min.js'))
```
RunDevelopment, line 63:
```suggestion
		? result.split(/\r?\n/g).filter(file => file.endsWith('.min.js'))
```

dangerfile.js
```diff
@@ -0,0 +1,120 @@
...
+		]);
+	}
+
+	markdown(`${getSummary(rows, totalMasterFileSize, totalFileSize)}
+
+## JS File Size Changes (gzipped)
+
+| file | master | pull | size diff | % diff |
+| --- | --- | --- | --- | --- |
+${rows.map(row => `| ${row.join(' | ')} |`).join('\n')}`);
```
RunDevelopment, line 114: 1. The summary is of the changed JS files, so it should be under that header.
1. I really think that the full list of changed files should be collapsed by default. This list can be really long for some (rare) PRs and will only contain one item for most (in which case the summary already contains the size change).
```suggestion
	markdown(`## JS File Size Changes (gzipped)

${getSummary(rows, totalMasterFileSize, totalFileSize)}

<details>

| file | master | pull | size diff | % diff |
| --- | --- | --- | --- | --- |
${rows.map(row => `| ${row.join(' | ')} |`).join('\n')}

</details>
`);
```

RunDevelopment (changes requested):
Just some minor nits and then I think we're done.

Also, please add this file to `.npmignore`. (Or maybe we could add a `files` section to `package.json` and ditch `.npmignore` instead? Might be easier looking forward.)

dangerfile.js
mAAdhaTTah: We're not looking for the larger of the two numbers here. We're using `||` because we want fallback should `from` be `0` but otherwise, we always want `from`.

mAAdhaTTah:
Ok, I've added your last suggestions, removed the failures, and squashed everything down. Are we good to go?

RunDevelopment (approved):

RunDevelopment:
@mAAdhaTTah This is something I wanted to ask: Why do you always squash the commit? I always "squash and merge" on GitHub anyway.

mAAdhaTTah:
@RunDevelopment By default, when GitHub squash & merges, the commit message is going to be a list of all the commits squashed together, which I don't really like as a commit message. I could rewrite the message as I merge, but in the past, I've forgotten to do that, and since I'm thinking about the commit message when I'm making the commit, I prefer to combine them into the one commit, with the message that I want, while I'm committing.

In this case in particular, because there was stuff in the commits I wanted to remove, I actually rolled back all of the commits locally and readded the commit with just the real changes.

Issue #2617, opened 2020-11-03
Safe usage with static imports without Babel


RunDevelopment:
**Motivation**
A lot of people seem to use Prism via ES6 [import declarations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and get it wrong. 

Each language definition has a list of languages it requires and optional dependencies. This information is an implementation detail and hidden from our users but also __required__ to make static imports work.

Our current solution for this problem is [the Babel plugin](https://github.com/mAAdhaTTah/babel-plugin-prismjs). However, users may not want or have the option to use this plugin, so I think that we should also offer an alternative solution.

**Description**
I really don't know how to solve this in a robust manner since we essentially have to _control the user's source code_.

Maybe a few ideas:

- A Prims CLI interface that generates an import file like this:
   `npx prismjs import-file --languages=html,php,java --js-out=./src/prism-importer.js`
    (Users will then import the generated file instead of our package.)

    The generated file will look like this:
    ```js
    import Prism from "prismjs/components/prism-core.min.js";
    import "prismjs/components/prism-markup.min.js";
    import "prismjs/components/prism-clike.min.js";
    import "prismjs/components/prism-markup-templating.min.js";
    import "prismjs/components/prism-php.min.js";
    import "prismjs/components/prism-java.min.js";
    export default Prism;
    ```

    The problem with this approach is that the generated file is __version-specific__ and may silently fail. The easiest solution for this is to include a version check in the generated file but I'm not sure how to do this efficiently.

- Another idea is to have the npm package include a Prism bundle that includes all languages to eliminate the need for multiple static imports. This completely throws away Prism's modularity but it's an easy-to-use solution for people who just want to try out Prism or don't care too much about bundle size.
   Also, this doesn't help with plugins since they have the same dependency problem.

joshgoebel:
> Another idea is to have the npm package include a Prism bundle that includes all languages ... easy-to-use solution for people who just want to try out Prism or don't care too much about bundle size.

It probably also works great for many on the server-side if they aren't paranoid about application start-up time since Node.js is quite fast at parsing code. Our Highlight.js NPM package distributes an `index.js` that loads EVERY language and our README includes instructions showing using it in that fashion or (for the more selective people) loading individual grammars.

[...]

Issue #2146, opened 2019-12-25
API for supported languages/plugins/etc.


RunDevelopment:
**Motivation**
Libraries sometimes need to know all languages/plugins/etc. Prism supports.

**Description**
A simple API to:
1. Check whether a given language/plugin/etc is supported.
2. Query all languages/plugins/etc.

My suggestion is to make this a part of <code>components<b>.js</b></code>. Something like:

```ts
components.supported(lang: string | string[]): boolean;
components.supported.languages: string[];
components.supported.plugins: string[];
```

The other properties of the `components` object are still private to Prism. By implementing the API in `components.js`, both browser and NodeJS devs have easy access to it.

**Alternatives**
Document `components.json` and declare it to be a part of Prism's public API. Right now, it's considered a private part of Prism, so we _could_ make it public.
This has the obvious downside that changes to `components.json` might be breaking and it takes a lot of freedom from Prism's core development team.

PR #2618, opened 2020-11-04
CLI bundler

### LINKED ISSUE #2404: Automatically downloaded customised files
**Motivation**
The feature where you can minify the code by selecting which languages you need support for before clicking 'download' is nice, but it would be great to do this automatically (e.g. to add to build tools).

Clicking the relevant checkboxes on the website generates a URL like `https://prismjs.com/download.html#themes=prism-funky&languages=markup+css+clike+javascript+haskell+protobuf+purebasic`

And we can evidently automatically create this URL. But then the "Download JS" and "Download CSS" buttons have to be pressed manually to trigger the actual download.

**Description**
And endpoint or similar where I could download the CSS and JS files from a customised URL directly e.g. 

GET https://prismjs.com/download.html#themes=prism-funky&languages=markup+css+clike+javascript+haskell+protobuf+purebasic?download=js

GET https://prismjs.com/download.html#themes=prism-funky&languages=markup+css+clike+javascript+haskell+protobuf+purebasic?download=css

**Alternatives**
Visiting the website manually or trying to reverse engineer the javascript that triggers the download button.

RunDevelopment:
This resolves #2404.

I added a small CLI bundler that is functionally equivalent to the download page. It can be used like this:
`npx prismjs bundle --languages=html,java,php --theme=coy --js-out=prism.js --css-out=prism.css`

API of the CLI:
```
PS C:\Users\micha\Git\prism> node .\bin\cli.js bundle --help
prismjs bundle

Create JS and CSS bundles for a specific configuration of
languages/plugins/themes.

Options:
      --version    Show version number                                 [boolean]
      --help       Show help                                           [boolean]
  -c, --config     The path of a JS/JSON configuration file that specifies the
                   languages, plugins, and theme of the bundle. The file has to
                   have the following format:
                   | {
                   |   // a list of languages (e.g. ["html", "php", "java"])
                   |   languages?: "all" | string[],
                   |   // a list of plugins (e.g. ["line-number", "toolbar"])
                   |   plugins?: "all" | string[],
                   |   // a theme id (e.g. "default", "coy")
                   |   theme?: string,
                   |   minified?: boolean,
                   | }                                                  [string]
      --languages  A list of languages to includes. Languages are separated by
                   comma (e.g. html,php,java). Use "--languages=all" to include
                   all languages.                                       [string]
      --plugins    A list of plugins to includes. Plugins are separated by comma
                   (e.g. line-numbers,toolbar). Use "--plugins=all" to include
                   all plugins.                                         [string]
      --theme      The name of the theme. If no theme is specified, Prism's
                   default theme will be used.                          [string]
      --minified   Whether languages and plugins should be minified. Use
                   --minified=false for development files.
                                                       [boolean] [default: true]
      --js-out     The output JS file(s) of the generated bundle. Existing files
                   will be overwritten.                                  [array]
      --css-out    The output CSS file(s) of the generated bundle. Existing
                   files will be overwritten.                            [array]
```

---

Aside from Prism's files, it has 2 additional dependencies:
1. yargs is used to parse arguments and handle documentation.
1. gzip-size is used to determine the compressed bundle size.

This is a problem. I incorrectly installed them as dev deps for now so that Prism still has 0 hard dependencies. I don't know what the best course of action for this is. 

There's also the question where we even want a CLI for Prism. (I think it would be nice.)

Some input would be nice.


siavashs:
@RunDevelopment is it possible to support both development and minified versions as output?

mAAdhaTTah:
> This is a problem. I incorrectly installed them as dev deps for now so that Prism still has 0 hard dependencies. I don't know what the best course of action for this is.

We could bundle up the cli into a single file with something like [ncc](https://github.com/vercel/ncc].

> There's also the question where we even want a CLI for Prism. (I think it would be nice.)

I'm not opposed per se but it is code we'll have to maintain for questionable value. You think there are a lot of people who'd find this useful? It's not a huge savings over downloading it from the site, and if you're going to run this from the command line several times, you'd be better off just installing it from npm and bundling it, I think.

RunDevelopment:
> We could bundle up the cli into a single file with something like ncc.

That looks REALLY good!

> You think there are a lot of people who'd find this useful?

Hard to say. It does make some things easier (e.g. updating Prism bundles) but idk if people will use it.

> It's not a huge savings over downloading it from the site, and if you're going to run this from the command line several times, you'd be better off just installing it from npm and bundling it, I think.

The "bundling it" part is the problem. Bundlers don't know in which order Prism's files have to be loaded because their dependencies are declared in `components.{json,js}` and not in the files themselves. We wouldn't need the download page and the babel plugin if it was that easy...

RunDevelopment:
I used ncc. We now have a `bin/src` folder that contains the CLI logic and a `bin/dist` folder that contains the bundled up CLI logic. `bin/dist` is ignored by git and `bin/src` is ignored by npm. The new `prepublishOnly` task will ensure that we won't accidentally publish a release without the CLI.

mAAdhaTTah:
> We wouldn't need the download page and the babel plugin if it was that easy...

My thinking is, as a PrismJS consumer, if I'm going to download Prism as a one-off, I will use the download page, but if I'm going to need to keep it updated over time, I'm going to use a bundler + npm and can use the babel plugin. I feel like the CLI sits in a space between "spinning up a quick project to play around with" (use the download page) and "building a serious application" (use the babel plugin) that isn't going to get much use.

RunDevelopment:
How about "building a serious application without babel"? We already had issues from people that don't want to use babel. CLI is a nice solution for those people.

But it's not like I don't get what you mean. In the worst case, this CLI will be something that gets basically no usage but we still have to maintain it. The reason why I want to somewhat push CLI is that it could be a potential solution for issues like #2617.

mAAdhaTTah:
> How about "building a serious application without babel"?

This could very well just be my bias, but... does anyone do this? Even looking at the "people aren't importing things correctly" issue, the user would be using ESModules, and I don't know how many people are using ESModules for serious applications without a bundler, at which point including babel is fairly straightforward.

RunDevelopment:
Well, with ES6+ features and ESModules coming close to being universally supported, there are fewer and fewer reasons to use Babel with each passing year. Projects with TypeScript don't need Babel. Projects with bundlers don't need Babel. 

I think that the main reason Babel is still so popular is that there is a small framework called React with nice custom syntax. It's the only reason I use Babel but I mainly do things in TypeScript, so things might be different on the JS side.

mAAdhaTTah:
> there are fewer and fewer reasons to use Babel with each passing year.

Given that the language will continue to evolve faster than the long tail of browsers will add support, I don't think this is true. My team could not easily figure out the exact set of syntax features we can use given our intended browser matrix. However, we can easily configure babel to either figure that out for us or transpile everything down to es5 lowest common denominator (which will be the case for any team that still supports IE11). This allows us to adopt new JS features as they get added to the language rather than when they're adopted by all our supported browsers. I expect this to be the case for a large number of teams for quite a while, React projects or otherwise, and I think this will be a feature of the ecosystem for the foreseeable future.

> Projects with bundlers don't need Babel.

This could be true, but I think at the point where you've got a bundler set up, and you want to integrate Prism, is it easier to integrate babel + the plugin or to integrate the CLI? I don't think configuring babel is significantly more work than setting up most bundlers in the first place (e.g. [parcel is zero-config for babel](https://parceljs.org/javascript.html#babel)), and you get to maintain your version through a system you're already using (package.json) rather than adopting an external mechanism for keeping your Prism version up-to-date (CLI).

---

Another thought: there's also a bit of ecosystem complexity involved in providing a third mechanism for consuming Prism. Instead of comparing the download page + babel plugin, you also have to consider the CLI and decide between them.

RunDevelopment:
I find your "ecosystem complexity" argument convincing. While I wouldn't mind third-part tooling (is the babel plugin thrid-party? Are you a third party @mAAdhaTTah???????? ;) ), I also think that it would be a bit of a problem if Prism provided X many ways to do essentially the same thing.

The problem is that `components.json`, the file that contains all the information needed to create bundles, [isn't part of Prism's public API](https://github.com/PrismJS/prism/pull/1998#issuecomment-565865896). If it was public, somebody could just make a package, call it `prismjs-cli`, and have the CLI in there. There already are a few libraries out there that use `components.{js,json}` because they need dependency information or just because they want to list all available languages. 

Maybe we should move towards stabilizing `components.{js,json}`? Right now, they contain dependency information but also stuff that is specific to the website. With stable and public `components.{js,json}`, there will be no need to have a CLI in this package.

mAAdhaTTah:
> is the babel plugin thrid-party? Are you a third party @mAAdhaTTah???????? ;

lol but on a serious note, I've actually wondered if it makes more sense to move the babel plugin into the PrismJS org, considering it's officially recc'd in the docs.

> There already are a few libraries out there that use components.{js,json} because they need dependency information or just because they want to list all available languages.

The babel plugin is one of those, and we created an API for that (`getLoader`). Right now, it's also using the `component.js` (even those it's technically private/not stable). Maybe it makes sense to expand `dependencies.js` to provide an API that allows us to implement both what the babel plugin + cli need?

I don't necessarily mind stabilizing `components.json` but we've seen some changes in that over the lifetime of the project and how we handle dependencies is a bit of a pain point so I kind of prefer having that flexibility in the internal representation.

RunDevelopment:
> I've actually wondered if it makes more sense to move the babel plugin into the PrismJS org, considering it's officially recc'd in the docs.

I don't really have strong opinions on this. The babel plugin would appear more official in the organization, I guess, but that's about the only argument I can see for moving it.

> Maybe it makes sense to expand dependencies.js to provide an API that allows us to implement both what the babel plugin + cli need?

That is probably the best solution but the name `dependencies.js` might be a bit misleading then. The minimum set of features I need to implement the CLI are:

- List all languages and plugins. (#2146)
- Get the id of the default theme (or have its name stabilized). 
- Get whether a component (lang/plugin/theme/etc.) has CSS and/or JS files.
- Get the minified/dev file paths of the CSS/JS files of a component.

(Looking at third-party libraries, it would also be good to be able to get the title and aliases of components as well.)

IMO, implementing those features as free functions in `dependencies.js` would be suboptimal. I'd rather have a class that takes the `components.json` data in the constructor. Something like: 
```js
new Components(require('components.json')).listAll("languages") // => ["markup", "css", ...]
```

One problem with this class is that the `getLoader` functionality should be part of the class but deprecating `getLoader` after only how-many releases is a bit...

mAAdhaTTah:
> I don't really have strong opinions on this. The babel plugin would appear more official in the organization, I guess, but that's about the only argument I can see for moving it.

Yeah, there's an "official blessing" aspect to it joining the org proper. It's not super important, so we can revisit it later.

> IMO, implementing those features as free functions in `dependencies.js` would be suboptimal. I'd rather have a class that takes the components.json data in the constructor.

Could you elaborate on why? I think if we're requiring consumers to pass `components.json` in via the constructor, we're kind of treating `components.json` as "public" in that it's something we're expecting consumers to interact with. I don't have a strong opinion on "class" vs "free functions" (although I tend to prefer functional approaches in my work) but requiring passing in the data structure I'm less sure about.

RunDevelopment:
>  if we're requiring consumers to pass components.json in via the constructor, we're kind of treating components.json as "public"

Uhmm. We are currently doing exactly that with `getLoader`... The current state with `getLoader` is that the function needs `components.json` to work but the user has to pass that data in. We only guarantee that `getLoader` can work with the data from `components.json` and nothing more. `components.json` may contain whatever.

We could also use `components.js` to hold the functionality we need.

mAAdhaTTah:
lol well guess I can toss that principle out.

Can we hang the functionality we want off that `getLoader` implementation? I know its API is not _great_, and I don't mind considering whatever implementation as a source of tech debt we'd pay off in v2, but I would def prefer avoiding multiple implementations that do similar things.

RunDevelopment:
The idea was to move `getLoader` into the new class/set of functions. The current `getLoader` implementation already does a lot the other functions will also have to do, so having everything in one place seems like a good idea.

---

Also, I just thought of a little trick to implement all of this in a backward-compatible way. The __public__ interfaces of `components.js` and `dependencies.js` looks like this:

```js
// components.js
var components = {}; // something
module.exports = components; 
```

```js
// dependencies.js
function getLoader(components, load, loaded) { ... }
module.exports = getLoader;
```

With an intended usage like this:

```js
const components = require('components.js');
const getLoader = require('dependencies.js');

const loader = getLoader(components, load, loaded);
```

How about we change it to this:

```js
// components.js
class Components {
  getLoader(load, loaded) { ... }
  // other functions (e.g. listAll, getTitle, etc.)
}
var components = new Components({ /* the JSON from `components.json` */ });
module.exports = components; 
```

```js
// dependencies.js
function getLoader(components, load, loaded) { 
  console.warn('Deprecated! Use `components.getLoader` instead.');
  return components.getLoader(load, loaded);
}
module.exports = getLoader;
```

RunDevelopment:
An ESM-based module system for languages and plugins is planned for v2 completely eliminating the need for this.

I'll close this now.

PR #2628, opened 2020-11-10
Add PromQL language support

arendjr:
This PR adds support for PromQL: https://prometheus.io/docs/prometheus/latest/querying/examples/


RunDevelopment (changes requested):
Thank you for the PR @arendjr!

I gave your PR a quick look and left you some comments. I'll give it a more in-depth review later.

Additional notes:
- We only use single-quoted strings
- Please quote all token names (e.g. `'comment'`, `'range-duration'`, etc) (like [here](https://github.com/PrismJS/prism/blob/master/components/prism-javascript.js)).
- Please use an IIFE to fix global namespace pollution (like [here](https://github.com/PrismJS/prism/blob/9df20c5ecef9a64bad180c05e7b32b750099026b/components/prism-protobuf.js#L1)).

components/prism-promql.js
```diff
@@ -0,0 +1,173 @@
...
+			lookbehind: true,
+			inside: {
+				"range-duration": {
+					pattern: /\b\d+[smhdwy]\b/i,
+					alias: "number",
+				},
+			},
+		},
+	],
+	number: /\b-?\d+(?:(?:\.\d*)?(?:[eE][+-]?\d+)?)?\b/,
```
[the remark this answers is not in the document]
arendjr: Good catch! I just used the Kausal blog post as a starting point.

arendjr:
Thanks for the thorough feedback already! I have addressed the styling issues, split keywords and functions as appropriate, simplified some things and added another comment to address one of your questions :) To be honest, I'm not sure about all the original motivations, as I reused some code from the Monaco highlighter and the blog post I referenced, but thanks to your review, I think it's pretty cleaned up now.

examples/prism-promql.html
```diff
@@ -0,0 +1,18 @@
+<h2>Examples</h2>
+<pre><code>
+# These examples are taken from: https://prometheus.io/docs/prometheus/latest/querying/examples/
```
RunDevelopment, line 3: Otherwise, the code block will start with an empty line.

```suggestion
<pre><code># These examples are taken from: https://prometheus.io/docs/prometheus/latest/querying/examples/
```

components/prism-promql.js
```diff
@@ -0,0 +1,148 @@
...
+				inside: {
+					'range-duration': {
+						pattern: /\b(?:\d+(?:[smhdwy]|ms))+\b/i,
+						alias: 'number',
+					},
+				},
+			},
+		],
+		'function': new RegExp('\\b(?:' + functions.join('|') + ')(?=\\s*\\()', 'i'),
+		'keyword': new RegExp('\\b(?:' + keywords.join('|') + ')\\b', 'i'),
```
RunDevelopment, line 143: After all keywords are matched, could we say that all functions are of the form `\b[a-zA-Z_]\w*(?=\s*\()`?

components/prism-promql.js
```diff
@@ -0,0 +1,173 @@
...
+Prism.languages.promql = {
+	comment: {
+		pattern: /^\s*#.*/m,
+	},
+	"context-aggregation": {
+		pattern: new RegExp("((?:" + vectorMatching.join("|") + ")\\s*)\\([^)]*\\)"),
+		lookbehind: true,
+		inside: {
+			"label-key": {
+				pattern: /\b[^,]*\b/,
```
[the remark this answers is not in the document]
RunDevelopment: I still don't understand this.

RunDevelopment:
Thank you for the quick response!

I still have some minor nits but it's overall it looks really good now.

components/prism-promql.js
arendjr: I think we can, yes. But I do think there is significant value in listing the functions explicitly, as it allows one to instantly see by its color when they have mistyped a function name.

components/prism-promql.js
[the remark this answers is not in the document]
arendjr: It's to highlight the comma-separated labels.

components/prism-promql.js
RunDevelopment: The main problem with the listing approach is that we have to maintain the list. New versions might add/remove functions and we'd have to follow suit. Lists also increase the file size of the language definition. The only upside I can see is the typo detection you mentioned but Prism isn't an editor.

I also want to point out that functions are [parsed generically by PromQL's parser](https://github.com/prometheus/prometheus/blob/cda52234eb5b2db1a7cc33cba030354da658e131/promql/parser/generated_parser.y#L329). I think that we should do the same.

arendjr: Fair enough. I can monkey-patch it back in for our own project, so I won't object to using a generic approach upstream :)

arendjr: Updated.

components/prism-promql.js
```diff
@@ -0,0 +1,99 @@
...
+				inside: {
+					'range-duration': {
+						pattern: /\b(?:\d+(?:[smhdwy]|ms))+\b/i,
+						alias: 'number',
+					},
+				},
+			},
+		],
+		'function': /\b[a-zA-Z_]\w*(?=\s*\()/i,
+		'keyword': new RegExp('\\b(?:' + keywords.join('|') + ')\\b', 'i'),
```
RunDevelopment, line 94: Please move `function` below `keyword`. This is just to make sure that we don't accidentally highlight a keyword as a function.

RunDevelopment (approved):
Please rebuild, so I can merge.

arendjr:
Done!

RunDevelopment:
Thank you for contributing @arendjr!

arendjr:
Thanks for your support! :rocket:

PR #2640, opened 2020-11-15
Added regex test coverage

RunDevelopment:
This adds a little regex test coverage tool/test. It tracks all matches (= the result of `/some regex/.exec(code)`) of all of Prism's regexes on all language tests. If a regex doesn't produce any non-null matches, it will be reported. Reported regexes are grouped by language, so the error message will look like this:

<details>

```
  1) Pattern test coverage
       Coverage
         - should cover all patterns in bison:
     AssertionError: 1 pattern(s) in bison are untested:
You can learn more about writing tests at https://prismjs.com/test-suite.html#writing-tests

Prism.languages.bison.bison.inside.number.pattern:
        /(^|[^@])\b(?:0x[\da-f]+|\d+)/gi
This pattern is completely untested. Add test files that match this pattern.
      at Context.<anonymous> (tests\coverage.js:98:13)
      at processImmediate (internal/timers.js:456:21)

  2) Pattern test coverage
       Coverage
         - should cover all patterns in gml:
     AssertionError: 5 pattern(s) in gml are untested:
You can learn more about writing tests at https://prismjs.com/test-suite.html#writing-tests

Prism.languages.gml.keyword:
        /\b(?:if|else|switch|case|default|break|for|repeat|while|do|until|continue|exit|…
This pattern is completely untested. Add test files that match this pattern.

Prism.languages.gml.number:
        /(?:\b0x[\da-f]+|(?:\b\d+\.?\d*|\B\.\d+)(?:e[+-]?\d+)?)[ulf]*/gi
This pattern is completely untested. Add test files that match this pattern.

Prism.languages.gml.operator:
        /[-+%=]=?|!=|\*\*?=?|\/\/?=?|<[<=>]?|>[=>]?|[&|^~]|\b(?:or|and|not|with|at|xor)\…
This pattern is completely untested. Add test files that match this pattern.

Prism.languages.gml.constant:
        /\b(?:self|other|all|noone|global|local|undefined|pointer_(?:invalid|null)|actio…
This pattern is completely untested. Add test files that match this pattern.

Prism.languages.gml.variable:
        /\b(?:x|y|(?:x|y)(?:previous|start)|(?:h|v)speed|direction|speed|friction|gravit…
This pattern is completely untested. Add test files that match this pattern.
      at Context.<anonymous> (tests\coverage.js:98:13)
      at processImmediate (internal/timers.js:456:21)
```

</details>

This also means that all languages without any tests will be reported (we had 3: GML, Ren'py, and Arduino).

The current coverage only reports completely untested patterns. We might want to expand this in the future, i.e. to assert that all keywords in keyword lists are tested. Those future applications are not part of this PR because it takes a lot of time to make all of Prism's >200 languages conform to a new standard.

The main motivation for this PR is code review. 
With this test, we will never merge a language with missing tests ever again.

---

I'm not quite done fixing all tests yet. I'll do the rest tomorrow or the day after.


RunDevelopment:
@mAAdhaTTah Could you please review this? I change a lot of files, so I kinda don't want to resolve merge conflicts for this. I would appreciate it if we could merge this quickly. 

The main reason so many language definitions change is that I found and fixed bugs. (As it turns out, untested code can contain bugs.) 
Most of the new code snippets in the new (and changed) test files are either from the official documentation of the respective programming language or from various sources (credited) I found while googling.

I also want to point out that the code coverage is by far not perfect. A regex (except for keyword lists) is considered covered if it has 1 match and since we don't know what parts of the pattern its matches cover, there is probably still a lot of untested code. I plan to address this problem in the future but the PR is already big enough for now.

mAAdhaTTah:
@RunDevelopment For the future, for big PRs like this, it would be helpful to ship it without whitespace or style changes. It would make it easier to review, as I wouldn't have to determine whether a change is style or functional.

tests/coverage.js
```diff
@@ -0,0 +1,241 @@
...
+	describe('Register all patterns', function () {
+		it('all', function () {
+			this.slow(10 * 1000);
+			// This will cause ALL regexes of Prism to be registered in the patterns map.
+			// (Languages that don't have any tests can't be caught otherwise.)
+			createInstance(ALL_LANGUAGES);
+		});
+	});
+
+	describe('Run all language tests', function () {
```
mAAdhaTTah, line 77: Are we re-running all of the language tests? We can't wrap the current test run?

RunDevelopment: > We can't wrap the current test run?

This will only work if *all* language tests are run. 

But in the end, coverage was easier to implement by rerunning all tests. The only point against this is performance, right?

mAAdhaTTah: Performance & code duplication.

Can we add the coverage checking to the base test suite?

RunDevelopment:
> For the future, for big PRs like this, it would be helpful to ship it without whitespace or style changes.

Sorry about that.

tests/coverage.js
RunDevelopment: I honestly don't think that's a good idea.

Coverage only uses our test files to gather data. There is no guarantee that this won't change/be expanded in the future. I don't want to marry two simple systems only to get one complex system.

If code dup is the problem, then I'll add a helper function that runs the whole test suite for us. The current test runner will then only parse CLI args and call the helper function. This might actually be the best solution since coverage doesn't actually care whether the test cases pass.

mAAdhaTTah: lol I don't know why it bothers me so much that we're running the test suite in one file, then rerunning the whole suite again in another file to collect coverage. Running the tests & collecting coverage have always gone hand-in-hand in other cases (e.g. running `nyc mocha` or `jest --coverage`), so my inclination is to combine or connect them, less from a perf or code dupe perspective and more of an testing arch perspective.

That said, strong opinion weakly held. I'm not sure I fully understand the downside here.

RunDevelopment: Yeah, I get your perspective. We usually use awesome test coverage tools like [Istanbul](https://istanbul.js.org/). These tools hook directly into NodeJS, so they can get detailed information about what functions/statements/branches are called/executed/takes and how many times. But getting this usage information it's not so easy for Prism.

Our goal is to figure out what patterns contributed to the token streams produced by the test files. Basically, we want to know what tokens each regex produced.

Right now, the coverage test *approximates* this by running all test files and remembering all matches of all (unique) regexes. This is not a correct solution as it overestimates the actual usage of our regexes. I.e. regex matches might lead to tokens that are subsequently discarded because of other greedy patterns. The correct solution is to examine the produced token stream and discard all regex matches that didn't produce a token present in the final output token stream of our test files.
The only reason I haven't implemented the correct solution is that it is hard to do and the approximation seemed good enough for now.

This is why I view the two tests as distinct. They have different goals but achieve them through similar methods, at least for now.

RunDevelopment:
It would be really nice if we could merge this. @mAAdhaTTah

PR #2643, opened 2020-11-19
SCSS: Add support for Sass modules

adamkudrna:
The aim of this PR is to support [Sass modules](https://sass-lang.com/blog/the-module-system-is-launched) that look like this (showing missing support in the screenshots):

image

image


components/prism-scss.js
```diff
@@ -41,7 +41,7 @@ Prism.languages.scss = Prism.languages.extend('css', {
 
 Prism.languages.insertBefore('scss', 'atrule', {
 	'keyword': [
-		/@(?:if|else(?: if)?|for|each|while|import|extend|debug|warn|mixin|include|function|return|content)/i,
+		/@(?:if|else(?: if)?|forward|for|each|while|import|use|extend|debug|warn|mixin|include|function|return|content)/i,
```
RunDevelopment, line 44: Let's add a `\b`, so we won't have partial highlighting for keywords like `@forward` ever again.

```suggestion
		/@(?:if|else(?: if)?|forward|for|each|while|import|use|extend|debug|warn|mixin|include|function|return|content)\b/i,
```

RunDevelopment (changes requested):
Great work @adamkudrna!

One small nit and this is ready to go.

components/prism-scss.js
adamkudrna: Fixed.

RunDevelopment (approved):

Issue #2651, opened 2020-11-20
Prism seems to not highlight HTML comments (at least when I tried it)


Neohiro79:
**You can see the problem in this JSFiddle here:**

> http://jsfiddle.net/neohiro79/5x13L0hm/47/

Is there something I've missed?

I even tried to give it a classname "comment" as proposed somewhere - but the HTML comments won't show up.

RunDevelopment:
Prism only highlights the text content of `code` elements. You have escape all `<` and `&` characters inside `code` elements or use our [Unescaped markup plugin](https://prismjs.com/plugins/unescaped-markup/).

Neohiro79:
Thanx a lot for the quick tip and answer! It finally worked once I've escaped all `<` and `>` characters:

http://jsfiddle.net/neohiro79/5x13L0hm/50/

Neohiro79:
Something what still bothers me, even if the [Unescaped markup plugin](https://prismjs.com/plugins/unescaped-markup/) is used:

What is the reason for naming the class `language-markup` if it is not "markup'ing the markup language called HTML at all?

It won't till a so called plugin "unescaped markup" is used - **BUT HERE COMES THE CAVEAT:** you can't use `script` tags inside this plugin, since it is based on putting things inside the `script tag` itself - and if you cannot use the alternative of the `<--` masking because you want exactly this 'seeing the comments'. 

**You can only use it in a weired combination like here:**

http://jsfiddle.net/neohiro79/5x13L0hm/59/

**It also puts some sort of weired additional 'tabspace' before the 'style' tags for some reason.**

see here: http://jsfiddle.net/neohiro79/5x13L0hm/61/

---

### The final question:

If this plugin can realise the `<--` and `-->` parts through some sort of `RegEx` algorithm, why can't this library realise and mask those `<--` and `-->` just from startup - so that one can use simply `language-markup` to finally get the library to do what is intended to do, in highlighting the markup correctly as expected?

Even more, what is the reason for the `comment` class, when even this won't work as expected?

**Am I the only one using comments in HTML code snippets?**

RunDevelopment:
Markup inside of `code` elements isn't treated as text by the browser. It's treated as markup and markup has meaning, so please don't put unescaped `script` tags inside `code` elements (unless they have `type="text/plain"`) because **the scripts are going to be executed**. Same for `style` and all other tags that have side effects. This why you're supposed to escape all `<` and `&` characters. 

[...]

PR #2652, opened 2020-11-21
Unescaped Markup: More doc regarding comments

RunDevelopment:
This makes the explanation of the usage of comments clearer and gives examples.

[Motivation](https://github.com/PrismJS/prism/issues/2651#issuecomment-731485802).

### DIFF
--- a/plugins/unescaped-markup/index.html
+++ b/plugins/unescaped-markup/index.html
@@ -32,7 +32,7 @@
 &lt;/script></code></pre>
 		</li>
 		<li>
-			Use a HTML-comment to escape your code:
+			Use an HTML-comment to escape your code:
 
 			<pre><code>&lt;pre class="language-markup">&lt;code>&lt;!--
 &lt;p>Example&lt;/p>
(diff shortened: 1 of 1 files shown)


plugins/unescaped-markup/index.html
```diff
@@ -24,15 +24,23 @@ <h1>How to use</h1>
...
+			Instead of using <code>&lt;pre>&lt;code></code> elements, use <code>&lt;script type="text/plain"></code>:
+
+			<pre><code>&lt;script type="text/plain" class="language-markup">
 &lt;p>Example&lt;/p>
 &lt;/script></code></pre>
 		</li>
-		<li>Use a HTML-comment to escape your code
-		<pre><code>&lt;pre class="language-markup">&lt;code>&lt;!--
+		<li>
+			Use a HTML-comment to escape your code:
```
mAAdhaTTah, line 35: English is weird, but I think this is correct ("aych" (H) sound is a vowel sound).

```suggestion
			Use an HTML-comment to escape your code:
```

mAAdhaTTah (approved):

RunDevelopment:
We regularly have some sort of typos and other mistakes. Is there a GitHub bot for some service like Grammarly?

RunDevelopment:
I looked into it and the best apps out there are simple spell checks. That might be enough but because of GH actions' restrictions, none will work with forks. Yay.

Let's just ignore my suggestion.

PR #2653, opened 2020-11-22
Update prism-pascal.js

SlawomirZalecki:
The directive class has been added because comments starting with {$ are directives


components/prism-pascal.js
```diff
@@ -5,9 +5,17 @@
 */
 
 Prism.languages.pascal = {
+	'asm': {
+		pattern: /(asm)([\s\S]+?)(end)/i,
```
RunDevelopment, line 9: One of the tests fails because of the unnecessary capturing groups, so ket's fix that and let's add word boundary assertions as well.
```suggestion
		pattern: /(asm\b)[\s\S]+?\bend/i,
```

components/prism-pascal.js
```diff
@@ -5,9 +5,17 @@
...
+		lookbehind: true,
+		inside:{keyword:/end/i,punctuation:/;$/i}
```
RunDevelopment, line 11: 1. Formatting.
1. The punctuation rule won't work. Did you mean `/;$/m`?
```suggestion
		inside: {
			'keyword': /end/i,
			'punctuation': /;$/i
		}
```

components/prism-pascal.js
```diff
@@ -5,9 +5,17 @@
...
+	},	
+	'directive': [
+		/\{\$[\s\S]+?\}/
+	],
```
RunDevelopment, line 15:
```suggestion
	'directive': /\{\$[\s\S]+?\}/,
```

components/prism-pascal.js
```diff
@@ -5,9 +5,17 @@
...
+		lookbehind: true,
+		inside:{keyword:/end/i,punctuation:/;$/i}
+	},	
+	'directive': [
+		/\{\$[\s\S]+?\}/
+	],	
 	'comment': [
 		/\(\*[\s\S]+?\*\)/,
-		/\{[\s\S]+?\}/,
+		/\{[^$][\s\S]+?\}/,
```
RunDevelopment, line 18:
```suggestion
		/\{[^$][\s\S]*?\}/,
```

RunDevelopment (changes requested):
Thank you for this PR @SlawomirZalecki!

Please add [tests](https://prismjs.com/test-suite.html) for the new features you added. You also have to run `npm run build` and commit all changed files to make the build pass.

components/prism-pascal.js
SlawomirZalecki: Thanks for your reply and suggestions

components/prism-pascal.js
SlawomirZalecki: It will be good, but the resulting html code is like this:

```
<span class="token keyword">asm</span>
  <span class="token asm">
    MOV AX,1234h
    MOV Number,AX
  <span class="token keyword">end</span>
  <span class="token punctuation">;</span>
</span>
```

In fact, I think it should be:

```
<span class="token keyword">asm</span>
<span class="token asm">
  MOV AX,1234h
  MOV Number,AX
</span>
<span class="token keyword">end</span>
<span class="token punctuation">;</span>
```

But I do not know how to do it.

components/prism-pascal.js
SlawomirZalecki: In fact, that's how it should be.
It was the first time I analyzed the prism.js code
I need it for my website and it wasn't working properly so I'm trying to improve.

components/prism-pascal.js
SlawomirZalecki: OK

components/prism-pascal.js
RunDevelopment: In that case, we can make the whole `asm` rule like this:

```js
'asm: {
    pattern: /(asm\b)[\s\S]+?(?=\bend;)/,
    lookbehind: true
}
```

SlawomirZalecki: Works perfect, thank you.

YoutacRSVARs (approved):

PR #2659, opened 2020-11-26
DataWeave: Adding support for DataWeave language

machaval:
This pull request adds support for [dataweave](https://docs.mulesoft.com/mule-runtime/4.3/intro-dataweave2) language. 

```dataweave
%dw 2.0
var a = "123"
---
a
```

As seen in this snippet github already supports this language


components/prism-dataweave.js
```diff
@@ -0,0 +1,48 @@
+(function (Prism) {
+    var keywords = /\b(?:match|input|output|ns|type|update|null|if|else|using|unless|at|is|as|case|do|fun|var|not|and|or)\b/;
+    
+    Prism.languages.dataweave = {	
+        "url": {
+            pattern:/(?:[A-z])+:\/\/(?:[A-z0-9/:\.\-_?=&])+|urn:(?:[A-z0-9:\-_\.?=&])+/
```
RunDevelopment, line 6: (Please apply the fix for `A-z` here as well.)

```suggestion
            pattern:/\b[A-z]+:\/\/[A-z0-9/:\.\-_?=&]+|\burn:[A-z0-9:\-_\.?=&]+/
```

components/prism-dataweave.js
```diff
@@ -0,0 +1,48 @@
...
+        },
+        'property': {
+            pattern: /(?:[A-z0-9_]+#)?(?:"(?:\\.|[^\\"\r\n])*"|(?:[A-z0-9_]+))(?=\s*(?::|@))/,
```
RunDevelopment, line 9: (Please apply the fix for `A-z` here as well.)

```suggestion
            pattern: /(?:[A-z0-9_]+#)?(?:"(?:\\.|[^\\"\r\n])*"|[A-z0-9_]+)(?=\s*[:@])/,
```

tests/languages/dataweave/comment_feature.test
```diff
@@ -0,0 +1,15 @@
...
+	/* Block comment */
+	/**
+	* MultiLine Comment
+	**/
+
+----------------------------------------------------
+
+[
+	["comment", "// Line comment \"\"\t"],
+	["regex", "/* Block comment */"],
```
RunDevelopment, line 11: Uhmm

components/prism-dataweave.js
```diff
@@ -0,0 +1,48 @@
...
+            pattern: /\/[^ ](?:[^\\\/\r\n]|\\[^\r\n])+\//,
+            greedy: true
+        },
+        'date': {
+            pattern: /\|[^ ](?:[^\\\|\r\n]|\\[^\r\n])+\|/,
+            greedy: true
+        },
+        'comment': [
+            {
+                pattern: /(^|[^\\])\/\*[\s\S]*?(?:\*\/|$)/,
```
RunDevelopment, line 29: This has to be greedy too.

Please add the snippet `/* "" */` to the tests as well.

RunDevelopment (changes requested):
Thank you for making this language definition @machaval!

There are a few things that have to be improved but overall solid work!

machaval:
Thanks for all the feedback I will do the required changes ASAP

tests/languages/dataweave/comment_feature.test
machaval: Good catch!

machaval:
All requested changes have been fixed now. Let me know if there is anything else you would like to be changed. Again thanks for all the comments you are the Regex master!

components/prism-dataweave.js
```diff
@@ -0,0 +1,47 @@
...
+        'property': {
+            pattern: /(?:[A-Za-z0-9_]+#)?(?:"(?:\\.|[^\\"\r\n])*"|[A-Za-z0-9_]+)(?=\s*[:@])/,
+            greedy: true
+        },
+        'string': {
+            pattern: /(["'`])(?:\\[\s\S]|(?!\1)[^\\])*\1/,
+            greedy: true
+        },
+        'mime-type': {
+            pattern: /(?:text|audio|video|application|multipart|image)\/(?:[A-Za-z-_+0-9])+/
```
RunDevelopment, line 15:
```suggestion
            pattern: /\b(?:text|audio|video|application|multipart|image)\/[\w+-]+/
```

components/prism-dataweave.js
```diff
@@ -0,0 +1,47 @@
+(function (Prism) {    
+    Prism.languages.dataweave = {	
+        'url': {
+            pattern:/\b(?:[A-Za-z])+:\/\/(?:[A-Za-z0-9/:\.\-_?=&])+|\burn:(?:[A-Za-z0-9:\-_\.?=&])+/
+        },
+        'property': {
+            pattern: /(?:[A-Za-z0-9_]+#)?(?:"(?:\\.|[^\\"\r\n])*"|[A-Za-z0-9_]+)(?=\s*[:@])/,
```
RunDevelopment, line 7:
```suggestion
            pattern: /(?:\w+#)?(?:"(?:\\.|[^\\"\r\n])*"|\w+)(?=\s*[:@])/,
```

components/prism-dataweave.js
```diff
@@ -0,0 +1,47 @@
+(function (Prism) {    
+    Prism.languages.dataweave = {	
+        'url': {
+            pattern:/\b(?:[A-Za-z])+:\/\/(?:[A-Za-z0-9/:\.\-_?=&])+|\burn:(?:[A-Za-z0-9:\-_\.?=&])+/
```
RunDevelopment, line 4:
```suggestion
            pattern:/\b[A-Za-z]+:\/\/[\w/:.?=&-]+|\burn:[\w:.?=&-]+/
```

components/prism-dataweave.js
```diff
@@ -0,0 +1,47 @@
...
+        },
+        'string': {
+            pattern: /(["'`])(?:\\[\s\S]|(?!\1)[^\\])*\1/,
+            greedy: true
+        },
+        'mime-type': {
+            pattern: /(?:text|audio|video|application|multipart|image)\/(?:[A-Za-z-_+0-9])+/
+        },       
+        'date': {
+            pattern: /\|[^ ](?:[^\\|\r\n]|\\[^\r\n])+\|/,
```
RunDevelopment, line 18: Why `[^ ]`? Is it really just the space character that is disallowed?
Also, is `||a|` valid then?

(Same for `regex`)

components/prism-dataweave.js
```diff
@@ -0,0 +1,47 @@
...
+                greedy: true
+            },
+            {
+                pattern: /(^|[^\\:])\/\/.*/,
+                lookbehind: true,
+                greedy: true
+            }
+        ],
+        'regex': {
+            pattern: /\/[^ ](?:[^\\\/\r\n]|\\[^\r\n])+\//,
```
RunDevelopment, line 34: 1. Too bad `/a/` isn't a valid regex. Or is it? ;)
1. Is `/abc[/]/` not a valid regex? In languages like JS, character classes are parsed slightly differently. Is that not the case here?

```suggestion
            pattern: /\/[^ ](?:[^\\\/\r\n]|\\[^\r\n])*\//,
```

components/prism-dataweave.js
```diff
@@ -0,0 +1,47 @@
...
+            greedy: true
+        },
+        'function': /\b[A-Za-z_]\w*(?=\s*\()/i,
+        'number': /-?\b\d+(?:\.\d+)?(?:e[+-]?\d+)?\b/i,
+        'punctuation': /[{}[\];(),.:@]/,        
+        'operator': /<<|>>|->|[<>~=]=?|!=|--?-?|\+\+?|\!|\?/,
+        'boolean': /\b(?:true|false)\b/,
+        'keyword': {
+            pattern: /\b(?:match|input|output|ns|type|update|null|if|else|using|unless|at|is|as|case|do|fun|var|not|and|or)\b/
+        }
```
RunDevelopment, line 44: Please simplify all rules that only have a `pattern` property.

```suggestion
        'keyword': /\b(?:match|input|output|ns|type|update|null|if|else|using|unless|at|is|as|case|do|fun|var|not|and|or)\b/
```

RunDevelopment (changes requested):
Thank you for the quick response @machaval!

I left you a few comments again. A few regexes and rules can be simplified and I also have some questions.

components/prism-dataweave.js
machaval: In dataweave if you want to put a `/` you need to  escape it `\/` `/abc[/]/` needs to be `/abc[\/]/`. Regarding spaces this is wrong. Also we dataweave doesn't support empty regex `//` is not a valid regex expression so shouldn't we keep the `+`

components/prism-dataweave.js
machaval: I changed the regex to something clearer `/\|[\w\-:+-]+\|/` and added more tests where it shows all the possible values

components/prism-dataweave.js
```diff
@@ -0,0 +1,45 @@
...
+        },
+        'string': {
+            pattern: /(["'`])(?:\\[\s\S]|(?!\1)[^\\])*\1/,
+            greedy: true
+        },
+        'mime-type': {
+            pattern: /\b(?:text|audio|video|application|multipart|image)\/[\w+-]+/
+        },       
+        'date': {
+            pattern: /\|[\w\-:+-]+\|/,
```
RunDevelopment, line 18:
```suggestion
            pattern: /\|[\w:+-]+\|/,
```

components/prism-dataweave.js
```diff
@@ -0,0 +1,45 @@
...
+            pattern: /(?:\w+#)?(?:"(?:\\.|[^\\"\r\n])*"|\w+)(?=\s*[:@])/,
+            greedy: true
+        },
+        'string': {
+            pattern: /(["'`])(?:\\[\s\S]|(?!\1)[^\\])*\1/,
+            greedy: true
+        },
+        'mime-type': {
+            pattern: /\b(?:text|audio|video|application|multipart|image)\/[\w+-]+/
+        },
```
RunDevelopment, line 16: Just the regex is enough.

```suggestion
        'mime-type': /\b(?:text|audio|video|application|multipart|image)\/[\w+-]+/,
```

components/prism-dataweave.js
```diff
@@ -0,0 +1,45 @@
+(function (Prism) {    
+    Prism.languages.dataweave = {	
+        'url': {
+            pattern:/\b[A-Za-z]+:\/\/[\w/:.?=&-]+|\burn:[\w:.?=&-]+/
+        },
```
RunDevelopment, line 5: Same here.
```suggestion
        'url': /\b[A-Za-z]+:\/\/[\w/:.?=&-]+|\burn:[\w:.?=&-]+/,
```

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @machaval!

PR #2677, opened 2020-12-22
Test page: Don't trigger ad-blockers with class

RunDevelopment:
I was wondering why the new "Share" button didn't show up. A selector `.share-wrapper` seems to be in one of the common block lists, so I just renamed the wrapper from "share-wrapper" to "link-wrapper".

### DIFF
--- a/test.html
+++ b/test.html
@@ -95,33 +95,33 @@ pre.show-tokens {
 	position: relative;
 }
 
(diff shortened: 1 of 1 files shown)


RunDevelopment:
The fix is very simple. I'll just merge this.

PR #2682, opened 2020-12-31
Latte: Fixed exponential backtracking

RunDevelopment:


joshgoebel:
Is the detector improving or this was just not dealt with before?

RunDevelopment:
This is one of the very few regexes that aren't checked by our detector. This was found using a CodeQL query.

joshgoebel:
I'll assume it's something weird about how markup templating works or something that it's "special"... so much of our regex are generated at run-time but since the tests are running at run-time we just hook into the FINAL prepared regex and run them thru the checker.

RunDevelopment:
Yeah, we just check the regexes of all grammars. Markup templating also needs one additional regex to detect the embedded language and that regex isn't checked.

One way around that might be to override the `exec`/`test` methods in `RegExp.prototype` to run the detector on all patterns used to tokenize Prism's test cases.

PR #2687, opened 2021-01-01
Add support for V langauge

taggon:
This PR adds support for [the V langauge](https://vlang.io/).


RunDevelopment (changes requested):
Thank you for making this PR @taggon!

I left you a few comments. Also, directives (e.g. [`#flag linux -lsdl2`](https://github.com/vlang/v/blob/master/doc/docs.md#flag), `#include "sqlite3.h"`) and [nested multiline comments](https://github.com/vlang/v/blob/master/doc/docs.md#comments) are not supported. Please add them. 

Apart from that, very solid work. Thank you!

taggon:
@RunDevelopment Thanks. I fixed all the problems you mentioned and also add support for directives(such as `#include` and `#flag`).

components/prism-v.js
```diff
@@ -0,0 +1,73 @@
...
+							pattern: /^\$\w[\s\S]*$/,
+							alias: 'variable'
+						},
+						'interpolation-punctuation': {
+							pattern: /^\${|}$/,
+							alias: 'punctuation'
+						},
+						'interpolation-expression': {
+							pattern: /[\s\S]+/,
+							inside: Prism.languages.v
```
RunDevelopment, line 26: I'm sorry that I suggested code that doesn't work. `Prism.languages.v` isn't defined yet, so `inside` will be set to `undefined`.

To fix this, please replace this line with:
```suggestion
							inside: null // see below
```

And add the following line after the `Prism.languages.extend` call.

```js
Prism.languages.v['string'][1].inside['interpolation'].inside['interpolation-expression'].inside = Prism.languages.v;
```

RunDevelopment (changes requested):
Thank you for your work! 

I'm sorry but one of my suggestions contained faulty code. My comment contains more information and a fix.

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @taggon!

PR #2688, opened 2021-01-02
Tests: Exhaustive pattern tests

RunDevelopment:
This improves the static pattern test to check every regex. 

### Method

This is achieved by running all test files for a given language and overwriting the `Prism.tokenize` function itself. The `tokenize` function will be replaced with a function that runs all checks on the grammar given to the function and then uses the original `tokenize` function. Assuming that our test files cover (nearly) all aspects of our languages, this approach will analyze all regexes used in grammar.

This runtime approach is necessary because some languages (e.g. PHP and Latte) use markup templating (MT) which changes the grammar with one that might not be reachable from `Prism.languages`.

However, even this isn't enough. MT also requires a regex to detect the embedded language ([example](https://github.com/PrismJS/prism/blob/1506f345b0c917b28047ec6e5b8b8dc81790ad98/components/prism-php.js#L306)). This pattern isn't passed to `Prism.tokenize`. 
To analyze these regexes for super-linear runtime, the prototype of the `RegExp` type itself is changed to replace the `exec` and `test` methods with methods that run the analysis for every regex that is used throughout the tests.

### Problem

The only problem with this approach is that it is slow.

To statically analyze all regexes, I need about 2s, and to run all test files, I need about another 2s on my machine. However, the exhaustive pattern test takes 9s on my machine.

This is because of the `RegExp` prototype magic that forces us to take the [slow path](https://v8.dev/blog/speeding-up-regular-expressions). Without the prototype magic, the whole thing only takes 5s on my machine.


RunDevelopment:
I will merge this. It really doesn't matter that the tests take longer now.

PR #2690, opened 2021-01-02
Added support for DOT

### LINKED ISSUE #2660: Language request: DOT (graph description language)
**Language**
"DOT is a graph description language. DOT graphs are typically files with the filename extension gv or dot." - [source](https://en.wikipedia.org/w/index.php?title=Graphviz&oldid=987303523)

**Additional resources**
[Wikipedia](https://en.wikipedia.org/wiki/DOT_(graph_description_language))

RunDevelopment:
This resolves #2660.

### Example

image

[source](ryandesign.com/canviz/graphs/dot/directed/ctext.gv.txt)


edukisto:
@RunDevelopment, I have some proposal.

RunDevelopment:
Sure, please do. I always appreciate feedback ;)

components/prism-dot.js
```diff
@@ -0,0 +1,77 @@
+// https://www.graphviz.org/doc/info/lang.html
+
+(function (Prism) {
+
+	var ID = '(?:' + [
+		// an identifier
+		/\b[a-zA-Z_][\w\x80-\x9F\xA1-\xFF]*/.source,
```
edukisto, line 7: An ID [can](https://graphviz.org/doc/info/lang.html) begin with a `[A-Z_a-z\x80-\xFF]` character. In case of using `[\x80-\xFF]` as a first character, the `\b` pattern can’t handle a word boundary.

components/prism-dot.js
```diff
@@ -0,0 +1,77 @@
...
+		// a number
+		/-?(?:\.\d+|\b\d+(?:\.\d*)?)/.source,
```
edukisto, line 9: I think, `ÿ123` (`ÿ` is \xFF) should be interpreted as a single ID.

components/prism-dot.js
```diff
@@ -0,0 +1,77 @@
...
+		// a double-quoted string
+		/"(?:[^\\"]|\\[\s\S])*"/.source,
```
edukisto, line 11: @RunDevelopment , what do you think about the following pattern?

```
"[^"\\]*(?:\\"?[^"\\]*)*"
```

I think, it’s a bit faster, but I’m not sure in correctness. Well, at least `pattern-tests.js` doesn’t complain :)

components/prism-dot.js
```diff
@@ -0,0 +1,77 @@
...
+	 * @param {string} flags
+	 * @returns {RegExp}
+	 */
+	function withID(source, flags) {
+		return RegExp(source.replace(/<ID>/g, function () { return ID; }), flags);
+	}
+
+	Prism.languages.dot = {
+		'comment': {
+			pattern: /\/\/.*|\/\*[\s\S]*?\*\/|^#.*/m,
```
edukisto, line 36: How about changing `\/\/` to `\/{2}`? The latter takes 1 step less.

edukisto: `\/\*[\s\S]*?\*\/` backtracks on every character inside a comment.

|Comment       |Steps|Backtracks|
|--------------|-----|----------|
|/\*\*/          |    5|         0|
|/\*1\*/         |    7|         1|
|/\*\*\*/         |    8|         1|
|/\*12\*/        |    9|         2|
|/\*\*\*\*/        |   11|         2|
|/\*123\*/       |   11|         3|
|/\*\*\*\*\*/       |   14|         3|
|/\*1234567890\*/|   25|        10|
|/\*\*\*\*\*\*\*\*\*\*\*\*/|   35|        10|

+3 steps for every asterisk and +2 steps for any other character.

As I see, the amount of steps/backtracks grows linearly. But anyway, I hope, and please correct me if I’m wrong, that

```
\/\*[^*]*\*+(?:[^*/][^*]*\*+)*\/
```

should always make a single backtrack. In most cases the pattern takes **7 steps**. The amount of steps grows linearly only on repeating “an asterisk – not an asterisk” substrings.

|Comment     |Steps|Backtracks|
|------------|-----|----------|
|/\*\*/      |    7|         1|
|/\*1\*/     |    7|         1|
|/\*\*\*/    |    7|         1|
|/\*12\*/    |    7|         1|
|...         |  ...|       ...|
|/\*\*1\*/   |   10|         1|
|/\*\*1\*\*/ |   10|         1|
|/\*\*1\*1\*/|   14|         1|

components/prism-dot.js
```diff
@@ -0,0 +1,77 @@
...
+		},
+		'attr-name': {
+			pattern: withID(/([\[;,\s])<ID>(?=\s*=)/.source),
+			lookbehind: true,
+			greedy: true,
+			inside: IDInside
+		},
+		'keyword': /\b(?:digraph|edge|graph|node|strict|subgraph)\b/i,
+		'compass-point': {
+			pattern: /(:\s*)(?:[ns][ew]?|[ewc])(?![\w\x80-\x9F\xA1-\xFF])/,
```
edukisto, line 60: > The compass point "_" specifies that an appropriate side of the port adjacent to the exterior of the node should be used, if such exists.

(from <https://graphviz.org/doc/info/attrs.html>).

> compass_pt : (n | ne | e | se | s | sw | w | nw | c | _)

(from <https://graphviz.org/doc/info/lang.html>)

An example:

```dot
digraph {
  a -> b:_;
}
```

components/prism-dot.js
RunDevelopment: You're right! I tested it and it's about 4x faster when no/few escapes are present (the common case) and 2x slower when the string contains only escapes. It's a lot faster in the common case, so that's a definite yes from me!

The correct version is `/"[^"\\]*(?:\\[\s\S][^"\\]*)*"/` btw.

components/prism-dot.js
RunDevelopment: I forgot one... Thanks for pointing that out!

components/prism-dot.js
RunDevelopment: > How about changing `\/\/` to `\/{2}`? The latter takes 1 step less.

Most regex engines actually desugar `[char]{x}` into `[char][char]...[char]` (x-many).

I don't how what tool you use to measure these steps but its underlying regex engine model may not be one used in browsers.

> `\/\*[\s\S]*?\*\/` backtracks on every character inside a comment.

Please be careful! Steps and backtracks are not the same as actual runtime. For those two patterns in particular, the simpler, seemingly-less-efficient one is actually faster.

<details>

```js
measure = fn => { let s = performance.now(); fn(); return performance.now() - s; };

var r1 = /\/\*[\s\S]*?\*\//;
var r2 = /\/\*[^*]*\*+(?:[^*/][^*]*\*+)*\//;

// nodejs repl
> measure(() => r1.test( '/*' + "*".repeat(1e6) ))
3.105300188064575
> measure(() => r2.test( '/*' + "*".repeat(1e6) ))
7.263500213623047

> measure(() => r1.test( '/*' + "a".repeat(1e6) ))
2.6819000244140625
> measure(() => r2.test( '/*' + "a".repeat(1e6) ))
4.4449992179870605

> measure(() => r1.test( '/*' + "*".repeat(1e6) + '*/' ))
1.992799997329712
> measure(() => r2.test( '/*' + "*".repeat(1e6) + '*/' ))
2.0360000133514404

> measure(() => r1.test( '/*' + "a".repeat(1e6) + '*/' ))
1.9930000305175781
> measure(() => r2.test( '/*' + "a".repeat(1e6) + '*/' ))
1.8396999835968018

> measure(() => r1.test( '/*' + "a*".repeat(1e6) ))
4.826600074768066
> measure(() => r2.test( '/*' + "a*".repeat(1e6) ))
49.58860111236572

> measure(() => r1.test( '/*' + "a*".repeat(1e6) + '*/' ))
3.439500093460083
> measure(() => r2.test( '/*' + "a*".repeat(1e6) + '*/' ))
44.13789987564087
```

</details>

I know. _Why is the less efficient one faster?_ My best guess is that the regex engine is just better at optimizing simple patterns.

components/prism-dot.js
RunDevelopment: I did this to guarantee O(n) runtime but you're right, it needs a better solution.

components/prism-dot.js
RunDevelopment: Fixed.

components/prism-dot.js
edukisto: > I don't how what tool you use to measure these steps

I use RegexBuddy (a trial version).

> > measure(() => r1.test( '/*' + "a*".repeat(1e6) ))
> 4.826600074768066
> > measure(() => r2.test( '/*' + "a*".repeat(1e6) ))
> 49.58860111236572

Wow.

> the regex engine is just better at optimizing simple patterns.

Ah, thank you very much!

components/prism-dot.js
edukisto: Glad to help!

PR #2701, opened 2021-01-06
[PHP] Numeral syntax improvements

Ayesh:
Improves numeral literal handling in the Prism PHP component in two ways:

1. Since PHP 7.4, it is possible to use [underscore numeric separator](https://php.watch/versions/7.4/underscore_numeric_separator). Prism correctly handles this for decimal numbers, but not for binary, octal, and hex numerals although PHP supports the underscore separator for binary, octal, and hex numeric literals as well.
2. In upcoming PHP 8.1, PHP supports an [explicit octal numeral prefix `0o`/`0O`](https://php.watch/versions/8.1/explicit-octal-notation). The existing `0` prefix is not removed.

This commit updates the PHP number regex to accommodate both cases above, and expands the tests for new syntax.

Thank you.


RunDevelopment (approved):

RunDevelopment:
Thank you very much for contributing @Ayesh!

Ayesh:
Thanks a lot for the quick responses and merging :) 🙏🏼

PR #2702, opened 2021-01-06
Markup: Added support for DOM event attributes

### LINKED ISSUE #2700: Feature request : Add Highlights for inline javascript into html tags under event-attributes
**To make html tags more clear**
Add Highlights for inline javascript into html tags under event-attributes

**Description**
```
<div onclick="this.style.color='red'">hello world</div>
```
As here the code must use javascript Highlighter inside `onclick` attributes

RunDevelopment:
This resolves #2700.

---

This is implemented via a new `addAttribute` function. It works similar to the `addInline` function and is also part of Markup. Right now, only CSS and JS use it.

JSX had to be changed because the special attributes collided with JSX' `script` token.


RunDevelopment:
Looking at the file size, it seems like the logic transfer from CSS to Markup worked well. The gain comes from the list of DOM event attributes in JS.

PR #2720, opened 2021-01-17
Docker: Improvements

RunDevelopment:
This makes a bunch of improvements to the Docker language definition.

The main change is that commands are now parsed as one token. This makes it easier to highlight specific parts of commands (like options) and produces fewer false positives than the previous method.  The only downside is that it is not easy to match a command whole so the pattern is a little more complex. 
I ran the new Docker grammar against some long and complex Dockerfiles from DockerHub and was not able to find any problems.  

Other changes:

- Some commands have options. Those are now highlighted.
- Variables are now supported.
- The `AS` keyword of `FROM` is now supported. 
- ["Punctuation"](https://github.com/PrismJS/prism/blob/4e7b2a82d733fc705bc4a50192fb1bbe552bc687/components/prism-docker.js#L11) was removed. I honestly do not have the slightest idea where this pattern came from. The [Dockerfile doc](https://docs.docker.com/engine/reference/builder/) doesn't mention anything in that regard. My best guess is that this pattern was supposed to highlight bash punctuation but this isn't a good idea because the argument may not be a bash command.


edukisto:
@RunDevelopment, just a small remark. I think, `AS` is not technically a Docker instruction, because

> ...each instruction is run independently, and causes a new image to be created...

For example,

```dockerfile
FROM alpine AS os
```

takes 1 step and generates 1 image.

```bash
sudo docker build .
sudo docker images
```

edukisto:
I mean, `AS` is OK, except the name. It’s confusing.

Oh, there is also a [`HEALTHCHECK CMD`](https://docs.docker.com/engine/reference/builder/#healthcheck) expression.

RunDevelopment:
You're right, the name isn't very fitting. I've been thinking and the name "instruction" isn't even correct for `FROM`, `CMD`, etc. because "instruction" refers to the whole thing, so instruction keyword + argument. `FROM` and so on are the keywords that start a specific instruction.

It might be most fitting to rename the `command` to `instruction` and to rename the current `instruction` to `keyword`. What do you think @edukisto?

---

`HEALTHCHECK` is a good point. Seems like I also missed `ONBUILD`.

---

Also, thanks for looking at this! Reviews really help!

edukisto:
I need some more time to answer.

edukisto:
**The first part. Thoughts on the `command` alias.**

I disagree with [this part](https://github.com/PrismJS/prism/blob/22cfc3bcc6d9ddb555581fa8bb14c8e2b04977be/components/prism-docker.js#L36-L39), because `ADD`, `ARG`, `CMD`, `ENTRYPOINT`, `RUN`, etc. are not called commands in the [docs](https://docs.docker.com/engine/reference/builder/).

I try to draw parallels between Dockerfile and assembly languages. `ADD`, `ARG`, `CMD`, etc. are similar to mnemonic identifiers of operations (see [Intel 80386](https://css.csail.mit.edu/6.858/2014/readings/i386.pdf) sections 1.3.3 and 2.4). Mathematical and logical operations are identified by operators, so I had an idea to propose the `operator` key. However, operator needs operands, but the [format of the Dockerfile](https://docs.docker.com/engine/reference/builder/#format) and corresponding BNFs contain another term, namely “argument”. Functions has arguments. Operation is a special case of function.

That’s why I suggest the `function` key. The Intel 80386 manual associates instructions with underlying CPU functions, too (see section 4.2).

edukisto:
> [An] instruction refers to the whole thing...

I agree!

edukisto:
**The second part. Thoughts on `FROM...AS` and `HEALTHCHECK CMD`.**

`FROM --platform=linux/amd64 python:latest AS bleeding_edge_python`

- `FROM` is a _function_ (?);
- `--platform` is a flag;
- `linux/amd64` is the flag value;
- `python:latest`, `AS`, `bleeding_edge_python` are instruction arguments;
- moreover, `AS` is a _keyword_, although not a reserved word. I agree with [this key](https://github.com/PrismJS/prism/blob/22cfc3bcc6d9ddb555581fa8bb14c8e2b04977be/components/prism-docker.js#L61) for `AS`.

Let me explain why I consider `AS` an argument of `FROM`. Both

```dockerfile
# An alias is missing!
FROM python:latest AS
```

and

```dockerfile
# AS is missing!
FROM python:latest bleeding_edge_python
```

give the following error message:

```bash
FROM requires either one or three arguments
```

Therefore, `AS` is **the second argument** of the three-argument form. [This](https://github.com/docker/engine/blob/master/vendor/github.com/moby/buildkit/frontend/dockerfile/instructions/parse.go#L302) is the best evidence.

`CMD` in `HEALTHCHECK CMD` is an [argument](https://github.com/docker/engine/blob/master/vendor/github.com/moby/buildkit/frontend/dockerfile/instructions/parse.go#L431-L463), too.

edukisto:
It looks like I agree with every existing key except [`command`](https://github.com/PrismJS/prism/blob/22cfc3bcc6d9ddb555581fa8bb14c8e2b04977be/components/prism-docker.js#L39) and [`operator`](https://github.com/PrismJS/prism/blob/22cfc3bcc6d9ddb555581fa8bb14c8e2b04977be/components/prism-docker.js#L57). AFAIK, a _line continuation character_ is just an [escape character](https://docs.docker.com/engine/reference/builder/#escape), so the `char` key is more appropriate than the `operator` key.

> Also, thanks for looking at this!

You are welcome!

RunDevelopment:
> **The first part. Thoughts on the `command` alias.**

> I disagree with [this part](https://github.com/PrismJS/prism/blob/22cfc3bcc6d9ddb555581fa8bb14c8e2b04977be/components/prism-docker.js#L36-L39), because `ADD`, `ARG`, `CMD`, `ENTRYPOINT`, `RUN`, etc. are not called commands in the [docs](https://docs.docker.com/engine/reference/builder/).

Agreed. I thought that `command` was previously used but nope, it wasn't. My bad. I'll remove it.

> That’s why I suggest the function key.

I don't think that's a good idea, purely for practical reasons. 

First of all, from your writing, it's not clear to me whether you suggest this also for the `instruction` rule or the `keyword` rules. Though it matters not because it's not a good idea in either case but for different reasons.

This `instruction` rule should not have a `function` alias because then the whole instruction will be highlighted in one color. This docker file can only contain comments and instructions, all text will be given a color different from the default text color. This is simply too much coloring.

The `keyword` rules should not have a `function` alias because of how themes implement coloring. Themes do not expect a token to have two _supported_ class names. 
Example: The default theme defines a color for [`keyword` tokens](https://github.com/PrismJS/prism/blob/04ef309c3045c7f23c78ce36cea72dec945edd6c/themes/prism.css#L115) and [`function` tokens](https://github.com/PrismJS/prism/blob/04ef309c3045c7f23c78ce36cea72dec945edd6c/themes/prism.css#L119). If a token has both classes, the color of the `function` token will be used because the CSS `function` rule was defined after the `keyword` rule. Other themes may define these rules in a different order, so highlighting between themes will likely not be consistent.

Maybe we could add something like `instruction-name` for those keywords?

---

> **The second part. Thoughts on `FROM...AS` and `HEALTHCHECK CMD`.**

You gave a very good explanation. I still think that they should be highlighted as keywords for two reasons:

1. They are mentioned as literal strings in the BNF grammar of their respective instructions.
2. It looks better. Syntax highlighting is supposed to help humans read code and I think that highlighting them as keywords helps to achieve that goal. While it may not be an accurate categorization, it is beneficial.

---

> a line continuation character is just an escape character

No, it isn't. Line continuations are removed. This is indirectly mentioned in the "Note on whitespace" in the [format section](https://docs.docker.com/engine/reference/builder/#format).

<details>

Let's look at the following docker file:

```dockerfile
FROM debian:buster

RUN echo "\
hello\
world"
```

It will result in the following output:

```
C:\fake\path>docker build .
Sending build context to Docker daemon  2.048kB
Step 1/2 : FROM debian:buster
 ---> e7d08cddf791
Step 2/2 : RUN echo "helloworld"
 ---> Running in aa37a128ebf0
helloworld
Removing intermediate container aa37a128ebf0
 ---> 969127bc2810
Successfully built 969127bc2810
```

As you can see, the `RUN` instruction is parsed as `RUN echo "helloworld"`.

This actually goes even further with the now deprecated empty continuation lines. The following docker file is equivalent to the above one:

```dockerfile
FROM debian:buster

RUN echo "\


hello\


world"
```

</details>

While the name `operator` is debatable, I did choose it for practical reasons. `operator` is widely supported by our themes and makes it clear that `\`s have a very special meaning in arguments.

edukisto:
> First of all, from your writing, it's not clear to me whether you suggest this also for the `instruction` rule or the `keyword` rules.

That’s true. To be clear:

```dockerfile
FROM python:latest AS bleeding_edge_python
```

generates

```html
<span class="token instruction command">
  <span class="token keyword">FROM</span>
  python:latest
  <span class="token keyword">AS</span>
  bleeding_edge_python
</span>
```

The `instruction` and these `keyword`’s are highlighted absolutely correctly. The only thing I disagree is `command`. I’ve been thinking about making the whole instruction a `function` instead of a `command`. It would imply a function (operation) call.

> This instruction rule should not have a `function` alias because then the whole instruction will be highlighted in one color.

Sad, but true :(

> The `keyword` rules should not have a `function` alias...

Yes, I’ve been thinking about making `FROM` an `operator` instead of a `keyword`, but it’s not an operator. I’ve been thinking about making `FROM` a `function` instead of a `keyword`, but it’s not a function itself, but a function/operation name in a call. I gave up these ideas yesterday, but didn’t explain it clearly enough.

> Maybe we could add something like `instruction-name` for those keywords?

Marking of `FROM`, `HEALTHCHECK`, etc. with `token instruction-name keyword` classes is a good idea!

> I still think that they should be highlighted as keywords...

Yes, `AS` in `FROM`, `CMD` and `NONE` in `HEALTHCHECK` are definetely arguments **and keywords**. I don’t suggest any changes in this part since [22cfc3b](https://github.com/PrismJS/prism/pull/2720/commits/22cfc3bcc6d9ddb555581fa8bb14c8e2b04977be). I just confirmed that `keyword` is the right choice.

> > a line continuation character is just an escape character
>
> No, it isn't. 

The [docs](https://docs.docker.com/engine/reference/builder/#escape) say it is :) 

> The `escape` directive sets the **character** used to **escape characters** in a Dockerfile.

By the way, we can redefine the escape character! A backtick (`) should be added [here](https://github.com/PrismJS/prism/blob/22cfc3bcc6d9ddb555581fa8bb14c8e2b04977be/components/prism-docker.js#L88).

A single backslash (or a single tick) escapes a single newline and doesn’t affect anything else. The behaviour is similar to the [line continuation escape character](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_02_01) in the POSIX shell language:

> The `<backslash>` and `<newline>` shall be removed...

For example,

```dockerfile
# escape=`
FROM alpine
COPY `t`
`e`
`s`
`t`
`.`
`t`
`x`
`t .
```

makes Docker search for `test.txt`. So I can’t see any difference between escaping newlines and escaping other characters.

> This actually goes even further with the now deprecated empty continuation lines...

Well, this is a good point. It makes backslash/backtick work as an “empty line removing” operator.

RunDevelopment:
> > > a line continuation character is just an escape character
> > 
> > No, it isn't.
> 
> The [docs](https://docs.docker.com/engine/reference/builder/#escape) say it is :)

Ahh, my bad. I read "escape character" as "character escape".

Yes, while the line continuation character is also the normal escape character, a line continuation is not a character escape. So I think it deserves special highlighting.

> By the way, we can redefine the escape character! 

I know and I hate it.

> A backtick (`) should be added [here](https://github.com/PrismJS/prism/blob/22cfc3bcc6d9ddb555581fa8bb14c8e2b04977be/components/prism-docker.js#L88).

Not just there. The entire instruction parsing has to be adjusted. It sounds dramatic but it's really just replacing all `\` characters with a ``[\\`]`` character class, so it's quite simple.

But I don't know whether we should actually do that. Doing the simple replacement means that we will get false positives for either escape mode. Rn, we don't support backticks but backslashes are handled correctly. 

I have never seen backticks being used in the wild and deemed them as unimportant. Their expressed purpose is to help writing docker files for Windows VMs but I haven't found any Windows docker files on Dockerhub that actually use them. Do we _have to_ support them?

edukisto:
> ...false positives for either escape mode.

Ah, I see... Now we have _two_ syntaxes: `dockerfile-backticks` and `dockerfile-backslashes`.

> Do we _have_ to support them?

Well, Microsoft [mentions](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-docker/manage-windows-dockerfile#escape-character) the `escape` directive, but yes, it turns out that they consistently use forward slashes as a directory separator in their Dockerfiles.

The work is not worth it.

Thank you very much for an interesting discussion!

RunDevelopment:
> Now we have two syntaxes

Exactly. Twice the work because of a feature nobody uses.

---

Thank you for the review and in-depth analysis!

PR #2723, opened 2021-01-18
Make copy-to-clipboard configurable with multiple attributes

### LINKED ISSUE #1438: Allow copy-to-clipboard texts to be translated
The [Copy to clipboard](https://prismjs.com/plugins/copy-to-clipboard/) plugin can show 3 _English_ messages to the user. There is no way to replace these messages with localized versions right now.

The goal of this issue is to add a mechanism to allow these messages to be translated.

---

See #1307 for details / approach.

edukisto:
Resolves #1438.

Makes the `copy-to-clipboard` plugin consider the following HTML attributes:

- `data-prismjs-copy`,
- `data-prismjs-copy-error`,
- `data-prismjs-copy-success`,
- `data-prismjs-copy-timeout`.


plugins/copy-to-clipboard/prism-copy-to-clipboard.js
```diff
@@ -35,13 +35,39 @@
 		head.appendChild(script);
 	}
 
+	function getSettings(start, defaults) {
```
RunDevelopment, line 38: From the method signature, I would have guessed that `defaults` is a readonly value. As it is right now, `getSettings` takes the `defaults` object, modifies it, and then returns it.

Either make this a pure function or don't return anything (and rename the function to better reflect its actual behaviour).

Also, please add a little JSDoc to explain the function. (Maybe even some types)

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
```diff
@@ -35,13 +35,39 @@
...
+		var settings = defaults;
+		for (var key in settings) {
+			var attr = prefix + key;
+			var element = start;
+			while (element && !element.hasAttribute(attr)) {
+				element = element.parentElement;
+			}
+			if (element !== null) {
+				var value = element.getAttribute(attr);
+				if (typeof value !== 'undefined') {
```
RunDevelopment, line 49: [`getAttribute`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute) will always return `null` or a string. It cannot return `undefined`.

The check is unnecessary anyway because we know that `element.hasAttribute(attr)` is `true` which means that `element.getAttribute(attr)` will return a string.

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
```diff
@@ -35,13 +35,39 @@
...
+		}
+		return settings;
+	}
+
 	Prism.plugins.toolbar.registerButton('copy-to-clipboard', function (env) {
+		var defaults = {
+			'copy': 'Copy',
+			'copy-error': 'Press Ctrl+C to copy',
+			'copy-success': 'Copied!',
+			'copy-timeout': 5000
```
RunDevelopment, line 62: Do we really need this? 

Also, `getSettings` doesn't parse numbers, so `setTimeout(..., settings['copy-timeout'])` may throw an error since [it's only defined for numbers](https://html.spec.whatwg.org/multipage/webappapis.html#windoworworkerglobalscope-mixin:dom-settimeout). What do we do if the attribute values isn't a valid number?

Maybe let's remove this option for now and add it later if needed.

RunDevelopment (changes requested):
Thank you for the PR @edukisto!

A few nits. Also, please add some documentation on how to use the feature you implemented.

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
mAAdhaTTah: We can simplify this further by moving the `defaults` up here and using them in scope. Agree with everything else.

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
mAAdhaTTah: FWIW, in Chrome at least, a numeric string works fine. A non-numeric string will run instantly. That said, I'm ok cutting it as a setting (cuz I don't think it'll need to be manipulated often) but if we decide to keep it, I'd be ok w/ no validation and users would need to make sure they do it right.

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
edukisto: OK, I’ll make it pure in a backward-compatible manner.

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
edukisto: Thanks! It should’ve been `null`. I’ll remove this line.

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
mAAdhaTTah: I'm ok with keeping it.

edukisto:
I had to replace `hyphenated-keys` with `underscored_keys` to make keys compatible with the `@param` syntax.

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
RunDevelopment: > We can simplify this further by moving the defaults up here and using them in scope.

I'm also for that. This makes the function even simpler.

```js
/**
 * Traverses up the DOM tree to find data attributes that override the default plugin settings.
 *
 * @param {Element} startElement An element to start from.
 * @returns {Settings} The plugin settings.
 *
 * @typedef {Record<"copy" | "copy-error" | "copy-success" | "copy-timeout", string | number>} Settings
 */
function getSettings(startElement) {
	/** @type {Settings} */
	var settings = {
		'copy': 'Copy',
		'copy-error': 'Press Ctrl+C to copy',
		'copy-success': 'Copied!',
		'copy-timeout': 5000
	};

	var prefix = 'data-prismjs-';
	for (var key in settings) {
		var attr = prefix + key;
		var element = startElement;
		while (element && !element.hasAttribute(attr)) {
			element = element.parentElement;
		}
		if (element) {
			settings[key] = element.getAttribute(attr);
		}
	}

	return settings;
}
```

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
```diff
@@ -1,4 +1,10 @@
-(function(){
+/**
+ * A DOM element.
+ * @external Element
```
RunDevelopment, line 3: This isn't necessary. You can just use `Element` from TypeScript's DOM lib. (See the `getSettings` version [here](https://github.com/PrismJS/prism/pull/2723/files#r560946722).)

RunDevelopment (changes requested):
Please add a few words on the [Copy to Clipboard plugin page](https://prismjs.com/plugins/copy-to-clipboard/) to explain how to use the new localization method. Maybe make it a new section? It doesn't have to be much. Just explain that these attributes are inherited from ancestor nodes, what attributes there are, and what they do.

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
edukisto: OK, I’ll follow the way.

Just one more point. `jsdoc` doesn’t parse this tag.

```javascript
/**
 * @typedef {Record<"copy" | "copy-error" | "copy-success" | "copy-timeout", string | number>} Settings
 */
```
`jsdoc` doesn’t parse this either.

```javascript
/**
 * @typedef {{"copy": String, "copy-error": String, "copy-success": String, "copy-timeout": Number}} Settings
 */
```

The only way to generate a list of properties is to use a complex type definition (the last example from [the table](https://jsdoc.app/tags-type.html#overview)):

```javascript
/**
 * @typedef Settings
 * @type {Object}
 * @property {String} copy A copy-to-clipboard message.
 * @property {String} copy-error A copying error message.
 * @property {String} copy-success A message for a successful copying.
 * @property {Number} copy-timeout A time-out for a state changing, ms.
 */

/** @type {Settings} */
```

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
edukisto: Got it. I just intended to separate the ECMAScript types (`Number`, `String`, `Object`) from the DOM types.

edukisto:
> Please add a few words on the Copy to Clipboard plugin page...

Sure, I remember [the first request](https://github.com/PrismJS/prism/pull/2723#pullrequestreview-570439373).

Thank you very much for your advices!

plugins/copy-to-clipboard/prism-copy-to-clipboard.js
RunDevelopment: Yes, JSDoc has a history of being slow to adopt TypeScript-style syntax and common TypeScript types (e.g. `Record`).

I personally don't worry about writing JSDoc-compatible JSDoc annotation as long as TS can parse them. JSDoc is just a static site generator and isn't even used for this file specifically. TypeScript, on the other hand, can use these types to detect mistakes and provide code completion.

Also, JSDoc might be able to parse it if we replace `Record<X, Y>` with `Object<X, Y>`.

plugins/copy-to-clipboard/index.html
```diff
@@ -20,14 +20,165 @@
...
+<section>
+	<h1>Settings</h1>
+
+	<p>By default, the plugin shows messages in English and sets a 5-second timeout after a click. You can use the following HTML5 data attributes to override the default settings:</p>
+
+	<ul>
+		<li><code class="token attr-name">data-prismjs-copy</code> — a copy-to-clipboard message;</li>
+		<li><code class="token attr-name">data-prismjs-copy-error</code> — a copying error message;</li>
+		<li><code class="token attr-name">data-prismjs-copy-success</code> — a message for a successful copying;</li>
+		<li><code class="token attr-name">data-prismjs-copy-timeout</code> — a timeout (in milliseconds) after copying. The value should be a non-negative integer.</li>
```
RunDevelopment, line 41: People that don't know the message cycle probably won't understand the settings from those descriptions alone. How about this:

```suggestion
		<li><code class="token attr-name">data-prismjs-copy</code> — default message displayed by Copy to Clipboard;</li>
		<li><code class="token attr-name">data-prismjs-copy-error</code> — a message displayed after failing copying, prompting the user to press <code>Ctrl+C</code>;</li>
		<li><code class="token attr-name">data-prismjs-copy-success</code> — a message displayed after a successful copying;</li>
		<li><code class="token attr-name">data-prismjs-copy-timeout</code> — a timeout (in milliseconds) after copying. Once the timeout passed, the success or error message will revert back to the default message. The value should be a non-negative integer.</li>
```

RunDevelopment (changes requested):
Very good and very detailed documentation! Yay!

A few minor nits and we're ready to go from what I see.

plugins/copy-to-clipboard/index.html
```diff
@@ -20,14 +20,165 @@
...
+		data-prismjs-copy-success="Скопировано!"
+	>
+		<pre><code class="language-c">int main() {
+	return 0;
+}</code></pre>
+
+		<pre><code class="language-js">console.log('Hello, world!');</code></pre>
+	</div>
+
+	<p>The next HTML document is in English, but some code blocks show messages in Russian and simplified Mainland Chinese. The other settings are set by default.</p>
```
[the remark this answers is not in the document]
edukisto: Right.

plugins/copy-to-clipboard/index.html
```diff
@@ -20,14 +20,165 @@
...
+&lt;/html&gt;</code></pre>
+
+	<p>Output:</p>
+
+	<div>
+		<pre><code class="language-js">console.log('Hello, world!');</code></pre>
+
+		<pre
+			lang="ru"
+			data-prismjs-copy="Скопировать"
```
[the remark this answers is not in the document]
edukisto: I do speak Russian, but not Chinese. I did my best to translate the messages into Chinese.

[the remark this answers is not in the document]
RunDevelopment: Ok. I was asking because I speak neither of those. Google-translating is all I can do to review here.

RunDevelopment (approved):
LGTM;

@mAAdhaTTah What about you?

mAAdhaTTah:
@RunDevelopment I'm into this but need some time to review. Will try to get to it this evening.

plugins/copy-to-clipboard/index.html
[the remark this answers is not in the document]
edukisto: My English-to-Russian translations match up with Google translations, except _Copy_ -> _Копировать_. If you don't mean a continuous (possibly repetitive) action, then it’s more natural to use the perfective aspect (_**С**копировать_). Although this is weak evidence, the result of back-translation matches the original text.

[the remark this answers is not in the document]
edukisto: 1. 复制 means “Copy”. You can see it in Word and [LibreOffice](https://github.com/LibreOffice/translations/blob/master/source/zh-CN/sw/messages.po#L25269-L25270) (be careful, the file is really huge). 复制链接 [means](https://github.com/LibreOffice/translations/blob/master/source/zh-CN/sw/messages.po#L25281-L25282) “Copy link”. 复制标题 [means](https://github.com/LibreOffice/translations/blob/master/source/zh-CN/sw/messages.po#L26463-L26464) “Copy heading”. 文本 [means](https://github.com/LibreOffice/translations/blob/master/source/zh-CN/sw/messages.po#L6814-L6815) “Text”. “Copy text” is most likely translated as 复制文本.

2. “按Ctrl+C复制” (按 means “press”, 复制 means “copy”) is googleable: [1](https://context.reverso.net/translation/english-chinese/Press+Ctrl+C+to+copy), [2](https://zhuanlan.zhihu.com/p/345036232), [3](https://cn.tidewaterschool.org/650602-how-to-select-copy-current-PJBHWV).

3. “你好，世界！” is directly taken from [Wikipedia](https://zh.wikipedia.org/wiki/Hello_World).

However, it is possible that I was mistaken with the translation of the message “Copied!”. I’ll recheck it today.

plugins/copy-to-clipboard/index.html
```diff
@@ -20,19 +20,171 @@
 
 <section>
 	<h1>How to use</h1>
-	<p>In addition to including the plugin file with your PrismJS build, ensure <a href="https://clipboardjs.com/">Clipboard.js</a> is loaded before the plugin.</p>
 
-	<p>The simplest way to include Clipboard.js is to use any of the
-		<a href="https://github.com/zenorocha/clipboard.js/wiki/CDN-Providers">recommended CDNs</a>. If you're using Browserify, Clipboard.js will be loaded automatically
+	<p>The plugin depends on <a href="https://clipboardjs.com/">clipboard.js</a> and the Prism <a href="https://prismjs.com/plugins/toolbar/">Toolbar</a> plugin. In addition to including the plugin file with your PrismJS build, ensure they are loaded before the plugin.</p>
```
mAAdhaTTah, line 24: Do we normally mention internal dependencies like this? Should we going forward? If we do, maybe we should link to the explanation of how that works in the main page?

plugins/copy-to-clipboard/index.html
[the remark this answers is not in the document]
mAAdhaTTah: @edukisto If you need some help translating, I have a Russian friend I could ask.

mAAdhaTTah (approved):
I love this. I think we're just waiting on the verification of the Russian "Copied!" translation and then we're gtg.

plugins/copy-to-clipboard/index.html
RunDevelopment: > internal dependencies

Toolbar or clipboard.js? 

Copy to clipboard will automatically load clipboard.js, so it's essentially just an implementation detail. I don't think it should be mentioned here.

As for Toolbar, all of Prism's components can have dependencies. We have Autoloader for languages but we don't offer an automatic way to load plugins right now. People that load plugins from CDN's by including files have to know this.

mAAdhaTTah: > Toolbar or clipboard.js?

The Toolbar.

> Copy to clipboard will automatically load clipboard.js, so it's essentially just an implementation detail. I don't think it should be mentioned here.

lol I was actually thinking the opposite. We don't include explanations of required internal dependencies in our docs generally because (imo) our expectation is they'll use one of the tools (website download or babel plugin) to generate their bundle. As an example, the previewers plugin [requires css-extras](https://github.com/PrismJS/prism/blob/master/components.json#L1398) but we don't mention it in the [docs](https://github.com/PrismJS/prism/blob/master/plugins/previewers/index.html). Clipboard.js is an external dependency which wouldn't be resolved by either of those approaches so imo should be mentioned.

RunDevelopment: > our expectation is they'll use one of the tools (website download or babel plugin)

> As an example, the previewers plugin requires css-extras but we don't mention it in the docs.

True but for plugins, we only have download and babel plugin. None work in combination with CDNs.

The best way to solve this might be to make Autoloader support loading plugins. We don't have to mention dependencies if Autoloader will handle them for our users.

> Clipboard.js is an external dependency which wouldn't be resolved by either of those approaches so imo should be mentioned.

But nobody has to resolve it in the first place because [the plugin itself will load clipboard.js](https://github.com/PrismJS/prism/blob/fd1081d256f2355b1c6f3690ea50282a9a907502/plugins/copy-to-clipboard/prism-copy-to-clipboard.js#L34). Why mention something nobody has to worry about?

mAAdhaTTah: > The best way to solve this might be to make Autoloader support loading plugins.

I'm open to this, although the autoloader functions by checking the language. We'd have to add a setting or data attribute to indicate plugins to load.

> But nobody has to resolve it in the first place because the plugin itself will load clipboard.js. Why mention something nobody has to worry about?

Because loading the plugin remotely is bad for performance compared to bundling it with the plugin, nor is it clear that using this plugin means it will be automatically loading additional scripts from external locations, which could have security implications. For both of these reasons, I definitely think it should be mentioned here.

mAAdhaTTah: I also want to be clear that this discussion shouldn't block the merge (once the Russian text is validated).

RunDevelopment: > security implications

True. That alone is enough to keep it.

> We'd have to add a setting or data attribute to indicate plugins to load.

That was my first idea as well. I'll make an issue.

> this discussion shouldn't block the merge

Definitely. 

---

Also, coming back to the original question: Do we want to keep the mentioning of Toolbar?

I don't have any hard feelings on that subject. I said that people need this information but we haven't had any people making issues about it yet, so it might not be a problem? Not sure.

mAAdhaTTah: I'm ok with leaving it.

plugins/copy-to-clipboard/index.html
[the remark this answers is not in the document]
edukisto: Thank you, @mAAdhaTTah, I know Russian very well, so Russian messages are correct for sure.

I got a little stuck with Chinese messages. I double-checked them today and now I have some evidence that they are correct too. (Be careful, linked files are huge.)

So, 文档已保存 [translates as](https://github.com/LibreOffice/translations/blob/cbe7554c6a4fddd40da90728699d7ff5d95cab89/source/zh-CN/helpcontent2/source/text/shared/guide.po#L3184-L3185) “The document is saved”. 文档 [translates as](https://github.com/LibreOffice/translations/blob/cbe7554c6a4fddd40da90728699d7ff5d95cab89/source/zh-CN/sw/messages.po#L13795-L13796) “document”. 保存 [translates as](https://github.com/LibreOffice/translations/blob/cbe7554c6a4fddd40da90728699d7ff5d95cab89/source/zh-CN/sw/messages.po#L17760-L17761) “save”. I have had doubts about using 已 or 已经 (we can see it [here](https://github.com/LibreOffice/translations/blob/cbe7554c6a4fddd40da90728699d7ff5d95cab89/source/zh-CN/cui/messages.po#L681-L682)), but then I found [this forum](https://web.archive.org/web/20210130220516/https://forum.wordreference.com/threads/%E5%B7%B2-%E5%B7%B2%E7%BB%8F.2861796/). It clarifies that 已 means “have/has been”, whereas 已经 means “have/has **already** been”. That’s why “The text is copied!” translates as “文本已复制！” ([here](https://github.com/PrismJS/prism/blob/f1cf15771ef8ff9b73e5ffb6831a253472b5b488/plugins/copy-to-clipboard/index.html#L179)).

OK, **the conclusion**. I’m almost sure that the Chinese translation is correct. Don’t worry about the Russian translation :)

plugins/copy-to-clipboard/index.html
RunDevelopment: Me too.

plugins/copy-to-clipboard/index.html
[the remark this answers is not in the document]
RunDevelopment: I think it's good then. The translators I tried all gave back what you'd expect. You cross-checked with multiple different sources and it all works out. 

I think it's correct. 

(And if it isn't then someone will get an easy _Good First Issue™_.)

RunDevelopment:
@mAAdhaTTah You think it's good to go?

mAAdhaTTah:
@RunDevelopment Yup! :)

@edukisto Thank you contributing!

edukisto:
@RunDevelopment and @mAAdhaTTah, thank you!

PR #2731, opened 2021-01-20
fix: bug #2614 class-name is now matched before keyword

### LINKED ISSUE #2614: [PHP] When code begins with class, class-name after it is not matched.
**Information**
- Language: PHP
- Plugins: none

**Description**
When code begins with `class | interface | trait` the class-name after it is not matched.
When an whitespace is added before `class` the class-name is matched correctly.

I know that correct php code starts with `<?php` and then first class-name is matched correctly.
But opening tag is often dropped in code snippets fot better readability (blogs, docs, ..) and then this is an issue.

**Code snippet**

<details>
<summary>The code being highlighted incorrectly. Class name `First` is not matched.</summary>

```
class First {}
class Second {}
```

</details>

TomPavelec:
Closes #2614


components/prism-php.js
```diff
@@ -33,6 +33,14 @@
 				'punctuation': /\\/
 			}
 		},
+		'class-name-definition': [
+			{
+				pattern: /(\b(?:class|interface|trait)\s+)\b[a-z_]\w*(?!\\)\b/i,
+				greedy: true,
```
RunDevelopment, line 39: I don't think greedy is necessary, is it?

components/prism-php.js
```diff
@@ -33,6 +33,14 @@
 				'punctuation': /\\/
 			}
 		},
+		'class-name-definition': [
```
RunDevelopment, line 36: The array isn't necessary.

components/prism-php.js
```diff
@@ -33,6 +33,14 @@
...
+			{
+				pattern: /(\b(?:class|interface|trait)\s+)\b[a-z_]\w*(?!\\)\b/i,
```
RunDevelopment, line 38: With `class|interface|trait` covered, you can remove them from the first `class-name` pattern. 

(GH doesn't allow me to make a comment there, so I'm making it here.)

RunDevelopment (changes requested):
Thank you for the PR @TomPavelec!

A few nits. Also, could you please add a test for the code snippet in #2614 to show that the bug has been fixed?

components/prism-php.js
TomPavelec: Fixed

components/prism-php.js
TomPavelec: Removed

components/prism-php.js
TomPavelec: Yes, I think that greedy can be dropped.

TomPavelec:
Thank you for quick response @RunDevelopment.

RunDevelopment (approved):

RunDevelopment:

Issue #2665, opened 2020-12-03
Hook request: "highlightelement-start"


LeaVerou:
I was just explaining to someone that even though Prism does not have an autodetect language, they can write a plugin to decide between the two languages they need to highlight, when a language is not specified. However, when I tried to find a good hook for this, I failed. 

Should we add a hook in the beginning of `Prism.highlightElement()`?

RunDevelopment:
I'm generally for this idea but

> when a language is not specified

this could be a problem. Right now, the language `none` stands for no language as in "I don't know the language" and as in "it's just plain text". We should split `none` into `none` (= "I don't know the language" or "language not specified") and `plain` (= plain text). Otherwise, language detectors will always be active for pain text.

Example:

```html
<!-- language is 'none' => try detect language -->
<code>int main() { return 0; }</code>

<!-- language is 'none' because it's plain text -->
<code class="language-none">This is *very* important.</code>
```

Say that "not specified" means that there is no `language-xxxx` class doesn't work because of inheritance. Users might use inheritance to set a page-wide default language and want to use auto-detection for a specific code block.

> Should we add a hook in the beginning of Prism.highlightElement()?

Where is "in the beginning"? Before or after we get the `textContent` ;)

From what I see, the easiest way to implement this would be to reorder `highlightElement` a little.

1. Use `getLanguage` to determine the language.
1. Get the `textContent`.
1. Create the `env` object.
1. Call this new hook.
1. Set the `className`s.
1. Rest as it is now starting with the `before-sanity-check` hook.

> "highlight-start"

From my perspective, this hook is specifically there to determine/adjust the language of an element. Why not call it something like "element-language" or just "language"?

LeaVerou:
> I'm generally for this idea but
> 
> > when a language is not specified
> 
> this could be a problem. Right now, the language `none` stands for no language as in "I don't know the language" and as in "it's just plain text". We should split `none` into `none` (= "I don't know the language" or "language not specified") and `plain` (= plain text). Otherwise, language detectors will always be active for pain text.
> 
> Example:
> 
> ```
> 
> <code>int main() { return 0; }</code>
> 
> 
> <code class="language-none">This is *very* important.</code>
> ```
> 
> Say that "not specified" means that there is no `language-xxxx` class doesn't work because of inheritance. Users might use inheritance to set a page-wide default language and want to use auto-detection for a specific code block.

[...]

PR #2738, opened 2021-01-25
Added support for plaintext

RunDevelopment:
This came up in #2665:

Right now, we use `language-none` (or just `none`) to mean "no language" but that has actually two meanings. "No language" could mean either plaintext or that the language is undefined/unknown. These two distinct concepts are both represented by `none`.

This is a problem when language detection comes into play. A code block with no (defined) language will get language `none` by default, so the language detector can do its job. But what happens for a plaintext code block? The user will use the `language-none` class to indicate plaintext. But these two cases are indistinguishable from the perspective of the language detector.

The simple solution for this problem is to split `none` into two languages: `none` and `plain`. 
From now on, `none` only represents an undefined/unknown language and `plain` will represent plaintext.

This solves the problem has some additional advantages as well. Since `none` has no grammar, plugins like [Show Invisibles](https://prismjs.com/plugins/show-invisibles/) cannot be used. `plain` will have an empty grammar, so plugins like SI will work.

---

I'm also thinking about giving `plain` some aliases like `plaintext`, `plain-text`, `text`, and `txt` but this will make the implementation less simple.


joshgoebel:
> I'm also thinking about giving plain some aliases like plaintext, plain-text, text, and txt but this will make the implementation less simple.

Just FYI in Highlight.js we named the grammar 'plaintext' and have aliases for `txt` and `text`.

RunDevelopment:
> > I'm also thinking about giving plain some aliases like plaintext, plain-text, text, and txt but this will make the implementation less simple.
> 
> Just FYI in Highlight.js we named the grammar 'plaintext' and have aliases for `txt` and `text`.

Thanks for the info! In that case, I'll add the aliases.

Issue #2735, opened 2021-01-25
Add support for PSL (PATROL Scripting Language)


bertysentry:
**Language**
PSL (PATROL Scripting Language) is the language used when programming for the BMC PATROL Agent.

**Additional resources**
[PSL Official Documentation at bmc.com](https://docs.bmc.com/docs/PATROLAgent/2008/patrol-script-language-overview-and-functions-938228568.html)
PSL has a C-like syntax and looks somehow similar to AWK with additional built-in functions.

Note: I will submit a PR to add support for this (creating the issue just so I reference it in my branch and PR)

PR #2739, opened 2021-01-25
Add support for PSL, PATROL Scripting Language (#2735)

bertysentry:
This adds support for PSL language, the scripting language used to write modules on the [PATROL Agent (BMC Software)](https://docs.bmc.com/docs/PATROLAgent/2008/patrol-script-language-overview-and-functions-938228568.html).

All tests pass :-)


bertysentry:
This resolves #2735

components/prism-psl.js
```diff
@@ -0,0 +1,28 @@
...
...ranges|get_text|gethostinfo|getpid|getpname|history_get_retention|history|index|int|is_var|intersection|isnumber|internal|in_transition|join|kill|length|lines|lock|lock_info|log|loge|log10|matchline|msg_check|msg_get_format|msg_get_severity|msg_printf|msg_sprintf|ntharg|num_consoles|nthargf|nthline|nthlinef|num_bytes|print|proc_exists|process|popen|printf|pconfig|poplines|pow|PslExecute|PslFunctionCall|PslFunctionExists|PslSetOptions|random|read|readln|refresh_parameters|remote_check|remote_close|remote_event_query|remote_event_trigger|remote_file_send|remote_open|remove|replace|rindex|sec_check_priv|sec_store_get|sec_store_set|set_alarm_ranges|set_locale|share|sin|sinh|sleep|sopen|sqrt|srandom|subset|set|substr|system|sprintf|sort|subset|snmp_agent_config|_snmp_debug|snmp_agent_stop|snmp_agent_start|snmp_h_set|snmp_h_get_next|snmp_h_get|snmp_set|snmp_walk|snmp_get_next|snmp_get|snmp_config|snmp_close|snmp_open|snmp_trap_receive|snmp_trap_ignore|snmp_trap_listen|snmp_trap_send|snmp_trap_raise_std_trap|snmp_trap_register_im|splitline|strcasecmp|str_repeat|trim|tail|tan|tanh|time|tmpnam|tolower|toupper|trace_psl_process|text_domain|unlock|unique|union|unset|va_arg|va_start|write)\b/,
```
RunDevelopment, line 24: Is this huge list really necessary? Wouldn't the approach languages like C and JavaScript take be better? Those languages use the function call syntax (`name()`) to detect function definitions and calls.

Not only will this drastically reduce the file size of the language, but it will also support user-defined functions.

This only has 2 problems:

1. This won't detect [`export function`](https://docs.bmc.com/docs/PATROLAgent/2008/export-938228605.html) statements, but that's easy to add (see suggested code).
2. [`foreach`](https://docs.bmc.com/docs/PATROLAgent/2008/foreach-938228607.html). This is difficult. We could either accept these false positives or we could try to match these variables before we match functions. Kinda like this:
	
	```js
	'foreach-variable': {
		pattern: /(\bforeach\s+(?:(?:\w+\b|"(?:\\.|[^\\"])*")\s+){0,2})[_a-zA-Z]\w*(?=\s*\()/,
		lookbehind: true,
		greedy: true
	},
	```
	I'm undecided as to what would be better.

```suggestion
	'function': [
		/\b[_a-z]\w*(?=\s*\()/i,
		{
			pattern: /(\bfunction\s+)[_a-zA-Z]\w*\b/,
			lookbehind: true
		}
	],
```

tests/languages/psl/heredoc_feature.test
```diff
@@ -0,0 +1,21 @@
+<<<ABC_1
+Text "string"
+# comment "string"
```
RunDevelopment, line 3:
```suggestion
Text "not a string"
# not a comment "not a string"
```

components/prism-psl.js
```diff
@@ -0,0 +1,28 @@
+Prism.languages.psl = {
+	'comment': [
+		{
+			pattern: /#.*/,
+			greedy: true
+		}
+	],
```
RunDevelopment, line 7: No need for the array. You can just inline the object literal.
```suggestion
	'comment': {
		pattern: /#.*/,
		greedy: true
	},
```

RunDevelopment (changes requested):
Thank you for the PR @bertysentry!

components/prism-psl.js
bertysentry: I really want to highlight these builtin functions, as they are different in nature from user-defined functions. These builtin functions are not part of a library made of user-defined functions, they are really part of the language definition itself.

I'm really in favor of highlighting these specifically. That's how it's currently done in VSCode, so that would be consistent.

bertysentry:
Thank you @RunDevelopment for the thorough and awesomely helpful review! I really appreciate it. I will take your suggestions into account and push the result later today.

components/prism-psl.js
```diff
@@ -0,0 +1,28 @@
...
+	'string': [{
+		pattern: /"(?:\\"|[^"\r\n]|[\r\n])*"/,
```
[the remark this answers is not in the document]
bertysentry: You, Sir, sure know your regular expressions! ;-)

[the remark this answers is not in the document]
bertysentry: Just to explain: the `\r\n` was introduced at some point in development because I wasn't sure the regex was running with "DOTALL" option.

components/prism-psl.js
```diff
@@ -0,0 +1,28 @@
...
+	'number': /\b(?:(?:0[xX][0-9a-fA-F]*)|(?:[0-9]+\.[0-9]+)|[0-9]+)\b/,
+	'operator': /[+\-~=\/*%&><\^!|\.]/,
+	'punctuation': /[\(\){}\[\]:;,?]/
```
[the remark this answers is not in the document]
bertysentry: Sorry about that newbie mistake!

components/prism-psl.js
RunDevelopment: I see. In that case, let's keep your current solution.

bertysentry: This is what it looks like in VSCode with the [psl-lang extension](https://marketplace.visualstudio.com/items?itemName=sentrysoftware.vscode-psl-lang&ssr=false#overview):
image

RunDevelopment: Oh, so the grammar for VSCode highlights them like this. In that case, let's do the same! 

I updated my suggestion. The new suggestion will tokenize all built-in functions as `builtin` and all user-defined functions as `function`. This should closely match VSCode's behaviour.

bertysentry:
@RunDevelopment Thank you again for the help on this! Code has been updated according to your suggestions.
(with one slight variation on user-defined functions)

components/prism-psl.js
```diff
@@ -0,0 +1,40 @@
...
+		lookbehind: true,
+		greedy: true
+	},
+	'user-function': {
+		pattern: /\b[_a-z][\w]*\b\s*\(/i,
+		alias: 'variable',
+		inside: {
+			'punctuation': /\(/
+		}
+	},
```
RunDevelopment, line 36: We won't have to tokenize the `(` if we use a lookahead.

```suggestion
	'user-function': {
		pattern: /\b[_a-z]\w*\b(?=\s*\()/i,
		alias: 'variable'
	},
```

components/prism-psl.js
```diff
@@ -0,0 +1,40 @@
...
+	'heredoc-string': {
+		pattern: /<<<([a-zA-Z_]\w*)[\r\n](?:.*[\r\n])*?\1\b/,
+		alias: 'string',
+		greedy: true
+	},
+	'keyword': /\b(?:__multi|__single|case|default|do|else|elsif|exit|export|for|foreach|function|if|last|line|local|next|requires|return|switch|until|while|word)\b/,
+	'constant': /\b(?:ALARM|CHART_ADD_GRAPH|CHART_DELETE_GRAPH|CHART_DESTROY|CHART_LOAD|CHART_PRINT|EOF|FALSE|False|false|NO|No|no|OFFLINE|OK|PSL_PROF_LOG|R_CHECK_HORIZ|R_CHECK_VERT|R_CLICKER|R_COLUMN|R_FRAME|R_ICON|R_LABEL|R_LABEL_CENTER|R_LIST_MULTIPLE|R_LIST_MULTIPLE_ND|R_LIST_SINGLE|R_LIST_SINGLE_ND|R_MENU|R_POPUP|R_POPUP_SCROLLED|R_RADIO_HORIZ|R_RADIO_VERT|R_ROW|R_SCALE_HORIZ|R_SCALE_VERT|R_SPINNER|R_TEXT_FIELD|R_TEXT_FIELD_LABEL|R_TOGGLE|TRIM_LEADING|TRIM_LEADING_AND_TRAILING|TRIM_REDUNDANT|TRIM_TRAILING|TRUE|True|true|VOID|WARN)\b/,
+	'variable': /\b(?:errno|exit_status|PslDebug)\b/,
+	'builtin-function': {
```
RunDevelopment, line 21: I not sure about the aliases you gave to built-in functions and user-defined functions. They are technically correct but they aren't consistent with how Prism uses them. In other languages, we use `builtin` to refer to either built-in types, functions, or (in rare cases) variables. Similarly, `variables` is usually used for _special_ variables like you did for `errno` and not "normal" variables and especially not function variables. They are always exceptions to these patterns but I try to keep the usage of token names consistent across languages.

For this language, I suggest the following: Keep the token names but change the aliases. The names are very descriptive, so I think they should be kept. 
`user-function` should definitely get a `function` alias and lose the `variable` alias. This will make it consistent with (most of) Prism's other languages.
I'm not sure about `builtin-function`. The `function` alias is fitting but so is `builtin`. I'm unsure as to whether `builtin` is best to be added as an alias or whether it should replace `function`.

Why did you choose these aliases in the first place?

components/prism-psl.js
bertysentry: No, you did not miss anything, you just have a better understanding of regexes than me ;-)

components/prism-psl.js
bertysentry: Good question!
At first, I went with the same as you suggested, i.e. `'builtin'` token for built-in functions, and `'function'` for user-defined functions.
The problem is that *builtin* and *string* tokens are colored the same way in most themes. So the semantic of a simple statement like below is not conveyed clearly (everything:
```perl
# PERL
print("Hello, World!");
```
Prism Default Theme, Dark and Funky:
image
image
image

Twilight and Tomorrow Night themes do use different colors, though:
image
image

(Note: in **prism-perl.js**, `print` is a *keyword*, not a *builtin*)

Anyway, so that's the reason why I bent the rules a little to have a better rendering. But I agree with your view on having a consistent display of various token types across different languages.

I you prefer me go revert back to "builtin" and "function", I will follow your suggestion (and make some CSS tweaks in my project). Just let me know!

RunDevelopment: Hmmm, `user-function`s should definitely have a `function` alias, I won't budge on that, but I see the problem with `builtin-function`s.

> (Note: in prism-perl.js, print is a keyword, not a builtin)

Does the `keyword` alias for `builtin-function` work for you?

Otherwise, I'd say that we just use the `builtin` and `function` alias for `builtin-function` and `user-function` respectively. The token names should make it rather simple for users to adjust themes.

bertysentry: I prefer to stick to `builtin` and `function` and not use `keyword` to avoid mixing things that are semantically different. If the user (like me) is unhappy with the color of a theme, it's pretty simple to customize the color for `builtin`.
Alright, it's settled, then! Pushing a new commit now.

components/prism-psl.js
```diff
@@ -0,0 +1,35 @@
...
...ranges|get_text|gethostinfo|getpid|getpname|history_get_retention|history|index|int|is_var|intersection|isnumber|internal|in_transition|join|kill|length|lines|lock|lock_info|log|loge|log10|matchline|msg_check|msg_get_format|msg_get_severity|msg_printf|msg_sprintf|ntharg|num_consoles|nthargf|nthline|nthlinef|num_bytes|print|proc_exists|process|popen|printf|pconfig|poplines|pow|PslExecute|PslFunctionCall|PslFunctionExists|PslSetOptions|random|read|readln|refresh_parameters|remote_check|remote_close|remote_event_query|remote_event_trigger|remote_file_send|remote_open|remove|replace|rindex|sec_check_priv|sec_store_get|sec_store_set|set_alarm_ranges|set_locale|share|sin|sinh|sleep|sopen|sqrt|srandom|subset|set|substr|system|sprintf|sort|subset|snmp_agent_config|_snmp_debug|snmp_agent_stop|snmp_agent_start|snmp_h_set|snmp_h_get_next|snmp_h_get|snmp_set|snmp_walk|snmp_get_next|snmp_get|snmp_config|snmp_close|snmp_open|snmp_trap_receive|snmp_trap_ignore|snmp_trap_listen|snmp_trap_send|snmp_trap_raise_std_trap|snmp_trap_register_im|splitline|strcasecmp|str_repeat|trim|tail|tan|tanh|time|tmpnam|tolower|toupper|trace_psl_process|text_domain|unlock|unique|union|unset|va_arg|va_start|write)\b/,
```
RunDevelopment, line 22: Just one last thing...

As I said, I really like the `builtin-function` name. It's very descriptive, so I think it should be included.

This should also make it easier for you to change the color of these tokens.
```suggestion
		pattern: /\b(?:acos|add_diary|annotate|annotate_get|asctime|asin|atan|atexit|ascii_to_ebcdic|batch_set|blackout|cat|ceil|chan_exists|change_state|close|code_cvt|cond_signal|cond_wait|console_type|convert_base|convert_date|convert_locale_date|cos|cosh|create|destroy_lock|dump_hist|date|destroy|difference|dget_text|dcget_text|ebcdic_to_ascii|encrypt|event_archive|event_catalog_get|event_check|event_query|event_range_manage|event_range_query|event_report|event_schedule|event_trigger|event_trigger2|execute|exists|exp|fabs|floor|fmod|full_discovery|file|fopen|ftell|fseek|grep|get_vars|getenv|get|get_chan_info|get_ranges|get_text|gethostinfo|getpid|getpname|history_get_retention|history|index|int|is_var|intersection|isnumber|internal|in_transition|join|kill|length|lines|lock|lock_info|log|loge|log10|matchline|msg_check|msg_get_format|msg_get_severity|msg_printf|msg_sprintf|ntharg|num_consoles|nthargf|nthline|nthlinef|num_bytes|print|proc_exists|process|popen|printf|pconfig|poplines|pow|PslExecute|PslFunctionCall|PslFunctionExists|PslSetOptions|random|read|readln|refresh_parameters|remote_check|remote_close|remote_event_query|remote_event_trigger|remote_file_send|remote_open|remove|replace|rindex|sec_check_priv|sec_store_get|sec_store_set|set_alarm_ranges|set_locale|share|sin|sinh|sleep|sopen|sqrt|srandom|subset|set|substr|system|sprintf|sort|subset|snmp_agent_config|_snmp_debug|snmp_agent_stop|snmp_agent_start|snmp_h_set|snmp_h_get_next|snmp_h_get|snmp_set|snmp_walk|snmp_get_next|snmp_get|snmp_config|snmp_close|snmp_open|snmp_trap_receive|snmp_trap_ignore|snmp_trap_listen|snmp_trap_send|snmp_trap_raise_std_trap|snmp_trap_register_im|splitline|strcasecmp|str_repeat|trim|tail|tan|tanh|time|tmpnam|tolower|toupper|trace_psl_process|text_domain|unlock|unique|union|unset|va_arg|va_start|write)\b/,
		alias: 'builtin-function'
```

bertysentry: You got it, chief! ;-)

components.json
```diff
@@ -910,6 +910,10 @@
 			"require": "clike",
 			"owner": "just-boris"
 		},
+		"psl": {
+			"title": "PSL",
```
RunDevelopment, line 914: Another thing: The title.

When googling "PSL" I could only find the [Property Specification Language](https://en.wikipedia.org/wiki/Property_Specification_Language), so let's not use the abbreviation.

```suggestion
			"title": "PATROL Scripting Language",
```

bertysentry:
@RunDevelopment All good? Again, thank you for the tremendous help! :-)

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @bertysentry!

bertysentry:
Thank you @RunDevelopment for your reactivity and your detailed and helpful review!

PR #2740, opened 2021-01-26
Adding Lawtex language support

LukasTCarvalho:
This adds support for the Lawtex language.


components/prism-lawtex.js
```diff
@@ -0,0 +1,32 @@
...
+      greedy: true
+    },
+    {
+      pattern: /(^|[^\\:])\/\/.*/,
+      lookbehind: true,
+      greedy: true
+    }
+  ],
+  'string': {
+    pattern: /(^|[^\\](?:\\\\)*)(")(?:\\[\s\S]|\$\([^)]+\)|\$(?!\()|`[^`]+`|(?!\2)[^\\`$])*\2/,
```
RunDevelopment, line 15:
```suggestion
    pattern: /(^|[^\\](?:\\\\)*)"(?:\\[\s\S]|\$\([^)]+\)|\$(?!\()|`[^`]+`|[^"\\`$])*"/,
```

components/prism-lawtex.js
```diff
@@ -0,0 +1,32 @@
...
+  'keyword': /\b(?:metainfo|head|body|foot|extra|where|extends|externs|template|list|struct|dependency|branch|topic|loop|period|if|else|elseif|foreach|print|use)\b/,
+  'important': /\b(?:this|value|id|mask|transient|itemsOnPage|language|style|type|default|description|tags|statement|overview|components|audience|inputs|functionalities|warnings|upgrades|title|name|label|order|mandatory|version|index|lower|upper|separator|request|element|help|tip|atomic|fields|loaders|options|key|operation|declarations|operations)\b/,
+  'logical': /\b(?:NOT|AND|OR|XOR|EXISTS|FORALL|IN)/,
+  'class-name': /\b(?:List|Object|String|Integer|Real|Boolean|Date|Time|Text|Document|Currency|Vector|Struct)\b/,
+  'function': /\b[A-Z].*?\b/,    
+  'boolean': /\b(?:true|false|null)\b/,
+  'operator': /(?:\[|\]|==|~=|<=|>=|<|>|\|)/,
+  'assignment': /=/,
+  'arithmetic': /(?:\-|\+|\&|\*|\/|%|\^)/,
+  'number': /(?:(?:0(?:x|X)[0-9a-fA-F]*)|(?:\+|-)?\b(?:(?:[0-9]+(?:\.[0-9]*)?)|(?:\.[0-9]+))(?:(?:e|E)(?:\+|-)?[0-9]+)?)(?:[LlFfUuDd]|UL|ul)?\b/
```
RunDevelopment, line 31: Changes:
1. `0x` probably isn't a valid number.
2. I added a `\b` before `0x`.
3. Made the whole regex ignore case.
4. Some other simplifications.

```suggestion
  'number': /(?:\b0x[0-9A-F]+|[+-]?\b(?:\d+(?:\.\d*)?|\.\d+)(?:E[+-]?\d+)?)(?:[LFUD]|UL)?\b/i
```

components/prism-lawtex.js
```diff
@@ -0,0 +1,32 @@
...
+  },
+  'keyword': /\b(?:metainfo|head|body|foot|extra|where|extends|externs|template|list|struct|dependency|branch|topic|loop|period|if|else|elseif|foreach|print|use)\b/,
+  'important': /\b(?:this|value|id|mask|transient|itemsOnPage|language|style|type|default|description|tags|statement|overview|components|audience|inputs|functionalities|warnings|upgrades|title|name|label|order|mandatory|version|index|lower|upper|separator|request|element|help|tip|atomic|fields|loaders|options|key|operation|declarations|operations)\b/,
+  'logical': /\b(?:NOT|AND|OR|XOR|EXISTS|FORALL|IN)/,
+  'class-name': /\b(?:List|Object|String|Integer|Real|Boolean|Date|Time|Text|Document|Currency|Vector|Struct)\b/,
+  'function': /\b[A-Z].*?\b/,    
+  'boolean': /\b(?:true|false|null)\b/,
+  'operator': /(?:\[|\]|==|~=|<=|>=|<|>|\|)/,
+  'assignment': /=/,
+  'arithmetic': /(?:\-|\+|\&|\*|\/|%|\^)/,
```
RunDevelopment, line 30: Is the distinction really necessary or could "operator" "assignment", "arithmetic", and "logical" be combined into one token? Is this useful for highlighting?

components/prism-lawtex.js
```diff
@@ -0,0 +1,32 @@
...
+    greedy: true,
+    inside: {
+      'escape': /\\./
+    }
+  },
+  'keyword': /\b(?:metainfo|head|body|foot|extra|where|extends|externs|template|list|struct|dependency|branch|topic|loop|period|if|else|elseif|foreach|print|use)\b/,
+  'important': /\b(?:this|value|id|mask|transient|itemsOnPage|language|style|type|default|description|tags|statement|overview|components|audience|inputs|functionalities|warnings|upgrades|title|name|label|order|mandatory|version|index|lower|upper|separator|request|element|help|tip|atomic|fields|loaders|options|key|operation|declarations|operations)\b/,
+  'logical': /\b(?:NOT|AND|OR|XOR|EXISTS|FORALL|IN)/,
+  'class-name': /\b(?:List|Object|String|Integer|Real|Boolean|Date|Time|Text|Document|Currency|Vector|Struct)\b/,
+  'function': /\b[A-Z].*?\b/,
```
RunDevelopment, line 26:
```suggestion
  'function': /\b[A-Z]\w*\b/,    
```

RunDevelopment (changes requested):
Thank you for the PR @looplexLukasC!

I did a quick review but I can't fully review your work without knowing the language. Could you please provide some documentation for Lawtex?

looplexLukasC:
Lawtex is a language developed by a Brazilian startup. The purpose of it is to make programming easier for law students or law professionals. The ideal of the startup is to make programming more accessible and facilitate the creating of programs that write and format legal documents. 

Since it is a Brazilian startup all the documentation that I have concerning the language is in Portuguese, so I'm not sure what I should share here.

components/prism-lawtex.js
looplexLukasC: Sorry I'm new to programming so I'm not sure of all the changes you have suggested. I'll try my best to evaluate and reply to them. 
This does seem to make the expression simpler since the language only allows for double quotes to open a string. Thank you.

components/prism-lawtex.js
looplexLukasC: The sintax highlighter used by the language (programmed into a Sublime package) considers `0x` a valid number even if it is appended to the end of a word (not fulfilling the `\b` requirement); however, I cannot think of a situation in which this would be used. So I would not be against the simplifications you have proposed.

components/prism-lawtex.js
looplexLukasC: No it is not! I simply copied the highlighter currently used. For some reason these three are declared separately but attributed to the same color.  Thank you.

components/prism-lawtex.js
looplexLukasC: Works perfectly!

RunDevelopment:
I'll close this now. Without documentation, I cannot review this. There's also the problem that searching for "lawtex" only yields a popular Tex package.

PR #2752, opened 2021-02-06
[PHP] Add support for `enum` in class-name_feature context

Ayesh:
[PHP 8.1 adds Enum support](https://php.watch/versions/8.1/enums#enum-syntax). This PR updates the class-like structure matching regex and its tests to accommodate the new `enum` structures.

Related: #2731


Ayesh:
I will keep this PR as a draft until the voting period for Enum proposal ends. It is very likely to pass, but I'd like to wait for the final vote off. Thank you.

tests/languages/php/class-name_feature.test
```diff
@@ -195,6 +197,13 @@ class Foo extends \Package\Bar implements App\Baz {}
 	]],
 	["punctuation", "{"],
 	["punctuation", "}"]
+
+	["keyword", "enum"],
+	["class-name-definition", "Foo"],
+	["keyword", "implements"],
+	["class-name", "Bar"],
+	["punctuation", "{"],
+	["punctuation", "}"],
```
RunDevelopment, line 206: The test token stream has to be valid JSON, so this comma is not allowed and there is a comma missing after `["punctuation", "}"]` before `["keyword", "enum"]`.

```suggestion
	["punctuation", "}"]

```

Ayesh: Thanks a lot for this. I think was the reason why some tests failed in CI. I force-pushed with this fixed🙏🏼 .

RunDevelopment (approved):
LGTM;
Looking forward to having this in Prism once the draft is through.

Ayesh:
RFC vote ended with 44:7 in favor. PHP 8.1 will have Enums 🎉🎊.

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @Ayesh!

PR #2755, opened 2021-02-11
Add support for Idris

KeenS:
This adds support for [Idris](https://www.idris-lang.org)


components.json
```diff
@@ -522,6 +522,11 @@
 			"title": "Icon",
 			"owner": "Golmote"
 		},
+                "idris": {
```
RunDevelopment, line 525: Please fix the indentation here. We use tabs.

RunDevelopment (changes requested):
Thank you for this PR @KeenS!

It seems like you based Idris on our Haskell language definition. Instead of copying the regexes, please the [`Prism.languages.extend`](https://prismjs.com/docs/Prism.languages.html#.extend) instead.

RunDevelopment:
Also, please do not modify `package-lock.json`. This the reason the CI fails.

Please revert your changes to `package-lock.json`, delete your local `node_modules` directory, and reinstall Prism's dependencies using `npm ci`. Rebuild Prism and commit all changes. The CI should then pass.

RunDevelopment:
@KeenS Please revert all changes to `package-lock.json` or else we can't merge this.

KeenS:
Sorry, I was careless. I removed the change to package-lock.json from git history and replaced spaces with tabs.

Though I don't think extending language syntax is a good way (changing Haskell  syntax shouldn't affect Idris'), I followed your advice

tests/languages/idris/keyword_feature.test
```diff
@@ -0,0 +1,17 @@
+case data do else if implementation
```
RunDevelopment, line 1: This doesn't look like it covers all keywords. Is there a reason for this?

components.json
```diff
@@ -522,6 +522,12 @@
 			"title": "Icon",
 			"owner": "Golmote"
 		},
+		"idris": {
+			"title": "Idris",
+			"alias": "idr",
+			"owner": "KeenS",
+			"require": ["haskell"]
```
RunDevelopment, line 529:
```suggestion
			"require": "haskell"
```

tests/languages/idris/keyword_feature.test
KeenS: Just because Idris has relatively larger size set of keywords. As lexing keywords is not so complex task, testing some of them is enough.

components.json
KeenS: Thank you :)

RunDevelopment (approved):

RunDevelopment:
Please rebuild to make the CI pass.

You also need to resolve the merge conflict. Since all conflicting files are generated ones, it should be enough to merge master and then rebuild to resolve the conflicts.

RunDevelopment:
Thank you for contributing @KeenS!

PR #2756, opened 2021-02-11
Add 'some' keyword for Swift

timroesner:
Swift 5.1 introduced opaque return types that make use of `some` as a keyword. 
This PR adds support to prism to highlight it as expected.


RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @timroesner!

PR #2770, opened 2021-02-18
feat: added support for PHP 8.1 enums

TomPavelec:


components/prism-php.js
```diff
@@ -13,6 +13,12 @@
 			alias: 'boolean'
 		},
 		/\b[A-Z_][A-Z0-9_]*\b(?!\s*\()/,
+		{
+			pattern: /(case\s+|::\s*)\b[a-z_]\w*\b(?!\s*\()/i,
```
RunDevelopment, line 17: I don't think this will work for `case NULL:`, does it? `NULL` is also constant but not an enum constant.

I also don't think that the alias `enum-constant` is a good fit. The `::` will also find regular class constants after all.

---

How about we split this into two patterns:

1. A pattern to find enum constant definitions. Something like `/(\bcase\s+)[a-z_]\w*(?=\s*[;=])/i`.
2. A pattern to find enum and class constants being used. Something like `/(::\s*)[a-z_]\w*(?!\s*\()/i`.

Since `::` constants are now supported, we should probably also highlight class constant definitions that don't follow the all-upper-case convention (e.g. `const foo = 0;`). Something like `/(\bconst\s+)[a-z_]\w*(?=\s*[;=])/i`.

TomPavelec: You are right. I've separated those patterns and changed aliases. (These two aliases are not necessary IMO)
And added support for lowercase class-constants.

RunDevelopment: > These two aliases are not necessary IMO

Then how about we remove them? I didn't say anything against the alias because I thought that you thought that they were beneficial.

With the aliases removed, we could combine the two patterns for better performance.

TomPavelec: Aliases removed, patterns combined.

RunDevelopment:

Issue #2035, opened 2019-09-01
Request: ColdFusion (CFML) syntax highlighting


MarieTay:
Would love the see the CFML (ColdFusion Markup Language) be supported, but tag syntax and scripting syntax.

Reference: https://cfdocs.org/

s-beji:
Since i saw the pull request from 2021 that mentions CFML was added and checked the download tool and CFScript was added like mentioned in the PR than we could close this issue and maybe update the docs to state more clearly cfml is supported since CFScript is not that obvious that it stands for CFML since in confusion there are more than just cfs files cfm if i am right.

I could also be wrong since i don't work directly with CFML but at my workplace we got our website build with CFML & Lucee but that's not my responsibility area i am responsible for In-house software that uses Blazor as Frontend that's why i am here because we got our own component lib and for demo examples we use prismJS

PR #2771, opened 2021-02-18
Added support for CFScript

mjclemente:
This partially deals with #2035, though does not resolve it entirely.

CFML has independent tag and script based syntaxes. This adds support for script syntax.

Here are some language references:

- https://cfdocs.org/script
- https://helpx.adobe.com/coldfusion/developing-applications/the-cfml-programming-language/extending-coldfusion-pages-with-cfml-scripting/the-cfscript-language.html
- https://www.petefreitag.com/cheatsheets/coldfusion/cfscript/


RunDevelopment (changes requested):
Good work @mjclemente!

A couple of minor things and then this is ready.

mjclemente:
Thanks for the fast, thorough, and really helpful review 😀 

I believe I've made all requested changes.

components.json
```diff
@@ -238,6 +238,12 @@
 			"require": "c",
 			"owner": "zeitgeist87"
 		},
+		"cfscript": {
+			"title": "CFML Script",
```
RunDevelopment, line 242: The docs consistently call this langauge "CFScript". I think we should too.
```suggestion
			"title": "CFScript",
```

RunDevelopment (changes requested):
Thank you for the quick response!

One more comment. Also, could you please add the link to https://cfdocs.org/script in a comment at the beginning of the file? This will be helpful to future contributors.

mjclemente:
Makes sense!

Changes made. Again, thanks for the prompt and thorough handling!

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @mjclemente!

mjclemente:
Thank you! 🎉

dawesi:
can we also create a branch of cfml for cfml/lucee (which boxlang also supports that syntax)

This is divergent from Adobe CFML/CFScript now also, as it containers many functions / helpers methods that are not in coldfusion and vice versa.

They publish the latest docs in json files zipped here.

https://docs.lucee.org/lucee-docs-json.zip

mjclemente:
Given that Lucee/Boxlang/Adobe ColdFusion are all CFML, my first thought would be that they could all be incorporated here. While there may be some different tags, I don't see any issues in including them all - they have more in common than different (at least Lucee CFML and Adobe). Not sure what functions are missing at this point.

PR #2773, opened 2021-02-20
Update elixir

gordalina:
Add support for @doc, @moduledoc, atom modules, function call tokens & raise keyword


RunDevelopment:
Could you please explain the motivation for this? The current highlighting seems to be quite good.

image

Code snippets for [here](https://hexdocs.pm/elixir/writing-documentation.html) and [here](https://elixir-lang.org/getting-started/module-attributes.html).

gordalina:
@RunDevelopment those two examples don't capture the extents of this change. I've put together a couple of examples to exemplify these changes.

Take for example [this codepen](https://codepen.io/gordalina/pen/eYBeJdM) and [this gist](https://gist.github.com/gordalina/4bef7ef64a2012bfeaa57b0a36c5f48b) - both have 3 examples, the codepen being the prismjs rendered one and the gist the one that github uses.

Prismjs does not parse @doc/@module doc as its own type, it currently returns string, which can't be syntax highlighted as a comment (see example of github).
This change also adds tokenization of modules and function calls, see in the gist example 3 line 25 and 26. Github tokenize `Application` and `put_all_env` separately, OTOH prismjs does not tokenize them, this PR tokenizes them.

This PR also adds the `raise` keyword.

RunDevelopment (changes requested):
Thanks for the clarification.

One thing aside from my comments:

The goal of the new `doc` rule is to enable GitHub-like highlighting, right? If that's the case, then you could also change the `attribute` rule to give `@doc` and `@moduledoc` a special alias. With that alias you could do GitHub-like highlighting in CSS like this:

```css
.token.attribute.special-alias,
.token.attribute.special-alias + .token.string {
  color: #888; /* or whatever color you want */
}
```

No need to add a new `doc` rule. 

The new `doc` rule also has a problem. Neither `doc` nor `doc-comment` are supported by Prism's standard themes. This means that doc attributes will remain unhighlighted despite being tokenized.

components/prism-elixir.js
```diff
@@ -36,6 +40,10 @@ Prism.languages.elixir = {
 		lookbehind: true,
 		alias: 'symbol'
 	},
+  'atom-module': {
+    pattern: /\b[A-Z][A-Z_a-z0-9]*\b/,
```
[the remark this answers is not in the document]
gordalina: This is intentional as it is meant to highlight modules, e.g.: `Elixir.Namespace` in `Elixir.Namespace.function()`
I tried calling it `module` which is what makes sense, but if I do, it does not tokenize properly, any suggestions here?

components/prism-elixir.js
```diff
@@ -1,4 +1,8 @@
 Prism.languages.elixir = {
+	'doc': {
+    pattern: /@(?:doc|moduledoc)\s+(?:("""|''')(?:\\[\s\S]|(?!\1)[^\\])+\1|.*)/,
```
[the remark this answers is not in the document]
gordalina: The `.*` matches all characters until end of line in this example: `@doc here goes some documentation`.
If there's a better way please let me know and I'll change it

gordalina:
> The goal of the new doc rule is to enable GitHub-like highlighting, right? If that's the case, then you could also change the attribute rule to give @doc and @moduledoc a special alias. With that alias you could do GitHub-like highlighting in CSS like this:

This makes a lot of sense, and it works with `@doc one single line` but when we have multiline docs `.attribute + .string` doesn't work as they're separated in different token lines. How do you propose to solve it?

```elixir
@moduledoc """
multiline
documentation
"""
```

I'll push the changes you requested

components/prism-elixir.js
```diff
@@ -1,4 +1,8 @@
...
+		alias: 'doc-comment'
```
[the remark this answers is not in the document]
gordalina: I realize it's not a comment - what alias (or not an alias at all) would you suggest?

RunDevelopment:
> doesn't work as they're separated in different token lines

I forgot that some libraries that use Prism under the hood do that. Prism doesn't split code into lines by itself, so I didn't account for this. Yes, my proposed alternative solution doesn't work.

components/prism-elixir.js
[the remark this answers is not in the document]
RunDevelopment: I apologize for my incredibly vague question.

I wanted to ask what the `.*` is supposed to tokenize. I'm not very familiar with Elixir, so I don't know what can and can't be done with attributes. I know that `.*` matches everything until the end of the line but is that correct?

It will correctly match all examples [here](https://hexdocs.pm/elixir/writing-documentation.html) and [here](https://elixir-lang.org/getting-started/module-attributes.html) but I wonder if that's enough.

For `@doc here goes some documentation`: How is `here goes some documentation` parsed? From what the documentation says, it sounds like it has to be a valid value, so is something like this possible?

```elixir
@doc deprecated: """Use Foo.bar/2 instead.
This function was removed for one reason or another.
"""
@doc since: "1.0.0" # some comment
```

components/prism-elixir.js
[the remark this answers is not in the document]
RunDevelopment: Let's remove the alias then. It isn't supported by our standard themes anyway.

components/prism-elixir.js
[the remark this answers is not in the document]
RunDevelopment: We do have the `namespace` class of modules and package names.

> I tried calling it `module` which is what makes sense, but if I do, it does not tokenize properly

That's because [another `module` rule](https://github.com/PrismJS/prism/blob/4d7d9b3879e2106e8c9facf04e3ebd73a66f0a1d/components/prism-elixir.js#L89) is inserted overwriting yours. Since your new pattern is essentially a superset of the old one, you could probably remove the old one.

On that note: Why does the old `module` pattern use `[A-Z][\w.\\]+`? I understand the `.` but I don't get the `\\`. Is that even correct?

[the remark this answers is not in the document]
gordalina: The only thing that comes to mind is elixir's default parameters, but it tokenizes correctly with or without `\\`.

```elixir
def get_time(timestamp \\ 0) do
  # ...
end
```

gordalina:
@RunDevelopment I've updated it with your requests and with a suggested direction on how to deal with @doc/moduledoc.

Here's what I think is the right approach with the right compromise
- Everything that's after `@attr <...>` is an elixir statement, so it should be tokenized as such. Case in point: `@doc since: "1.3.0"`
- The elixir guide on writing documentation points out that it should be `@(doc|moduledoc) "string"` so I suggest we only tokenize those as docs and the remainder as normal attributes. (see included doc_feature.test)

Let me know what you think.

RunDevelopment:
Sounds good! Let's go with that.

components/prism-elixir.js
```diff
@@ -55,7 +62,10 @@ Prism.languages.elixir = {
 		alias: 'variable'
 	},
 	'number': /\b(?:0[box][a-f\d_]+|\d[\d_]*)(?:\.[\d_]+)?(?:e[+-]?[\d_]+)?\b/i,
-	'keyword': /\b(?:after|alias|and|case|catch|cond|def(?:callback|exception|impl|module|p|protocol|struct|delegate)?|do|else|end|fn|for|if|import|not|or|require|rescue|try|unless|use|when)\b/,
+	'keyword': /\b(?:after|alias|and|case|catch|cond|def(?:callback|exception|impl|module|p|protocol|struct|delegate)?|do|else|end|fn|for|if|import|not|or|raise|require|rescue|try|unless|use|when)\b/,
+	'function': {
+		pattern: /\b[_a-zA-Z]\w*[?!]?(?:(?=\s*(?:\.\s*)?\()|\/\d+)/
+	},
```
RunDevelopment, line 68:
```suggestion
	'function': /\b[_a-zA-Z]\w*[?!]?(?:(?=\s*(?:\.\s*)?\()|\/\d+)/,
```

components/prism-elixir.js
```diff
@@ -1,4 +1,7 @@
 Prism.languages.elixir = {
+	'doc': {
+		pattern: /@(?:doc|moduledoc)\s+("{1,3}|'{1,3})(?:\\[\s\S]|(?!\1)[^\\])+\1/
+	},
 	'comment': /#.*/m,
```
RunDevelopment, line 5: Comment has to be greedy or else `# @doc "abc"` will not be matched correctly.

```suggestion
	'comment': {
		pattern: /#.*/m,
		greedy: true
	},
```

components/prism-elixir.js
```diff
@@ -1,4 +1,7 @@
 Prism.languages.elixir = {
+	'doc': {
+		pattern: /@(?:doc|moduledoc)\s+("{1,3}|'{1,3})(?:\\[\s\S]|(?!\1)[^\\])+\1/
```
RunDevelopment, line 3: `"{1,3}|'{1,3}` is not a good idea.

```suggestion
		pattern: /@(?:doc|moduledoc)\s+(?:("|')(?:\\(?:\r\n|[\s\S])|(?!\1)[^\\\r\n])*\1|("""|''')[\s\S]*?\2)/
```

components/prism-elixir.js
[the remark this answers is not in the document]
RunDevelopment: In that case, `\` was probably incorrect.

tests/languages/elixir/capture_feature.test
```diff
@@ -7,7 +7,12 @@ fun = &(&1 + 1)
 
 [
 	"fun ", ["operator", "="],
-	["capture", "&Math.zero?/1"],
```
RunDevelopment, line 10: What do we do about captures? Having `&Math.zero?/1` and `&is_function/1` be highlighted differently is quite inconsistent.

tests/languages/elixir/capture_feature.test
```diff
@@ -7,7 +7,12 @@ fun = &(&1 + 1)
 
 [
 	"fun ", ["operator", "="],
-	["capture", "&Math.zero?/1"],
+	["operator", "&"],
+	["module", "Math"],
+	["punctuation", "."],
+	"zero?",
+	["operator", "/"],
+	["number", "1"],
```
RunDevelopment, line 15: Shouldn't this be highlighted as a function?

On that note: Wouldn't it be better to just highlight `zero?` as a function instead of `zero?/1`?

components/prism-elixir.js
```diff
@@ -1,4 +1,7 @@
 Prism.languages.elixir = {
+	'doc': {
```
RunDevelopment, line 2: I agree with [the general direction you want to take](https://github.com/PrismJS/prism/pull/2773#issuecomment-784707821). However, I think that we still need to highlight `@doc|@moduledoc` and the string value. The easiest way to do that is an `inside` grammar like this:

```js
inside: {
  'attribute': /^@\w+/,
  'string': /['"][\s\S]+/
}
``` 

Thoughts?

components/prism-elixir.js
gordalina: This pattern breaks multiline strings


I've pushed one that works - by my regex foo isn't great - so feel free to suggest a better one.

tests/languages/elixir/capture_feature.test
gordalina: I think we should highlight `zero?` and `is_function` as functions and `/1` as operator + number.
I'll push a fix for that

tests/languages/elixir/capture_feature.test
gordalina: I've updated the `function` tokenizer and removed capture as it was a subset of function

gordalina:
@RunDevelopment I've addressed all your comments and it's ready for review.

components/prism-elixir.js
RunDevelopment: > This pattern breaks multiline strings

My bad. I messed up the order. I'll suggest a pattern that works...

components/prism-elixir.js
```diff
@@ -1,5 +1,15 @@
 Prism.languages.elixir = {
-	'comment': /#.*/m,
+	'doc': {
+		pattern: /@(?:doc|moduledoc)\s+("|'|"""|''')(?:\\[\s\S]|(?!\1)[^\\])+\1/,
```
RunDevelopment, line 3: Alright, that one should work!

```suggestion
		pattern: /@(?:doc|moduledoc)\s+(?:("""|''')[\s\S]*?\1|("|')(?:\\(?:\r\n|[\s\S])|(?!\2)[^\\\r\n])*\2)/,
```

gordalina:
@RunDevelopment I've updated it, should be ready to go 🤞

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @gordalina!

PR #2779, opened 2021-03-01
Fix INI

### LINKED ISSUE #2775: Several highlighting bugs in INI
**Information**

- Language: INI
- Plugins: none

**Description**

Does someone still need INI? :)

Anyway, I’ve made several tests with C (using [GetPrivateProfileSectionNames](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getprivateprofilesectionnames) and [GetPrivateProfileString](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getprivateprofilestring) from Win32 API), [PHP](https://www.php.net/manual/en/function.parse-ini-file.php), and [Python](https://docs.python.org/3.8/library/configparser.html).

1. Comments. The dollar sign is useless in `/^[ \t]*[;#].*$/m`. It is also possible that the dot is harmful, because it matches any character except `\n`, `\r`, `\u2028`, or `\u2029` (the `m` flag doesn’t change the behavior), whereas `\u2028` and `\u2029` can occur inside comments.

    I confirm that `#` and `;` comment operators are allowed only at the beginning of a line (besides PHP).

    **Example 1.1**

    ```ini
    [My#Section;Name]
    my#;key=my#;value
    ```

    - the section name is `My#Section;Name`;
    - the key is `my#;key`;
    - the value is `my#;value`.

2. `selector` should be renamed to `namespace`, because sections act like namespaces.

    **Example 2.1**

    ```ini
    [MySection1]
    mykey=myvalue1

    [MySection2]
    mykey=myvalue2
    ```

    Parsers allows to get `mykey` values from both `MySection1` and `MySection2`. The second key doesn’t override the first one, so these values are `myvalue1` and `myvalue2`, respectively.

3. Outer brackets should not be included in section names. If multiple square brackets are given, only the first one (from the left) should be considered a paired punctuation mark.

    **Example 3.1**

    ```ini
    [foo[bar]
    ```

    The section name is `foo[bar`.

    **Example 3.2**

    ```ini
    [foo]bar]
    ```

    The section name is `foo`.

4. Section names, keys, values can contain **any** whitespaces (`\s`). Beginning and trailing whitespaces (`[ \f\t\v]`, not `[ \t]`) on each line should be considered irrelevant.

    **Example 4.1**

    ```ini
     [   My  Section ]  
       my  key  = my  value   
    ```

    - the section name is `My␣␣Section`, not `␣[␣␣␣My␣␣Section␣]`;
    - the key is `my␣␣key`;
    - the value is `my␣␣value`, not `=␣my␣␣value␣␣␣`.

5. Quotes preserve inner spaces. `"` and `'` are allowed, but they are not interchangeable. Parses remove quotes around values.

    **Example 5.1**

    ```ini
     [  " My  Section   " ]  
      " my  key   " =  "   my  value " 
    ```

    - the section name is `"␣My␣␣Section␣␣␣"`;
    - the key is `"␣my␣␣key␣␣␣"`;
    - **the value is `␣␣␣my␣␣value␣`** (without quotes).

6. If multiple equal signs are given, only the first one (from the left) should be considered a key-value separator. All other equal signs should be included in the value.

    **Example 6.1**

    ```ini
    mykey==myvalue
    ```

    The key is `mykey`. The value is `=myvalue`.

    **Example 6.2**

    ```ini
    foo===bar==baz=qux
    ```

    The key is `foo`. The value is `==bar==baz=qux`.

7. AFAIK, [escaping](https://en.wikipedia.org/wiki/INI_file#Escape_characters) is not implemented in the Win32 API. It is not implemented in Prism too. Should it be implemented?

**Notes**

1. PHP caveats:

    - doesn’t allow `#` as a comment operator;
    - allows inline comments after section names and values;
    - doesn’t allow spaces _before_ `[`;
    - doesn’t trim whitespaces around section names (e. g. `[␣My␣␣Section␣␣␣]` is interpreted as `␣My␣␣Section␣␣␣`; same for quoted section names);
    - doesn’t support quoted keys.

2. Python caveats:

    - doesn’t trim whitespaces around section names;
    - trims the `\s` (not only `[ \f\t\v]`) whitespaces around keys.

**Code snippet**

[Test page](https://prismjs.com/test.html#language=ini&text=%23%20Foo.%20Here%20goes%20U%2B2028%3A%20%E2%80%A8.%20Bar.%20Baz.%0A%3B%20Foo.%20Here%20goes%20U%2B2029%3A%20%E2%80%A9.%20Bar.%20Baz.%0A%0A%23%20The%20name%20of%20the%20section%20is%20My%E2%90%A3%E2%90%A3Section.%0A%23%20Moreover%2C%20it%20is%20not%20a%20selector%2C%20but%20a%20namespace.%0A%20%5B%20%20%20My%20%20Section%20%5D%20%20%0A%0A%23%20The%20name%20of%20the%20section%20is%20%22%E2%90%A3My%E2%90%A3%E2%90%A3Section%E2%90%A3%E2%90%A3%E2%90%A3%22.%0A%23%20Moreover%2C%20it%20is%20not%20a%20selector%2C%20but%20a%20namespace.%0A%20%5B%20%20%22%20My%20%20Section%20%20%20%22%20%5D%20%20%0A%0A%23%20Names%20are%20My%22Section%2C%20My%22Section%22Name%2C%20%22My%22Section%22%2C%20%22My%22Section%22Name%22.%0A%5BMy%22Section%5D%0A%5BMy%22Section%22Name%5D%0A%5B%22My%22Section%22%5D%0A%5B%22My%22Section%22Name%22%5D%0A%0A%23%20Only%20the%20leftmost%20equal%20sign%20should%20be%20considered%20a%20key-value%20separator.%0Amykey%3D%3Dmyvalue%0Afoo%3D%3D%3Dbar%3D%3Dbaz%3Dqux%0A%0A%23%20Keys%20can%20contain%20any%20whitespaces.%0A%20%20%20my%20%20key%20%20%3D%20my%20%20value%20%20%20%0A%20%20%22%20my%20%20key%20%20%20%22%20%3D%20%20%22%20%20%20my%20%20value%20%22%20)

<details>

```ini
# Foo. Here goes U+2028:  . Bar. Baz.
; Foo. Here goes U+2029:  . Bar. Baz.

# The name of the section is My␣␣Section.
# Moreover, it is not a selector, but a namespace.
 [   My  Section ]  

# The name of the section is "␣My␣␣Section␣␣␣".
# Moreover, it is not a selector, but a namespace.
 [  " My  Section   " ]  

# Names are My"Section, My"Section"Name, "My"Section", "My"Section"Name".
[My"Section]
[My"Section"Name]
["My"Section"]
["My"Section"Name"]

# Only the leftmost equal sign should be considered a key-value separator.
mykey==myvalue
foo===bar==baz=qux

# Keys can contain any whitespaces.
   my  key  = my  value   
  " my  key   " =  "   my  value " 
```

</details>

edukisto:
Sorry for being late! This PR resolves #2775. I’ve made the module mimic the [behavior](https://github.com/PrismJS/prism/issues/2775#issuecomment-787477723) of Win32 API.


edukisto:
I’ve also replaced `constant` with `attr-name`.

components/prism-ini.js
```diff
@@ -1,11 +1,37 @@
...
-	'selector': /^[ \t]*\[.*?\]/m,
-	'constant': /^[ \t]*[^\s=]+?(?=[ \t]*=)/m,
-	'attr-value': {
-		pattern: /=.*/,
+	'comment': {
+		pattern: /(^[ \f\t\v]*)[#;][^\n\r]*/m,
+		lookbehind: true
+	},
+	'header': {
+		pattern: /(^[ \f\t\v]*)\[[^\n\r\]]*\]?/m,
```
RunDevelopment, line 7: The last `\]` is optional because of the [`[foo` case](https://github.com/PrismJS/prism/issues/2775#issuecomment-787477723), right?

components/prism-ini.js
```diff
@@ -1,11 +1,37 @@
...
 		inside: {
-			'punctuation': /^[=]/
+			'section-name': {
+				pattern: /(\[[ \f\t\v]*)[^ \f\n\r\t\v\]]+(?:[ \f\t\v]+[^ \f\n\r\t\v\]]+)*/,
+				lookbehind: true,
+				alias: 'selector'
+			},
+			rest: {
+				'punctuation': /\[|\]/
+			}
```
RunDevelopment, line 17: `rest` will (in a sense) inline its grammar. You don't need `rest` in this case.

```suggestion
			'punctuation': /\[|\]/
```

components/prism-ini.js
```diff
@@ -1,11 +1,37 @@
...
+		pattern: /(^[ \f\t\v]*)[#;][^\n\r]*/m,
+		lookbehind: true
+	},
+	'header': {
+		pattern: /(^[ \f\t\v]*)\[[^\n\r\]]*\]?/m,
+		lookbehind: true,
 		inside: {
-			'punctuation': /^[=]/
+			'section-name': {
+				pattern: /(\[[ \f\t\v]*)[^ \f\n\r\t\v\]]+(?:[ \f\t\v]+[^ \f\n\r\t\v\]]+)*/,
```
RunDevelopment, line 11: 1. The `header` pattern already assures us `[^\n\r\]]`, so we don't have exclude `\r\n` here.
2. A `^` at the start of the pattern isn't necessary here for correctness but it allows the regex engine to better optimize the pattern.

```suggestion
				pattern: /(^\[[ \f\t\v]*)[^ \f\t\v\]]+(?:[ \f\t\v]+[^ \f\t\v\]]+)*/,
```

components/prism-ini.js
```diff
@@ -1,11 +1,37 @@
...
+		lookbehind: true,
+		alias: 'attr-name'
+	},
+	'value': {
+		pattern: /(=[ \f\t\v]*)[^ \f\n\r\t\v]+(?:[ \f\t\v]+[^ \f\n\r\t\v]+)*/,
+		lookbehind: true,
+		alias: 'attr-value',
+		inside: {
+			'inner-value': {
+				pattern: /("|').+(?=\1)/,
```
RunDevelopment, line 31: I thought that it was only recognized as a "string" if it started and ended with quotes (of the same kind). Is that also the case if it's in the middle? E.g. `foo=b"a"r`

RunDevelopment (changes requested):
Great work @edukisto!

I have a few questions and suggestions. 

Also, please add a comment with a link to your detailed explanation of the differences in INI parsers. And please also mention that Prism's INI implements the Win32 behaviour. This will make it easier for others to understand the code.

components/prism-ini.js
edukisto: Yes. This is how [`GetPrivateProfileSectionNames`](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getprivateprofilesectionnames) works, although this behavior is undocumented.

components/prism-ini.js
edukisto: Right! Thanks!

components/prism-ini.js
edukisto: The parsers I’ve checked treat quotation marks in `"b"a"r"`, `"bar`, `b"a"r`, `ba"r`, `bar"` in almost the same way. The only difference is that Python doesn’t truncate outer _paired_ quotation marks.

Quotes are used to create values with leading and/or trailing spaces.

Parsers consistently strip off leading and trailing spaces. A quotation mark (or any other printable character) before leading spaces or after trailing spaces turns those spaces into internal ones. **These quotes do not have to be paired**: `"␣bar`, `bar␣"`. Of course, the quotation marks affect only adjacent spaces. There can be as many quotes as you like.

The only difference between parsers is that Win32 API and PHP truncate _outer paired_ quotation marks.

components/prism-ini.js
edukisto: 1. This is a bright idea.
2. This is another bright idea.

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @edukisto!

edukisto:
Thank you for your help and guidance, @RunDevelopment!

PR #2784, opened 2021-03-05
Copy to clipboard: Removed ClipboardJS dependency

### LINKED ISSUE #2286: Local script for copy clipboard plugin
**Motivation**
Use the plugin in offline mode.

**Description**
An option to append the js to the download package instead of inserting script element into the header.

**Alternatives**
Download the script manually and remove the header injection.

### LINKED ISSUE #2565: WebPack and /prism/plugins/copy-to-clipboard/prism-copy-to-clipboard.js
Webpack is not able to compile\merge **prism-copy-to-clipboard.js** into a single file.
- Prism version: [e.g. 1.14, latest from the download page, etc.]
- Plugins: [prism-copy-to-clipboard.js]
- Environment: [Webpack]

Does the latest version of Prism from the [download page](https://prismjs.com/download.html) also have this issue? **Yes**
If you're using Prism with Node or Webpack, please also see the [usage section](https://prismjs.com/index.html#basic-usage). **Looked but did not solve**

**Description**
Running Webpack produces and error 
```
ERROR in ./wwwroot/lib/prism/plugins/copy-to-clipboard/prism-copy-to-clipboard.js
Module not found: Error: Can't resolve 'clipboard' in 'C:\ws\DocHost\wwwroot\lib\prism\plugins\copy-to-clipboard'
 @ ./wwwroot/lib/prism/plugins/copy-to-clipboard/prism-copy-to-clipboard.js 14:18-38
 @ ./wwwroot/scripts/vendor-index.js
```
**Example**
I used Visual Studio 2019 but I do not think it is a factor.
Attached [vendor-index .txt](https://github.com/PrismJS/prism/files/5269562/vendor-index.txt) is my vendor index file
Command line `webpack --watch --mode=development`

When I comment out line 15
`ClipboardJS = require('clipboard');`
Webpack works

### LINKED ISSUE #2783: [BUG] clipboard js bundling - still uses cdn (privacy violation + prevents offline)
**Information:**
- Prism version: download page
- Plugins: clipboard
- Environment: browser

Does the latest version of Prism from the [download page](https://prismjs.com/download.html) also have this issue? Yes

**Description**

When checking the clipboardjs plugin on the downloads page, the resulting bundle still attempts to contact a thirdparty website. This is unexpected, a problem for offline use, a privacy attack, and a security risk.

I recommend removing the plugin from the list to avoid misleading security-relevant behavior, or removing the surprise remote code execution.

Edit: See also https://github.com/PrismJS/prism/issues/2286 . I view this as a security bug , which was what caught our eye.

RunDevelopment:
This removes Copy to clipboard's dependency on [ClipboardJS](https://clipboardjs.com/). 

ClipboardJS has been replaced with a few functions that behave similarly. The actual copying is implemented by [the Async + fallback combo here](https://stackoverflow.com/a/30810322/7595472). This alone should give us very good browser compatibility (works in IE11). 

Older FF versions do not support copying text to the clipboard at all. For these browsers, we use the same trick as ClipboardJS: we just select the text to copy and ask the user to press Ctrl+C. It's implemented slightly differently, however. ClipboardJS creates a temporary invisible element and selects the text of that but we can just select the text of the `<code>` element we want to copy (makes everything a lot easier). 
This may scroll the code block but I didn't give this much consideration since it will only affect a very small portion of our users.

Prism now has 0 required (runtime) dependencies on other libraries.

---

This resolves #2783
This resolves #2565
This resolves #2286


RunDevelopment:
The +38B in bundle size should be viewed in the context that we don't have to download the 3kB of ClipboardJS anymore.

PR #2796, opened 2021-03-08
Added support for Generic log files

### LINKED ISSUE #2054: Add support for log files
**Language**
Generic log files (such as *nix log files, logs generated by Makefile, etc.)

**Additional resources**
An example implementation for vim-polyglot: https://github.com/sheerun/vim-polyglot/blob/master/syntax/log.vim

RunDevelopment:
This resolves #2054.

---

Since "generic log file" is very imprecise, I opted to defeat the issue with test data. I added a few test files that start with an `_`. These are log files from my machine or other places. The idea is that if all of these logs look good, then the language definition is usable.


joshgoebel:
This is a very interesting idea. :)

PR #2798, opened 2021-03-10
Language Nevod

dmitry-surkov:
Add support for [Nevod](https://nevod.io) language used for pattern-based text search.


components/prism-nevod.js
```diff
@@ -0,0 +1,79 @@
...
+		pattern: /(?:@search\s+|#)[a-zA-Z0-9\-.]+(?:\.\*)?\s*(?=;)/,
+		inside: {
+			'keyword': /@search\b/,
+			'name': /#?.*(?=;)/
+		}
+	},
+	'basic-reference': {
+		pattern: /\b(?:Word|Punct|Symbol|Space|LineBreak|Start|End|Alpha|AlphaNum|Num|NumAlpha|Blank|WordBreak|Any)(?:\(.*\)|\b)/,
+		inside: {
+			'name': /.+(?=[(\b])/,
```
RunDevelopment, line 58: `\b` inside character classes means backspace. Are you sure this is correct?

components/prism-nevod.js
```diff
@@ -0,0 +1,79 @@
...
+		},
+		{
+			pattern: /'(?:''|[^'])*'(?!')!?\*?/,
+			greedy: true,
+			inside: {
+				'modifiers': /!$|!\*$|\*$/,
+			},
+		}
+	],
+	'keyword': /@(?:inside|outside|having|search|where)\b/,
```
RunDevelopment, line 19: I would like to suggest the following to make the whole grammar a lot simpler: Match keywords after patterns that include them.

I other languages, we use a design pattern that roughly works like this:

```js
{
	// string, comments, whatever
	
	'function': {
		pattern: /(\bfunction\s+)\w+/,
		lookbehind: true
	},
	'class-name': {
		pattern: /(\bclass\s+)\w+/,
		lookbehind: true
	},
	'keyword': /\b(?:function|class|some|other|keywords)\b/

	// other tokens that don't need/include keywords
}
``` 

This has the advantage that tokens that are _defined by_ keywords don't have to tokenize these keywords, so there is only one `keyword` pattern. (You see this design pattern being used in [Ruby](https://github.com/PrismJS/prism/blob/a68f1fb6e23143ada9676fa6f66af72a3b1eebe8/components/prism-ruby.js#L17).)

components/prism-nevod.js
```diff
@@ -0,0 +1,79 @@
...
+					'attr-name': /[a-zA-Z0-9\-.]+/,
+					'attr-hidden-mark': /~/,
+					'attr-punctuation': /[,]/,
+					'attr-bracket': /[()]/,
+				},
+			},
+		},
+	},
+	'search': {
+		pattern: /(?:@search\s+|#)[a-zA-Z0-9\-.]+(?:\.\*)?\s*(?=;)/,
```
RunDevelopment, line 49: The last `\s*` should be included in the lookahead. Tokens should generally not include leading or trailing spaces.

components/prism-nevod.js
```diff
@@ -0,0 +1,79 @@
...
+	'semi-colon': /;/,
+	'attr-capture': {
+		pattern: /[a-zA-Z0-9\-.]+\s*:/,
+		inside: {
+			'attr-name': /[a-zA-Z0-9\-.]+/,
+			'colon': /:/,
+		},
+	},
+	'punctuation': /[,+_]|\.{2,3}/,
+	'name': /\S+/
```
RunDevelopment, line 78: Is this supposed to capture a general identifier?

components/prism-nevod.js
```diff
@@ -0,0 +1,79 @@
...
+	},
+	'search': {
+		pattern: /(?:@search\s+|#)[a-zA-Z0-9\-.]+(?:\.\*)?\s*(?=;)/,
+		inside: {
+			'keyword': /@search\b/,
+			'name': /#?.*(?=;)/
+		}
+	},
+	'basic-reference': {
+		pattern: /\b(?:Word|Punct|Symbol|Space|LineBreak|Start|End|Alpha|AlphaNum|Num|NumAlpha|Blank|WordBreak|Any)(?:\(.*\)|\b)/,
```
RunDevelopment, line 56: I'm not quite sure about the `\(.*\)`. It seems to be a little too simple to work. Those are token parameters, right? Should the parameters not be tokenized (e.g. `Alpha(2-10, TitleCase)`)? Wouldn't parameters automatically get sensible highlighting if we didn't include them here?

components/prism-nevod.js
```diff
@@ -0,0 +1,79 @@
...
+		inside: {
+			'name': /.+(?=[(\b])/,
+			'params': /\(.*\)/,
+		},
+	},
+	'quantifier': /\b\d+(?:\+|-\d+)?\b/,
```
RunDevelopment, line 62: The `\+` is a problem because of the ending `\b`. Since `+` is a non-word character, the `\b` requires that it is followed by a word character (e.g. `+r` or `+4`). Is this correct?

components/prism-nevod.js
dmitry-surkov: Fixed as suggested.

components/prism-nevod.js
dmitry-surkov: Fixed and improved.

components/prism-nevod.js
dmitry-surkov: Fixed to capture general identifier.

components/prism-nevod.js
dmitry-surkov: I have refined the pattern to detect the built-in pattern with the parameters.

components/prism-nevod.js
dmitry-surkov: This pattern is rewritten.

components/prism-nevod.js
```diff
@@ -0,0 +1,112 @@
...
+		inside: {
+			'string-attrs': /!$|!\*$|\*$/,
+		},
+	},
+	'namespace': {
+		pattern: /(@namespace\s+)[a-zA-Z0-9\-.]+(?=\s*{)/,
+		lookbehind: true,
+	},
+	'pattern': {
+		pattern: /(@pattern\s+)?#?[a-zA-Z0-9\-.]+(?:\s*(?:[(]\s*(?:(?:~\s*)?[a-zA-Z0-9\-.]+\s*(?:,\s*(?:(?:~\s*)?[a-zA-Z0-9\-.]*))*)[)]))?(?=\s*=)/,
```
RunDevelopment, line 15: 1. This pattern contains a few unnecessary non-capturing groups.
2. You were very careful to allowing `\s*` between token but it seems like you forgot a `\s*` after the last `[a-zA-Z0-9\-.]*`. Am I correct to assume that you initially had a `\s*` there but removed it because of the exponential/polynomial backtracking test?

```suggestion
		pattern: /(@pattern\s+)?#?[a-zA-Z0-9\-.]+(?:\s*[(]\s*(?:~\s*)?[a-zA-Z0-9\-.]+\s*(?:,\s*(?:~\s*)?[a-zA-Z0-9\-.]*)*[)])?(?=\s*=)/,
```

components/prism-nevod.js
```diff
@@ -0,0 +1,112 @@
...
+			},
+			'standard-pattern-attr': {
+				pattern: /[a-zA-Z0-9\-.]+/,
+				alias: 'builtin',
+			},
+			'punctuation': /[,()]/,
+		},
+	},
+	'quantifier': {
+		pattern: /\b\d+\s*(?:\+|-\s*\d+)?(?=\W)/,
```
RunDevelopment, line 63: 1. The first `\s*` is a bit of a problem because the string ``"1\ \ \ "`` (`1` followed by 3 spaces) will be matched as `1\ \ ` (`1` followed by 2 spaces).
2. The `(?=\W)` means that there has to be a character after the quantifier. Prism should also be able to highlight small code snippets, so the end of the string should also be valid here.

(Same for the other `quantifier` pattern)

```suggestion
		pattern: /\b\d+(?:\s*\+|\s*-\s*\d+)?(?!\w)/,
```

components/prism-nevod.js
```diff
@@ -0,0 +1,112 @@
...
+					'operator': {
+						pattern: /~/,
+						alias: 'attr-hidden-mark',
+					},
+				},
+			},
+		},
+	},
+	'search': {
+		pattern: /(@search\s+|#)[a-zA-Z0-9\-.]+(?:\.\*)?(?=\s*;)/,
```
RunDevelopment, line 39: Isn't the `#` part of the name? If yes, then this can be done like this:

```suggestion
		pattern: /(?:(@search\s+)|#)[a-zA-Z0-9\-.]+(?:\.\*)?(?=\s*;)/,
```

tests/languages/nevod/keyword_feature.test
```diff
@@ -0,0 +1,13 @@
+@inside @outside @having
+@search @where
```
RunDevelopment, line 2: Please add `@require`, `@namespace`, and `@pattern`.

components/prism-nevod.js
dmitry-surkov: 1. Thank you, will fix.
2. Yes, this was the reason.

components/prism-nevod.js
dmitry-surkov: `@pattern A = Word;` and `#A = Word;` - are pattern definitions, but
`@search A;` and `#A;` - are not pattern definitions, but are search targets.

components/prism-nevod.js
dmitry-surkov: Fixed.

components/prism-nevod.js
dmitry-surkov: Fixed, thank you.

tests/languages/nevod/keyword_feature.test
dmitry-surkov: Test improved to check also for missed keywords.

components/prism-nevod.js
dmitry-surkov: In the second case (`#A;`) the `#` is not part of the name. The idea behind `#` is to simplify one-string examples where the pattern is defined and used for search.

dmitry-surkov:
@RunDevelopment, seems I have resolved all the issues. Do you see any other issues that prevent this PR from merge? Btw, I have also merged all changes from the `PrismJS:master` to the `nezaboodka:language-nevod` to keep PR up to date.

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @dmitry-surkov!

PR #2804, opened 2021-03-15
Αdding support for Q# language

fedonman:
This is more like a stripped version of C# with some minor modifications to match Q# style. Code highlighting is pretty good so far, although some improvements are already identified (like special functions, and integration with C# / Python host programs)


RunDevelopment:
Thank you for the PR @fedonman!

Right now, your Q# is still too C#-like (e.g. Q# doesn't have verbatim strings or multiline comments). There are also a few missing keywords and operators.

Fortunately, Microsoft provides a [full ANTRL4 grammar for Q#](https://github.com/microsoft/qsharp-language/tree/main/Specifications/Language/5_Grammar), so missing keywords and operators shouldn't be a problem. This grammar is extremely useful, so I recommend using it (it has the answer to literally all questions regarding Q# syntax).

And please add more tests. Q# is a simple language compared to C# but it is not trivial, so please test keywords, operators, and the `class-name` patterns you added.

fedonman:
Thank you for the grammar link! It really helps.

I have:

- modified keywords list
- modified regex expressions for numbers and punctuation
- removed multiline comments / strings
- removed verbatim strings
- added 3 more test cases

components/prism-qsharp.js
```diff
@@ -0,0 +1,168 @@
...
+		// keywords which are used to declare a type
+		typeDeclaration: 'newtype',
+		// all other keywords
+        other: 'Adjoint adjoint and apply as auto body borrow borrowing Controlled controlled distribute elif else fail fixup for function if in internal intrinsic invert is let mutable namespace new not open operation or repeat return self set until use using while within'
+    } 
+	// keywords
+	function keywordsToPattern(words) {
+		return '\\b(?:' + words.trim().replace(/ /g, '|') + ')\\b';
+	}
+	var typeDeclarationKeywords = keywordsToPattern(keywordKinds.typeDeclaration);
```
RunDevelopment, line 56: There are quite a few unused variables here. Please remove all of them. (About half of all variables here are unused.)

After doing that, you'll see that you can merge `keywordKinds.typeDeclaration` and `keywordKinds.other`. Please do that.

components/prism-qsharp.js
```diff
@@ -0,0 +1,168 @@
...
+						lookbehind: true,
+						inside: {
+							'punctuation': /^:/
+						}
+					},
+					'punctuation': /^\{|\}$/,
+					'expression': {
+						pattern: /[\s\S]+/,
+						alias: 'language-csharp',
+						inside: Prism.languages.csharp
```
RunDevelopment, line 149: You should probably change that :)

components/prism-qsharp.js
```diff
@@ -0,0 +1,168 @@
...
+	var tuple = replace(/\(<<0>>+(?:,<<0>>+)+\)/.source, [tupleElement]);
+	var typeExpression = replace(/(?:<<0>>|<<1>>)(?:\s*(?:\?\s*)?<<2>>)*(?:\s*\?)?/.source, [tuple, identifier, array]);
+
+	var typeInside = {
+		'keyword': keywords,
+		'punctuation': /[<>()?,.:[\]]/
+	};
+
+	// strings & characters
+	var character = /'(?:[^\r\n'\\]|\\.|\\[Uux][\da-fA-F]{1,8})'/.source; // simplified pattern
```
RunDevelopment, line 78: Q# doesn't have characters as C# does, so this and everything that has to do with characters can be removed.

components/prism-qsharp.js
```diff
@@ -0,0 +1,168 @@
...
+			pattern = pattern.replace(/<<self>>/g, function () { return '(?:' + pattern + ')'; });
+		}
+		return pattern.replace(/<<self>>/g, '[^\\s\\S]');
+	}
+
+	// https://docs.microsoft.com/en-us/azure/quantum/user-guide/language/typesystem/
+	// https://github.com/microsoft/qsharp-language/tree/main/Specifications/Language/5_Grammar
+	var keywordKinds = {
+		// keywords which represent a return or variable type
+		type: 'Unit Int BigInt Double Bool String Qubit Result Range true false Zero One Pauli PauliI PauliX PauliY PauliZ Adj Ctl',
```
RunDevelopment, line 46: Could you please sort these alphabetically as well?

components/prism-qsharp.js
```diff
@@ -0,0 +1,168 @@
...
+			{
+				pattern: RegExp(character),
+				greedy: true,
+				alias: 'character'
+			}
+		],
+		'class-name': [
+			{
+				// open Microsoft.Quantum.Canon;
+				pattern: re(/(\bopen\s+)<<0>>(?=\s*;)/.source, [identifier]),
```
RunDevelopment, line 101: You can also [alias imports](https://github.com/microsoft/qsharp-language/tree/main/Specifications/Language/1_ProgramStructure).

Also, do we need the lookahead? Can you have `open` or `as` as variable names?

```suggestion
				pattern: re(/(\b(?:as|open)\s+)<<0>>(?=\s*(?:;|as\b))/.source, [identifier]),
```

RunDevelopment (changes requested):

components/prism-qsharp.js
fedonman: accepted this change

components/prism-qsharp.js
fedonman: Done

components/prism-qsharp.js
fedonman: Correct. I removed this.

components/prism-qsharp.js
fedonman: You are right, I had forgotten about these. They are removed now.

components/prism-qsharp.js
```diff
@@ -0,0 +1,168 @@
...
+            {
+				// namespace Quantum.App1;
+				pattern: re(/(\bnamespace\s+)<<0>>(?=\s*{)/.source, [identifier]),
+				lookbehind: true,
+				inside: typeInside
+			},
+		],
+		'keyword': keywords,
+		'number': /(?:0(?:x[\da-f_]*[\da-f]|b[01_]*[01]|o[0-7]*[0-7])|(?:\B\.\d*|\b\d+(?:\.\d*)?)(?:e[-+]?\d+)?)l?\b/i,
+		'operator': /and=|or=|<[-=]|[-=]>|[*^=\-!+\/%=]=?|>>>=?|<<<=?|\^\^\^=?|\|\|\|=?|&&&=?|w\/=?|\.\.\.|~~~/,
```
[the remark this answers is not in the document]
fedonman: I edited the operators regex to follow grammar, and added the corresponding test

components/prism-qsharp.js
```diff
@@ -0,0 +1,168 @@
...
+			},
+            {
+				// namespace Quantum.App1;
+				pattern: re(/(\bnamespace\s+)<<0>>(?=\s*{)/.source, [identifier]),
+				lookbehind: true,
+				inside: typeInside
+			},
+		],
+		'keyword': keywords,
+		'number': /(?:0(?:x[\da-f_]*[\da-f]|b[01_]*[01]|o[0-7]*[0-7])|(?:\B\.\d*|\b\d+(?:\.\d*)?)(?:e[-+]?\d+)?)l?\b/i,
```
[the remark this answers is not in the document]
fedonman: You are right. Change accepted as is.

components/prism-qsharp.js
fedonman: Ops :)

components/prism-qsharp.js
```diff
@@ -0,0 +1,155 @@
...
+			pattern: /and=|or=/,
+			alias: 'operator'
+		}
+	});
+
+	// string interpolation
+	var formatString = /:[^}\r\n]+/.source;
+
+	// single line
+	var sInterpolationRound = nested(replace(/[^"'/()]|\/(?!\*)|\/\*(?:[^*]|\*(?!\/))*\*\/|<<0>>|\(<<self>>*\)/.source, [regularString]), 2)
```
RunDevelopment, line 116: Q# doesn't have multiline comments, so we can simplify this. Also, the string pattern changed, so we have to remove the `'` character.

```suggestion
	var sInterpolationRound = nested(replace(/[^"()]|<<0>>|\(<<self>>*\)/.source, [regularString]), 2)
```

components/prism-qsharp.js
```diff
@@ -0,0 +1,155 @@
...
+			'string': /[\s\S]+/
+		};
+	}
+
+	Prism.languages.insertBefore('qsharp', 'string', {
+		'interpolation-string': {
+			pattern: re(/(^|[^@\\])\$"(?:\\.|\{\{|<<0>>|[^\\"{])*"/.source, [sInterpolation]),
+			lookbehind: true,
+			greedy: true,
+			inside: createInterpolationInside(sInterpolation, sInterpolationRound),
```
RunDevelopment, line 149: Please inline the function call. The function is only used once now, so there's no reason to keep it around.

components/prism-qsharp.js
```diff
@@ -0,0 +1,155 @@
...
+	Prism.languages.insertBefore('qsharp', 'number', {
+		'range': {
+			pattern: /\.\./,
+			alias: 'operator'
+		}
+	});
+	
+	Prism.languages.insertBefore('qsharp', 'keyword', {
+		'andOrEqual': {
+			pattern: /and=|or=/,
```
RunDevelopment, line 107: Maybe we should highlight `and`, `or`, and `not` as operators instead? That way, `and` and `and=` are highlighted consistently. We do this in a few other languages as well but I will also accept if you keep it as is.

components/prism-qsharp.js
```diff
@@ -0,0 +1,155 @@
...
+	function keywordsToPattern(words) {
+		return '\\b(?:' + words.trim().replace(/ /g, '|') + ')\\b';
+	}
+	var keywords = RegExp(keywordsToPattern(keywordKinds.type + ' ' + keywordKinds.other));
+
+	// types
+	var generic = nested(/<(?:[^<>;=+\-*/%&|^]|<<self>>)*>/.source, 2);
+	var name = /@?\b[A-Za-z_]\w*\b/.source;
+	var genericName = replace(/<<0>>(?:\s*<<1>>)?/.source, [name, generic]);
+	var identifier = replace(/(?!<<0>>)<<1>>(?:\s*\.\s*<<1>>)*/.source, [keywordKinds.other, genericName]);
```
RunDevelopment, line 60: You forgot something there.

```suggestion
	var identifier = replace(/(?!<<0>>)<<1>>(?:\s*\.\s*<<1>>)*/.source, [keywordsToPattern(keywordKinds.other), genericName]);
```

components/prism-qsharp.js
```diff
@@ -0,0 +1,155 @@
...
+    } 
+	// keywords
+	function keywordsToPattern(words) {
+		return '\\b(?:' + words.trim().replace(/ /g, '|') + ')\\b';
+	}
+	var keywords = RegExp(keywordsToPattern(keywordKinds.type + ' ' + keywordKinds.other));
+
+	// types
+	var generic = nested(/<(?:[^<>;=+\-*/%&|^]|<<self>>)*>/.source, 2);
+	var name = /@?\b[A-Za-z_]\w*\b/.source;
```
RunDevelopment, line 58: The `@` is C#-specific.

```suggestion
	var name = /\b[A-Za-z_]\w*\b/.source;
```

components/prism-qsharp.js
```diff
@@ -0,0 +1,155 @@
...
+
+	var typeInside = {
+		'keyword': keywords,
+		'punctuation': /[<>()?,.:[\]]/
+	};
+
+	// strings
+	var regularString = /"(?:\\.|[^\\"\r\n])*"/.source;
+
+	Prism.languages.qsharp = Prism.languages.extend('clike', {
```
RunDevelopment, line 70: Please override the comments. C-like support for single-line and multiline comments.

```suggestion
	Prism.languages.qsharp = Prism.languages.extend('clike', {
		'comment': /\/\/.*/,
```

components/prism-qsharp.js
```diff
@@ -0,0 +1,155 @@
...
+		return pattern.replace(/<<self>>/g, '[^\\s\\S]');
+	}
+
+	// https://docs.microsoft.com/en-us/azure/quantum/user-guide/language/typesystem/
+	// https://github.com/microsoft/qsharp-language/tree/main/Specifications/Language/5_Grammar
+	var keywordKinds = {
+		// keywords which represent a return or variable type
+		type: 'Adj BigInt Bool Ctl Double false Int One Pauli PauliI PauliX PauliY PauliZ Qubit Range Result String true Unit Zero',
+		// all other keywords
+        other: 'Adjoint adjoint and apply as auto body borrow borrowing Controlled controlled distribute elif else fail fixup for function if in internal intrinsic invert is let mutable namespace new newtype not open operation or repeat return self set until use using while within'
```
RunDevelopment, line 48: Please use tabs for indentation and trim all trailing whitespaces.

RunDevelopment (changes requested):
I think this will be the last round.

components/prism-qsharp.js
fedonman: Done

components/prism-qsharp.js
fedonman: Done

components/prism-qsharp.js
fedonman: Done

components/prism-qsharp.js
fedonman: Done

components/prism-qsharp.js
fedonman: I removed `and`, `or`, and `not` from keywords and added them to operators instead. Now, everything is highlighted from `operators` regex.

components/prism-qsharp.js
fedonman: Done

components/prism-qsharp.js
```diff
@@ -0,0 +1,145 @@
...
+	function keywordsToPattern(words) {
+		return '\\b(?:' + words.trim().replace(/ /g, '|') + ')\\b';
+	}
+	var keywords = RegExp(keywordsToPattern(keywordKinds.type + ' ' + keywordKinds.other));
+
+	// types
+	var generic = nested(/<(?:[^<>;=+\-*/%&|^]|<<self>>)*>/.source, 2);
+	var name = /\b[A-Za-z_]\w*\b/.source;
+	var genericName = replace(/<<0>>(?:\s*<<1>>)?/.source, [name, generic]);
+	var identifier = replace(/(?!<<0>>)<<1>>(?:\s*\.\s*<<1>>)*/.source, [keywordsToPattern(keywordKinds.other), genericName]);
```
RunDevelopment, line 60: Since `identifier` is only used by the classname patterns, we can do the following:

1. Renaming. The Antlr4 grammar calls our `name` [`identifier`](https://github.com/microsoft/qsharp-language/blob/a51b436235fdb9cdbab4dc43089e1a46303933ac/Specifications/Language/5_Grammar/QSharpLexer.g4#L149) and our `identifier` [`qualifiedName`](https://github.com/microsoft/qsharp-language/blob/a51b436235fdb9cdbab4dc43089e1a46303933ac/Specifications/Language/5_Grammar/QSharpParser.g4#L13).
2. The Q# patterns are actually a lot simpler than the C# ones. E.g. no generics, no keywords.

When using my suggestion, please keep in mind that the variable `identifier` has been renamed to `qualifiedName`.

```suggestion
	var identifier = /\b[A-Za-z_]\w*\b/.source;
	var qualifiedName = replace(/<<0>>(?:\s*\.\s*<<0>>)*/.source, [identifier]);
```

components/prism-qsharp.js
```diff
@@ -0,0 +1,145 @@
...
+
+	var typeInside = {
+		'keyword': keywords,
+		'punctuation': /[<>()?,.:[\]]/
+	};
+
+	// strings
+	var regularString = /"(?:\\.|[^\\"\r\n])*"/.source;
```
RunDevelopment, line 68: Strings are allowed to contain line breaks.

```suggestion
	var regularString = /"(?:\\.|[^\\"])*"/.source;
```

components/prism-qsharp.js
```diff
@@ -0,0 +1,145 @@
...
+							pattern: /[\s\S]+/,
+							alias: 'language-qsharp',
+							inside: Prism.languages.qsharp
+						}
+					}
+				},
+				'string': /[\s\S]+/
+			}
+		}
+	});
```
RunDevelopment, line 141: Q#'s string interpolation grammar is a lot simpler than C#'s one, so we can simplify this.

```suggestion
	// single line
	var interpolationExpr = nested(replace(/\{(?:[^"{}]|<<0>>|<<self>>)*\}/.source, [regularString]), 2);

	Prism.languages.insertBefore('qsharp', 'string', {
		'interpolation-string': {
			pattern: re(/\$"(?:\\.|<<0>>|[^\\"{])*"/.source, [interpolationExpr]),
			greedy: true,
			inside: {
				'interpolation': {
					pattern: re(/((?:^|[^\\])(?:\\\\)*)<<0>>/.source, [interpolationExpr]),
					lookbehind: true,
					inside: {
						'punctuation': /^\{|\}$/,
						'expression': {
							pattern: /[\s\S]+/,
							alias: 'language-qsharp',
							inside: Prism.languages.qsharp
						}
					}
				},
				'string': /[\s\S]+/
			}
		}
	});
```

RunDevelopment (changes requested):
Last review round, I think.

Please take a look at my comments and resolve the merge conflicts.

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @fedonman!

PR #2810, opened 2021-03-17
Improve support for Dart classes & generics

brianegan:
- No longer depend on keywords such as `new` or `extends`
  - Detect PascalCase class names instead
  - Support generic parameters as well
  - Properly highlights class constructors
  - Added tests

| Before  | After |
| ------------- | ------------- |
| Screenshot from 2021-03-17 14-37-42 | Screenshot from 2021-03-17 14-39-15 |
| Screenshot from 2021-03-17 14-38-05  | Screenshot from 2021-03-17 14-39-06  |


components/prism-dart.js
```diff
@@ -1,24 +1,67 @@
...
+	Prism.languages.insertBefore('dart','class-name',{
+		'generics': {
+			pattern: /<(?:[\w\s,.&?]|<(?:[\w\s,.&?]|<(?:[\w\s,.&?]|<[\w\s,.&?]*>)*>)*>)*>/,
+			inside: {
+				'class-name': className,
+				'keyword': keywords,
+				'punctuation': /[<>(),.:]/,
+				'operator': /[?&|]/
+			}
+		},
```
RunDevelopment, line 65: You copied this from Java, right? That's not a problem but please adjust the `punctuation` and `operator` patterns in it. There are a few characters in there that it simply cannot contain.

Also, this a problem that Java has as well. It will incorrectly detect generics in code like `if (a<b&&c>d)`. Maybe we could solve this by requiring that generics are preceded by a class name? Kinda like this: 

```js
{
  pattern: /(\b[A-Z]\w*\s*)<(?:...)*>/,
  lookbehind: true,
  inside: { ... }
}
```

RunDevelopment (changes requested):
Thank you for the PR @brianegan!

One small nit and then gtg.

components/prism-dart.js
RunDevelopment: > `var names = <String>{};`

Forget what I said...

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @brianegan!

brianegan:
You're welcome, thanks for the review and merging :)

PR #2814, opened 2021-03-18
C++: Added generic functions and made `::` punctuation

RunDevelopment:
This adds support for generic function invocations (e.g. `std::make_unique<int>(4)`). 

I also changed the token type of double colon `::` from operator to punctuation. I choose to do this because this token is very common and has similar semantics as a dot `.` in other languages.


tests/languages/cpp/base-clause_feature.test
```diff
@@ -57,31 +61,34 @@ class service : private Transport // comment
...
 	["punctuation", ";"],
+
 	["keyword", "class"],
 	["class-name", "Y"],
 	["operator", ":"],
 	["base-clause", [
 		["keyword", "virtual"],
-		" baz",
-		["operator", "::"],
+		["class-name", "baz"],
```
RunDevelopment, line 70: Note to self: This is unexpected. Investigate.

PR #2818, opened 2021-03-19
PHP: Fixed methods with keyword names + minor improvements

### LINKED ISSUE #2817: Keywords should be colored as methods, when used as methods.
**Information**
- Language: PHP
- Plugins: none

**Description**

In PHP there are keywords, which can't be used as class names or variable names, but they can be used as method names.

For example, `match` and `foreach` are keywords, but they can be used as `$var->match()` and `$var->forEach()` methods. 

For example GitHub and PhpStorm colors this code corectly.

**Code snippet**

[Test page](https://prismjs.com/test.html#language=php&text=%24variable-%3Eforeach()%3B%20%2F%2F%20this%20%22foreach%22%20is%20a%20method%0A%24variable-%3Emethod()%3B%20%20%2F%2F%20this%20is%20also%20a%20method%0A%0Aforeach%20(%24list%20as%20%24value)%20%7B%20%2F%2F%20this%20%22foreach%22%20is%20a%20keyword%0A%7D)

<details>

```php
$variable->foreach(); // this "foreach" is a method
$variable->method();  // this is also a method

foreach ($list as $value) { // this "foreach" is a keyword
}
```

</details>

RunDevelopment:
This resolves #2817.

---

Apart from the obvious fix, I also made some minor improvements to the code.


Danon:
@RunDevelopment Hey, perhaps I should've mentioned. Instance methods are called with `->` operator, but static methods are called with `::` operator. So I believe you need one pattern for function definition, one for instance method call and one for static method call.

Here's a Prism test page: https://prismjs.com/test.html#language=php&text=%24variable-%3Eforeach()%3B%20%2F%2F%20this%20%22foreach%22%20is%20a%20method%0A%24variable-%3Emethod()%3B%20%20%2F%2F%20this%20is%20also%20a%20method%0A%0Aforeach%20(%24list%20as%20%24value)%20%7B%20%2F%2F%20this%20%22foreach%22%20is%20a%20keyword%0A%7D%0A%0ATest%3A%3Aforeach()%3B%0ATest%3A%3Amethod()%3B%0A

```php
$variable->foreach(); // this "foreach" is a method
$variable->method();  // this is also a method

foreach ($list as $value) { // this "foreach" is a keyword
}

Test::foreach();  // this is "foreach" static method
Test::method();  // this is "method" static method

# Exceptions
$variable->class(); // this "class" should be interpreted as "keyword"
Test::class; // This "class" should still be a keyword, just as it is now. Please, don't break it.
```

RunDevelopment:
Wait, so `class` is always a keyword? [Docs says](https://www.php.net/manual/en/reserved.keywords.php):

> As of PHP 7.0.0 these keywords are allowed as property, constant, and method names of classes, interfaces and traits, except that `class` may not be used as constant name.

So `class` can't be a constant but you also listed a method named `class` under exceptions.

Danon:
> Wait, so class is always a keyword? 

Yes, `class` is always a keyword.

> So class can't be a constant but you also listed a method named class under exceptions.

I meant that since PHP 7.4, you can use all keywords as method names, with the exception of `class` which can't.

Issue #309, opened 2014-07-30
Add contributing docs


apfelbox:
We need docs on how to contribute in general, how the newly introduced gulp workflow works and how you install all of it.

Docs for #269

QuantumInformation:
+1 for gitter

anxbt:
i can generate readme.md,to show how to clone/setup the project locally.can u assign me this issue.

PR #2825, opened 2021-03-23
docs: add gitter badge in README

ersachin3112:
Add glitter badge in readme as suggested in https://github.com/PrismJS/prism/issues/309#issuecomment-50861379

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Build Status](https://github.com/PrismJS/prism/workflows/CI/badge.svg)](https://github.com/PrismJS/prism/actions)
 [![npm](https://img.shields.io/npm/dw/prismjs.svg)](https://www.npmjs.com/package/prismjs)
+[![Gitter chat](https://badges.gitter.im/LeaVerou/prism.png)](https://gitter.im/LeaVerou/prism)
 
 Prism is a lightweight, robust, and elegant syntax highlighting library. It's a spin-off project from [Dabblet](https://dabblet.com/).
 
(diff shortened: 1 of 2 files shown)


RunDevelopment:
Thank you for the PR @ersachin3112!

However, I'm not sure whether we still use gitter... There is also the Github discussion tab which (while not enable in this repo) seems to have a similar purpose but __a lot__ more structured.

I would like to know what the other maintainers think. 
@LeaVerou @mAAdhaTTah @Golmote

Personally, I'd rather enable the GH discussion tab instead of reviewing gitter.

mAAdhaTTah:
Yeah, I agree. Gitter has been quiet since 2018 and I would prefer to not have to install another chat app. GitHub discussions would be good, if we think it's needed.

LeaVerou:
I thought Github Discussions were more forum than chat? Anyhow, I'm fine with whatever you decide!

Another option could be a Slack channel.

RunDevelopment:
> I thought Github Discussions were more forum than chat

Good point. Yes, it's not a perfect replacement but I think that something forum-like would be better. 

Many of the issues we currently get are in this Q&A form. Opening a GH discussions tab could lead to people asking their questions over there instead of opening an issue. This would neatly separate actual issues from questions.

What do you think? Want to give GitHub discussions a try? If it doesn't work out, we can just close it again.

LeaVerou:
Nothing to lose by giving it a try!

ersachin3112:
Yeah, Enabling GitHub discussion sounds good

RunDevelopment:
I enabled the discussions tab. I also added a link to our issue templates, so people can more easily find it.

image

---

Since we now have GitHub Discussions, I will close this PR for now. Feel free to ask to reopen this if gitter is still needed or if we closed GitHub Discussions.

PR #2826, opened 2021-03-23
Copy to Clipboard: iFrame fix

### LINKED ISSUE #2824: Current Prism version breaks "Copy to Clipboard Button" when used in iFrame without permission "clipboard-write"
**Information:**
- Prism version: Current
- Plugins: Copy to Clipboard Button
- Environment: Chrome (in iFrame)

**Description**
After updating to a new version of Prism by using the downloader from https://prismjs.com/download.html the "Copy to Clipboard Button" functionality is broken. The copy button worked with an older version of Prism (1.22.0) that was downloaded some time ago from the same URL.
IMHO, this has to do with the changes in #2784 as this replaces the copy to clipboard logic. The current code only tests if "navigator.clipboard" exists and if yes, it uses navigator.clipboard.writeText() but this could result in the following error if the iFrame does not have the right permissions:

`Uncaught (in promise) DOMException: The Clipboard API has been blocked because of a permissions policy applied to the current document. See https://goo.gl/EuHzyv for more details.`

In the old implementation https://clipboardjs.com/ was used which seems to either does not rely on navigator.clipboard.writeText() or check if the API is not blocked.

RunDevelopment:
This fixes #2824.

If `clipboard.writeText` fails for any reason, the implementation will now use the fallback method before. This should work around the permission problem in iFrames. I am aware that what I just said means: "this will circumvent browser security measurements," but there's not much I can do about it. The API that is supposed to be used to request access (in a secure way) is still a working draft and only supported by one browser engine.

Even though libraries like clipboard.js only use the fallback method, I think we should still keep `clipboard.writeText`. The fallback uses deprecated non-standard methods, so I believe that it's a good idea to have `clipboard.writeText` around in case a browser removes `execCommand` in the future.


plugins/copy-to-clipboard/prism-copy-to-clipboard.js
```diff
@@ -63,7 +63,10 @@
 	/** @param {CopyInfo} copyInfo */
 	function copyTextToClipboard(copyInfo) {
 		if (navigator.clipboard) {
-			navigator.clipboard.writeText(copyInfo.getText()).then(copyInfo.success, copyInfo.error);
+			navigator.clipboard.writeText(copyInfo.getText()).then(copyInfo.success, function() {
```
sohamsshah, line 66: This looks good 👍
You can add error catching block for promises.

PR #2831, opened 2021-03-25
Added ESLint with minimal config

RunDevelopment:
**Motivation:**
#1796 seems to have become stale because of disagreement over indentation style and the sheer number of changes (look at the number of conflicts).

**Description:**
This PR adds ESLint with a minimal config. This means that the ESLint config only contains checks for possible errors and best practices. It __does not contain style rules__.

The idea is to get ESLint now and add style rules later. That way, we can benefit from error checking now and can discuss the style we want to enforce in detail in dedicated PRs. Each dedicated PR will only add/change one rule (or a small number of rules), so we can move quickly and review each rule change individually.

**Future PRs:**
There are a few PRs I will make as soon as this one got merged:

1. A PR to re-enabling disabled rules with TODOs. A few rules are disabled because fixing them (while possible) isn't easy and because I wanted to keep the diff of this PR as small as possible.
2. A PR to enforces our string quote style.
3. A PR to enforce semicolons.

## Changes

- Added ESLint + config
- Added GH actions CI job for linting
- Fixed all linting errors (while trying to keep the diff as minimal as possible)


LeaVerou (approved):
LGTM! I think that's a good strategy, add minimal ESLint at first, then add rules gradually.

LeaVerou:
Btw at some point we should rewrite all these `var` into `let` and `const`.

RunDevelopment:
> Btw at some point we should rewrite all these `var` into `let` and `const`.

Guess I'll make a PR for the [`no-var`](https://eslint.org/docs/rules/no-var) as well after this :)

joshgoebel:
> Guess I'll make a PR for the no-var as well after this :)

Welcome to the future, Prism.js!  Glad to see this happening.

LeaVerou:
> > Btw at some point we should rewrite all these `var` into `let` and `const`.
> 
> Guess I'll make a PR for the [`no-var`](https://eslint.org/docs/rules/no-var) as well after this :)

I am envious of your energy! 😁

PR #2832, opened 2021-03-26
Accept hyphens in bash function names

hmarr:
While not a standard feature of POSIX shells, several popular shells including bash and zsh support hyphens in function names ([source](https://unix.stackexchange.com/questions/168221/are-there-problems-with-hyphens-in-functions-aliases-and-executables)).


RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @hmarr!

PR #2842, opened 2021-03-31
AQL: Update for ArangoDB 3.8

Simran-B:
- Add K_PATHS and WINDOW keywords
- Support binary and hexadecimal number literals

https://www.arangodb.com/docs/devel/release-notes-new-features38.html


RunDevelopment:
Thank you for contributing @Simran-B!

PR #2845, opened 2021-04-03
Ruby: Added missing regex flags

RunDevelopment:
This fixes #2841.


sampatbadhe (commented):
changes looks good to me 👌

PR #2851, opened 2021-04-07
small regex changes for better generic log support

dkanada:
I encountered a few small issues while trying out the new generic log support.

Levels: Serilog uses three letter acronyms for their log levels
Exception: Many programming languages would benefit from highlighted exceptions (Java and C# for example)
UUID: Microsoft uses GUIDs which can be valid without dashes
Date: Serilog and Python output datetimes without a `T` character by default
Time: RFC 3339 allows `+00:00` for the timezone portion

There were a few larger problems I encountered that I'd like to get some input on before changing the order and existence of the current rules.

1. File paths will often contain UUIDs so I'd like to move the path and hash rules above the UUID rule.
2. The string matching is *extremely* greedy and was stealing almost all tokens that would have otherwise matched UUID, path, IP, or URL patterns. I solved it by removing that pattern entirely, but I figured that might be too drastic.
3. Any paths with spaces in the name weren't caught by the `file-path` rule. The logfile I was testing used strings to avoid path ambiguity (further exacerbating the string problem) but it doesn't seem like that rule accounts for this situation.


dkanada:
I figured this would cause issues with any tests haha, let me know which changes make sense and I can fix the problems.

RunDevelopment:
Thank you for the PR @dkanada!

Lots of good ideas in here.

I'll go through your proposed changes one by one. I will only comment on whether these changes should be made not how they were/will be made. (Comments on the implementation will follow.)

In general, please try to share examples if you can. There is no one log format, so I have no idea what the logs you try to highlight look like.

> Levels: Serilog uses three letter acronyms for their log levels

Accepted.

> Exception: Many programming languages would benefit from highlighted exceptions (Java and C# for example)

From my testing, Java stack traces looked ok but I haven't tested C# exceptions. Could you please provide a few stack traces (just all in one file) that don't look good?

> UUID: Microsoft uses GUIDs which can be valid without dashes

Accepted.

> Date: Serilog and Python output datetimes without a `T` character by default

Accepted.

> Time: RFC 3339 allows `+00:00` for the timezone portion

Accepted.

> 1. File paths will often contain UUIDs so I'd like to move the path and hash rules above the UUID rule.

File paths should be tokenized as such. IMO we should rather tokenize UUIDs _inside_ file paths.

> 2. The string matching is _extremely_ greedy and was stealing almost all tokens that would have otherwise matched UUID, path, IP, or URL patterns. I solved it by removing that pattern entirely, but I figured that might be too drastic.

Strings are commonly used to store this kind of data. I don't really see why would want to not highlight strings. Do you have an example?

> 3. Any paths with spaces in the name weren't caught by the `file-path` rule. The logfile I was testing used strings to avoid path ambiguity (further exacerbating the string problem) but it doesn't seem like that rule accounts for this situation.

Yes, that is intentional. The problem is that some log formats use tabs or spaces to separate arguments. The separator between a path and an error message might be a single space. A trade-off between false positives and false negatives had to be made and I decided on false negatives.

components/prism-log.js
```diff
@@ -12,27 +12,31 @@ Prism.languages.log = {
 
 	'level': [
 		{
-			pattern: /\b(?:ALERT|CRIT|CRITICAL|EMERG|EMERGENCY|ERR|ERROR|FAILURE|FATAL|SEVERE)\b/,
+			pattern: /\b(?:ERR|ALERT|CRIT|CRITICAL|EMERG|EMERGENCY|ERR|ERROR|FAILURE|FATAL|SEVERE)\b/,
```
RunDevelopment, line 15: We sort word lists alphabetically.

components/prism-log.js
```diff
@@ -93,7 +97,7 @@ Prism.languages.log = {
 		alias: 'number'
 	},
 	'time': {
-		pattern: /\b\d{1,2}:\d{1,2}:\d{1,2}(?:[.,:]\d+)?(?:\s?[+-]\d{2,4}|Z)?\b/,
+		pattern: /\b\d{1,2}:\d{1,2}:\d{1,2}(?:[.,:]\d+)?(?:\s?[+-]\d{2}:\d{2}|Z)?\b/,
```
RunDevelopment, line 100: Your change means that `... +0004` isn't accepted anymore.

```suggestion
		pattern: /\b\d{1,2}:\d{1,2}:\d{1,2}(?:[.,:]\d+)?(?:\s?[+-]\d{2}(?:\d{0,2}|:\d{2})|Z)?\b/,
```

components/prism-log.js
```diff
@@ -66,7 +70,7 @@ Prism.languages.log = {
 	},
 
 	'uuid': {
-		pattern: /\b\w{8}-\w{4}-\w{4}-\w{4}-\w{12}\b/,
+		pattern: /\b\w{8}-?\w{4}-?\w{4}-?\w{4}-?\w{12}\b/,
```
RunDevelopment, line 73: This is something I did wrong, but it will really be a problem now: we really need to enforce hexadecimal digits.

```suggestion
		pattern: /\b[0-9a-f]{8}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{12}\b/i,
```

components/prism-log.js
```diff
@@ -12,27 +12,31 @@ Prism.languages.log = {
...
 		},
 		{
-			pattern: /\b(?:FINER|FINEST|TRACE|VERBOSE)\b/,
+			pattern: /\b(?:TRC|VRB|FINER|FINEST|TRACE|VERBOSE)\b/,
 			alias: ['trace', 'comment']
 		}
 	],
 
+	'exception': {
+		pattern: /(.*EXCEPTION.*)(\n\s+at\s+.+)+/im
```
RunDevelopment, line 37: Welcome to the world of exponential backtracking! I will be your guide.

But seriously: this pattern has an exponential worst-case runtime. What kind of log is this pattern supposed to match? Could you give me an example?

RunDevelopment (changes requested):

components/prism-log.js
dkanada: Yeah, I noticed this rule was catching random groups of 32 alphanumeric characters now. I'll switch it to hexadecimal.

components/prism-log.js
dkanada: I knew this rule would be a problem 😆 It was mainly just to start a discussion. Lots of program logs contain multiline exceptions for when things really go wrong, and my goal here was to highlight the *entire* exception and really make it stand out.

dkanada: I'm on mobile at the moment but you can see an example exception in the logs of [this](https://github.com/jellyfin/jellyfin/issues/2355) issue.

components/prism-log.js
RunDevelopment: I just noticed, it already contains `ERR`. 

This is why these things are sorted.

components/prism-log.js
RunDevelopment: What is the benefit of capturing the whole exception? The current highlighting looks ok (not perfect but ok) to me.

image

dkanada: It's definitely fine right now, but I believe that exception is highlighted with `property` instead of a unique token for errors or exceptions, which would be nice. The multiline check was specifically just to wrap the entire section for highlighting.

dkanada:
Regarding the string issues, I have no problems with matching strings at all. However, it seemed like the string matches were blocking any further matches for UUID or other more specific tokens. I don't personally want to highlight strings in my logs, but I definitely want to use a special color for IP addresses. I thought the UUID was overriding paths but I can't reproduce that one now so you can ignore that comment.

```
file-path and possible uuid are only matched as string

Output #0, hls, to '/data/transcodes/275395c7ef978782d9755ae338e67cb7.m3u8':
  Metadata:
    encoder         : Lavf58.45.100

url and ip-address are only matched as string

[{timestamp}] [INF] [205] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "{ip}" closed
[{timestamp}] [WRN] [255] Jellyfin.Server.Middleware.ResponseTimeMiddleware: Slow HTTP Response from "{url}" to "{ip}"
```

dkanada:
It actually might not be an issue with strings but simply the special token matching - perhaps quotes aren't getting counted as word boundaries. I feel like wrapping variables in quotes is a relatively common practice though, so this doesn't seem unreasonable.

One more item I wanted to bring up was the properties, which match a trailing colon. Was that implemented to improve YAML and key-value output? Perhaps it would be better to match the colons as operators and remove them from the property rule with a non-capturing group.

components/prism-log.js
dkanada: Removed the extra `ERR` and sorted the other terms alphabetically.

components/prism-log.js
RunDevelopment: > a unique token for errors or exceptions, which would be nice

That is not possible without assuming a stack trace format. This error-message-plus-stack-trace format with `at` is somewhat shared by C#, Java, and JS but there are other formats as well like the ones used in C++, Rust, or Python. It's not necessarily a problem that we miss some languages tho.

Also, is it even useful to tokenize the entire error?

RunDevelopment:
> it seemed like the string matches were blocking any further matches for UUID or other more specific tokens

<sup>This is something I wanted to include in my first comment but forgot to.</sup> 

The solution to this is to highlight IP-addresses, UUIDs, etc _inside_ strings. Prism has the `inside` property for this ([example](https://github.com/PrismJS/prism/blob/3786f3965cb433cb16d9a2f467337916c28eb8db/components/prism-c.js#L24)).

> [...] properties, which match a trailing colon [...] Perhaps it would be better to match the colons as operators and remove them from the property rule [...]?

Good idea. Go for it! (But please use a lookahead.)

> Was that implemented to improve YAML and key-value output?

Yes, many logs contain key-value pairs using this pattern. But I actually never tested it on yaml files.

components/prism-log.js
dkanada: I think it's useful, but since the performance issues are pretty clear, I will concede that it's probably not worth adding to this ruleset.

dkanada:
I decided against the property colon change in the end. One last issue is that the hash rule is designed in a way that currently conflicts with the new GUID changes. How do you want that resolved? Also, it seems the build CI is failing since the `.min` file isn't updated, am I expected to update that manually or should I ignore it for now?

components/prism-log.js
```diff
@@ -83,7 +83,7 @@ Prism.languages.log = {
 
 	'date': {
 		pattern: RegExp(
-			/\b\d{4}[-/]\d{2}[-/]\d{2}T(?=\d{1,2}:)/.source +
+			/\b\d{4}[-/]\d{2}[-/]\d{2}(?=[T\s]\d{1,2}:)/.source +
```
RunDevelopment, line 86: The `T` is part of the date.

```suggestion
			/\b\d{4}[-/]\d{2}[-/]\d{2}(?:T(?=\d{1,2}:)|(?=\s\d{1,2}:))/.source +
```

components/prism-log.js
```diff
@@ -93,7 +93,7 @@ Prism.languages.log = {
 		alias: 'number'
 	},
 	'time': {
-		pattern: /\b\d{1,2}:\d{1,2}:\d{1,2}(?:[.,:]\d+)?(?:\s?[+-]\d{2,4}|Z)?\b/,
+		pattern: /\b[T\s]?\d{1,2}:\d{1,2}:\d{1,2}(?:[.,:]\d+)?(?:\s?[+-]\d{2}:?\d{2}|Z)?\b/,
```
RunDevelopment, line 96: Since `T` is part of the date, it shouldn't be part the time.

```suggestion
		pattern: /\b\d{1,2}:\d{1,2}:\d{1,2}(?:[.,:]\d+)?(?:\s?[+-]\d{2}:?\d{2}|Z)?\b/,
```

RunDevelopment (changes requested):
> Am I expected to update that manually

Yes. Run `npm run build` please.

> the hash rule is designed in a way that currently conflicts with the new GUID changes

I haven't thought of that... This is definitely a problem but how are we going to resolve it?

The string `01234567890123456789012345678901` (32 hex digits) could be a hash (e.g. an MD5 hash) or a GUID without hyphens. What is it?

I say we drop unhyphenated GUIDs. UUIDs and GUIDs are supposed to be hyphenated anyway. Let's require the hyphens again.

components/prism-log.js
RunDevelopment: Performance isn't an issue. For every exponentially backtracking regex, we can find an equivalent fast regex. Don't worry, I'll help you find it :)

Now that I think about it, if we could reliably find stack trace and identify the programming language (Java/JS/C#/etc.), then we could use our JavaStackTrace and JS Stack trace language to highlight them. That would be pretty cool.

For now, could you test on your side whether the following works:

```js
	'exception': {
		pattern: /(^|[^\w.])[\w.]*(?:Exception|Error):.*(?:(?:\r\n?|\n)[ \t]*(?:at[ \t].+|\.{3}.*|Caused by:.*))+/,
		greedy: true,
	},
```

I only tested it for Java stack traces and JS stack traces but it should work for C# stack traces as well.

dkanada: This works relatively well, but still trips up on the following section.

```
System.ArgumentNullException: Value cannot be null. (Parameter 'item')
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.InvokeInnerFilterAsync()
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextResourceFilter>g__Awaited|24_0(ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
```

dkanada:
I'm poking at the `inner` rule, but I'm still a bit unclear about its usage. As one example, would the UUID rule need to be duplicated at the root level and the `inner` block for strings?

components/prism-log.js
RunDevelopment: I didn't know C# stack traces had this feature... Here's an updated version of the above pattern:

```
	'exception': {
		pattern: /(^|[^\w.])[\w.]*(?:Exception|Error):.*(?:(?:\r\n?|\n)[ \t]*(?:at[ \t].+|\.{3}.*|Caused by:.*|--- .*))+/,
		greedy: true,
	},
```

RunDevelopment:
Yes, you need to duplicate the patterns you want inside strings. Luckily, there are variables :) In the case of log files, it should probably be done kinda like this:

```js
var uuid = { pattern: ..., ... }; // the UUID pattern
var url = /.../; // the URL pattern

Prism.languages.log = {
	'string': {
		// Single-quoted strings must not be confused with plain text. E.g. Can't isn't Susan's Chris' toy
		pattern: /"(?:[^"\\\r\n]|\\.)*"|'(?![st] | \w)(?:[^'\\\r\n]|\\.)*'/,
		greedy: true,
		inside: {
			'url': url,
			'uuid': uuid
		}
	},
	...,
	'url': url,
	...,
	'uuid': uuid,
	...
};
```

Please don't forget to wrap the whole thing into an IIFE or else we pollute the global namespace.

components/prism-log.js
dkanada: That did the trick, should I add this to the ruleset?

RunDevelopment: Sure! Isn't that what you wanted in the first place?

dkanada:
After updating the tests for the exception rule, I noticed URLs and other tokens inside the exception were no longer getting matched, which might not be desirable. Since I have somewhat lost my spare time recently, I'd rather get the current changes merged than let the PR grow stale while I consider what to do about the inner matching.

RunDevelopment:
> Since I have somewhat lost my spare time recently, I'd rather get the current changes merged than let the PR grow stale while I consider what to do about the inner matching.

Perfectly understandable. 

Thank you for contributing!

PR #2858, opened 2021-04-14
Updated PHP list of keywrods with PHP 7.4 'fn' keyword

Danon:
In PHP 7.4 there comes a new keyword, 'fn'.

I added it to the theme, I checked, and it works correctly:
 - As a lambda (checked)
 - As a function declaration (checked)
 - As a function call (checked)
 - As a classname (checked)
 - As a namespace (checked)


RunDevelopment:
Please rebuild Prism by running `npm run build`.

Apart from that, this PR is gtg.

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @Danon!

PR #2868, opened 2021-04-19
glimmer grammar

NullVoxPopuli:
not ready for review -- throwing this up so I don't forget about  it


components/prism-glimmer.js
```diff
@@ -0,0 +1,290 @@
...
+					},
+				},
+				...Punctuation,
+				...String,
+			},
+		},
+	};
+
+
+	// if (Prism.languages.markup) {
```
NullVoxPopuli, line 222: saving these for another PR as this one is pretty large

components.json
```diff
@@ -457,6 +457,23 @@
...
+				"hbs": "Handlebars",
+				"handlebars": "Handlebars",
+				"htmlbars": "HTMLBars"
+			},
+			"optional": [
+				"javascript"
+			],
+			"modify": [
+				"javascript"
+			]
```
RunDevelopment, line 475: `modify` implies `optional`.

```suggestion
			"modify": "javascript"
```

components.json
```diff
@@ -457,6 +457,23 @@
...
 		},
+		"glimmer": {
+			"title": "Glimmer / Ember",
+			"require": "markup-templating",
+			"owner": "NullVoxPopuli",
+			"alias": ["hbs", "handlebars", "htmlbars"],
+			"aliasTitles": {
+				"hbs": "Handlebars",
+				"handlebars": "Handlebars",
+				"htmlbars": "HTMLBars"
```
RunDevelopment, line 468: Why? We already have a language for Handlebars.

components.json
```diff
@@ -457,6 +457,23 @@
 			"title": "Git",
 			"owner": "lgiraudel"
 		},
+		"glimmer": {
+			"title": "Glimmer / Ember",
+			"require": "markup-templating",
```
RunDevelopment, line 462: You also directly use `markup` in your language definition.

I know that `markup-templating` requires `markup`, so `markup` is transitively required but it's better to be explicit.

```suggestion
			"require": ["markup", "markup-templating"],
```

components/prism-glimmer.js
```diff
@@ -0,0 +1,290 @@
+/* eslint-disable no-use-before-define */
+/*tri
+* Much of this implementation is copied from highlightjs-glimmer
+*
+*/
+function setup(Prism) {
```
RunDevelopment, line 6: This will pollute the global namespace. Please use an IIFE. [Example](https://github.com/PrismJS/prism/blob/80d475e4d8ddc8c5d744a4143b3078545d72d6aa/components/prism-css.js#L1).

components/prism-glimmer.js
```diff
@@ -0,0 +1,290 @@
...
+	const NUMBER = /\b0x[\dA-Fa-f]+\b|(?:\b\d+(?:\.\d*)?|\B\.\d+)(?:[Ee][+-]?\d+)?/;
+
+	const STRING =  new RegExp(regex.either(
+		/"[^{"]+/, /"/,
+		/'[^{']+/, /'/,
+		/"[^"]+"/,
+		/'[^']+'/,
+	));
+
+	const Arg = {
```
RunDevelopment, line 51: Please use lowerCamelCase for variables.

RunDevelopment (changes requested):
Thank you for making this language definition @NullVoxPopuli!

Please revert the changes to `package-lock.json` and convert your code into ES5. Luckily, most of the spread objects only contain one key, so getting rid of the ES2018 object spread should be quite simple. Please follow [this example](https://github.com/PrismJS/prism/blob/80d475e4d8ddc8c5d744a4143b3078545d72d6aa/components/prism-java.js#L9).

I also left you a few comments.

I will fully review this PR after these issues have been resolved.

components/prism-glimmer.js
RunDevelopment: Looking forward to it.

But please remove the commented-out code and make something like a TODO comment instead.

components.json
NullVoxPopuli: most ember/glimmer developers have been using `hbs` as a backstop for lack of better syntax support. By including the glimmer syntax instead of the handlebars syntax in their bundles, they'll get new highlights without having to change any of their code

components/prism-glimmer.js
NullVoxPopuli: I think making all variables lowerCamelCase would make things harder to reason about
-- I needed more conventions:

`PascalCase` - spreadable pre-configured set up grammar-key to grammar
`SHOUT_CASE` - Regex or other simpler things

Even with this, I don't think I'm 100% consistent

NullVoxPopuli:
> convert your code into ES5

Can we add a transpiler to gulp?

components/prism-glimmer.js
NullVoxPopuli: why not use modules and convert to IIFE during build for CDNs

RunDevelopment:
I agree with what @mAAdhaTTah [said on that matter](https://github.com/PrismJS/prism/pull/2873#issuecomment-826135845).

> As for ES6, I believe we need IE11 support for our unminified code as well, so I think that's a no go until the V2 rewrite.

So please write ES5 code.

components/prism-glimmer.js
RunDevelopment: PascalCase is usually reserved from classes. The JS community is pretty consistent with that. Please don't use it for variables.

As for upper snake case/shout case, this is usually used for constant (e.g. enum constants or magic numbers). Using it for regexes a little strange but I'm not against it.

components.json
RunDevelopment: Aliases are supposed to be either 1) alternative names referring to the same language (e.g. `js` and `javascript`) or 2) other languages that use the same (or only a subset of the) syntax (e.g. `svg` and `markup`).

I've seen that the Ember docs use `handlebar` but this is something they have to fix. We only provide languages (and syntax highlighting, I guess). Please remove the aliases.

However, these aliases aren't bad just misplaced. Do you want to make a separate PR adding these aliases to Handlebar? Otherwise, I'll do that.

components/prism-glimmer.js
RunDevelopment: Because of our download page. The "bundler" implemented by our download page only concatenates source files. It doesn't parse JS or anything, it's as dumb as it gets...

The "bundler" is pretty old and hasn't seen updates for years now. The only reason we haven't replaced it with a real bundler is that it just works (there haven't been any issues with it since the day I joined the project). The only inconvenience is that we have to use IIFEs.

However, this will change in v2. We plan to use ESM in our source files, so we'll need a real bundler then.

NullVoxPopuli:
I'll save ya the trouble of reviewing this further and just release my own npm package for this. I have a pretty good build setup that I did with highlight.js, seen here: https://github.com/NullVoxPopuli/highlightjs-glimmer/blob/main/scripts/build.cjs

thanks for the comments though!

RunDevelopment:
>thanks for the comments though!

You're welcome!

If you have any questions, feel free to ask us anytime on your [discussions page](https://github.com/PrismJS/prism/discussions).

PR #2873, opened 2021-04-24
Swap uglify for terser

NullVoxPopuli:
While working on: https://github.com/PrismJS/prism/pull/2868
I noticed that the existing build system does not support `const` or `let`.

This PR removes uglify,  which does not support ES6+, and replaces it with terser via https://www.npmjs.com/package/gulp-terser

Depending on Prism's browser support policy this will force a breaking change.

This PR is split into two commits:
1. the meaningful changes
2. npm run build


RunDevelopment:
> I noticed that the existing build system does not support `const` or `let`.

This isn't a technical limitation but an intentional choice. Prism v1.x is an ES5 project to support older browsers such as IE11. (I know that IE11 actually supports some ES6 features (such as `let` and `const`) but it is a lot easier to enforce ES5 than it is to enforce the specific subset of ES6 supported by IE11.) However, we do plan to drop IE11 in v2.0 (see #1578) but that might still take a while.

Therefore:

- This change cannot be part of Prism v1.x. We cannot drop support for IE11 without changing the major version.
- As for v2.0: using Terser is only one step we have to undertake in order to modernize our codebase. This PR will not help much with these tasks.

For these reasons, I will now close this PR.

---

@NullVoxPopuli 
I'm sorry to reject your PR but I cannot allow ES6+ features yet. Please use ES5 equivalents (such `var` instead of `let` and `const`) while the Prism team figures out a way to modernize this codebase.

NullVoxPopuli:
all good. do you have thoughts on transpiling down to ES5?

mAAdhaTTah:
Interesting that switching to Terser made all of the builds larger as well. If we continued to support our given syntax set but the minifier produced better results, I'd be in support but it seems like Terser produces worse results across the board for our codebase.

NullVoxPopuli:
> Interesting that switching to Terser made all of the builds larger as well

afaict, This is only because the variable `Prism` is no longer abbreviated to one letter, since it's a global

RunDevelopment:
Yes, terser is a little bit worse than uglify for ES5 but not by much. When talking about minified ES6+ code, it can do a bit more.

joshgoebel:
Well I was going to say terser can typically do much better than it's defaults (and this may still be true), but for us the difference is only 1% (300 bytes)... so I guess it all depends.  It's also possible we don't have it tuned as finely as it could be as well. :)

mAAdhaTTah:
I don't think that's the case because all of the languages are wrapped in IIFEs and pass in Prism, so they're mostly not referencing the global Prism. I only mention this because my general thing is keeping the bundle small, so if you wanna spend some time configuring Terser to produce smaller bundles, I'm all for it.

As for ES6, I believe we need IE11 support for our unminified code as well, so I think that's a no go until the V2 rewrite.

PR #2878, opened 2021-04-28
Support for RPG Free format

gianluca1606:
So i was finally able to create the pull request now, thanks for michael for the help.

Any suggestions or changes are welcome and i will try my best to do it.

The changes are pretty trackable since i only created a few files and builded the project.

It would be amazing if the pull requests gets accepted since there is no RPG syntax highlight for free from other libraries/tools.


RunDevelopment:
Thank you for the PR @gianluca1606!

I'll review it shortly.

PR #2911, opened 2021-05-21
fix: better tag selection

Chunjee:


RunDevelopment:
@Chunjee Could you please explain what issue you are trying to fix and how your fix works?

Chunjee:
Incorrect tag selection (first red box):
image

Post fix:
image

RunDevelopment:
I see, in that case, your fix is not correct. Try [this example](https://prismjs.com/test.html#language=autohotkey&text=%3B%20test%20code%0A%0AA.invert(%7B1%3A%20%22a%22%2C%202%3A%22A%22%7D)%0A%3B%20%3D%3E%20asdasd%0A%0ANormalLabel%3A%0A%3B%20something).

I have a question: are labels line-based? So can there be any no-space, no-comments after the label on the same line?

If they are, then your problem can be fixed like this:

```diff
@@ -11,8 +11,13 @@ Prism.languages.autohotkey = {
                }
        ],
+       'tag': {
+               // labels
+               pattern: /^([ \t]*)[^\s,`":]+(?=:[ \t]*$)/m,
+               lookbehind: true
+       },
        'string': /"(?:[^"\n\r]|"")*"/m,
-       'tag': /^[ \t]*[^\s:]+?(?=:(?:[^:]|$))/m, //labels
        'variable': /%\w+%/,
        'number': /\b0x[\dA-Fa-f]+\b|(?:\b\d+(?:\.\d*)?|\B\.\d+)(?:[Ee]-?\d+)?/,
```

Chunjee:
I am afraid that comments are allowed after a label; and in some unique circumstances, before.

After:
```autohotkey
validLabel: ; a comment
```

Before:
```autohotkey
/* multiline comment
*/ validLabel:
```

RunDevelopment:
> comments are allowed after a label; and in some unique circumstances, before.

Perfect. In that case, you can use the code I commented above.

Chunjee:
Using the code in the comments appears to have the same issue:
image

RunDevelopment:
No, it doesn't.

image

You probably forgot to remove the old `tag` (see diff above).

Chunjee:
Thank you, after some more experimentation I think the issue was a duplicate string definition.

PR #2914, opened 2021-05-23
Updated prism-jsx.min.js to match the latest prism.jsx.js version

jerehut:
The minified file was outdated so I minified the current prism.jsx.js. Some of the letters used for minification changed (i -> t), I hope that's not an issue?

Includes changes from #2836 #2862 #2881.


mAAdhaTTah:
It already is up to date. You may not be using the same version of uglify. Install with `npm ci` to make sure you're matching the same version and you should see the same output as what's on `master`.

That said, if there's a new version of uglify that's producing better results, we should probs upgrade and run it over everything. If you want to PR that, we'd be open to it.

PR #2921, opened 2021-05-28
Adding basic support for mathematica

msollami:
Support for the wolfram language:
- Followed "new language"  tutorial
- Added html examples and test files
- Initial work highlights basic common expressions including numbers, system-level symbols, comments, operators, and strings


components/prism-wolfram.js
```diff
@@ -0,0 +1,24 @@
+Prism.languages.wolfram = {
+	'comment': {
+		pattern: /(?:\(\*[\w\'\s\r\n\*]*\*\))/
+	},
```
RunDevelopment, line 4: 1. You can inline the regex.
2. Unnecessary non-capturing group.
2. `[\w\'\s\r\n\*]*`: typically, languages allow pretty much all characters inside comments. Wouldn't `\(\*[\s\S]*?\*\)` work better?

Also, asking out of curiosity, does the language support nested comments? I.e. `(* comment (* nested *) *)`.

```suggestion
	'comment': /\(\*[\s\S]*?\*\)/,
```

components.json
```diff
@@ -1346,6 +1346,14 @@
 			"require": "markup",
 			"owner": "Golmote"
 		},
+		"wolfram": {
+			"title": "Wolfram language",
+			"alias": "wl",
+			"aliasTitles": {
+				"wl": "Mathematica"
```
RunDevelopment, line 1353: Wouldn't it be better to have another alias `mathematica` for this alias title?

RunDevelopment (changes requested):
Thank you for making this language definition @msollami!

I left you a few comments with things to improve.

Also, I have a question about `class-name` and `regex`: Are these really class names/regexes? I don't know Mathematica but the patterns don't look like that to me.

components/prism-wolfram.js
```diff
@@ -0,0 +1,24 @@
...
+	'class-name': {
+		pattern: /(?:(?:\w+\`)+\w*)/i,
+		lookbehind: false
+	},
+	'string': {
+		pattern: /(")(?:\\.|(?!\1)[^\\\r\n])*\1/i,
```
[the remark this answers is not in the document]
msollami: Ah, good suggestion!

components/prism-wolfram.js
```diff
@@ -0,0 +1,24 @@
...
+	'class-name': {
+		pattern: /(?:(?:\w+\`)+\w*)/i,
+		lookbehind: false
+	},
```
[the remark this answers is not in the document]
msollami: @RunDevelopment 
1 & 2. Thanks!
3. You're correct there's no classes, but I wanted to add color to show contexts, which are formatted as follows (each has at least one backtick):   contextpaths`use`backticks`...`subcontextName`

components/prism-wolfram.js
```diff
@@ -0,0 +1,24 @@
...
+	'string': {
```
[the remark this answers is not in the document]
msollami: @RunDevelopment Ok will do!

components/prism-wolfram.js
msollami: @RunDevelopment Yes it supports nested comments, anything within the enclosing (* *) is considered a comment

components.json
msollami: @RunDevelopment Ah I thought it had to do with the file extension... sure that makes sense. Is there a way to tell the prism system which extensions a language uses?

components/prism-wolfram.js
```diff
@@ -0,0 +1,24 @@
...
...sterWolframID|$RequesterWolframUUID|$ResourceSystemBase|$ResourceSystemPath|$RootDirectory|$ScheduledTask|$ScriptCommandLine|$ScriptInputString|$SearchLanguage|$SecuredAuthenticationKeyTokens|$ServiceCreditsAvailable|$Services|$SessionID|$SetParentLink|$SharedFunctions|$SharedVariables|$SoundDisplay|$SoundDisplayFunction|$SourceLink|$SSHAuthentication|$SubtitleDecoders|$SubtitleEncoders|$SummaryBoxDataSizeLimit|$SuppressInputFormHeads|$SynchronousEvaluation|$SyntaxHandler|$System|$SystemCharacterEncoding|$SystemCredentialStore|$SystemID|$SystemMemory|$SystemShell|$SystemTimeZone|$SystemWordLength|$TargetSystems|$TemplatePath|$TemporaryDirectory|$TemporaryPrefix|$TestFileName|$TextStyle|$TimedOut|$TimeUnit|$TimeZone|$TimeZoneEntity|$TopDirectory|$TraceOff|$TraceOn|$TracePattern|$TracePostAction|$TracePreAction|$UnitSystem|$Urgent|$UserAddOnsDirectory|$UserAgentLanguages|$UserAgentMachine|$UserAgentName|$UserAgentOperatingSystem|$UserAgentString|$UserAgentVersion|$UserBaseDirectory|$UserBasePacletsDirectory|$UserDocumentsDirectory|$Username|$UserName|$UserURLBase|$Version|$VersionNumber|$VideoDecoders|$VideoEncoders|$VoiceStyles|$WolframDocumentsDirectory|$WolframID|$WolframUUID)\b/,
```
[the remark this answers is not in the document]
msollami: @RunDevelopment  These are the the symbols in the System` context, a strength of this language is that it has an extremely large builtin library of commands. I guess we don't need to color them.  

User-defined symbols can contain numbers but must start with letter, but the builtins are 7108 symbols that the system predefines (Names["System`*"] // Length == 7108) 

I'm not sure what the best practice is here...

components/prism-wolfram.js
RunDevelopment: > Yes it supports nested comments

Ah, ok. In that case, it's the same [comment syntax as Applescript](https://github.com/PrismJS/prism/blob/ad9878ad39cd7afaab9e3a67f99fef3d8167b28c/components/prism-applescript.js#L4). You can probably just copy-paste the regex.

components.json
RunDevelopment: We do file extensions via aliases too. You can have one alias `wl` and one `mathematica`.

components/prism-wolfram.js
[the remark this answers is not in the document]
RunDevelopment: In that case, you should give the token a proper name for the language construct you are trying to highlight and use an alias to assign it a color. [Example](https://github.com/PrismJS/prism/blob/ad9878ad39cd7afaab9e3a67f99fef3d8167b28c/components/prism-abnf.js#L16).

components/prism-wolfram.js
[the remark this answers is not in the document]
msollami: @RunDevelopment To answer your question - regex is not really a regex - mathematica has it's own internal pattern matcher, but it's very similar. Mathematica has what's called a Blank (_), BlankSequence (__), and BlankNullSequence (___), which are indicated by one 1 two or three underscores respectively.  They can be named and typed:

a_List  is a pattern named a that matches an expression with the head List, e.g. List[x,y]
b__Integer  is a pattern named b that matches any sequence of at least 1 Integer, e.g. (1,2,3,4)
___ matches any sequence of zero or more expressions

components/prism-wolfram.js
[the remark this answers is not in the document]
RunDevelopment: Maybe we could limit it to builtins that handle control flow (e.g. `If`)? After all, we don't have to highlight every builtin function.

> but must start with letter

But a few in the regex start with `$`?

components.json
msollami: Awesome, I'll add that.

components/prism-wolfram.js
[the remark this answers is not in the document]
RunDevelopment: Please don't re-use threads like this... It took me a minute to figure out which question you were referring to...

To it's a pattern-matching syntax. Cool feature but not a regex. I think we should do the same as with `class-name`. Pick a descriptive token name (maybe `pattern`?) and use `regex` as an alias.

components/prism-wolfram.js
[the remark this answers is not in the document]
msollami: Yes adding the $ prefix is their convention for system defined global variables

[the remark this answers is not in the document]
msollami: Ah I misspoke, I meant to say valid variable names _can't start with a number_ but other than that you can use arbitrary unicode.

components/prism-wolfram.js
msollami: Oh nice, I'll update to that.

components/prism-wolfram.js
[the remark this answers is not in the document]
RunDevelopment: > Yes adding the $ prefix is their convention for system defined global variables

Would that be something worth highlighting on its own?

I'm imaging something like `'variable': /\$\w+/`.

components.json
```diff
@@ -1346,6 +1346,11 @@
 			"require": "markup",
 			"owner": "Golmote"
 		},
+		"wolfram": {
+			"title": "Wolfram language",
+			"alias": ["wl", "mathematica"],
```
RunDevelopment, line 1351: Why did you remove the alias title? It was really good.

```suggestion
			"alias": ["wl", "mathematica"],
			"aliasTitles": {
				"mathematica": "Mathematica"
			},
```

components/prism-wolfram.js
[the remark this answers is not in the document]
msollami: Ok. I see aliases for names of languages but not for the keys of patterns? How would I set 'pattern' as an alias for 'regex'?

components.json
msollami: Ah sorry, adding them back.

components/prism-wolfram.js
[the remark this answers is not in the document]
msollami: Mathematica's front end doesn't do that, but it might be nice feature to add here anyway!

components.json
RunDevelopment: No need to apologize. It's not like I'm your boss lol

components/prism-wolfram.js
```diff
@@ -0,0 +1,24 @@
...
+	'context': {
+		pattern: /\w+`+\w*/,
+		alias: 'class-name'
+	},
+	'blank': {
+		pattern: /\b\w+\_{1,3}\b/,
```
RunDevelopment, line 14: `\w` includes the `_` character. This means that this regex is equivalent to `\b\w+_\b`. Is this still correct or is this a bug?

components/prism-wolfram.js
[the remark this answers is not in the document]
RunDevelopment: It would also reduce the file size again. Thanks to you cutting down the list of builtins, we are down to about 2kBs.

With all variables matched by a simple regex, like the one above, we could probably go down to 1kB.

components/prism-wolfram.js
msollami: Good catch. So to be a pattern must have at least one trailing `_` but can have up to three `_`, so
maybe `/\b[^_]\w*\_{1,3}\b/`?

RunDevelopment: > at least one trailing `_` but can have up to three `_`

In that case, you want `\b\w+_\b`.

msollami: Ok I just add that.

msollami:
@RunDevelopment Thanks for your review & help, this looks good to me now.

components/prism-wolfram.js
```diff
@@ -0,0 +1,29 @@
...
...sterWolframID|$RequesterWolframUUID|$ResourceSystemBase|$ResourceSystemPath|$RootDirectory|$ScheduledTask|$ScriptCommandLine|$ScriptInputString|$SearchLanguage|$SecuredAuthenticationKeyTokens|$ServiceCreditsAvailable|$Services|$SessionID|$SetParentLink|$SharedFunctions|$SharedVariables|$SoundDisplay|$SoundDisplayFunction|$SourceLink|$SSHAuthentication|$SubtitleDecoders|$SubtitleEncoders|$SummaryBoxDataSizeLimit|$SuppressInputFormHeads|$SynchronousEvaluation|$SyntaxHandler|$System|$SystemCharacterEncoding|$SystemCredentialStore|$SystemID|$SystemMemory|$SystemShell|$SystemTimeZone|$SystemWordLength|$TargetSystems|$TemplatePath|$TemporaryDirectory|$TemporaryPrefix|$TestFileName|$TextStyle|$TimedOut|$TimeUnit|$TimeZone|$TimeZoneEntity|$TopDirectory|$TraceOff|$TraceOn|$TracePattern|$TracePostAction|$TracePreAction|$UnitSystem|$Urgent|$UserAddOnsDirectory|$UserAgentLanguages|$UserAgentMachine|$UserAgentName|$UserAgentOperatingSystem|$UserAgentString|$UserAgentVersion|$UserBaseDirectory|$UserBasePacletsDirectory|$UserDocumentsDirectory|$Username|$UserName|$UserURLBase|$Version|$VersionNumber|$VideoDecoders|$VideoEncoders|$VoiceStyles|$WolframDocumentsDirectory|$WolframID|$WolframUUID)\b/,
```
RunDevelopment, line 8: Can't we remove all `$XXX` variables from this now that we have `global-variable`?

```suggestion
	'keyword': /\b(?:Abs|AbsArg|Accuracy|Block|Do|For|Function|If|Manipulate|Module|Nest|NestList|None|Return|Switch|Table|Which|While)\b/,
```

RunDevelopment:
Sorry, it was quite late for me yesterday. 

I agree that it's pretty much gtg now. We can merge this after the lining errors and my last comment have been addressed.

components/prism-wolfram.js
msollami: Ah yes, we don't need those now.

msollami:
> I agree that it's pretty much gtg now. We can merge this after the lining errors and my last comment have been addressed.

I added a new line at the end of file; changed tabs to spaces; and removed trailing whitespace, so the linter should be happy now.

RunDevelopment:
> Is there a way I can trigger the action to run the linter?

No, I have to trigger this manually every time.

components.json
```diff
@@ -1346,6 +1346,16 @@
 			"require": "markup",
 			"owner": "Golmote"
 		},
+		"wolfram": {
+			"title": "Wolfram language",
+			"alias": ["mathematica", "nb", "wl"],
+			"aliasTitles": {
+				"mathematica": "Mathematica",
+				"wl": "Wolfram Language",
```
RunDevelopment, line 1354: It's the same as the title of `wolfram`, so this can be omitted.

msollami:
> npm run lint

Good to know! Ok, it now passes the linter checks locally.

components.json
msollami: But the L is capitalized?

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @msollami!

PR #2924, opened 2021-05-31
ESLint: Added regexp plugin

RunDevelopment:
This PR adds the [`eslint-plugin-regexp` plugin](https://github.com/ota-meshi/eslint-plugin-regexp). This plugin statically analyses regex patterns and usages to find potential problems and suggest improvements. (Full disclaimer: I am a maintainer of the plugin.)

Right now, I only enabled a few rules to handle common things in code review (all warnings) and a few rules to detect (potential) errors.

---

@LeaVerou @Golmote @mAAdhaTTah 

There is another really good rule I am unsure about adding, the `regexp/no-dupe-disjunctions` rule. This rule actually finds more than a dozen bugs in our regexes (not fixed here) and even points out likely causes of exp backtracking directly and live in the IDE. 

However, the rule is slow. Right now, it increases the total linting time 5x (`npm run lint`), from 5s to 25s on my machine. This doesn't seem to be a problem for live code editing/viewing.

Should I also enable the `regexp/no-dupe-disjunctions` rule?


mAAdhaTTah:
This is kinda awesome, and I love how it makes every language slightly smaller. As for the other rule, maybe open in a separate PR and discuss there? Would be good to see the impact on the code.

RunDevelopment:
> I love how it makes every language slightly smaller

In that case, we might also want to add the [`regexp/order-in-character-class` rule](https://ota-meshi.github.io/eslint-plugin-regexp/rules/order-in-character-class.html) which sorts the elements of character classes. This will make it easier for gzip to compress our regexes.

> As for the other rule, maybe open in a separate PR and discuss there? Would be good to see the impact on the code.

Good idea. There are also a few other rules I would like to add. I'll make PRs for each.

PR #2927, opened 2021-06-03
add implementation for several new tokens in graphql lang

JaKXz:
Hey Prism team, I'd love to get your feedback on this set of new tokens for GraphQL and see if we can't get it to be worthy to merge upstream - I will likely be using this implementation for a relatively big project at work coming soon :) I've tried to add some more "integration" tests to cover the extra tokens that I need and hopefully this helps others along the way [and I think it fixes a couple bugs as well].

Here are the highlights:

- Adding these extra tokens for `language-graphql` enables folks to apply more visual language to GraphQL code samples. 
  - I have tried to add aliases to the more generic/older/previous implementation tokens as appropriate. This way you as a dev have choices to style more parts of the GraphQL code being presented.
  - The names of the tokens are roughly chosen based on the official graphql documentation / what seemed to make sense.
- Fixed [or at least added a test for] the handling of the `{ order(id: 4) { id, name, product } }` GQL query shorthand so it is not all rendered as plain with punctuation.
- Corrected the ordering of tokens and the `constant` vs `class-name` Regexes 

I'd be happy to live pair on this if this is something that's interesting to you! :)


joshgoebel:
> Added a lookahead feature to prism-core that allows you to confirm a token by what is on the right of it.

Can't this be done with regex alone?

JaKXz:
> > Added a lookahead feature to prism-core that allows you to confirm a token by what is on the right of it.
> 
> Can't this be done with regex alone?

@joshgoebel yes, but because of the way matches are used in prism-core [as far as I understand], I wasn't able to get it to work without changing the underlying logic in `matchPattern`. see: https://github.com/PrismJS/prism/blob/f3dd72c15fb1e58cdbf2f4a5c2c885c3a3e85057/components/prism-core.js#L967-L975

joshgoebel:
And it calls `pattern.exec` to run the regex again the text, not sure what am I missing...? What was the exact issue you were having with a simple regex look-ahead?

JaKXz:
> And it calls `pattern.exec` to run the regex again the text, not sure what am I missing...? What was the exact issue you were having with a simple regex look-ahead?

I needed access to the part of the match that didn't include the punctuation being used for checking what kind of token it is... which is kinda brittle. But for [these tokens](https://github.com/PrismJS/prism/pull/2927/files#diff-1782b75624f0928ff1fb90b3059c848afd7377c2ddb6e7213d657fb36e016973R76-R83) I needed the element without the punctuation, and that was always in `match[1]`.

components/prism-graphql.js
```diff
@@ -27,17 +37,50 @@ Prism.languages.graphql = {
...
+	'keyword': /\b(?:directive|enum|extend|fragment|implements|input|interface|on|repeatable|scalar|schema|subscription|type|union)\b/,
 	'operator': /[!=|&]|\.{3}/,
+	'property-query': {
+		pattern: /\w+(\s*\()/,
+		lookahead: true
+	},
+	'object': {
+		pattern: /\w+(\s*{)/,
+		lookahead: true
+	},
```
joshgoebel, line 83: Why do you need a capture group at all...?  Look-ahead doesn't have to expand the match.

```suggestion
	'object': {
		pattern: /\w+(?=\s*{)/
	},
```

components/prism-graphql.js
```diff
@@ -27,17 +37,44 @@ Prism.languages.graphql = {
...
+		alias: 'function'
+	},
+	'keyword-mutation': {
+		pattern: /\b(?:mutation)\b/,
+		alias: 'keyword'
+	},
+	'keyword-query': {
+		pattern: /\b(?:query)\b/,
+		alias: 'keyword'
+	},
```
RunDevelopment, line 73: I don't think this is necessary. We have [a plugin](https://prismjs.com/plugins/highlight-keywords/) just for this purpose.

components/prism-graphql.js
```diff
@@ -27,17 +37,44 @@ Prism.languages.graphql = {
...
+		greedy: true
 	},
 	'fragment': {
 		pattern: /(\bfragment\s+|\.{3}\s*(?!on\b))[a-zA-Z_]\w*/,
 		lookbehind: true,
 		alias: 'function'
 	},
-	'keyword': /\b(?:directive|enum|extend|fragment|implements|input|interface|mutation|on|query|repeatable|scalar|schema|subscription|type|union)\b/,
+	'property-mutation': {
+		pattern: /\b(mutation.*(?:[\n\r\u2028\u2029][\n\s]*)?)\b\w+/g,
```
RunDevelopment, line 52: 1. I commend you for acknowledging that `.` is equivalent to `[^\n\r\u2028\u2029]` but we can make this pattern a little simple using a line assertion `$`. This works because `/$/m` is equivalent to `/(?!.)/`.
2. `[\n\s]`: `\n` is included by `\s`.
3. Prism will automatically add the `g` flag if necessary. No need to explicitly add it.

```suggestion
		pattern: /\b(mutation.*$\s+)\b\w+/m,
```

components/prism-graphql.js
```diff
@@ -27,17 +37,44 @@ Prism.languages.graphql = {
 		pattern: /[a-z_]\w*(?=\s*(?:\((?:[^()"]|"(?:\\.|[^\\"\r\n])*")*\))?:)/i,
 		greedy: true
 	},
+	'constant': /\b(?!ID\b)[A-Z][A-Z_\d]*\b/,
 	'class-name': {
-		pattern: /(\b(?:enum|implements|interface|on|scalar|type|union)\s+|&\s*)[a-zA-Z_]\w*/,
-		lookbehind: true
+		pattern: /(\b(?:enum|implements|interface|on|scalar|type|union)\s+|&\s*|:\s*|\s*\[)[A-Z_]\w*/,
```
RunDevelopment, line 42: `\s*\[`: Is this in the wrong order or is this intentional?

Assuming that it is intentional: You can remove the leading `\s*`.

```suggestion
		pattern: /(\b(?:enum|implements|interface|on|scalar|type|union)\s+|&\s*|:\s*|\[)[A-Z_]\w*/,
```

components/prism-graphql.js
```diff
@@ -18,7 +18,17 @@ Prism.languages.graphql = {
 	},
 	'number': /(?:\B-|\b)\d+(?:\.\d+)?(?:e[+-]?\d+)?\b/i,
 	'boolean': /\b(?:true|false)\b/,
+	'variable-input': {
+		pattern: /(mutation.*)\$[\w_]+/i,
```
RunDevelopment, line 22: 1. Keywords are case-sensitive, so no `i` flag.
2. We need a word boundary before `mutation`.
3. `[\w_]`: `_` is included by `\w`

```suggestion
		pattern: /(\bmutation.*)\$\w+/,
```

components/prism-graphql.js
```diff
@@ -18,7 +18,17 @@ Prism.languages.graphql = {
...
+		alias: 'variable',
+		lookbehind: true
+	},
 	'variable': /\$[a-z_]\w*/i,
+	'atom-input': {
+		pattern: /(mutation.*:\s*)\b[A-Z_]\w*\b!?/,
```
RunDevelopment, line 28:
```suggestion
		pattern: /(\bmutation.*:\s*)\b[A-Z_]\w*\b!?/,
```

components/prism-graphql.js
```diff
@@ -27,17 +37,44 @@ Prism.languages.graphql = {
 		pattern: /[a-z_]\w*(?=\s*(?:\((?:[^()"]|"(?:\\.|[^\\"\r\n])*")*\))?:)/i,
 		greedy: true
 	},
+	'constant': /\b(?!ID\b)[A-Z][A-Z_\d]*\b/,
 	'class-name': {
-		pattern: /(\b(?:enum|implements|interface|on|scalar|type|union)\s+|&\s*)[a-zA-Z_]\w*/,
-		lookbehind: true
+		pattern: /(\b(?:enum|implements|interface|on|scalar|type|union)\s+|&\s*|:\s*|\s*\[)[A-Z_]\w*/,
+		lookbehind: true,
+		greedy: true
```
RunDevelopment, line 44: I'm pretty sure that this `greedy: true` is unnecessary.

RunDevelopment (changes requested):
Thank you for the PR @JaKXz! Very good work! 

Except for the new `keyword-*` token, I think the new tokens are a good idea. I only excluded the `keyword-*` tokens because we already have [a plugin for this](https://prismjs.com/plugins/highlight-keywords/).

Your general approach seems quite good but there is one thing that scares me: `mutation.*`. The `.*` seems brittle. Could please explain the `variable-input`, `atom-input`, and `property-mutation` tokens in more detail?

Regex-wise, I found a few things to improve/fix.

Regarding the `--language` option, this should be its own PR. Do you want to make one or shall I?

---

Also, thank you @joshgoebel for handling the lookahead feature! I would have said the same.

tests/pattern-tests.js
```diff
@@ -38,7 +39,7 @@ for (const languageIdentifier in testSuite) {
 
 
 for (const lang in languages) {
-	if (lang === 'meta') {
+	if (lang === 'meta' || (!!argv.language && lang !== argv.language)) {
```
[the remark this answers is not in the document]
JaKXz: yep! Here's a separate PR: https://github.com/PrismJS/prism/pull/2929

components/prism-graphql.js
JaKXz: I see, that's pretty cool. How do I say that the `graphql` language requires this plugin? Or is that up to the end user?

I ask because I would like to keep the integration tests I have passing, but I can concede it's reasonable that the plugin is separately tested so if someone just imports that plugin they can count on it working.

components/prism-graphql.js
JaKXz: I can't take credit for knowing the equivalence... the patterns tests pointed this out to me as a fix for what I had before... 😂 fixed now, thanks!

components/prism-graphql.js
JaKXz: yep, thank you!

components/prism-graphql.js
JaKXz: Yes, it was intentional -- thanks for the suggestion, it works!

components/prism-graphql.js
RunDevelopment: > Or is that up to the end user?

It's supposed to be on the end user.

Technically you _could_ require the plugin but please don't. The plugin will add these keyword aliases for each and every language. GraphQL would essentially force this behavior onto other languages.

> if someone just imports that plugin they can count on it working.

Yup, that's the idea. The user just has to load the file and it'll work.

> I would like to keep the integration tests I have passing

With "integration tests" you mean the `.test` files?

If so, sorry but that's won't work because the test files contain the token names. It's really easy to update test files tho (`npm run test:languages -- --update`), so this shouldn't be a problem.

JaKXz: > The plugin will add these keyword aliases for each and every language. 

Right, I would love to isolate it to graphql if possible:

```
	Prism.hooks.add('wrap', function wrapGraphqlKeywords(env) {
		if (env.type === 'keyword' && env.language === 'graphql') {
			env.classes.push('keyword-' + env.content);
		}
	});
```

but it didn't work when I put it in the `components/prism-graphql.js` file, even in an IIFE. I'm missing some understanding about how plugins work in the system, could you explain that to me? :)

JaKXz: Or, is it just not worth it? Probably that eh? 😅

JaKXz:
> Could please explain the variable-input, atom-input, and property-mutation tokens in more detail?

Good question @RunDevelopment -- these tokens are for visually showing the relationships between the inputs to a GraphQL mutation specifically. My team wants to have a specific theme that visually relates the `variable-input` and the `atom-input`, and we have other visualizations that will match the scheme we use here.

`property-mutation` is just to say that this mutation is one of the properties available in the global `Mutation` object of your GraphQL server. So having access to style it is just to denote that.

They are specific to the mutation keyword, hence the `mutation.*` but I am definitely open to better solutions. Here's a sneak peek of what we're going for:
image

components/prism-graphql.js
RunDevelopment: > Just to confirm, to add the plugin, I could just require it like:

In NodeJS? Yes.

> but it didn't work

Oh, it does work. `env.classes` is essentially the list of aliases, so you are adding a special alias for each keyword. However, our test file format doesn't include aliases, so they won't show up.

That is one reason why it doesn't work. The other is that the `wrap` hook is called after tokenization. So any and all changes a `wrap` hook does do not affect our test files which only store tokens.

You have to update the tests either way...

JaKXz: Ah. Thanks for the explanation!

RunDevelopment:
> Here's a sneak peek (WIP) of what we're going for:

That does look really good but I don't think that this is possible with regexes alone. Especially the tracking of the `input-variable`.

That being said, Prism has hooks for this reason! We can use the `after-tokenize` hook to modify the GraphQL token stream however we wish.

My idea is to remove the specialized `variable-input` and tokenized with the "normal" `variable` token instead. We then have a `after-tokenize` hook that goes through the token stream to detect the list of input variables and then gives a special alias to all of them.

The same idea can also be applied to `atom-input` and `property-mutation`. This would get rid of all these scary `.*`s and we won't be constraint by what regexes can and can't do.

Do you want me to write a little `after-tokenize` hook?

JaKXz:
> That does look really good

Thanks! 🙌🏽 

> Do you want me to write a little after-tokenize hook?

Go for it, yeah! Thanks for helping me get up to speed w/ all the lifecycle methods Prism has :)

JaKXz:
Sorry for the weird unverified commit, but, I added another "integration" test for ya :)

components/prism-graphql.js
JaKXz: I came up with a quick and dirty way to write some "e2e tests", would love to know what you think :) https://github.com/JaKXz/prism/pull/1/files

JaKXz: ^with those tests in I've resolved this 👍🏽

JaKXz:
@RunDevelopment do you think this is mergeable without the `after-tokenize` hook?

RunDevelopment:
> do you think this is mergeable without the `after-tokenize` hook?

Sorry, I hadn't had the time yet. I'll make you a PR with the hook tomorrow, so we can get rid of the `.*s`.

Apart from that, I think this PR is good to go.

JaKXz:
Awesome! No rush at all really... I actually got some more requirements today so maybe we'll have some merging work to do.

components/prism-graphql.js
```diff
@@ -18,7 +18,16 @@ Prism.languages.graphql = {
...
 	'number': /(?:\B-|\b)\d+(?:\.\d+)?(?:e[+-]?\d+)?\b/i,
 	'boolean': /\b(?:true|false)\b/,
+	'variable-input': {
+		pattern: /(\bmutation.*)\$\w+/,
+		alias: 'variable',
+		lookbehind: true
+	},
 	'variable': /\$[a-z_]\w*/i,
+	'atom-input': {
+		pattern: /[A-Z]\w*Input(?=!?.*$)/m,
```
JaKXz, line 28: Managed to simplify this pattern a bit to get rid of the `mutation.*` [and fix a test that slipped by because I missed the diff until I looked at the snapshot 😅 ] but yeah I'm not sure if I can refactor the `variable-input` and `property-mutation` tokens like this. 

If someone can piece together a better pattern from those tests I would be very impressed.

JaKXz: @RunDevelopment I found in my testing that we do need this one to get the input type definitions correct... I'll add it back with a test.

RunDevelopment: Could you share the example that was wrong?

JaKXz: Here's the failing example:

```
input OrderInput {
	id: ID!
	customAttributes: [AttributeInput!]
	customer: Customer
	localizationExtensions: [LocalizationExtensionInput!]
	metafields: [MetafieldInput!]
	note: String!
	shippingAddress: MailingAddressInput
	tags: [String!]
}
```

and I added a failing test for it in my [`tests/e2e-example` branch](https://github.com/JaKXz/prism/pull/1/files#diff-bd57f122459de3b3ce8bc9288301d18ae02ce2ceac7e4937a4039c3cd7f8bb24R26) which passes if I put this pattern back in.

JaKXz: This is fixed in my branch https://github.com/JaKXz/prism/pull/1 -- I won't merge that while we're not on board with the move to snapshot e2e testing :) thanks for the discussion and all the help though!

PR #2934, opened 2021-06-05
Tests: Automatically normalize line ends

RunDevelopment:
Also discussed [here](https://github.com/PrismJS/prism/pull/2932#discussion_r645623467), The parser for `.test` files will now automatically convert all file contents to CRLF line ends. This ensures that all languages are tested with the more difficult line ends.

This lead to the discovery (and fix) of a bug in False.

This PR contains 4 things:

- The parser now converting line ends.
- `.test` files no longer being declared as `binary` in `.gitattributes`.
- The fix for False.
- The changes for all changed test files. Some test files used LF line ends which were reflected in the JSON.

This PR **does not** convert the line ends of existing `.test` files to LF. This means that we now have a bunch of CRLF files in the repo. Should I convert them?

@mAAdhaTTah @joshgoebel


joshgoebel:
> This means that we now have a bunch of CRLF files in the repo. Should I convert them?

I'm personally in favor of unix line endings - whichever one that is - and being consistent seems appropriate.

RunDevelopment:
Advice for reviewers: Don't open the "Files" tab on Github, your tab will freeze.

RunDevelopment:
Ok, no. I reverted that. I did a small test and these EOL changes will conflict will literally every PR that changes a test file.

PR #2943, opened 2021-06-13
ESLint: Added `regexp/sort-alternatives` rule

RunDevelopment:
This adds a rule that enforces lists of words to be sorted.

Note that this rule is extremely conservative. It will only reorder alternatives if it is 100% safe to do so.

This rule is pretty nice for Prism because sorted lists are easier to read and maintain. As an added bonus, they are also easier to compress for Gzip. I often ask contributors to sort word lists and now I don't have to do this anymore, which is nice.


JaKXz:
If this is `eslint --fix`able that would be amazing

RunDevelopment:
[It is](https://ota-meshi.github.io/eslint-plugin-regexp/rules/sort-alternatives.html) ;)

The only reason I haven't merged this yet is that I'm currently working on making this rule smarter.

PR #2945, opened 2021-06-13
JS Templates: Added SQL templates

RunDevelopment:
This resolves the issue brought up by [this comment](https://github.com/PrismJS/prism/issues/1930#issuecomment-860224348).


JaKXz (approved):
LGTM :)

karlhorky:
Thanks! Will this be a new minor (`1.24.0`)?

RunDevelopment:
Yes.

karlhorky:
Great, looks like it's working with `prism@1.24.0` 🎉

```js
const users = await sql`
  SELECT * FROM users WHERE id = ${id}
`;
```

Screen Shot 2021-07-04 at 17 57 21

In order to achieve this, I had to load the following languages (in this order):

1. SQL
2. JS Templates
3. JavaScript

PR #2947, opened 2021-06-13
Made Match Braces and Custom Class compatible

### LINKED ISSUE #2933: Plugin match-braces does not work with a custom prefix from custom-class plugin
**Information:**
- Prism version: latest from the download page
- Plugins: match-braces, custom-class
- Environment: Browser

**Description**
When using the plugin `match-braces` alongside `custom-class` plugin it works fine, unless you declare a custom prefix.

e.g.
```js
Prism.plugins.customClass.prefix('pr-');
```

Most likely it's because [this line](https://github.com/PrismJS/prism/blob/master/plugins/match-braces/prism-match-braces.js#L113) is using the wrong selectors.

So, instead of:
```js
var punctuation = Array.prototype.slice.call(code.querySelectorAll('span.token.punctuation'));
```

It should be something like:
```js
var punctuation = Array.prototype.slice.call(code.querySelectorAll(`span.${prefix}token.${prefix}punctuation`));
```
I'm not sure how to get the prefix though, but I believe it's stored somewhere.

**Example**
[JSFiddle](https://jsfiddle.net/wj20fuqy/1/)

RunDevelopment:
This makes the two plugins compatible with each other.

CC will now expose an `apply` method which applies all transformations to one class name. MB will detect whether CC is loaded and use its transformation function if loaded.

Since MB will dynamically detect CC, the dependency doesn't have to be declared in `components.json`. The two plugins can be loaded in any order.

---

This resolves #2933.


plugins/custom-class/prism-custom-class.js
```diff
@@ -62,6 +62,17 @@
...
+		/**
+		 * Applies the current mapping and prefix to the given class name.
+		 *
+		 * @param {string} className A single class name.
+		 * @param {string} language The language of the code that contains this class name.
+		 *
+		 * If the language is unknown, pass `"none"`.
+		 */
+		apply: function (className, language) {
+			return prefixString + (mapper ? mapper(className, language) : className);
```
mAAdhaTTah, line 75: Can we reuse `apply` within the plugin to reduce duplication?

RunDevelopment: Done

PR #2950, opened 2021-06-13
Liquid: Added Markup support, missing tokens, and other improvements

RunDevelopment:
This reimplements Liquid using Markup templating and adds missing tokens, keywords, and operators.

---

closes #2919.


components/prism-liquid.js
```diff
@@ -1,12 +1,35 @@
...
+		pattern: /^\{(?:\{\{|[%\{])-?|-?(?:\}\}|[%\}])\}$/,
+		alias: 'punctuation'
+	},
+	'string': {
+		pattern: /"[^"]*"|'[^']*'/,
+		greedy: true
+	},
+	'keyword': /\b(?:as|assign|break|capture|case|comment|contains|continue|decrement|echo|else|elsif|endcapture|endcase|endcomment|endfor|endif|endraw|endtablerow|endunless|for|if|in|include|increment|limit|liquid|offset|range|raw|render|reversed|tablerow|unless|when|with)\b/,
+	'function': {
+		pattern: /(^|[\s;|&])(?:append|prepend|capitalize|cycle|cols|abs|at_least|at_most|ceil|compact|concat|date|default|divided_by|downcase|escape|escape_once|first|floor|join|last|lstrip|map|minus|modulo|newline_to_br|plus|remove|remove_first|replace|replace_first|reverse|round|rstrip|size|slice|sort|sort_natural|split|strip|strip_html|strip_newlines|times|truncate|truncatewords|uniq|upcase|url_decode|url_encode|paginate)(?=$|[\s;:|&])/,
```
JaKXz, line 12: I would love to be able to get all the function names from the `liquid` gem somehow..

JaKXz (approved):
This is fantastic, thank you Michael!

components/prism-liquid.js
RunDevelopment: Is there a way to detect functions by syntax?

RunDevelopment: I figured it out.

RunDevelopment:
Ok, this is now a complete overhaul of Liquid. I added every feature I could find and derived the regexes from the current Liquid parser.

It should cover everything and everything _should_ be correct.

@JaKXz Could you please see where I missed something/whether something is highlighted incorrectly?

components/prism-liquid.js
```diff
@@ -1,12 +1,61 @@
...
+	'number': /\b\d+(?:\.\d+)?\b/,
+	'operator': /[!=]=|<>|[<>]=?|[|?:=-]|\b(?:and|or|contains(?=\s))\b/,
+	'punctuation': /[.,\[\]()]/
 };
+
+Prism.hooks.add('before-tokenize', function (env) {
+	var liquidPattern = /\{%\s*comment\s*%\}[\s\S]*?\{%\s*endcomment\s*%\}|\{(?:%[\s\S]*?%|\{\{[\s\S]*?\}\}|\{[\s\S]*?\})\}/g;
+	var insideRaw = false;
+
+	Prism.languages['markup-templating'].buildPlaceholders(env, 'liquid', liquidPattern, function (match) {
```
JaKXz, line 42: If I were to wish for something it would be that the error message would tell you that you need to have `prism-markup-templating` loaded, something like this

```suggestion
	if (!Prism.languages['markup-templating']) {
		throw new Error('You need to load prismjs/components/prism-markup-templating to use the Liquid lang'); 
	}

	Prism.languages['markup-templating'].buildPlaceholders(env, 'liquid', liquidPattern, function (match) {
```

JaKXz (commented):
Looks great in my testing but I have one suggestion

components/prism-liquid.js
RunDevelopment: Languages don't load their own dependencies. It's not their responsibility to ensure the correct load order. If Prism components aren't loaded correctly, they are allowed to fail for any reason.

I get that good error messages are nice in these situations but we have a hundred languages with dependencies like this. If all of them checked their dependencies and threw an error like this, then Prism would carry a lot of dead weight for a lot of users (namely: all users that correctly use Prism, so hopefully all users).

We want Prism to be as light-weight as possible and this just doesn't make the cut.

JaKXz: hm. Interesting. Is the assumption that everyone is loading _every_ language/component?

RunDevelopment: No, only the ones wanted + their dependencies. Why?

JaKXz: > No, only the ones wanted + their dependencies. Why?

In that case, I don't think this statement is true:

> If all of them checked their dependencies and threw an error like this, then Prism would carry a lot of dead weight for a lot of users

It would certainly add a few bytes here and there. However, I think the improved developer experience is well worth the trade.

RunDevelopment: > However, I think the improved developer experience is well worth the trade.

I disagree. Instead of telling you "you should have loaded these files", the components should just load their dependencies themselves. This is how it will be done in v2.0. So we won't have to worry about this for much longer.

Until then, I would like to carry on as usual. I mean, it seems to have worked so far ;)

JaKXz: Haha that’s fair. I didn’t want to suggest figuring out dependency injection but glad that’s in the works!

JaKXz (approved):

PR #2951, opened 2021-06-15
ESLint: Enabled `no-dupe-disjunctions` rule

RunDevelopment:
This enables the `regexp/no-dupe-disjunctions` rule. This is the performance-heavy rule I talked about in #2924.

> However, the rule is slow. Right now, it increases the total linting time 5x (`npm run lint`), from 5s to 25s on my machine. This doesn't seem to be a problem for live code editing/viewing.

The only commit of this PR will be just enabling the rule. This will allow us to see the problems the rule finds. I will add the fixes for these reports in another PR. This allows merging the fixes even if we should decide against this rule.

---

@LeaVerou @Golmote @mAAdhaTTah


LeaVerou:
I don't understand, if it makes linting this slow, how is it not a problem for live linting? Do editors not run it?

RunDevelopment:
Editors do run it. It's not a problem for editors because they only run ESLint on open files. The rule is slow as in: "it takes about 10ms per regex". This time is noticeable when linting all of our 2500 regexes but not when it's only the 10~20 regexes per language definition.

LeaVerou:
Ah I see. Thanks for the explanation! I'm fine with this.

RunDevelopment:
In that case, I'll merge the fixes and then this rule.

PR #2953, opened 2021-06-15
JS: Added support for import assertions

RunDevelopment:
This adds support for Stage 3 proposals [Import Assertions](https://github.com/tc39/proposal-import-assertions) and [JSON modules](https://github.com/tc39/proposal-json-modules).


JaKXz (approved):
:tada:

PR #2959, opened 2021-06-21
ESLint: Use cache

RunDevelopment:
ESLint has a caching functionality. Might as well use it since linting got slower after #2951.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ hide-*.js
 node_modules
 .idea/
 .DS_Store
+.eslintcache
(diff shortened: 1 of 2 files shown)


mAAdhaTTah:
Should we save + restore the cache in CI?

RunDevelopment:
Nah, I don't want people to suddenly have to commit the ESLint cache as well. 

Plus, the cache might be outdated or otherwise invalid (the cache assumes that ESLint rules + config are constant) so it's actually good that the CI doesn't use the cache.

mAAdhaTTah:
They wouldn't have to commit the cache, same way they don't have to commit the npm cache, but if the eslint cache is assuming things about the rules, then yeah, let's leave it.

mAAdhaTTah (approved):

PR #2964, opened 2021-06-22
fix definition-query and definition-mutation tokens

JaKXz:


tests/languages/jsx!+graphql+js-templates/graphql_inclusion.test
```diff
@@ -0,0 +1,223 @@
...
+			["property-query", "nodes"],
+			["punctuation", "("],
+			["attr-name", "ids"],
+			["punctuation", ":"],
+			["variable", "$ids"],
+			["punctuation", ")"],
+			["punctuation", "{"],
+
+			["operator", "..."],
+			["keyword", "on"],
```
JaKXz, line 150: @RunDevelopment I noticed this would be tokenized to `definition-mutation` while I was working on the template string issue (#2957)

RunDevelopment (approved):

RunDevelopment:
Good catch. Thank you for fixing this @JaKXz!

PR #2970, opened 2021-06-26
C#: Added context check for `from` keyword

### LINKED ISSUE #2968: Keyword "from" is not a valid keyword in some languages
**Information:**
- Prism version: latest from the CDN.
- Plugins: none
- Environment: Browser

**Description**
Prism incorrectly highlights the "from" word as a keyword in some languages which are not supposed to be highlighted. An example of this is Javascript and C#.

Although in C# is kinda tricky since it is a keyword when using a from-in clause:
```cs
func(from: "America", to: "Asia"); // Prism incorrectly highlights "from" as a keyword

from element in list; // no issues here
```

**Example**
[JSFiddle](https://jsfiddle.net/30shef45/1/)

RunDevelopment:
This fixes #2968.

image


mAAdhaTTah (approved):

RunDevelopment:
Thanks for the review @mAAdhaTTah. I'll merge this after releasing a patch for #2973.

PR #2975, opened 2021-07-01
check existence of document: static site generator breaking on nested code blocks

songololo:
I don't understand exactly what would have changed, but recently the `gridsome` static site generator started breaking when processing nested syntax code blocks. This presumably occurred with the latest release.

It can be fixed by adding a check for whether `document` exists in the current context.

See:
https://github.com/PrismJS/prism/issues/2976
https://github.com/gridsome/gridsome/issues/1524

### DIFF
--- a/components/prism-markdown.js
+++ b/components/prism-markdown.js
@@ -351,11 +351,13 @@
 			}
 		} else {
 			// get the textContent of the given env HTML
-			var tempContainer = document.createElement('div');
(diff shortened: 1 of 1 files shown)


RunDevelopment:
Thank you for the PR @songololo!

However, this solution doesn't work for servers. I'm currently working on a solution and will make PR shortly.

Issue #2976, opened 2021-07-01
static site generator breaking on nested code blocks


songololo:
**Information:**
v1.23 and v1.24 both caused this behaviour.

**Description**
I don't understand exactly what would have changed, but recently the `gridsome` static site generator started breaking when processing nested syntax code blocks. This presumably occurred with the latest release.

It can be fixed by adding a check for whether `document` exists in the current context.
https://github.com/PrismJS/prism/pull/2975

Additional description and examples in the related `gridsome` repo:
https://github.com/gridsome/gridsome/issues/1524

RunDevelopment:
Duplicate of #2973.

Issue #2974, opened 2021-06-29
Copy to clipboard button has double shadow


Aetf:
**Information:**
- Prism version: 1.24.0
- Plugins: toolbar, copy-to-clipboard
- Environment: browser

**Description**
The copy-to-clipboard toolbar button has double shadows.

The toolbar  css styles both `button` and `span`, not the direct child: https://github.com/PrismJS/prism/blob/master/plugins/toolbar/prism-toolbar.css#L46

**Example**

Screenshot of my website

![image](https://user-images.githubusercontent.com/1519759/123870995-1d051880-d901-11eb-80d9-33f02b634176.png)

Also a minimal example on jsfiddle

https://jsfiddle.net/vo6a5rxw/5/

PR #2977, opened 2021-07-01
Markdown: Fixed markdown not working in NodeJS

### LINKED ISSUE #2973: Unable to run in node when using markdown
See https://github.com/PrismJS/prism/commit/2b355c98720b7c4f697200d666ddfdcfc3350dbe#diff-12f505384949e8ade50a6337e923bbe2ac037b951fb003617889a165758b4d3c

This commit and other commit include `document`, then it makes prism unable to run under node environment.

This is absolutly a breaking change in prismjs, and should be consider as a bug.

RunDevelopment:
Markdown crashed on Node. The code-block-highlighting relyed on browser DOM to get the text content of an HTML string (see `wrap` hook function).

The fix employed by this PR is to replace the DOM-based approach with a custom function: `textcontent(html: string)`. The function is implemented using Prism's markup tag regex and some custom logic to unescape HTML entities I stole (not literally) from lowdash. It works pretty well on Node and in browsers (notably IE11).

I also contemplated detecting whether we have DOM access and using the DOM approach when available. I decided against this because:

1. Consistent behavior. Markdown working just fine in the browser but not on NodeJS is exactly what I tried to fix.
2. Security. It's unexpected that Markdown inserts anything into the DOM. This might open the door for XSS attacks. 

---

@mAAdhaTTah @LeaVerou @Golmote

This fix is urgent. Please review ASAP.

I want to release the fix right after it got merged.

---

This fixes #2973.


LeaVerou:
This is not a regression, right? If not, why is it urgent?

RunDevelopment:
It is a regression. Markdown doesn't work on Node anymore. See #2973 and #2976.

LeaVerou:
What introduced this regression? `git blame` shows that these lines have existed for at least 5 months?

RunDevelopment:
The linked commit did introduce the regression. We just haven't had a release for 6 months, so it went unnoticed.

LeaVerou:
I see. I'm not a huge fan of stripping HTML with regexes (it worries me about potential security issues) but I suppose our best bet is to merge this in and remove the regression as soon as possible, while we figure out a better long-term fix.

RunDevelopment:
> I'm not a huge fan of stripping HTML with regexes 

Neither am I.

I'll open a separate issue for a sound way to get the text content of an HTML string.

RunDevelopment:
@LeaVerou, can I take your comment as approval? Just want to make sure :)

I plan on releasing a patch right after merging this.

RunDevelopment:
Actually, I want the patch to include a fix for #2974 as well.

LeaVerou (approved):

meteorlxy:
@RunDevelopment If #2974 could not be merged soon, would you mind to release a patch for this fix first?

RunDevelopment:
@meteorlxy v1.24.1 is now available. It includes both fixes.

PR #2978, opened 2021-07-02
components: add hoon programming language

matildepark:
We're using this downstream on https://urbit.org (in a site using next.js and remark-prism), so passing this language definition upstream. Derived from our [tmLanguage definition](https://github.com/urbit/urbit/blob/master/extras/hoonSublime/hoon.tmLanguage) and the documentation.


components.json
```diff
@@ -529,6 +529,10 @@
 			],
 			"owner": "danielgtaylor"
 		},
+    "hoon": {
```
RunDevelopment, line 532: Please use tabs for indentation.

components/prism-hoon.js
```diff
@@ -0,0 +1,42 @@
...
+  ],
+  "class-name": [
+    {
+      pattern: /@([a-z])?([A-Za-z0-9-]*[A-Za-z0-9])?/,
+      lookahead: true,
+      greedy: true
+    },
+    {
+      pattern: /\*/
+    }
```
RunDevelopment, line 25: Instead of `{ pattern: /regex/ }` you can just use `/regex/`. Please use the simple regex notation wherever possible.

RunDevelopment (changes requested):
Thank you for the PR @matildepark!

Please fix the linting error and make the tests pass. 

One reason that the tests don't pass is that you used capturing groups extensively. Except in the case of lookbehinds, you can replace all capturing groups with non-capturing groups. This should fix most of the problems our tests find.

Also, please add language tests for Hoon. We regularly improve our matching algorithm and tooling. Language tests are essential to make sure that nothing breaks. Please add tests that cover every regex in Hoon. 
Step 6 in [this guide](https://prismjs.com/extending.html#creating-a-new-language-definition) explains how to add tests. For a more detailed explanation, see [the test suite documentation](https://prismjs.com/test-suite.html#writing-tests).

If you run into problems or have any questions, feel free to ask anytime. I'm here to help.

matildepark:
Should resolve, I think — @RunDevelopment

components/prism-hoon.js
```diff
@@ -0,0 +1,30 @@
+Prism.languages.hoon = {
+	constant: /(?:%\.n)|(?:%\.y)|(?:%[\w-]+\w?)/,
+	comment: {
```
RunDevelopment, line 3: Please use quoted keys for all token names. So `constant` and `comment` have to be changed.

```suggestion
	'comment': {
```

components/prism-hoon.js
```diff
@@ -0,0 +1,30 @@
+Prism.languages.hoon = {
+	constant: /(?:%\.n)|(?:%\.y)|(?:%[\w-]+\w?)/,
```
RunDevelopment, line 2: This can be simplified.
```suggestion
	constant: /%(?:\.[ny])|[\w-]+)/,
```

components/prism-hoon.js
```diff
@@ -0,0 +1,30 @@
...
+		pattern: /::.*/,
+		lookahead: true,
+		greedy: true
+	},
+	'function-name': /(?:\+[-+] {2})?(?:[a-z]([a-z0-9-]*[a-z0-9])?)/,
+	'class-name': [
+		{
+			pattern: /@(?:[a-z])?(?:[A-Za-z0-9-]*[A-Za-z0-9])?/,
+			lookahead: true,
+			greedy: true
```
RunDevelopment, line 13: This doesn't need to be greedy.

components/prism-hoon.js
```diff
@@ -0,0 +1,30 @@
...
+	constant: /(?:%\.n)|(?:%\.y)|(?:%[\w-]+\w?)/,
+	comment: {
+		pattern: /::.*/,
+		lookahead: true,
+		greedy: true
+	},
+	'function-name': /(?:\+[-+] {2})?(?:[a-z]([a-z0-9-]*[a-z0-9])?)/,
+	'class-name': [
+		{
+			pattern: /@(?:[a-z])?(?:[A-Za-z0-9-]*[A-Za-z0-9])?/,
```
RunDevelopment, line 11: This can be simplified.

Also, I don't think it's right that just `@` is a class name or is it?

```suggestion
			pattern: /@(?:[A-Za-z0-9-]*[A-Za-z0-9])?/,
```

components/prism-hoon.js
```diff
@@ -0,0 +1,30 @@
+Prism.languages.hoon = {
+	constant: /(?:%\.n)|(?:%\.y)|(?:%[\w-]+\w?)/,
+	comment: {
+		pattern: /::.*/,
+		lookahead: true,
+		greedy: true
+	},
+	'function-name': /(?:\+[-+] {2})?(?:[a-z]([a-z0-9-]*[a-z0-9])?)/,
```
RunDevelopment, line 8: 1. `function-name` is a non-standard name not supported by any themes. Please use `function` instead.
2. Shouldn't the `(?:\+[-+] {2})?` be a lookbehind?

components/prism-hoon.js
```diff
@@ -0,0 +1,30 @@
...
+			pattern: /"(?:[^"])*"/,
+			lookahead: true,
+			greedy: true
+		},
+		{
+			pattern: /'(?:[^'])*'/,
+			lookahead: true,
+			greedy: true
+		}
+	],
```
RunDevelopment, line 28: It's more performant to combine the two.

```suggestion
	'string': {
		pattern: /"[^"]*"|'[^']*'/,
		greedy: true
	},
```

RunDevelopment (changes requested):
I have a few minor suggestions and I think I found a bug.

components/prism-hoon.js
matildepark: Just `@` is an atom, the most basic data structure in Nock, and is therefore valid to reference along any other aura (or typed atom). (As for why class name — placement of Urbit-specific terms into syntax highligher-adjacent terms I believe I took from our vim definition, but I can't recall at the moment.)

components/prism-hoon.js
matildepark: I think I used Rust's definition as reference there. I can change it.

Wouldn't a lookbehind be looking for `  ++`? You need two spaces between any rune any its arguments, so we want `++  `.

matildepark: The inline code on GH shortened the starting two spaces in that first inline code block.

matildepark:
I believe everything addressed again; it's failing a test on identifiers, which I'm a little confused about; you can't name something `foo_123` in hoon; it's generally kabob case, and `_` is used in runic constructions (or our 'keywords'). Should Hoon be excluded from that? Also added the same example of the cipher, as I think it includes most of the features you'd highlight in one example.

matildepark:
Oh, right, I also reverted the change in  https://github.com/PrismJS/prism/pull/2978/commits/3982e750fb12d0d2eab5339ec54d502f71bdef6c because it was a broken definition.

RunDevelopment:
> Oh, right, I also reverted the change in [3982e75](https://github.com/PrismJS/prism/commit/3982e750fb12d0d2eab5339ec54d502f71bdef6c) because it was a broken definition.

Indeed. I'll make a new suggestion.

components/prism-hoon.js
```diff
@@ -0,0 +1,21 @@
+Prism.languages.hoon = {
+	'constant': /(?:%\.n)|(?:%\.y)|(?:%[\w-]+\w?)/,
```
RunDevelopment, line 2:
```suggestion
	'constant': /%(?:\.[ny]|[\w-]+)/,
```

components/prism-hoon.js
RunDevelopment: I meant the lookbehind like this:

```
'function': {
  pattern: /(\+[-+] {2})?(?:[a-z]([a-z0-9-]*[a-z0-9])?)/,
  lookbehind: true
}
```

The `\+[-+] {2}` is not part of the function name itself, or is it?

RunDevelopment:
> it's failing a test on identifiers, which I'm a little confused about; you can't name something `foo_123` in hoon; it's generally kabob case, and `_` is used in runic constructions (or our 'keywords'). Should Hoon be excluded from that? 

Yeah, just exclude it [here](https://github.com/PrismJS/prism/blob/151121cd7d436feb985396b9f89208271ab2c6d9/tests/identifier-test.js#L44). I think you only need to disable `word` and `template`.

components/prism-hoon.js
matildepark: You can define an arm by writing `++  name` in a core; and then refer to it (if it's in the environment) as `name`. Now maybe you consider `++` as a rune (or "keyword" here), but thus far in practice we group the arm name alongside its runic constructor in our pre-existing syntax highlighters, so it's all grouped here.

components/prism-hoon.js
```diff
@@ -0,0 +1,21 @@
+Prism.languages.hoon = {
+	'constant': /%(?:\.[ny]|[\w-]+)/,
+	'comment': {
+		pattern: /::.*/,
+		lookahead: true,
+		greedy: true
+	},
+	'function': /(?:\+[-+] {2})?(?:[a-z]([a-z0-9-]*[a-z0-9])?)/,
```
RunDevelopment, line 8: You missed a capturing group :)

```suggestion
	'function': /(?:\+[-+] {2})?(?:[a-z](?:[a-z0-9-]*[a-z0-9])?)/,
```

components/prism-hoon.js
RunDevelopment: Ok. In that case, a lookbehind isn't necessary.

matildepark:
Oh, lol, it's because I changed it to function-name. OK, one sec ...

RunDevelopment:
You can let the test runner update all tests with `npm run test:languages -- --update`.

RunDevelopment:
Nice, tests pass. Just an `npm run build` and this is ready.

matildepark:
Built!

RunDevelopment (approved):

RunDevelopment:
Thank you @matildepark!

It's ready as is but we are currently fixing 2 bugs that came up after the last release. I'll merge this PR right after the patch for these 2 bugs is released, maybe in 1 or 2 days.

matildepark:
No worries, thanks for the patience.

components/prism-hoon.js
```diff
@@ -0,0 +1,21 @@
+Prism.languages.hoon = {
+	'constant': /%(?:\.[ny]|[\w-]+)/,
+	'comment': {
+		pattern: /::.*/,
+		lookahead: true,
```
RunDevelopment, line 5: Sorry, I missed this one @matildepark.

We don't have something called `lookahead`, we only have `lookbehind`. Please remove the `lookahead: true` here and in `class-name`. It doesn't do anything.

RunDevelopment (changes requested):

RunDevelopment:
Thank you for contributing @matildepark!

PR #2990, opened 2021-07-09
.ino is main file extension for arduino files

Outreach-ChemEng-UofU:


joshgoebel:
Interesting, does Prism do aliases like this?

Outreach-ChemEng-UofU:
> Interesting, does Prism do aliases like this?

I looked at the python language and that's how they handled py and python

joshgoebel:
Ruby also. `Prism.languages.rb = Prism.languages.ruby;`

Perhaps you are right. :)  Just wasn't aware of this.

RunDevelopment (changes requested):
Thank you for the PR @Outreach-ChemEng-UofU!

To add an alias, you also have to register it in [`components.json`](https://github.com/PrismJS/prism/blob/bb93fac0409f768e4c5d6d4b9d2310840f2c2866/components.json#L132) using the `"alias"` property. You also have to rebuild Prism by running the command `npm ci && npm run build`.

---

@joshgoebel Yes, this is how Prism currently does aliases.

RunDevelopment:
Thank you for contributing @Outreach-ChemEng-UofU!

PR #2991, opened 2021-07-11
Adding keyword 'record' 'init' 'nullable' (new syntax of C# 8, 9)

kmycode:
New syntax keyword is added on c# 8, 9.
This pull request is adding keywords `record`, `init` and `nullable`.

### nullable (C# 8)
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8

### record, init (C# 9)
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9

### example
```cs
#nullable enable

class TestClass
{
  public string Keyword { get; init; }
}

record TestModel(string Name, uint Id, int Age);
```

PR #2993, opened 2021-07-11
C#: Added `with` keyword & improved record support

RunDevelopment:
As promised in #2991, here are some improvements for C# records.

- Added support for the `with` keyword.
- Type lists for records (inheritance) are now supported.
- Added support for the [`new()` generic constraint](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-constraint). I simply forgot to add it. `new()` was added in C# 7.3, I believe.


JaKXz (approved):
I haven't worked with C# in a few years, so take with a grain of salt, but this makes sense to me 👍🏽

PR #2994, opened 2021-07-11
Sass: Fixed issues with CSS Extras

### LINKED ISSUE #2992: Sass class selectors use inline-color plugin
**Information**
- Language: Sass
- Plugins: inline-color

**Description**
When prism renders Sass using `inline-color` plugin, if a class selector has a color name, the inline color preview is displayed.

**Code snippet**
[Test page](https://prismjs.com/test.html#language=sass&text=.lucidagrande-normal-black-14px%20%0A%20%20color%3A%20var(--black)%0A%20%20font-family%3A%20var(--font-family-lucidagrande)%0A%20%20font-size%3A%20var(--font-size-m)) - though it doesn't display usage of plugins.

image

```
.lucidagrande-normal-
black-14px 
  color: var(--black)
  font-family: var(--font-family-lucidagrande)
  font-size: var(--font-size-m)
```

**Expected behavior**
Class names / any other selector should not use inline-color plugin.

RunDevelopment:
This fixes #2992.

Sass did not expect CSS to be modified by CSS Extras. Sass 1) did not have an optional dependency for `css-extras` and it did not declare its tokens as `greedy: true`. Both of these issues have been fixed. I also improved the boundary checks of some CSS Extras tokens to make sure they can't be part of other identifiers.


JaKXz (approved):
LGTM :)

PR #2998, opened 2021-07-15
Liquid: add all objects from Shopify reference

JaKXz:
@RunDevelopment another quick one for ya :) thanks again for the suggestion!


RunDevelopment (approved):

RunDevelopment:
Thanks, @JaKXz!

PR #3002, opened 2021-07-19
hoon: fix mixed-case aura tokenisation

matildepark:
In Hoon, we have "auras", which are [categories of atoms](https://urbit.org/docs/hoon/reference/stdlib/4o/#aura), which are themselves just positive integers. These look like `@da` but are sometimes, mixed-case, lowercase then uppercase, like `@tD` in the Caesar cipher example. These were incorrectly parsed, and are fixed in this PR. Compare the test diffs for an example.


components/prism-hoon.js
```diff
@@ -8,6 +8,7 @@ Prism.languages.hoon = {
 	'class-name': [
 		{
 			pattern: /@(?:[A-Za-z0-9-]*[A-Za-z0-9])?/,
+			greedy: true
```
RunDevelopment, line 11: `greedy: true` is one way to fix it. However, `greedy: true` is more expensive performance-wise and there is an even simpler fix: Just swap the order of `function` and `class-name`.

Swapping them should have the same result without the performance hit.

matildepark:
Fair enough! Patched.

RunDevelopment (approved):

RunDevelopment:
Thank you very much @matildepark!

PR #3004, opened 2021-07-21
Added support for new language for ASCII diagrams (aka: dial)

pylover:
[Dial](https://github.com/pylover/dial) is a language specially designed to render ASCII diagrams.

Currently, only sequence diagrams are supported, but the roadmap is to support two more types of diagrams: `fork` and `class` diagrams.

The Dial can also run flawlessly inside the browsers using the awesome project: [Brython](https://github.com/brython-dev/brython).


RunDevelopment:
Thank you for the PR @pylover!

I'm sorry to say that I don't think we can accept this PR.

We try to only add languages that are useful to a portion of our users. Judging from the stars of your repo, Dial language highlighting will only be used by a dozen of people total. I don't think that we can justify the maintenance cost of adding a language that will only be used by a dozen people.

There's also the fact that Dial is only 2 months old. Programming/configuration languages tend to change over time. As they get used and abused, people request new features, find bugs, and suggest improvements. Dial is too young to expect any form of stability yet (API and syntax). We also have no insurance that Dial won't be killed/depracted/forgotten in the next two months.

Another problem is that there is not language specification/documentation. This makes it hard to even review this PR because I would have to understand the Dial language parser.

For the abovementioned reasons, I will reject and close this PR.

PR #3005, opened 2021-07-22
∀ is a keyword and alias for forall in PureScript

toastal:


RunDevelopment (changes requested):
Thank you for the PR @toastal!

Please rebuild Prism (`npm run build` after `npm ci`) and add the new keyword to the [keyword test](https://github.com/PrismJS/prism/blob/master/tests/languages/purescript/keyword_feature.test) (and [update the test](https://prismjs.com/test-suite.html#writing-tests-update)). Otherwise, the PR is gtg!

toastal:
> Please rebuild Prism (`npm run build` after `npm ci`) and add the new keyword to the [keyword test](https://github.com/PrismJS/prism/blob/master/tests/languages/purescript/keyword_feature.test) (and [update the test](https://prismjs.com/test-suite.html#writing-tests-update)). Otherwise, the PR is gtg!

I think we are good? Ran for me on Node 14.x locally

tests/languages/purescript/keyword_feature.test
```diff
@@ -18,30 +18,32 @@ primitive
...
+	["keyword", "instance"],
+	["keyword", "let"],
+	["keyword", "module"],
+	["keyword", "newtype"],
+	["keyword", "of"],
+	["keyword", "primitive"],
+	["keyword", "then"],
+	["keyword", "type"],
+	["keyword", "where"],
+	"\r\n∀"
```
RunDevelopment, line 46: And we can see that your addition doesn't work.

I'll make a suggestion on how to fix it.

components/prism-purescript.js
```diff
@@ -1,5 +1,5 @@
 Prism.languages.purescript = Prism.languages.extend('haskell', {
-	'keyword': /\b(?:ado|case|class|data|derive|do|else|forall|if|in|infixl|infixr|instance|let|module|newtype|of|primitive|then|type|where)\b/,
+	'keyword': /\b(?:ado|case|class|data|derive|do|else|forall|if|in|infixl|infixr|instance|let|module|newtype|of|primitive|then|type|where|∀)\b/,
```
RunDevelopment, line 2: The problem is the `\b`s. They are necessary for ASCII keywords but won't work for ∀. `\b` means "asserts a word boundary". Since ∀ isn't a word character, the `\b`s require ∀ to be preceded and followed by a word character (e.g. `a∀a`).


```suggestion
	'keyword': /\b(?:ado|case|class|data|derive|do|else|forall|if|in|infixl|infixr|instance|let|module|newtype|of|primitive|then|type|where)\b|∀/,
```

RunDevelopment:
Thank you for contributing @toastal!

PR #3011, opened 2021-07-26
extend line highlight to full line when nested inside scroll container

Bkucera:
before:

after:
image

- [ ]  test


plugins/line-highlight/prism-line-highlight.js
```diff
@@ -195,6 +195,8 @@
 				});
 			}
 
+			line.style.width = pre.scrollWidth + 'px';
```
RunDevelopment, line 198: This mutates the DOM, so it has to be pushed to `mutateActions`.

```suggestion
			mutateActions.push(function () {
				line.style.width = pre.scrollWidth + 'px';
			});
```

RunDevelopment (changes requested):
Thank you for the PR @Bkucera!

RunDevelopment:
Also, please rebuild Prism by running `npm ci && npm run build`.

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @Bkucera!

Issue #3021, opened 2021-07-31
Add missing keywords for Swift


AvdLee:
**Information**
- Language: Swift
- Plugins: 'none'

**Description**
Some of Swift's keywords are missing. This issue is here for myself to make a collection over the upcoming week, after which I'm happy to implement this myself. If you're reading this, feel free to open up a PR adding support for these keywords!

- [ ] `associatedtype`
- [ ] `@autoclosure`
- [ ] `@discardableResult`
- [ ] `@dynamicMemberLookup`
- [ ] `fileprivate`
- [ ] `@resultBuilder`
- [ ] `@unknown`

Also, not sure if this code is currently highlighted wrong as well:

### Available
Current:
![image](https://user-images.githubusercontent.com/4329185/127739855-e38a7bba-9d9c-433c-a8b6-e553f0e69f0a.png)

Expected:
![image](https://user-images.githubusercontent.com/4329185/127739869-70d5ab98-717c-404d-9cf7-9f8e69fdd897.png)

RunDevelopment:
> I'm happy to implement this myself.

Thank you very much!

> Also, not sure if this code is currently highlighted wrong as well:

What does `#available` mean? Is that part of the function(?) name? I search for this but could only find stuff related to parameters.

AvdLee:
> What does #available mean? Is # part of the function(?) name? I search for this but could only find stuff related to parameters.

It's an availability API to make the compiler happy when using latest API's while also supporting older iOS/macOS versions.

It has the same syntax of functions, which I guess is the reason for its current highlighting. 

In Swift, it's called a declaration attribute: https://docs.swift.org/swift-book/ReferenceManual/Attributes.html

RunDevelopment:
But attributes start with an `@`. [This post](https://stackoverflow.com/questions/32761511/difference-between-available-and-available-in-swift-2-0/32761650) explains the difference. Maybe we should highlight `@available` and `#available` differently?

Also, attributes seem to be very generic in Swift (devs can even define their own). It might be better to have a generic attribute pattern (e.g. `/@\w+/`) instead of the [fixed list](https://github.com/PrismJS/prism/blob/679539ecd04c3b41d6f8ab10c9d62b27c06a8de4/components/prism-swift.js#L22) we currently have.

AvdLee:
[...]

PR #3022, opened 2021-07-31
Swift: Major improvements

RunDevelopment:
I made a number of changes to Swift:

1. Added support for nested comment (`/* /* foo */ */`).
1. Added support for multi-line strings (`""" foo """`) and raw strings (`#" foo "#`)
1. Removed single-quoted strings. Swift does not and has never supported single-quoted strings.
1. Correct tokenization for all operators + punctuation.
1. Added "omit" rule for `_`.
1. Added support for statement labels. (`foo: while ...`)
1. Removed builtin functions/properties. They just looked out of place and provided no real value.
1. Improved function detection.
1. Naming-based type detection. All type-like structures (classes, structs, protocols, etc.) typically have UpperCamelCase names.

Edit:

10. Support directives.
11. Support all/custom attributes.
12. Updated keywords.

---

@AvdLee 
I hope this helps with #3021.

Also, I'm not familiar with Swift and not sure about the naming-based class detection. Can we assume that everything starting with an upper-case letter and containing a lower-case letter is a class name?


AvdLee:
> Also, I'm not familiar with Swift and not sure about the naming-based class detection. Can we assume that everything starting with an upper-case letter and containing a lower-case letter is a class name?

It's fair to say yes, but Swift doesn't enforce us doing so. I haven't seen any code examples though in which developers defined classes starting with a lowercase!

AvdLee:
> @AvdLee
> I hope this helps with #3021.

Really impressive PR! Thanks a lot, I'm learning a lot about this library as well now 🙂  
I'm sure we're getting closer and closer to better support!

Btw, if you like to explore how Prism is behaving with Swift, you can check out my articles here: avanderlee.com
All highlighted code is backed by Prism 💪

RunDevelopment:
@AvdLee I think I got everything from #3021 now.

Also, in your article about [dependency injection](https://www.avanderlee.com/swift/dependency-injection/), you presented this attribute: `@Injected(\.networkProvider)`. What does the `\.` mean?

AvdLee:
> @AvdLee I think I got everything from #3021 now.

Dang, that's amazing! Thanks a ton 💪 

> Also, in your article about dependency injection, you presented this attribute: @Injected(\.networkProvider). What does the \. mean?

It's an accessor to a key path reference. See "Key-Path Expression" in this document: https://docs.swift.org/swift-book/ReferenceManual/Expressions.html

RunDevelopment:
> It's an accessor to a key path reference.

Seems like I missed that part while scanning through the docs. Thank you!

PR #3057, opened 2021-08-30
Core: Fixed type error on null

### LINKED ISSUE #3052: Null access for empty matches with greedy matching
So I just found this bug. If a greedy token creates an empty match `m` at the end of the string (`m = re.exec(text); m[0] == "" && m.index == text.length`), then `matchGrammar` will throw a TypeError. [This line](https://github.com/PrismJS/prism/blob/4e9338acee10bfc68eb650f18c3517ba5f543a3c/components/prism-core.js#L941) throws the error.

What happens is that the greedy matching algorithm tries to find the string or token with the index of `m`. Since there cannot be an empty string or token at the end of the token stream, we will instead try to access the string value of the tail node of the linked list. That tail node has a value of `null`, boom, TypeError.

Example:

```js
Prism.languages.test = {
  'oh-no': {
    pattern: /$/,
    greedy: true
  }
}

Prism.highlight('foo', Prism.languages.test, 'test');
```

image

---

This is not a high prior bug. This bug has been in Prism ever since greedy matching was introduced and nobody noticed up until now.

RunDevelopment:
This fixes #3052.

I fixed the problem by ignoring all matches that cause the problem. I think this is ok because empty tokens at the end of the string don't have any value anyways.


tests/core/greedy.js
```diff
@@ -105,4 +105,19 @@ describe('Greedy matching', function () {
 		});
 	});
 
+	it('issue3052', function () {
+		// this is to test for a bug where greedy tokens where matched like non-greedy ones if the token stream ended on
+		// a string
```
RunDevelopment, line 110:
```suggestion
		// If a greedy pattern creates an empty token at the end of the string, then this token should be discarded
```

PR #3064, opened 2021-09-02
Added support for ASP.NET Razor

### LINKED ISSUE #1225: Add support for ASP.NET Razor to languages
Razor views mix HTML with a specialised C# or VB.NET style syntax.

This would need two languages to be supported:

- cshtml
- vbhtml

RunDevelopment:
This resolves #1225.

image

---

Man, this was difficult. CSHTML is incredibly context-sensitive, so I had to write something pretty close to a parser. I think it turned out ok in the end but the whole language definition is still pretty complex.


RunDevelopment:
Also, since MS seems to let VBHTML die, I will not add support for it. 

I was unable to find sufficient documentation for the language variant. MS has [exactly one page](https://docs.microsoft.com/en-us/aspnet/web-pages/overview/getting-started/introducing-razor-syntax-vb) of documentation for VBHTML last updated 7 years ago. However, from that page, I can see that VBHTML is even more context-sensitive than CSHTML and will most likely require an even more complex grammar.

I do not wish to put time and effort into creating and maintaining the grammar of a poor-documented, dying, incredibly complex language.

PR #3080, opened 2021-09-15
Ensure Previewers popup is visible across themes

### LINKED ISSUE #3074: Code appears on top of Previewers popup when Twilight theme is used
**Information:**
- Prism version: Current deployment on prismjs.com
- Plugins: Previewers
- Environment: Browser? I guess? But I think it's anywhere Previewers is used with Twilight. You can see it on https://prismjs.com/plugins/previewers/

**Description**
Using the Previewers plugin with the Twilight theme selected results in the code hovering on top of the preview popup (screenshot below). I have not really extensively tested this, I was just playing around on the Prism website and happened to notice it. Of the default themes, this only happens with the Twilight theme, the others are fine.

**Example**
image

**Resolution**
I opened DevTools and changed the `z-index` of `.token` to `0`, and the problem was resolved.

https://github.com/PrismJS/prism/blob/d216e602f38e3ff9acec69962b40ada07322e076/themes/prism-twilight.css#L155-L159

image

Of course, seeing the note that the tokens were purposely assigned a `z-index` of `1` probably means that we shouldn't resolve it by simply changing it to `0`. We can probably set a `z-index` for the popup instead, but I'm not sure what implications it would have down the line, especially for users who assign `z-index` to other things on their site.

I can help to resolve this, but I do want to be sure, because I've been lurking a bit and have seen a lot of mention of "backwards compatibility" and "breaking change" and a possible Prism V2 in future.

Cheers!

hoonweiting:
Some themes set a positive z-index for some of their parts, such as the `.token`s in Twilight. This makes the Previewers popup appear underneath the code, which is silly. This PR sets a higher z-index for the popup itself (10), so theme creators can still play around with z-index in their themes, without obscuring the popup if the Previewers plugin were to be used.

This closes #3074.

### DIFF
--- a/plugins/previewers/prism-previewers.css
+++ b/plugins/previewers/prism-previewers.css
@@ -13,6 +13,7 @@
 	width: 32px;
 	height: 32px;
 	margin-left: -16px;
+	z-index: 10;
 
 	opacity: 0;
 	-webkit-transition: opacity .25s;


hoonweiting:
I kinda forgot this isn't prism-themes, so I'm not sure if I need to run `npm run build`?

RunDevelopment (approved):

RunDevelopment:

PR #3081, opened 2021-09-15
Increase selector specificities of plugin overrides in Twilight theme

hoonweiting:
Prompted by https://github.com/PrismJS/prism-themes/pull/141. Also removes redeclared properties, those that are the same as those in the plugins' default stylesheets.

### DIFF
--- a/themes/prism-twilight.css
+++ b/themes/prism-twilight.css
@@ -140,11 +140,6 @@ code[class*="language-"]::selection, code[class*="language-"] ::selection {
 	cursor: help;
 }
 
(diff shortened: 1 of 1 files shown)


RunDevelopment (changes requested):
Thanks for the PR @hoonweiting!

One request: Please also remove the `pre[data-line]` on line 143. It's a duplicate of Line Highlight's rule.

RunDevelopment (approved):

RunDevelopment:

PR #3088, opened 2021-09-20
Document disableWorkerMessageHandler

jakearchibald:
We're going to use this in SVGOMG, so I figured it should be documented 😄


RunDevelopment:
I didn't even know we had this option. Thank you for documenting it!

Unfortunately, you documented the wrong file. `prism.js` is generated. Please document `components/prism-core.js` instead, run `npm ci && npm run build`, and commit all changes.

jakearchibald:
Done!

components/prism-core.js
```diff
@@ -49,6 +49,27 @@ var Prism = (function (_self) {
 		 * @public
 		 */
 		manual: _self.Prism && _self.Prism.manual,
+		/**
+		 * By default, if Prism is in a web worker, it assumes its in a worker it created itself, so it uses
```
RunDevelopment, line 53:
```suggestion
		 * By default, if Prism is in a web worker, it assumes that it is in a worker it created itself, so it uses
```

RunDevelopment (changes requested):
Sorry for the delay @jakearchibald!

jakearchibald:
Updated

components/prism-core.js
```diff
@@ -49,6 +49,27 @@ var Prism = (function (_self) {
 		 * @public
 		 */
 		manual: _self.Prism && _self.Prism.manual,
+		/**
+		 * By default, if Prism is in a web worker, it assumes that it is in a worker it created itself, so it uses
+		 * `addEventListener` to communicate with its parent instance. However, if you're using Prism manually in your
+		 * own worker, you don't want it to do this.
+		 *
```
RunDevelopment, line 56: Our linter really doesn't like trailing spaces.
```suggestion
		 *
```

jakearchibald:
Fixed

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @jakearchibald!

PR #3097, opened 2021-09-27
Use `\d` for `[0-9]`

valtlai:
Like #1107. Also `[\d.]` for `[0-9.]` and `[\d_]` for `[0-9_]`.


RunDevelopment:
Thank you for the PR @valtlai!

While I do agree that `\d` is better than `[0-9]`, I don't think that replacing `0-9` within character classes like `[0-9a-f]` with `\d` improves readability.

Also, we should probably use the [`regexp/prefer-d`](https://ota-meshi.github.io/eslint-plugin-regexp/rules/prefer-d.html) rule to enforce this. I opened an issue for this: ota-meshi/eslint-plugin-regexp#341

RunDevelopment:
My addition to `regexp/prefer-d` was just merged and will be part of the next release.

valtlai:
I edited this PR to only replace `[0-9]` with `\d`, `[0-9.]` with `[\d.]` and `[0-9_]` with `[\d_]`, so character classes like `[0-9a-z]` are kept as is.

RunDevelopment (approved):

RunDevelopment:

PR #3098, opened 2021-09-27
Use tabs in package(-lock).json

valtlai:
This is more consistent with the most of the code base, and npm will preserve the indentation style.


RunDevelopment:
Thank you for the PR @valtlai!

Both of these files are generated and managed by npm. We can't choose the indentation because every `npm i <something>` will change it back spaces. So it's impractical to maintain tabs.

For this reason, I will reject this PR.

valtlai:
That’s not true: npm *does* maintain the indentation style, i.e. running `npm i <something`, for example, won’t change tabs back to spaces. I tested it with both npm v6 and v7.

RunDevelopment:
Oh, I found an old issue where people were angry that it didn't. They must have changed it since then.

valtlai:
Yeah, I’ve probably used npm too little time since I didn’t know this hasn’t always worked like today.&nbsp;😄

RunDevelopment (approved):

RunDevelopment:

Issue #2849, opened 2021-04-06
Document all token types


RunDevelopment:
**Motivation**
Themes depend on Prism producing tokens with specific types (or aliases). Right now, we do not guarantee or document any of those types.

**Description**
Document all standard token types (e.g. `keyword`, `comment`, ...). It should explain the general concept behind each token type and give at least one example.

The documentation should also include how languages are embedded. [Example](https://github.com/PrismJS/prism/blob/3786f3965cb433cb16d9a2f467337916c28eb8db/components/prism-shell-session.js#L37).

We should also guarantee that these concepts within languages are guaranteed to use these token types. (E.g. we guarantee that keywords always have a `keyword` class.) It might sound like we already do this but this is not the case right now. (E.g. we have many languages with operator keywords (e.g. `not` in Python) that do not have a `keyword` class.)

joshgoebel:
>  (E.g. we have many languages with operator keywords (e.g. not in Python) that do not have a keyword class.)

https://github.com/PrismJS/prism/blob/master/components/prism-python.js#L54

What am I missing or not understanding about that statement?

RunDevelopment:
> Looks like not is a keyword to me?

Oops. Wrong language. I meant SQL. Thanks for pointing that out!

joshgoebel:
SQL has a `keyword` class also... but not is an `operator` there.  Are you talking about resolving this inconsistency between grammars?

https://github.com/PrismJS/prism/blob/master/components/prism-sql.js#L19

RunDevelopment:
> Are you talking about resolving this inconsistency between grammars?

Yes, that too. My main intention was to enable themes to decide whether operator keywords are to be highlighted as keywords or operators. Right now, Prism languages make that decision (and that pretty inconsistently as you pointed out) by assigning them one type (either `keyword` or `operator`). In the case of operator keywords, the best solution would probably be to use an alias for keywords, so we get a CSS class name of `.keyword.operator`.

joshgoebel:
> .keyword.operator

How would you encode that to HTML then? Nested tags?  I dislike that because it makes `operator` ambiguous on it's own you then have to use something like `pre > operator` to scope it to the top level, but for us that doesn't even work because operators don't always have to be top-scope... we might legitimately have an operator inside of some other scope (other than keyword)... so then the order of the CSS becomes super important.  And even then you could see breakage because say `.operator` defines a background while `.keyword.operator` does not.  

Is that a bug? A feature?

[...]

Issue #354, opened 2014-09-01
PrismJS.com download page doesn't include Git language CSS


ccoley:
The download page on the website doesn't add the CSS rules for the Git language. Is this a bug or am I supposed to go get the proper CSS from @lgiraudel manually?

joshgoebel:
> I understood the principle of aliases but it sounds weird to me to use an alias just for a matter of color. I'm fine with aliasing "git comment" by "comment" but what should be the alias of "git sha1" or "git diff coordinates" ?

_You're not aliasing colors._  You're aliasing semantics.

- Your language might call them "annotations", you'd alias that to "comment".
- Your language might call them "slots", perhaps you'd alias to "variable"
- You have "sha_commit", perhaps your alias that to "tag" or "important"

It is a difficult problem when nothing that exists seems to be a good match, but that's a hard problem to solve - since you can't just have 1000 different CSS attributes in every theme - at some point you have to agree on comment elements with semantic meaning.

PR #3104, opened 2021-10-03
Document standard tokens and provide examples

hoonweiting:
This PR creates a new page on the website called `tokens.html`, where the standard token names are listed in a definition list, with at least one example provided for each definition. I'm only familiar with like, maybe ten languages at best, so I would **really** appreciate anyone's and everyone's feedback on the definitions! Also, more ideas for examples are very welcome!!

I also have a few questions in particular:
1. Should `inserted` and `deleted` be collapsed into one entry? This will also allow us to use a combined example to showcase insertions and deletions, which seems more legit, lol. Also, should Brainfuck be showcased as an example of using `inserted` and `deleted`?
2. Should `attr-name` and `attr-value` be collapsed into one entry as well?
3. In particular, I would like feedback on `symbol`, `namespace`, `prolog`, and `cdata`, because I honestly have no idea what they are, and copied them off google, haha.

And one more thing, I could have sworn I saw in a comment somewhere about "aliasing semantics" when it comes to token names. But for some reason I can't seem to find it right now. I do want to include something along those lines in the introductory blurb to encourage contributors to alias non-standard tokens with semantically-similar standard tokens! Just really not sure how to phrase it.

This hopes to resolve the first half of #2849.


tokens.html
```diff
@@ -0,0 +1,313 @@
...
+x || y
+z ? x : y</code></pre></dd>
+
+		<dt><code>variable</code></dt>
+		<dd>An object that is assigned and holds a specific value.
+		<pre><code class="language-less">@nice-blue: #5B83AD;
+@light-blue: lighten(@nice-blue, 20%);</code></pre></dd>
+
+		<dt><code>constant</code></dt>
+		<dd>The name of a a variable that cannot be modified later on.
```
RunDevelopment, line 137: That's not entirely true. I can't find the relevant discussion anymore but basically: We view `constant`s are program constants. So not every `const` JS variable is a constant. We also want the constant to be static (= live as long as the program) similar to [Rust const](https://doc.rust-lang.org/std/keyword.const.html) and C `#define` macros constants.

Frankly, I'm not sure how to best communicate the concept of a constant that we want to highlight here.

However, I know that the word "variable" doesn't fit. 

```suggestion
		<dd>The name of a constant.
```

tokens.html
```diff
@@ -0,0 +1,313 @@
...
+		<pre><code class="language-javascript">const PI = 3.14159;</code></pre></dd>
+
+		<dt><code>property</code></dt>
+		<dd>An attribute/characteristic.
```
RunDevelopment, line 141: See JSON.

```suggestion
		<dd>An attribute/characteristic or object/map key.
```

tokens.html
```diff
@@ -0,0 +1,313 @@
...
+			[ 0, -4, -8, -12 ]
+		]
+	},
+	// we even support comments
+	"error": null,
+	"status": "Ok"
+}</code></pre></dd>
+
+		<dt><code>punctuation</code></dt>
+		<dd>Punctuation that is not contained within a string, such as brackets.
```
RunDevelopment, line 163: > "that is not contained within a string"

Where does this come from? Or asked differently: what is punctuation contained in a string?

tokens.html
```diff
@@ -0,0 +1,313 @@
...
+	'''
+	copy = sorted(pool)
+	size = len(copy)
+	if size % 2 == 1:
+		return copy[(size - 1) / 2]
+	else:
+		return (copy[size/2 - 1] + copy[size/2]) / 2</code></pre></dd>
+
+		<dt><code>important</code></dt>
+		<dd>Anything that is important.
```
RunDevelopment, line 177:
```suggestion
		<dd>Anything that is important and needs special highlighting.
```

tokens.html
```diff
@@ -0,0 +1,313 @@
...
+		<dt><code>bold</code></dt>
+		<dd>Bolded text. Mostly found in document-markup languages.
+		<pre><code class="language-markdown">**I am bolded text!**</code></pre></dd>
+
+		<dt><code>italic</code></dt>
+		<dd>Italicised text. Mostly found in document-markup languages.
+		<pre><code class="language-markdown">*I am italicised text!*</code></pre></dd>
+
+		<dt><code>tag</code></dt>
+		<dd>The markup version of a keyword, to indicate when an element begins and ends.
```
RunDevelopment, line 212: A tag is quite different from a keyword. It's its own concept, so I think a short generic description is most fitting.

```suggestion
		<dd>A markup tag (e.g. HTML and XML tags).
```

tokens.html
```diff
@@ -0,0 +1,313 @@
...
+		<pre><code class="language-markup">&lt;?xml version="1.0" encoding="utf-8"?&gt;
+&lt;svg&gt;&lt;/svg&gt;</code></pre></dd>
+
+		<dt><code>doctype</code></dt>
+		<dd>Document type declaration, specific to markup languages.
+		<pre><code class="language-markup">&lt;!DOCTYPE html&gt;
+&lt;html&gt;&lt;/html&gt;</code></pre></dd>
+
+		<dt><code>cdata</code></dt>
+		<dd>Character data, specific to markup languages. Deprecated as of DOM4.
```
RunDevelopment, line 240: It may be deprecated for modern websites but it's still going strong in XML.

```suggestion
		<dd>Character data, specific to markup languages.
```

tokens.html
```diff
@@ -0,0 +1,313 @@
...
+		<pre><code class="language-markup">&lt;video allowfullscreen controls&gt;
+	&lt;source src="hello_world.mp4" type="video/mp4" /&gt;
+&lt;/video&gt;</code></pre></dd>
+
+		<dt><code>attr-value</code></dt>
+		<dd>The value or argument that is passed to a property in a markup tag.
+		<pre><code class="language-markup">&lt;p id="greeting"&gt;Hello World!&lt;/p&gt;</code></pre></dd>
+
+		<dt><code>namespace</code></dt>
+		<dd>Used to provide uniquely named elements and attributes in XML documents.
```
RunDevelopment, line 226: This is also used by e.g. Java for package names.

Examples: [C#](https://prismjs.com/test.html#language=csharp&text=namespace%20Foo.Bar%20%7B%7D%0Ausing%20Foo.Bar%3B), [Java](https://prismjs.com/test.html#language=java&text=%2F%2F%20Java%0Aclass%20Foo%20extends%20foo.bar.Foo%20%7B%0A%09java.util.List%3Cfoo.bar.Foo.Bar%3E%20bar(foo.bar.Baz%20bat)%20throws%20java.lang.IOException%20%7B%0A%09%09throw%20new%20java.lang.UnsupportedOperationException(%22Not%20implemented%22)%3B%0A%09%7D%0A%7D), [Rust](https://prismjs.com/test.html#language=rust&text=use%20std%3A%3Acollections%3A%3AHashMap%3B%0Ause%20std%3A%3Afmt%3B%0Ause%20std%3A%3Aiter%3B%0Ause%20std%3A%3Aresult%3B%0Ause%20std%3A%3Async%3A%3AArc%3B%0A%0Ause%20regex_syntax%3A%3Ahir%3A%3A%7Bself%2C%20Hir%7D%3B%0Ause%20regex_syntax%3A%3Ais_word_byte%3B%0Ause%20regex_syntax%3A%3Autf8%3A%3A%7BUtf8Range%2C%20Utf8Sequence%2C%20Utf8Sequences%7D%3B%0A%0Ause%20crate%3A%3Aprog%3A%3A%7B%0A%20%20%20%20EmptyLook%2C%20Inst%2C%20InstBytes%2C%20InstChar%2C%20InstEmptyLook%2C%20InstPtr%2C%20InstRanges%2C%0A%20%20%20%20InstSave%2C%20InstSplit%2C%20Program%2C%0A%7D%3B%0A%0Ause%20crate%3A%3AError%3B%0A%0Atype%20Result%20%3D%20result%3A%3AResult%3CPatch%2C%20Error%3E%3B%0Atype%20ResultOrEmpty%20%3D%20result%3A%3AResult%3COption%3CPatch%3E%2C%20Error%3E%3B).

```suggestion
		<dd>Used to provide uniquely named elements and attributes in XML documents. Outside of markup languages, it is used to tokenize the package/namespace part of identifiers.
```

tokens.html
```diff
@@ -0,0 +1,313 @@
...
+		],
+		"series": [
+			[ 0, 1, 2, 3 ],
+			[ 0, -4, -8, -12 ]
+		]
+	},
+	// we even support comments
+	"error": null,
+	"status": "Ok"
+}</code></pre></dd>
```
RunDevelopment, line 160: This is a little too long IMO. Maybe we could say: 8 lines max per example code block?

```suggestion
		<pre><code class="language-json">{
	"data": {
		"labels": ["foo", "bar"],
	},
	"error": null,
	"status": "Ok"
}</code></pre></dd>
```

RunDevelopment (changes requested):
Great work @hoonweiting!

> Should inserted and deleted be collapsed into one entry?
> Should attr-name and attr-value be collapsed into one entry as well?

I think that's a good idea for both.

> In particular, I would like feedback on symbol, namespace, prolog, and cdata

You got all of them right. Just one thing needs to be added to `namespace` (see comment).

> I could have sworn I saw in a comment somewhere about "aliasing semantics"

Aliasing semantics work like this: If you have a token that is more specific, you can use a custom token name and give it a standard token name as an alias. E.g. (take from Rust):

```js
'function-definition': {
	pattern: /(\bfn\s+)\w+/,
	lookbehind: true,
	alias: 'function'
},
```

Some languages also abuse aliasing to get a particular highlighting color but that shouldn't be done.

tokens.html
hoonweiting: Ah hmm, I guess that makes sense. I suppose we will have to rely on examples to help illustrate it then.

tokens.html
```diff
@@ -0,0 +1,313 @@
...
+	<dl>
+		<dt><code>keyword</code></dt>
+		<dd>Pre-defined and reserved words.
+		<pre><code class="language-javascript">for (let x of y) {
+	// Code here
+	return z;
+}</code></pre></dd>
+
+		<dt><code>builtin</code></dt>
+		<dd>Functions/Methods that are available out of the box.
```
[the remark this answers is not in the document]
hoonweiting: Haha alright! I will google more and update it when I do.

tokens.html
```diff
@@ -0,0 +1,313 @@
...
+2147483647
+0o177
+0b100110111
+3
+79228162514264337593543950336
+0o377
+0x100000000
+0xdeadbeef
+3.14
+10.
```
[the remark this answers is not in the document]
hoonweiting: Oh speaking of the length of examples, what should I do about this one? I honestly just copied and pasted it from the examples page to catch as many cases as possible, and to be honest, I had no idea Python supports imaginary numbers!

tokens.html
hoonweiting: Hmm. I think I wanted to say that `punctuation` shouldn't capture punctuation within strings, such as in the Python example (reproduced below for convenience):

```
def median(pool):
	'''Statistical median to demonstrate doctest.
	>>> median([2, 9, 9, 7, 9, 2, 4, 5, 8])
	7
	'''
	copy = sorted(pool)
	size = len(copy)
	if size % 2 == 1:
		return copy[(size - 1) / 2]
	else:
		return (copy[size/2 - 1] + copy[size/2]) / 2
```

But yeah, it seems pretty unnecessary on second thought, since it would take conscious effort to actually define the punctuation within a string as `punctuation`.

Can't quite think of a way to make it sound better at this point. How did people ever write dictionaries? Hahaha

tokens.html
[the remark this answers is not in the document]
RunDevelopment: No need. I think keeping it simple is best.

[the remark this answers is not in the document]
RunDevelopment: Let's just make it "Functions/Methods/Classes/Types".

tokens.html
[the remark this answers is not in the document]
RunDevelopment: I think 3 or 4 numbers are enough.

tokens.html
RunDevelopment: Tokens can't _overlap_, so I don't think that we need to specify this here. 

It's also a little wrong because tokens can be _nested_ (`inside`) and some languages support string interpolation which gives rise to string interpolation punctuation. So there can be punctuation inside strings, just not the same one as outside strings.

I think we can just go with: "Punctuation such as brackets, parentheses, commas, and more."

tokens.html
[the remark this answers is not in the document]
hoonweiting: Alright, thanks!

tokens.html
hoonweiting: Ah nested, that's the word!

> I think we can just go with: "Punctuation such as brackets, parentheses, commas, and more."

Sounds good!

hoonweiting:
Alright, most of the changes have been made, just need to add more examples, particularly for `constant`.

---

Also I have three examples for `inserted` and `deleted` now (all taken from the examples page), but I have no idea why there's no vertical margin between them (unlike other tokens with multiple examples). Upon opening dev tools, I see that the style is being applied to the `<div class="code-toolbar">` tag directly, so I'm seeing `<div class="code-toolbar" style="margin: 0px;">` for all the diff/git codeblocks, compared to `<div class="code-toolbar" style="margin: 8px 0px;">` in the non-last `url`, `property`, and `important` (basically the other tokens with multiple examples) codeblocks.

On further testing it seems that it's a problem of having two codeblocks with the same language next to each other, but this is not a perfect explanation either (css with css, and markup with markup is fine, but diff/markdown/python are not, of those I just tried).

And the above two paragraphs only hold true in Chrome (and Edge). In Firefox, none of them have vertical spacing. I don't know what's going on tbh, lol. Commenting out the Toolbar and Show Language plugins seemed to return the margins consistently, so I'll guess the problem lies somewhere there.

However, not sure if all this will be a different story when published online, since not all elements are loading for me locally (like the Prism logo with that download button). I doubt it, but I can't say for certain.

I thought this was gonna be a short paragraph but I guess not! Maybe I should open a new issue for this.

RunDevelopment:
> Also I have three examples for inserted and deleted now

Sounds a bit much. Maybe we could just use the middle example? A git diff should be understood by most people.

> since not all elements are loading for me locally (like the Prism logo with that download button)

`npm start` opens a small local server for the website.

> Maybe I should open a new issue for this.

Please do. That sounds like a general problem with Toolbar.

hoonweiting:
> Maybe we could just use the middle example?

Sure thing!

> npm start opens a small local server for the website.

Oh damn, I didn't know there was something like this. This is fantastic, thank you!

> That sounds like a general problem with Toolbar.

Okay, I will do so after more testing. Inb4 it's me forgetting to escape some tags or making some other typo somewhere, haha.

hoonweiting:
Also if anyone is interested, here's the comment on "aliasing semantics" that I finally managed to find: https://github.com/PrismJS/prism/issues/354#issuecomment-731504292

hoonweiting:
I don't know if I'm completely happy with this, the examples feel pretty inconsistent to me, some are foobar types, some are very toy examples, others are more legit. But that's honestly probably not so important in the grand scheme of things, so maybe it's much ado about nothing.

I do think the introductory blurb needs work though. I'm not sure if it sounds a little repetitive, or if it's unclear.

tokens.html
```diff
@@ -0,0 +1,335 @@
...
+	&lt;source src="hello_world.mp4" type="video/mp4" /&gt;
+&lt;/video&gt;</code></pre></dd>
+
+		<dt><code>namespace</code></dt>
+		<dd>Used to provide uniquely named elements and attributes in XML documents. Outside of markup languages, it is used to tokenize the package/namespace part of identifiers.
+		<pre><code class="language-markup">&lt;html:p foo:bar="baz" foo:weee&gt;&lt;/html:p&gt;</code></pre>
+		<pre><code class="language-csharp">namespace Foo.Bar {}
+using Foo.Bar;</code></pre>
+		<pre><code class="language-java">class Foo extends foo.bar.Foo {
+	java.util.List<foo.bar.Foo.Bar> bar(foo.bar.Baz bat) throws java.lang.IOException {
```
RunDevelopment, line 227:
```suggestion
	java.util.List&lt;foo.bar.Foo.Bar> bar(foo.bar.Baz bat) throws java.lang.IOException {
```

tokens.html
```diff
@@ -0,0 +1,335 @@
...
+
+use crate::prog::{
+		EmptyLook, Inst, InstBytes, InstChar, InstEmptyLook, InstPtr, InstRanges,
+		InstSave, InstSplit, Program,
+};
+
+use crate::Error;
+
+type Result = result::Result<Patch, Error>;
+type ResultOrEmpty = result::Result<Option<Patch>, Error>;</code></pre></dd>
```
RunDevelopment, line 249: Let's shorten this a little.

```suggestion
use std::fmt;
use std::sync::Arc;</pre></dd>
```

tokens.html
```diff
@@ -0,0 +1,335 @@
...
+	<div class="intro" data-src="assets/templates/header-main.html" data-type="text/html"></div>
+
+	<h2>Prism tokens</h2>
+	<p>Prism identifies tokens in your code, which are in turn styled by CSS to produce the syntax highlighting. This page provides an overview of the standard tokens and corresponding examples.</p>
+</header>
+
+<section id="standard-tokens">
+	<h1>Standard tokens</h1>
+
+	<p>When defining a new language, you will need to provide token names for each 'type' of code, such as keywords and operators, so Prism can highlight code accordingly. It is recommended to make use of the following standard tokens to ensure that highlight code will be highlighted, as Prism's themes (both official and non-official) only guarantee coverage for these standard tokens.</p>
```
RunDevelopment, line 30: 1. Prism doesn't highlight code, our themes do.
2. Slight change for better clarity.

```suggestion
	<p>When defining a new language, you will need to provide token names for each 'type' of code, such as keywords and operators, so Prism's themes can assign colors (and other styles) accordingly. Prism's themes (both official and non-official) only guarantee coverage for these standard tokens, so it is recommended to make use of the following standard tokens to ensure that code will be highlighted.</p>
```

tokens.html
```diff
@@ -0,0 +1,335 @@
...
+
+	<p>Sometimes, a language might use a particular name to refer to certain pieces of code, but which is not one of Prism's standard token names, such as <code>function-defintion</code>. Since <code>function-definition</code> is not a standard token, you might want to alias it with a standard token such as <code>function</code>, which is semantically similar in meaning, and will ensure that Prism's themes will highlight it.</p>
```
RunDevelopment, line 32: Maybe we could move that into its own section "Non-standard tokens"?

tokens.html
```diff
@@ -0,0 +1,335 @@
...
+
+	<dl>
```
RunDevelopment, line 34: Regarding the definition list: I think a table would be better. I'll send you a PR.

tokens.html
hoonweiting: Makes sense!

PR #3105, opened 2021-10-03
Elm: Recognise unicode escapes as valid Char

hoonweiting:
[Elm recognises unicode escapes from `'\u{0000}'` to `'\u{10FFFF}'` as valid Char.](https://package.elm-lang.org/packages/elm/core/latest/Char) Not sure if the regex I modified is the best way to do it though!


components/prism-elm.js
```diff
@@ -1,7 +1,7 @@
 Prism.languages.elm = {
 	'comment': /--.*|\{-[\s\S]*?-\}/,
 	'char': {
-		pattern: /'(?:[^\\'\r\n]|\\(?:[abfnrtv\\']|\d+|x[0-9a-fA-F]+))'/,
+		pattern: /'(?:[^\\'\r\n]|\\(?:[abfnrtv\\']|\d+|x[0-9a-fA-F]+|u\{(?:[A-F0-9]{4,5}|10[0-9A-F]{4})\}))'/,
```
RunDevelopment, line 4: Elm also allows `'\u{ffff}'` and we don't have to count the number of hex digits exactly.

```suggestion
		pattern: /'(?:[^\\'\r\n]|\\(?:[abfnrtv\\']|\d+|x[0-9a-fA-F]+|u\{[0-9a-fA-F]+\}))'/,
```

RunDevelopment (changes requested):
Thanks for the PR @hoonweiting!

Only one nit!

components/prism-elm.js
hoonweiting: Ah great catch on the lowercase! But I'm not sure about the counting, because Elm only recognises specifically `'\u{0000}'` to `'\u{10ffff}`'. Then again, Prism is a highlighter, not a linter...?

hoonweiting: Okay I think it makes sense, yes.

RunDevelopment: Prism is also used for editing by some, so highlighting incomplete chars is advantageous.

RunDevelopment (approved):

RunDevelopment:

PR #3108, opened 2021-10-04
package.json: Add engines.node field to ensure the node.js version >=10

shfshanyue:


RunDevelopment:
The requirement of NodeJS >=10 from the README only applies to our test suite and tooling. Prism itself works all the way down to NodeJS >=6. We just can't test that because our tests don't run.

shfshanyue:
Thansks, I see. So do we need to change it to NodeJS >= 6?

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @shfshanyue!

PR #3126, opened 2021-10-05
Python: Recognise walrus operator

hoonweiting:
[Python 3.8 introduced the assignment expression operator](https://www.python.org/dev/peps/pep-0572/), better known as the walrus operator. However, I'm not sure if `...|!=|:=|...` is preferred over `...|[!:]=|...`.


RunDevelopment (approved):
LGTM; I'll merge this after GH Actions come back online.

Thank you @hoonweiting!

> However, I'm not sure if `...|!=|:=|...` is preferred over `...|[!:]=|...`.

Both are fine.

PR #3127, opened 2021-10-05
TAP: Conform to quoted properties style

RunDevelopment:


RunDevelopment:
I'll merge this now since it's style-only.

PR #3142, opened 2021-10-07
Python: Add `match` and `case` (soft) keywords

hoonweiting:
[Python 3.10 (just released on Monday!) introduces structural pattern matching](https://www.python.org/dev/peps/pep-0634/), aka switch-case statements, except that in Python it's [match-case](https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching). It's currently a [soft keyword](https://docs.python.org/3.10/reference/lexical_analysis.html#soft-keywords), together with `_` which kinda acts like `default` in JavaScript. This PR adds `match`, `case`, and `_` (if `:` comes right after it) as keywords.

`*` can be used as a wildcard, and `_` can also be used within lists and dictionaries to represent the idea of `default`. But I'm not sure if I should try to match `_` when it is used in such situations? I feel that there'll be quite a few cases to consider (even if we only consider code that follows the style guide), and I don't know how Prism normally deals with this.


components/prism-python.js
```diff
@@ -51,7 +51,7 @@ Prism.languages.python = {
 			'punctuation': /\./
 		}
 	},
-	'keyword': /\b(?:and|as|assert|async|await|break|class|continue|def|del|elif|else|except|exec|finally|for|from|global|if|import|in|is|lambda|nonlocal|not|or|pass|print|raise|return|try|while|with|yield)\b/,
+	'keyword': /\b(?:_(?=:)|and|as|assert|async|await|break|case|class|continue|def|del|elif|else|except|exec|finally|for|from|global|if|import|in|is|lambda|match|nonlocal|not|or|pass|print|raise|return|try|while|with|yield)\b/,
```
RunDevelopment, line 54:
```suggestion
	'keyword': /\b(?:_(?=\s*:)|and|as|assert|async|await|break|case|class|continue|def|del|elif|else|except|exec|finally|for|from|global|if|import|in|is|lambda|match|nonlocal|not|or|pass|print|raise|return|try|while|with|yield)\b/,
```

PR #3148, opened 2021-10-10
Tests: Added `.html.test` files for replace `.js` language tests

RunDevelopment:
This is a branch of #3147 and adds support for HTML `.test` files called `.html.test`. It is just an extension of the current test format.

The expected HTML code is formatted using Prettier (good-looking and readable format) and can be inserted automatically using the `--update` flag (just like for all other `.test` files). This makes it super easy to add and update `.html.test` files.

Support for `.js` language tests has been removed since they are no longer necessary. `.html.test` files have the same functionality and are easier to use.


JaKXz (approved):
Nice!

PR #3152, opened 2021-10-11
ESLint: Added `regexp/prefer-{plus,question,star}-quantifier` rules

RunDevelopment:
One of the few rules of eslint-plugin-regexp that don't change any of our regexes.


JaKXz (approved):
😎

PR #3158, opened 2021-10-21
Check js-templates in case it was already loaded

romulof:
If `js-templates` is loaded twice, it will break because `Prism.languages.javascript['template-string']` will be changed from a regexp to an Array.

Would not happen in normal scenarios, but it  breaks the optimized module loader I'm writing and the extra check would not harm function and the bundle size impact is minimal.

### DIFF
--- a/components/prism-js-templates.js
+++ b/components/prism-js-templates.js
@@ -2,6 +2,10 @@
 
 	var templateString = Prism.languages.javascript['template-string'];
 
+	if (Array.isArray(Array)) {
+		templateString = templateString[templateString.length - 1]
+	}
+
 	// see the pattern in prism-javascript.js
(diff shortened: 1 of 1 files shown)


RunDevelopment:
> If `js-templates` is loaded twice

That's the problem. You can't just load a language twice, Prism's languages aren't guaranteed to be okay with that. To be fair, most languages will continue to work but it's not guaranteed. 

If you want to load a language again, you have to honor its dependencies to other languages. In this case, JS templates modifies JS, so JS has to be reloaded too (assuming that only JS and JS templates are loaded, it gets more complex when more languages are involved).

For more information about dependencies, see [this section of our docs](https://prismjs.com/extending.html#dependencies).

---

Since this PR is solving a problem that will only occur if Prism is used incorrectly, I will reject this PR.

PR #3177, opened 2021-11-04
Website: Improved readability

### LINKED ISSUE #3175: prismjs.com has poor readability
That background image provides a very poor contrast.

image

image

RunDevelopment:
Fixes #3175.

image

image

### DIFF
--- a/assets/style.css
+++ b/assets/style.css
@@ -130,7 +130,8 @@ header, footer {
 	padding: 30px -webkit-calc(50% - 450px); /* Workaround for bug */
 	padding: 30px calc(50% - 450px);
 	color: white;
-	text-shadow: 0 -1px 2px black;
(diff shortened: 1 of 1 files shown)


not-my-profile:
Thanks, that is definitely better! However I still find the text below the headings to be a bit hard to read. Probably mainly due to the narrow font.

RunDevelopment:
I agree. It's more readable with a larger font size.

image

It's not perfectly readable (it's hard with such a high-contrast background) but I think it's okay-ish.

not-my-profile:
How about something like `linear-gradient( transparent, rgba(0, 0, 0, 0.7) ),url("https://prismjs.com/assets/img/spectrum.png") fixed;`? That is making the background image less distracting by overlaying a transparent gradient?

RunDevelopment:
Very good idea @not-my-profile!

image

not-my-profile:
LGTM!

PR #3179, opened 2021-11-06
Website: Improve readability of features

### LINKED ISSUE #3175: prismjs.com has poor readability

not-my-profile:
Fixes #3175.

Before:

image

After:

image


RunDevelopment:
As I said in #3177, very good idea.

However, I'll close this because it doesn't get the font size right. The sizes of paddings (among others) changed here, so I prefer #3177 where I fix all of that.

PR #3190, opened 2021-11-09
Adding a PrismJS React and HTML tutorial link

catchsrinivas:
Adding a PrismJS React and HTML tutorial link


RunDevelopment:
THanks for the article @catchsrinivas!

Unfortunately, I can't accept it as is since it contains a few mistakes:

1. "How to add Prism.js syntax highlighting to any website? > Step 4:" 

    The shown code blocks are a little wrong. This is HTML (not React), so the spaces between `<pre>` and `<code>` will be displayed on the page. In plain HTML, there must not be spaces and/or line breaks between the opening and closing `<pre>` and `<code>` tags. (It doesn't matter in React because React will automatically remove any spaces and line breaks.)

2. "Get PrismJS working in React"

    The React section isn't correct. Since you aren't using the babel plugin, `import Prism from "prismjs"` will just import a Prism instance with HTML, CSS, and JS loaded. Nothing else. Importing other languages isn't easy without the babel plugin.

    Also, selecting other languages before downloading the CSS won't do anything because languages don't have associated CSS. Also also, downloading CSS isn't necessary anyway with the babel plugin.

catchsrinivas:
@RunDevelopment Thanks for the feedback. I have made the necessary changes in the article for both HTML and React tutorials and also added a sample working demo project in github. Could you please review and approve?

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @catchsrinivas!

PR #3259, opened 2021-12-02
Twig: Improved tokenization

RunDevelopment:
Changes:

- Use Markup templating
- Remove non-standard use of `property` token
- Simplify delimiter tokens
- Removed non-standard use of `tag` token


JaKXz (approved):
Excellent!

PR #3264, opened 2021-12-02
Zig: Use standard `char` token

RunDevelopment:
Changes:

- Added `char` token
- Renamed `builtin-types` -> `builtin-type`


hoonweiting (commented):
Looking at the [keyword reference for the latest stable version (0.8.1)](https://ziglang.org/documentation/0.8.1/#Keyword-Reference), I think there are three new keywords: `anyframe`, `anytype`, and `nosuspend`. Since this PR is focused on using the `char` token and does not touch the `keyword` token at all, should adding these keywords be part of another PR instead? If so, then I think this PR is ready to go!

RunDevelopment:
> I think there are three new keywords: `anyframe`, `anytype`, and `nosuspend`

Good catch! I made a separate PR for that (#3279).

PR #3266, opened 2021-12-04
NSIS: Update Regex pattern for variables

idleberg:
Unless at the beginning of a variable name, variables in NSIS may contain the `.`-character ([see example](https://github.com/kichik/nsis/blob/master/Contrib/Modern%20UI%202/Interface.nsh#L11-L25)).

**Changes:**

- update Regex pattern for variables to support dots
- add test for the above

PR #3267, opened 2021-12-04
NSIS: Update Regex for constants pattern

idleberg:
Similar to #3266, the constant patterns needs to be updated to support `!`-characters. While this is, arguably, highly unreadable, it's valid code shown in this [example](https://nsis.sourceforge.io/Check_if_a_file_exists_at_compile_time) (take note of the line `${!defineifexist} var_name file_name`).

**Changes:**

- update Regex pattern for constants to support exclamation marks
- add test for the above


RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @idleberg!

PR #3270, opened 2021-12-05
C#: Added `char` token

RunDevelopment:


hoonweiting (commented):
What a coincidence! I was looking at this exact same thing a couple of hours ago too.

I did have a question though, and that is, should `character` have the quantifier `{1,8}` at the end?

https://github.com/PrismJS/prism/blob/baa95cab5f1f69aa4b965d7335b657b2a2931959/components/prism-csharp.js#L81-L84

I ask because I remember you asked me to remove it from #3105. I'm not familiar enough with C# (not that I knew Elm either!), so should the quantifier be removed here to make it consistent with Elm, or be left as it is for some special reason?

RunDevelopment:
I used `{1,8}` to allow unfinished characters for a better typing experience. This is also the reason why I asked you to change Elm's char quantifier.

I though `[0-9a-fA-F]+` was the best choice for Elm because many other languages with a `\u{0000}` escape syntax allow arbitrarily long hex numbers to represent the code point. I don't know much about Elm, so I assumed it's the same here (Elm docs aren't clear about this).

C# on the hand only allows `\x00`, `\u0000`, and `\U00000000` as hex escape sequences. So 2, 4, or 8 hex digits. Since the maximum is 8, I went with `{1,8}` instead of `+`.

hoonweiting:
Ahh I see, thank you for the explanation! This PR looks good to me!

PR #3272, opened 2021-12-05
editorconfig: Swap out `property` for `key`; alias with `attr-name`

hoonweiting:
This PR replaces the `property` token with `key`, and aliases it to the standard token `attr-name`. This PR also replaces the `string` alias with `attr-value` for the `value` token. These changes bring editorconfig in line with INI and systemd.


RunDevelopment (approved):

RunDevelopment:
Thanks @hoonweiting!

PR #3275, opened 2021-12-07
HTTP: Added support for special header value tokenization

RunDevelopment:
This PR changes how HTTP tokenizes header lines and allows it to give special tokenization to specific headers. This is used to include CSP, HPKP, and HSTS tokenization in HTTP.


Tuneabq (commented):
👍👍

Issue #3265, opened 2021-12-03
Enforce consistent practices for greedy patterns


RunDevelopment:
**Motivation**
Greedy matching is hard and interacts in non-obvious ways with non-greedy matching. This has caused issues in the past, like the one fixed in #2680.

One best practice for grammars containing both greedy and non-greedy tokens is to ensure that there are no non-greedy tokens before any greedy token. Let's call a grammar of that form a normalized grammar or grammar in normal form.

Normalized grammars are immune to the unintuitive interactions between greedy and non-greedy patterns. Hence, they are immune to a whole range of highlighting bugs. Unfortunately, a lot of our current languages do not follow this best practice.

**Description**
The goals of this issue are to:

1. Find out whether every grammar can feasibly be brought into normal form.
2. If yes, add automated tests that enforce normalized grammars.

These goals can be achieved incrementally. The first steps should be to find, fix, and test for specific patterns where grammars only slightly deviate from normal form.

PR #3280, opened 2021-12-14
Added support for Odin

### LINKED ISSUE #3204: Odin Programming Language
**Language**
Odin is a general-purpose programming language with distinct typing, built for high performance, modern systems, and built-in data-oriented data types. The Odin Programming Language, the C alternative for the joy of programming. The Data-Oriented Language for Sane Software Development.

**Additional resources**
https://odin-lang.org/
https://odin-lang.org/docs/overview/
https://github.com/odin-lang/Odin

Syntax:
```js
Prism.languages.odin = Prism.languages.extend('clike', {
	'string': {
		pattern: /(["'`])(?:\\[\s\S]|(?!\1)[^\\])*\1/,
		greedy: true
	},
	'keyword': /\b(?:asm|auto_cast|bit_set|break|case|cast|context|continue|defer|distinct|do|dynamic|else|enum|fallthrough|for|foreign|if|import|in|map|matrix|not_in|or_else|or_return|package|proc|return|struct|switch|transmute|typeid|union|using|when|where)\b/,
	'boolean': /\b(?:_|false|true|nil)\b/,
	'number': /(?:\b0[xh][0-9a-fA-F_]+|(?:\b0b[01_])|(?:\b0o[0-7_])|(?:\b0d[0-9_])|(?:\b0z[0-9abAB_])|(?:\b\d+(?:\.\d*)?|\B\.\d+)(?:e[-+]?\d+)?)[ijk]?/i,
	'operator': /[+-\/%~!=<>]=?|%%=?|---|\?|\|(?:=|\|=?)?|&(?:=|&=?|\~=?)?|>(?:>=?|=)?|<(?:<=?|=)?|:\s*=|:\s*:|\.\.[=<]?|->|[@#$]/,
	'builtin': /\b(?:abs|align_of|cap|clamp|complex|conj|expand_to_tuple|hadamard_product|imag|jmag|kmag|len|matrix_flatten|max|min|offset_of|offset_of_by_string|outer_product|quaternion|real|size_of|soa_unzip|soa_zip|swizzle|transpose|type_info_of|type_of|typeid_of|any|b(?:8|16|32|64)|bool|byte|c?string|complex(?:32|64|128)|f(?:16|32|64)(?:le|be)?|quaternion(?:64|128|256)|rawptr|rune|[ui](?:8|16|32|64|128)(?:le|be)?|u?int(?:ptr)?\b/
});
delete Prism.languages.odin['class-name'];
```

edukisto:
Resolves #3204.

1. I think, the only pattern that can be inherited from `clike` is `function`.

2. I see no point in highlighting the names of built-in procedures (`abs`, `align_of`, `u128le`, etc.), since they are ordinary shadowable identifiers.

3. The [manual](https://github.com/odin-lang/Odin/wiki/Keywords-and-Operators) explicitly defines brackets and parentheses as operators. I’ll move them to the punctuation section if you like.

4. I’ve added the following features which present both in the compiler and @gingerBill’s [post](https://github.com/PrismJS/prism/issues/3204#issue-1061211436), but remain undocumented:

    - the `asm` keyword;
    - the `0d`, `0h`, and `0z` number prefixes;
    - the `j` and `k` number suffixes.

5. No [nested multiline comments](https://odin-lang.org/docs/overview/#comments) :( See `tests/languages/odin/failed_multiline_comment_feature.test`.


components/prism-odin.js
```diff
@@ -0,0 +1,61 @@
...
+		return RegExp(pattern.replace(/<escapes>/g, escapes), flags);
+	}
+
+	Prism.languages.odin = {
+		/**
+		 * The current implementation of multiline comments doesn't support nesting.
+		 */
+		'comment': [
+			{
+				pattern: /\/\*[\s\S]*?(?:\*\/|$)/,
```
RunDevelopment, line 14: > No nested multiline comments :(

Turn the frown upside down :) Let's add support for 1 level of nested.

```suggestion
				pattern: /\/\*(?:[^/*]|\/(?!\*)|\*(?!\/)|\/\*[\s\S]*?(?:\*\/|$))*(?:\*\/|$)/,
```

components/prism-odin.js
```diff
@@ -0,0 +1,61 @@
...
+				greedy: true
+			},
+			{
+				pattern: /#![^\n\r]*/,
```
RunDevelopment, line 18: Hash bangs are only allowed as the start of the file, right?

```suggestion
				pattern: /^#!.*/,
```

components/prism-odin.js
```diff
@@ -0,0 +1,61 @@
...
+				greedy: true
+			},
+			{
+				pattern: /\/\/[^\n\r]*/,
```
RunDevelopment, line 22:
```suggestion
				pattern: /\/\/.*/,
```

components/prism-odin.js
```diff
@@ -0,0 +1,61 @@
...
+				greedy: true
+			}
+		],
+
+		/**
+		 * Should be found before strings because of '"'"- and '`'`-like sequences.
+		 */
+		'char': {
+			pattern: withEscapes(/'(?:<escapes>|[^\n\r'\\])'/.source),
```
RunDevelopment, line 31: 1. This also need to be greedy to behave correctly. You can read more about this here: #3265.
2. Handling escape sequences explicitly like this actually has some disadvantages:
    1. We need to maintain the list of all possible escape sequences.
    2. We can't highlight incomplete or just slightly wrong characters. Highlighting slightly wrong characters is useful because it makes for a better experience while typing.
    For those reason, I would suggest using a simpler pattern for escape sequences like this:

```suggestion
			pattern: /'(?:\\(?:.|[0Uux][0-9A-Fa-f]{1,6})|[^\n\r'\\])'/,
			greedy: true,
```

The idea behind `[0Uux][0-9A-Fa-f]{1,6}` is that this matches any complete and incomplete numerical escape sequence.

components/prism-odin.js
```diff
@@ -0,0 +1,61 @@
...
+			inside: {
+				'symbol': withEscapes(/<escapes>/.source)
```
RunDevelopment, line 33: Why not just do this? Same for string.

```suggestion
				'symbol': escapes
```

components/prism-odin.js
```diff
@@ -0,0 +1,61 @@
...
+			}
+		},
+
+		'string': [
+			{
+				pattern: /`[^`]*`/,
+				greedy: true
+			},
+			{
+				pattern: withEscapes(/"(?:<escapes>|[^\n\r"\\])*"/.source),
```
RunDevelopment, line 43: Same as with `char`, handling escapes explicitly has downsides.

```suggestion
				pattern: /"(?:\\.|[^\n\r"\\])*"/,
```

components/prism-odin.js
```diff
@@ -0,0 +1,61 @@
...
+			}
+		],
+
+		'number': /(?:\b0(?:b[01_]+|d[\d_]+|h_*(?:(?:(?:[\dA-Fa-f]_*){8}){1,2}|(?:[\dA-Fa-f]_*){4})|o[0-7_]+|x[\dA-F_a-f]+|z[\dAB_ab]+)\b|(?:\b\d+(?:\.\d*)?|\B\.\d+)(?:[Ee][+-]?\d*)?[ijk]?)/,
+
+		'boolean': /\b(?:_|false|nil|true)\b/,
+
+		'keyword': /\b(?:asm|auto_cast|bit_set|break|case|cast|context|continue|defer|distinct|do|dynamic|else|enum|fallthrough|for|foreign|if|import|in|map|matrix|not_in|or_else|or_return|package|proc|return|struct|switch|transmute|typeid|union|using|when|where)\b/,
+
+		'function': /\b\w+(?=\()/,
```
RunDevelopment, line 57: Can there be spaces between the function name and `(`?

tests/languages/odin/comment_feature.test
```diff
@@ -0,0 +1,90 @@
...
+/*' '*/
+/*'\a'*/
+/**/
+/*//*/
+/*0*/
+/*``*/
+/*`\a`*/
+/*false*/
+/*if*/
+Not a comment /* A comment */
```
RunDevelopment, line 17: What? Why would this not be a comment?

tests/languages/odin/comment_feature.test
```diff
@@ -0,0 +1,90 @@
...
+//""
+//#!
+//' '
+///**/
+////
+//0
+//``
+//false
+//if
+Not a comment // A comment
```
RunDevelopment, line 43: Why would this not be a comment?

components/prism-odin.js
```diff
@@ -0,0 +1,61 @@
...
+
+		'operator': /\+\+|---?|->|\.\.[<=]?|(?:&~|[-!*+/=~]|[%&<>|]{1,2})=?|[#$(),.:;?@\[\]^{}]/
```
RunDevelopment, line 59: We typically differentiate between operators and punctuation. 

Operators are e.g. addition, division, bitwise or, comparison, variable assignment, or anything else that creates/changes data. We are typically pretty broad with we allow being operators, so when in doubt, make it an operator.

Punctuation are symbols that glue together other pieces of syntax. E.g. separators, delimiters, and brackets for grouping.

---

Also, what is `#`?

```suggestion
		'operator': /\+\+|---?|->|\.\.[<=]?|(?:&~|[-!*+/=~]|[%&<>|]{1,2})=?|[#$?@^]/,
		'punctuation': /::|[(),.:;\[\]{}]/
```

components/prism-odin.js
```diff
@@ -0,0 +1,61 @@
...
+				greedy: true,
+				inside: {
+					'symbol': withEscapes(/<escapes>/.source)
+				}
+			}
+		],
+
+		'number': /(?:\b0(?:b[01_]+|d[\d_]+|h_*(?:(?:(?:[\dA-Fa-f]_*){8}){1,2}|(?:[\dA-Fa-f]_*){4})|o[0-7_]+|x[\dA-F_a-f]+|z[\dAB_ab]+)\b|(?:\b\d+(?:\.\d*)?|\B\.\d+)(?:[Ee][+-]?\d*)?[ijk]?)/,
+
+		'boolean': /\b(?:_|false|nil|true)\b/,
```
RunDevelopment, line 53: I still think that `_` should be a keyword. It's a discard operator, right?

RunDevelopment (changes requested):
That you for the PR @edukisto!

components/prism-odin.js
edukisto: I have now double-checked the work of the compiler. `#!` works exactly the same as `//`. This is implicitly [stated](https://odin-lang.org/docs/spec/#comments) in the [outdated] specification.

components/prism-odin.js
edukisto: `.` doesn’t match `\u2028` and `\u2029`. The Odin compiler assumes that `#!`- and `//`-comments end with either `\n` or `\r`, but not `\u2028` or `\u2029`.

components/prism-odin.js
edukisto: `#` starts a tag [compiler] directive. The [manual] is flooded with hashed, but you can still find `#align`, `#assert`, `#no_bounds_check`, etc. there.

tests/languages/odin/comment_feature.test
edukisto: I meant that the part **before** slashes is not a part of the comment. `"\r\nNot a comment ", ["comment", "// A comment"]` is the correct result.

tests/languages/odin/comment_feature.test
edukisto: See the next comment.

components/prism-odin.js
edukisto: Yeah, I’ll fix this pattern.

components/prism-odin.js
edukisto: So no, `#!` can be placed anywhere.

components/prism-odin.js
edukisto: I should wrap my head around this pattern :)

components/prism-odin.js
edukisto: It was a shameless and mindless copy-paste. Thank you!

components/prism-odin.js
edukisto: I personally think of operators as functions. In this sense, an operator converts values of its parameters (or operands) into a result. `[(),.:;\[\]{}]` do not produce any values, so they are punctuation marks.

If we do not take into account the [opinion](https://github.com/odin-lang/Odin/wiki/Keywords-and-Operators) of the author, then I will have to make more corrections, because

- `---` is not an operator, too. It’s an identifier for an undefined value. Can’t figure out how to classify it;

- `..`, `..<`, `..=` are punctuation marks, since they denote intervals just like a `:`;

- `->` is just a punctuation mark. It points to the type of the return value;

- `$` indicates a “constant parameter”, which makes the procedure reject non-constant arguments. This thing is a tricky one. `$` is not related to `const` in C. (In Odin you cannot assign a value to a parameter anyway. By the way, you can’t take its address either.) `$` doesn’t modify the value. It just forbids variables. Is `$` an operator or a punctuation sign?

- `@` indicates a decorator. Not an operator.

`?` is a type assertion operator. `&` is an operator, since it returns the address of the value. What do you think about `^` (pointer type)?

components/prism-odin.js
edukisto: I need a couple of days to answer.

components/prism-odin.js
RunDevelopment: > `#` starts a tag [compiler] directive. The [manual] is flooded with hashed, but you can still find `#align`, `#assert`, `#no_bounds_check`, etc. there.

In that case, we should probably have a token `'directive': { pattern: /#\w+/, alias: 'property' }`.

> * `---` is not an operator, too. It’s an identifier for an undefined value. Can’t figure out how to classify it;

This should probably be its own token. E.g. `'undefined': { pattern: /---/, alias: '<alias>' }` with an 'operator' or maybe even 'important' alias.

> * `..`, `..<`, `..=` are punctuation marks, since they denote intervals just like a `:`;

I would say that they create a new range, so they are operators.

Regarding `:`: They are used for slices and `:` is just a separator to separate the low and high bounds. So I think `:` should be punctuation.

> * `->` is just a punctuation mark. It points to the type of the return value;

Agreed.

> * `$` indicates a “constant parameter”, which makes the procedure reject non-constant arguments. This thing is a tricky one. `$` is not related to `const` in C. (In Odin you cannot assign a value to a parameter anyway. By the way, you can’t take its address either.) `$` doesn’t modify the value. It just forbids variables. Is `$` an operator or a punctuation sign?

`$` marks a "variable" as a compile-time constant, right? From what I see, they are pretty much the same as C++ templates.

Doesn't that make `$` hugely significant? Maybe we should even give it its own token with an `important` alias?

> * `@` indicates a decorator. Not an operator.

Good point, let's make it punctuation.

> `?` is a type assertion operator. `&` is an operator, since it returns the address of the value. What do you think about `^` (pointer type)?

I would say that `^` operates on a type to create a pointer type, so I would say that it's an operator.

components/prism-odin.js
RunDevelopment: No worries. Take your time. :)

components/prism-odin.js
edukisto: > `$` marks a "variable" as a compile-time constant, right?

Yes, only constants can be used as arguments.

> I would say that `^` operates on a type to create a pointer type, so I would say that it's an operator.

I completely agree with you.

> So I think `:` should be punctuation.

Yes.

> > `..`, `..<`, `..=`
> 
> I would say that they create a new range, so they are operators.

True! They create new values. How could I miss that?

components/prism-odin.js
edukisto: It is a discard, er... _thing_ :) It seems to me that `_` behaves like a _write-only variable_. Both in Odin and C# you can “assign” a value to the discard variable, but you can’t get its value, because it doesn’t have any value. The C# guide [defines](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards) “discards” as variables.

```odin
package main

import "core:fmt"

main :: proc() {
  _ = 42 // Works!
  fmt.println(_) // '_' cannot be used as a value
}
```

Note that the approach to retrieve the value of `_` is a compile-time error, but it is not named a “syntax error”. By contrast, if you use a keyword or an operator as a left-side operand in an assignment expression, Odin shows a **syntax** error message.

components/prism-odin.js
edukisto: > $ marks a "variable" as a compile-time constant, right? From what I see, they are pretty much the same as C++ templates.

I found the following snippet in the Odin overview (and fixed quite a bit):

```odin
package main

import "core:mem"

my_new :: proc($T: typeid) -> ^T {
  return (^T)(mem.alloc(size_of(T), align_of(T)))
}

main :: proc() {
  ptr := my_new(int)
}
```

This confirms that we can use `$` to mark a variable, which holds a type name. So yes, in this case the operator is similar to `typename` in C++.

components/prism-odin.js
edukisto: The redeclaration of `_` causes a compile-time error, but the error message doesn’t have a syntax error label.

```odin
package main

main :: proc() {
  _: int // No new declarations on the lhs
}
```

The declaration of variables called `+` and `return` causes a compile-time error. The error message has a syntax error label:

```odin
package main

main :: proc() {
  +: int      // Syntax Error: Expected an operand
  return: int // Syntax Error: Expected an operand
}
```

components/prism-odin.js
edukisto: Whoops! Some checks were not successful. I should test this fix locally.

edukisto: It turns out that `[\s\S]*?` causes exponential backtracking. I’ve replaced it with `(?:\*(?!\/)|[^*])*`.

components/prism-odin.js
edukisto: 1. Done. I’ve made the `char` pattern greedy.
2. Work in progress.

components/prism-odin.js
edukisto: Done. See f7569f0

components/prism-odin.js
edukisto: Done. See babdd8c

components/prism-odin.js
RunDevelopment: Oh, you're right. My bad. Sorry.

components/prism-odin.js
RunDevelopment: Hmmm, I'm still in favor of making this a keyword, or maybe an operator. It's just that `boolean` is definitely not fitting. I'm somewhat against `variable` because this token name is typically reserved for variables you can read from.

components/prism-odin.js
```diff
@@ -0,0 +1,90 @@
...
+		},
+
+		'arrow': {
+			pattern: /->/,
+			alias: 'punctuation'
+		},
+
+		'operator': /\+\+|--|\.\.[<=]?|(?:&~|[-!*+/=~]|[%&<>|]{1,2})=?|[?^]/,
+
+		'punctuation': /[(),.:;@\[\]{}]/
```
RunDevelopment, line 88: image

`::` should be one token.

```suggestion
		'punctuation': /::|[(),.:;@\[\]{}]/
```

components/prism-odin.js
```diff
@@ -0,0 +1,90 @@
...
+				}
+			}
+		],
+
+		'directive': {
+			pattern: /#\w+/,
+			alias: 'property'
+		},
+
+		'number': /(?:\b0(?:b[01_]+|d[\d_]+|h_*(?:(?:(?:[\dA-Fa-f]_*){8}){1,2}|(?:[\dA-Fa-f]_*){4})|o[0-7_]+|x[\dA-F_a-f]+|z[\dAB_ab]+)\b|(?:\b\d+(?:\.\d*)?|\B\.\d+)(?:[Ee][+-]?\d*)?[ijk]?)/,
```
RunDevelopment, line 60: image
```suggestion
		'number': /(?:\b0(?:b[01_]+|d[\d_]+|h_*(?:(?:(?:[\dA-Fa-f]_*){8}){1,2}|(?:[\dA-Fa-f]_*){4})|o[0-7_]+|x[\dA-F_a-f]+|z[\dAB_ab]+)\b|(?:\b\d+(?:\.(?!\.)\d*)?|\B\.\d+)(?:[Ee][+-]?\d*)?[ijk]?)/,
```

components/prism-odin.js
```diff
@@ -0,0 +1,90 @@
...
+
+		'boolean': /\b(?:_|false|nil|true)\b/,
+
+		'keyword': /\b(?:asm|auto_cast|bit_set|break|case|cast|context|continue|defer|distinct|do|dynamic|else|enum|fallthrough|for|foreign|if|import|in|map|matrix|not_in|or_else|or_return|package|proc|return|struct|switch|transmute|typeid|union|using|when|where)\b/,
+
+		'procedure': {
```
RunDevelopment, line 66: It would be nice if we could also find procedure definitions.

image

Could please add the following token before `keyword`?

```js
'procedure-definition': {
  pattern: /\b\w+(?=\s*::\s*proc\b)/,
  alias: ['procedure', 'function']
}
```

Also, is `\s` ok or does it have to be `[ \t]` here as well?

components/prism-odin.js
edukisto: Yes, `variable` would look especially weird because no other variables are highlighted (we can’t distinguish between variables and constants).

components/prism-odin.js
edukisto: Yes, sure! I should have done this already.

`\s` is too universal everywhere in Odin. The specification [defines](https://odin-lang.org/docs/spec/#tokens) only four whitespace characters. Moreover, the current implementation forbids newlines before the first colon for some reason.

So the most consistent implementation is

```odin
\b\w+(?=[ \t]*:[ \n\r\t]*:[ \n\r\t]*proc\b)
```

or

```odin
\b\w+(?=[ \t]*(?::[ \n\r\t]*){2}proc\b)
```

I’m not sure if this is intentional and not a bug. Maybe we should ask @gingerBill for advice.

components/prism-odin.js
edukisto: > `::` should be one token.

I don’t think so. The following is possible:

```odin
x : int : 42
```

This notation is similar to the `name : type = value` variable notation.

components/prism-odin.js
RunDevelopment: Wow. There can [spaces between colons](https://github.com/odin-lang/Odin/blob/0548db423067bce16d45af651819bf56feb5d411/src/tokenizer.cpp#L1261). I did not expect that. I take back my comment about `::` having to be one token.

The second pattern you suggested looks good :)

components/prism-odin.js
RunDevelopment: Agreed.

components/prism-odin.js
edukisto: I’ve made it a keyword. See 7a292f2.

components/prism-odin.js
edukisto: Is there a need to distinguish between declarations and calls? See cd99873, please.

And I think it is better to find procedure names **after** searching for keywords, but before Booleans (including `nil`) because the latter ones can be function names. `_` can’t. Wow.

Unfortunately, `proc` had to be moved to a separate block.

RunDevelopment: > Is there a need to distinguish between declarations and calls?

Yes, there is. As you saw yourself, the procedure definition pattern needs to be able to see the `proc` keywords, so it has to be matched before the `proc` keyword is. However, the call syntax overlaps with some keyword, so the call pattern has to be after **all** keywords.

Right now, `cross :: proc(a, b: Vector3) -> Vector3` will not be highlighted correctly. (Please add this snippet to the procedure test.)

If the procedure definition pattern is matched before all keywords and the procedure call is matched after all keywords, then everything will be highlighted correctly.

I know that it's a little weird to split the procedure patterns, but it's necessary from a technical standpoint. We also have to do this in a few other languages, so it's not like Odin will be a first in that regard :)

RunDevelopment (approved):

RunDevelopment:
Please resolve the merge conflicts, and then we can merge this. Do the following steps:

- Merge the master branch of this repo into this branch. 
- Run `npm ci && npm run build`. This will resolve the conflicts.
- Commit all changes.

This should work. Don't be surprised by any minified files changing. We recently switched to a different minifier.

RunDevelopment:
I just merged #3424, so I'll close this now.

Thank you for contributing @edukisto!

PR #3293, opened 2022-01-06
Scala: Added support for interpolated strings

### LINKED ISSUE #3284: [Feature request] Add code highlighting for String interpolation in Scala
I would like to know if it is possible to add highlighting of code that is part of Scala's string interpolation:
```
  val data = lines.map(_.toInt).sliding(2).toList
  println(s"We have ${data.size} pairs")

```
The part between `${` and `}` in a string that starts with `s"` should be highlighted as if it is actual code.

RunDevelopment:
This resolves #3284.

image


JaKXz (approved):
LGTM!

Issue #3170, opened 2021-10-28
[Tracking issue] Token types guarantee


hoonweiting:
This is a tracking issue for the second half of #2849!

Long long list of languages ahead:

- [ ] abap
- [ ] abnf
- [ ] actionscript
- [x] ada (#3381)
- [x] agda
- [x] al
- [ ] antlr4
- [ ] apacheconf
- [x] apex
- [x] apl
- [x] applescript (#3182)
- [x] aql (#3183)
- [x] arduino
- [x] arff
- [ ] asciidoc
- [x] asm6502 (#3184)
- [x] asmatmel
- [ ] aspnet
- [x] autohotkey (#3384, #3385)
- [x] autoit
- [x] avisynth
- [x] avro-idl (#3185)
- [ ] bash
- [x] basic
- [x] batch
- [x] bbcode
- [x] bicep
- [x] birb
- [ ] bison
- [ ] bnf
- [x] brainfuck
- [x] brightscript
- [x] bro (#3186)
- [ ] bsl
- [x] c (#3207)
- [ ] cfscript
- [x] chaiscript
- [ ] cil
- [x] clike
- [x] clojure (#3188)
- [ ] cmake
- [x] cobol
- [x] coffeescript
- [x] concurnas (#3189)
- [ ] coq
- [x] cpp
- [x] crystal (#3194)
- [x] csharp (#3270)
- [ ] cshtml
- [x] csp (#3276)
- [x] css
- [x] css-extras
- [x] csv
- [x] cypher (#3195)
- [x] d (#3196)
- [x] dart (#3197)
- [x] dataweave
- [ ] dax
- [ ] dhall
- [x] diff
- [ ] django
- [ ] dns-zone-file
- [x] docker
- [x] dot
- [ ] ebnf
- [x] editorconfig (#3272, #3305)
- [x] eiffel (#3198)
- [x] ejs
- [ ] elixir
- [x] elm
- [x] erb
- [x] erlang (#3382)
- [x] etlua
- [x] excel-formula (#3391)
- [ ] factor
- [x] false
- [x] firestore-security-rules
- [x] flow (#3390)
- [x] fortran
- [ ] fsharp
- [ ] ftl
- [x] gap
- [x] gcode (#3205)
- [x] gdscript
- [x] gedcom (#3386)
- [ ] gherkin
- [ ] git
- [x] glsl
- [x] gml
- [x] gn
- [x] go (#3208)
- [x] graphql
- [ ] groovy
[...]

PR #3308, opened 2022-01-24
AutoIt: Allow hyphen in directive

hoonweiting:
This PR slightly modifies the `directive` pattern in AutoIt to allow for hyphens, therefore allowing for `#include-once` (the only AutoIt directive with a hyphen in its name) to be recognised as a directive.

Reported [here](https://github.com/PrismJS/prism/issues/3170#issuecomment-1020048041)


RunDevelopment (approved):

RunDevelopment:
Thank you for the bug fix @hoonweiting!

PR #3311, opened 2022-01-27
Core: Added better error message for missing grammars

RunDevelopment:
As mentioned in #3310. The error message users get when incorrectly loading grammars is quite non-descriptive. I added a little error that says that there is no grammar for the input language.


JaKXz (approved):
👍🏽

PR #3312, opened 2022-01-28
Add span with command-line-(input|output) class to command-line plugin lines

at055612:
Currently it is not possible to distinguish between input and output, unless all input has some kind of highlighting on it.  This change wraps each line with a span and class to indicate whether it is input or output.

Reasons for needing to do this so that it is possible to:
* Style the output to look different to the input, e.g. lighter colour
* Set `user-select: none;` for output lines
* Maybe add a toggle to show/hide output lines.

For example

```html
<pre class="command-line language-bash" data-user="root" data-host="host1" data-filter-output="(out)" tabindex="0">
	<code class="language-bash">
		<span class="command-line-prompt">
			<span data-user="root" data-host="host1"></span>
			<span></span>
			<span data-user="root" data-host="host1"></span>
			<span></span>
		</span>
		<span class="command-line-input">
			<span class="token builtin class-name">echo</span>
			<span class="token string">"hello"</span> plain 
			<span class="token string">"world"</span>
		</span>
		<span class="command-line-output">hello plain world</span>
		<span class="command-line-input">
			<span class="token builtin class-name">echo</span>
			<span class="token string">"
				<span class="token variable">
					<span class="token variable">$(</span>
					<span class="token function">date</span>
					<span class="token variable">)</span>
				</span>"
			</span>
		</span>
		<span class="command-line-output">Fri 28 Jan 21:06:47 GMT 2022</span>
	</code>
</pre>
```

The above HTML was from
image


plugins/command-line/prism-command-line.js
```diff
@@ -130,9 +130,14 @@
...
-		for (var i = 0, l = outputLines.length; i < l; i++) {
+		for (var i = 0, l = codeLines.length; i < l; i++) {
+			// Add spans to allow distinction of input/output text for styling
 			if (outputLines.hasOwnProperty(i)) {
-				codeLines[i] = outputLines[i];
+				codeLines[i] = '<span class="command-line-output">'
+					+ outputLines[i] + '</span>';
+			} else {
+				codeLines[i] = '<span class="command-line-input">'
+					+ codeLines[i] + '</span>';
```
RunDevelopment, line 140: Instead of inventing new class names, we could use the token used by [Shell session](https://github.com/PrismJS/prism/blob/master/components/prism-shell-session.js).
```suggestion
				codeLines[i] = '<span class="token output">'
					+ outputLines[i] + '</span>';
			} else {
				codeLines[i] = '<span class="token command">'
					+ codeLines[i] + '</span>';
```

RunDevelopment (changes requested):
Sorry for the delay @at055612!

The PR in itself looks good. I just suggest using different CSS classes in my comment.

at055612:
Accepted your css class name change.  I have re-generated the minified js.

RunDevelopment:
Thank you for contributing @at055612!

at055612:
Thanks for merging this.

Issue #2948, opened 2021-06-13
Easy embedded and recursive languages


RunDevelopment:
**Motivation**
Right now, we use `inside` grammars for embedded languages. 

This forces us to declare every embedded language as a dependency (required or optional). This increases the number of dependencies a lot. Optional embedded languages are the primary purpose of optional dependencies.

This idea comes from Highlight.JS' `sublanguage`, as pointed out [here](https://github.com/PrismJS/prism/issues/2880#issuecomment-830634254).

**Description**
Allow `inside` to be a language id (string). `Prism.tokenize` will then resolve the language at runtime.

This trivially solves the problem of recursive languages (example [1](https://github.com/PrismJS/prism/blob/99f3ddcdc1519ebb26b1772369ea52c09d2657eb/components/prism-php.js#L241), [2](https://github.com/PrismJS/prism/blob/99f3ddcdc1519ebb26b1772369ea52c09d2657eb/components/prism-markup.js#L70)). It also simplifies the dependencies because languages can be resolved at runtime.

**Breaking Change**
The only problem is that this is a breaking change. Some plugins change grammars (although the arguably should not). These plugins have to be changed to support the new grammar format.

RunDevelopment:
Implemented in v2. Both `inside` and `rest` can now be a string that references a language.

PR #3314, opened 2022-02-01
[WIP] advanced nesting inside liquid

JaKXz:
@RunDevelopment may I get your feedback on this approach to get other languages tokenized inside liquid? I have a feeling there's a better way but I can't remember where to look to find a better example of what we're trying to achieve here 😅 

I wanted to share this with you before I continue down this path of trying to get nested `json` to work with the `schema` keyword, e.g.

```liquid
{% schema %}
  {
    "json is highlighted properly": [
      "world. cool huh?",
      true,
    ],
    "nested": {
      "object": "properties"
    },
    "invalid": type
    "missing": "comma"
  }
{% endschema %}
```

Thanks!

### DIFF
--- a/components.json
+++ b/components.json
@@ -781,7 +781,7 @@
 		},
 		"liquid": {
 			"title": "Liquid",
-			"require": "markup-templating",
+			"require": ["markup-templating", "javascript"],
 			"owner": "cinhtau"
 		},
 		"lisp": {
--- a/components/prism-liquid.js
+++ b/components/prism-liquid.js
@@ -1,7 +1,26 @@
-Prism.languages.liquid = {
+Prism.languages.liquid = {};
+Object.assign(Prism.languages.liquid, {
 	'comment': {
 		pattern: /(^\{%\s*comment\s*%\})[\s\S]+(?=\{%\s*endcomment\s*%\}$)/,
 		lookbehind: true
+	},
+	'script': {
+		pattern: /<script>[\s\S]*<\/script>|\{% javascript %\}[\s\S]*\{% endjavascript %\}/,
+		greedy: true,
+		inside: {
+			'tag': /<script>|<\/script>|\{% javascript %\}|\{% endjavascript %\}/,
+			'javascript': {
+				alias: ['js', 'javascript', 'language-javascript'],
+				pattern: /\S+(?:\s+\S+)*/,
+				inside: {
+					'liquid': {
+						pattern: /\{%.*%\}|\{\{.*\}\}/,
+						inside: Prism.languages.liquid,
+					},
+					rest: Prism.languages.javascript,
+				}
+			},
+		}
 	},
 	'delimiter': {
 		pattern: /^\{(?:\{\{|[%\{])-?|-?(?:\}\}|[%\}])\}$/,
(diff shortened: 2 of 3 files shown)


components.json
```diff
@@ -781,7 +781,7 @@
 		},
 		"liquid": {
 			"title": "Liquid",
-			"require": "markup-templating",
+			"require": ["markup-templating", "javascript"],
```
RunDevelopment, line 784: JS should probably be an optional dependency.

tests/languages/liquid/nested_feature.test
```diff
@@ -0,0 +1,151 @@
...
+	["liquid", [
+		["script", [
+			["tag", "<script>"],
+
+			["javascript", [
+				["comment", "// The drop is highlighted because liquid will render inside the comment."],
+
+				["comment", "// As such, it isn't a \"real\" comment w.r.t. liquid."],
+
+				["comment", "// comment {{ product.value }}"],
```
RunDevelopment, line 39: From the comments above, I guess that this is supposed to be a liquid section, right?

RunDevelopment (changes requested):
A general question: Are `{% javascript %}` and `{% schema %}` custom tags (or similar)? I don't see them documented anywhere

components/prism-liquid.js
```diff
@@ -1,8 +1,27 @@
-Prism.languages.liquid = {
+Prism.languages.liquid = {};
+Object.assign(Prism.languages.liquid, {
```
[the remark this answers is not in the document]
JaKXz: In the lines below there’s a spot where we’re referring to the Prism.languages.liquid object in an `inside` block. How would you suggest getting around this?

tests/languages/liquid/nested_feature.test
JaKXz: Yeah, definitely still a work in progress! Is there a better way of reusing the existing delimiters and extending markup/markup-templating?

JaKXz:
The `schema` tag is documented here: https://shopify.dev/themes/architecture/sections/section-schema

tests/languages/liquid/nested_feature.test
RunDevelopment: Unfortunately, there isn't right now. I have a PR open adding general templating functions that would allow us to correctly handle these nested cases, but that PR will have to wait for v2 and ESM.

Seems like you have to make a tradeoff here. We can either tokenize JS correctly and miss some Liquid sections, or tokenize Liquid correctly and miss some JS. I would tend to tokenizing Liquid correctly since that's the main goal here. This is also easier to do since we just have to move the `liquid` token a little:

```js
		inside: {
			'tag': /.../,
			'liquid': {
				pattern: /\{%.*%\}|\{\{.*\}\}/,
				inside: Prism.languages.liquid,
			},
			'javascript': {
				alias: ['js', 'javascript', 'language-javascript'],
				pattern: /\S+(?:\s+\S+)*/,
			},
		}
```

components/prism-liquid.js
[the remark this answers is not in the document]
mAAdhaTTah: I'm not sure how we deal with it elsewhere, but a getter would work.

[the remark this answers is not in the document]
RunDevelopment: We typically do it like [this](https://github.com/PrismJS/prism/blob/82d0ca15e85574e14d5714cdf2b989a5a9a36143/components/prism-brightscript.js#L18).

I also want to mention #2948. This would allow you to write `inside: 'liquid'`.

JaKXz:
Thanks for the feedback friends :) I am going to close this for now to reduce noise since I don't have the bandwidth atm.

PR #3320, opened 2022-02-03
added missing PureBasic keyword and fixed constants with $ at the end

HeX0R101:
Added missing keyword "ForEver" and constants with a $ at the end (e.g. #NULL$) were shown wrong.


HeX0R101:
Sorry for all those force-pushings, I didn't touch prism for a long time, seems I forgot how to do things right.
Got it now (I think)

RunDevelopment:
Thank you for the PR @HeX0R101!

Could please you also add a test for constants with a `$` at the end?

> Sorry for all those force-pushings

Don't worry about it :)

HeX0R101:
Done!

RunDevelopment (approved):

PR #3326, opened 2022-02-09
Command Line: Add support for line continuation

at055612:
**Comand line plugin**

Adds the following data attributes to allow you to show multi-line commands that have line continuation characters/strings.

* `data-continuation-prompt`
* `data-continuation-str`

This effectively changes the prompt for the continued lines as show in the updated help page for the plugin.  If `data-continuation-str` is not set then the plugin will behave as before, though maybe it should default to `\` for posix shell.

image


plugins/command-line/index.html
```diff
@@ -49,6 +49,8 @@ <h1>How to use</h1>
 	</dl>
 
 	<p>Optional: To automatically present some lines as output, you can prefix those lines with any string and specify the prefix using the <code class="language-markup">data-filter-output</code> attribute on the <code class="language-markup">&lt;pre></code> element. For example, <code class="language-markup">data-filter-output="(out)"</code> will treat lines beginning with <code class="language-markup">(out)</code> as output and remove the prefix.</p>
+
+  <p>Optional: For multi-line commands you can specify the <code class="language-markup">data-continuation-str</code> attribute on the <code class="language-markup">&lt;pre></code> element. For example, <code class="language-markup">data-continuation-str="\"</code> will treat lines ending with <code class="language-markup">\</code> as being continued on the following line. Continued lines will have a prompt as set by the attribute <code class="language-markup">data-continuation-prompt</code> or a default of <code class="language-markup">&gt;</code>.</p>
```
RunDevelopment, line 53: Let's use tabs here like everywhere else.

```suggestion
	<p>Optional: For multi-line commands you can specify the <code class="language-markup">data-continuation-str</code> attribute on the <code class="language-markup">&lt;pre></code> element. For example, <code class="language-markup">data-continuation-str="\"</code> will treat lines ending with <code class="language-markup">\</code> as being continued on the following line. Continued lines will have a prompt as set by the attribute <code class="language-markup">data-continuation-prompt</code> or a default of <code class="language-markup">&gt;</code>.</p>
```

plugins/command-line/prism-command-line.js
```diff
@@ -200,3 +212,4 @@
 	});
 
 }());
+// vim: set tabstop=2 shiftwidth=2 noexpandtab:
```
RunDevelopment, line 215: ?

plugins/command-line/prism-command-line.js
```diff
@@ -79,6 +62,21 @@
...
+
+		/** @type {int[]} */
+		var continuationLineIdxs = commandLine.continuationLineIdxs = [];
+		var lineContinuationStr = pre.getAttribute('data-continuation-str');
+
+		// Identify code lines that are a continuation line and thus don't need
+		// a prompt
+		if (lineContinuationStr && codeLines.length > 1) {
+			for (var j = 1; j < codeLines.length; j++) {
+				if (codeLines.hasOwnProperty(j - 1) && codeLines[j - 1].endsWith(lineContinuationStr)) {
```
RunDevelopment, line 74: [IE11 doesn't support `endsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith).

Also, we should probably support optional spaces after the line continuation string.

plugins/command-line/prism-command-line.js
```diff
@@ -168,15 +166,29 @@
...
+			promptLine = '<span data-user="' + user + '" data-host="' + host + '"></span>';
+		}
+
+		var continuationLineIdxs = commandLine.continuationLineIdxs || [];
+		var continuationPromptText = getAttribute('data-continuation-prompt', '>');
+		var continuationPromptLine = '<span data-continuation-prompt="' + continuationPromptText + '"></span>';
+
+		// Assemble all the appropriate prompt/continuation lines
+		for (var j = 0; j < rowCount; j++) {
+			if (continuationLineIdxs.includes(j)) {
```
RunDevelopment, line 187: IE11 doesn't support [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes). Please use `indexOf`.

plugins/command-line/prism-command-line.js
```diff
@@ -79,6 +62,21 @@
 		}
 
 		var codeLines = env.code.split('\n');
+
+		/** @type {int[]} */
+		var continuationLineIdxs = commandLine.continuationLineIdxs = [];
```
RunDevelopment, line 67: Please don't abbreviate "indexes".

Also, this array is essentially used as a set of indexes. So why not use a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)?

RunDevelopment (changes requested):
Thank you for the cool PR @at055612!

The general approach looks good to me. I just a few minor nits here and there.

plugins/command-line/index.html
at055612: Sorry about that, my editor defaults to expanding tabs.

plugins/command-line/prism-command-line.js
at055612: It's a vim modeline (https://vim.fandom.com/wiki/Modeline_magic).  I put it in while editing to ensure tab indents and was in two minds whether to take it out.  Useful for any other vim users.  Happy to remove.

plugins/command-line/prism-command-line.js
at055612: Will fix the endsWith.

Not sure I agree about supporting optional spaces.  Pretty sure in bash if you have anything between the `\` and the `\n` it is not a continuation.  May well be the same for powershell but am not a powershell user.  The aim with one of these blocks is to be able to copy/paste the content into a terminal and run it so if it contains extra spaces that break the continuation they won't be able to paste it and run it.  If someone has some other shell with a continuation that needs a space at the end then `data-continuation-str` supports that.

plugins/command-line/prism-command-line.js
at055612: Will rename.
I don't do very much js so overlooked `Set` as a thing, also partly thought for the small number of lines involved it would not be much of a hit to see if it was in an array.  Happy to fix.

at055612:
While I was in there I also tweaked the styling for the prompts and output so they use opacity instead of a colour which seems to work better with the different themes.  

image
image

I also prevented `user-select` on the shell output so you can copy just the commands which I think is what most people would want to do.  Should have done this as part of the previous PR I did, sorry.
image

plugins/command-line/prism-command-line.js
RunDevelopment: We already have a linter in place to ensure proper formatting. No need for any vim magic :)

I also want to say that I think it's a really bad idea to include editor configurations into source code files.

plugins/command-line/prism-command-line.js
RunDevelopment: I mostly wanted to help users that type in text manually (not copy'n'pasting it), but you're right. Also, if user really need this feature, then people will request it and adding it later is pretty easy.

RunDevelopment:
I agree with the color change but not with the `user-select: none`. 

People not wanting to select the output is a pretty big assumption. If they really didn't want the output to be selectable, then they could still add that themselves. However, it's a no-go for a syntax highlighter to make the highlighted non-selectable by default.

I also want to point out that the color change should have been its own PR. The change is completely independent of the line continuation feature but I can only merge it once we have line continuations ready. You can keep the color change in this PR, but, in the future, please try to make many smaller PRs instead of one big one.

plugins/command-line/prism-command-line.js
```diff
@@ -79,6 +72,22 @@
 		}
 
 		var codeLines = env.code.split('\n');
+
+		/** @type {int[]} */
+		var continuationLineIndicies = commandLine.continuationLineIndicies = new Set();
```
RunDevelopment, line 77: The linting error here is my bad. The linter is configured to only allow ES5 code, and ES5 doesn't include `Set`. I'll add an exception for stuff that IE11 supports.

RunDevelopment: I just updated the ESLint config. Just update this PR and the tests will pass.

at055612: Doh! Didn't spot that in time.  Just changed the code to use a sparse array of booleans to mark the positions.  Are you happy with that rather than reverting back to a Set?

at055612:
>I agree with the color change but not with the user-select: none.

>People not wanting to select the output is a pretty big assumption. If they really didn't want the output to be selectable, then they could still add that themselves. However, it's a no-go for a syntax highlighter to make the highlighted non-selectable by default.

I thought it might be contentious.  I will remove it and add some words to the html page saying how to do it you want to.

>I also want to point out that the color change should have been its own PR. The change is completely independent of the line continuation feature but I can only merge it once we have line continuations ready. You can keep the color change in this PR, but, in the future, please try to make many smaller PRs instead of one big one.

Yes, sorry, was a bit naughty but I figured it was a tiny change that I could sneak in.  Noted.

plugins/command-line/prism-command-line.js
at055612: Now back to a `Set`

plugins/command-line/prism-command-line.css
```diff
@@ -31,3 +32,12 @@
 .command-line-prompt > span[data-prompt]:before {
 	content: attr(data-prompt);
 }
+
+.command-line-prompt > span[data-continuation-prompt]:before {
+	content: attr(data-continuation-prompt);
+}
+
+.command-line span.token.output {
+  /* Make shell output a bit lighter to distinguish them from shell commands */
```
RunDevelopment, line 41: Please use tabs here as well.
```suggestion
	/* Make shell output a bit lighter to distinguish them from shell commands */
```

plugins/command-line/prism-command-line.js
```diff
@@ -4,6 +4,16 @@
...
 
+	// Polyfill for pre-ES6 Support, i.e. Internet Explorer
+	if (!String.prototype.endsWith) {
+		String.prototype.endsWith = function (search, this_len) {
+			if (this_len === undefined || this_len > this.length) {
+				this_len = this.length;
+			}
+			return this.substring(this_len - search.length, this_len) === search;
+		};
+	}
```
RunDevelopment, line 15: Not a good idea. People will not expect this random plugin to polyfill one specific string function. This plugin having the side effect of polyfilling *some* functions is very surprising, this is why libraries generally don't the polyfilling themselves.

Instead, just make a little function `endsWith` tries to use the built-in `endsWith` and falls back to a simple implementation if not available. Kinda like the `startsWith` function a few lines below here.

plugins/command-line/prism-command-line.css
at055612: :facepalm: Sorry.

RunDevelopment (approved):

at055612:
@RunDevelopment Thanks for the merge.  Sorry for all the back/forth.

PR #3328, opened 2022-02-10
ESLint: Allow `Map` and `Set` in ES5 code

RunDevelopment:
See [this reply](https://github.com/PrismJS/prism/pull/3326#discussion_r803569302) for context.

Briefly, IE11 partially supports `Set` and `Map`, so we should be allowed to use them throughout our code base. However, ESLint is configured to only allow ES5 code, so it rejects `Set` and `Map` as undefined variables. This PR adds an exception to our ESLint config to declare `Set` and `Map` as defined.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -131,7 +131,10 @@ module.exports = {
 				worker: true
 			},
 			globals: {
-				'Prism': true
+				'Prism': true,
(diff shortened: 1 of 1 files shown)


RunDevelopment:
Since this is a very small and required by #3326, I will merge this after all tests pass.

PR #3335, opened 2022-02-11
fix: error "Element is not defined" in SSR

WuQiuYang:
sometimes,  not work in ssr,  Element is not defined


RunDevelopment (approved):

RunDevelopment:
Wait, I take my approval back.

Why do you get this error in the first place? We already have a `typeof document === 'undefined'` check to only run those plugins if a DOM is present. Why does this check not work for you?

RunDevelopment (changes requested):
<sup>(A "request changes" comment to revoke my approval.)</sup>

WuQiuYang:
> 
For some reasons, I defined document in the node environment, so typeof document !== 'undefined'  ,  emm...

RunDevelopment:
I see. In that case, I will reject this PR. The problem is not with Prism's detection of a DOM (which clearly specifies the `document` variable), but with your code defining the `document` variable.

The easiest solution for you might be to do a little shuffle before loading Prism to ensure that the `document` variable is undefined.

```js
const oldDocument = global.document;
global.document = undefined;
require('prismjs');
global.document = oldDocument;
```

Another solution would be to not load these plugins in the first place. May I ask how you load these plugins?

PR #3341, opened 2022-02-16
Command Line: Escape markup in command line output

### LINKED ISSUE #3340: Command Line plugin not escaping output HTML
**Information:**
- Prism version: 13b56a9
- Plugins: Command Line
- Environment: Browser

**Description**
If the command output includes html reserved characters, they are not escaped and thus are treated as HTML.
This is because the command output is not subject to highlighting and thus does not get escaped by Prism.js.

I have a fix.  Will submit a PR shortly.

**Example**
If you have this HTML
```html
<pre class="command-line" data-filter-output="(out)"><code class="language-bash">
echo '&lt;span style=&quot;font-size: 2em&quot;&gt;foobar&lt;/span&gt;'
(out)&lt;span style=&quot;font-size: 2em&quot;&gt;foobar&lt;/span&gt;
</code></pre>
```
Then you get this output

image

But what we want is:

image

at055612:
Fixes #3340


RunDevelopment:
You just found and fixed a security vulnerability that has been in Prism for years. Thank you!

The bug has been introduced in #856, almost 4 years ago, and was first released [v1.14.0](https://github.com/PrismJS/prism/releases/tag/v1.14.0).

I'll release a new version of make an advisory after merging this PR.

RunDevelopment (approved):

at055612:
Happy to help.

How do the release tags relate to the js/css downloads on the prism site?  The last release was January I think but if I download the css/js from the site now it includes my recent PRs which went in since the Jan release.  Is the site based on master?

I think it may be better if the site was built from a tagged release then the version number in the downloaded css/js would always correspond to a fixed point in the code.

RunDevelopment:
> How do the release tags relate to the js/css downloads on the prism site?

Not at all. As you guess, they get everything straight from master.

> I think it may be better if the site was built from a tagged release then the version number in the downloaded css/js would always correspond to a fixed point in the code.

I agree, but that's not easy to implement on top of GitHub pages. However, we actually need this for another feature, so it is planned that you will eventually be able to select a specific version on the download page.

PR #3349, opened 2022-02-20
Added readonly keyword to PHP language

frankdejonge:
Ref: https://www.php.net/manual/en/language.oop5.properties.php#language.oop5.properties.readonly-properties


RunDevelopment:
Thank you for the PR @frankdejonge!

Please also [add the new keyword](https://prismjs.com/test-suite.html#writing-tests-update) to the [keyword test](https://github.com/PrismJS/prism/blob/master/tests/languages/php/keyword_feature.test). Also, please run `npm ci && npm run build` and commit the generated files as well.

PR #3352, opened 2022-02-22
Command line: Increased prompt opacity

RunDevelopment:
While reviewing switching between themes on the [Command line plugin page](https://prismjs.com/plugins/command-line/), I noticed that the prompt is very hard to read for some themes. This is because the prompt only has an opacity of 40%. While this may have been in line with the previous fixed color (see #3326), I think it's too transparent.

This PR changes the opacity to 70%. This ensures that the prompt is readable in all themes. (Except Funky, but Funky has problems with Command Line in general, so let's ignore it.)

---

@at055612, this might interest you.


at055612:
Agree with this change, also makes it consistent with the opacity of the output lines.

PR #3365, opened 2022-03-05
Keep Markup: Restore original nodes instead of clones

Golmote:
Up until now, the Keep Markup plugin used to clone the markup nodes. This is problematic when you have a reference to a markup node before highlighting: it would be lost after highlighting.

These changes make it so that the plugin keeps the actual original markup nodes in the highlighted output.


plugins/keep-markup/prism-keep-markup.js
```diff
@@ -38,14 +38,21 @@
 				return;
 			}
 
+			// Replace element with a clone
+			var clone = element.cloneNode(false);
+			element.parentNode.replaceChild(clone, element);
+			while (element.firstChild) {
+				clone.appendChild(element.firstChild);
+			}
```
RunDevelopment, line 46: Why is this necessary?

From what I see, the `before-highlight` hook doesn't change the current DOM tree. (At least, it doesn't in a way that really matters to Prism.) We also know that inserting the kept DOM nodes happens [after insert](https://github.com/PrismJS/prism/blob/3bd8fdb15a24a8ce901a7690256752b168dd6288/components/prism-core.js#L590). Since insert will automatically detach the nodes we indent to keep, why is the cloning here necessary? Couldn't we leave the DOM tree unchanged in this hook, assume that the nodes have been detached in `after-highlight`, and then just use the original nodes without ever cloning them?

Golmote: @RunDevelopment I did this so that data contains only detached empty nodes. Later, when we're inserting the node back into the highlighted tree, its content needs to be discarded anyway, so we can [append the extracted range](https://github.com/PrismJS/prism/blob/d50d37c4bfcb92348fe67eba8c38c036aa31c85f/plugins/keep-markup/prism-keep-markup.js#L110).

We could store directly the element with its children, and only empty it just before we append the range. But we'd store unneeded information in data.

plugins/keep-markup/prism-keep-markup.js
```diff
@@ -100,8 +107,8 @@
 						var range = document.createRange();
 						range.setStart(nodeState.nodeStart, nodeState.nodeStartPos);
 						range.setEnd(nodeState.nodeEnd, nodeState.nodeEndPos);
-						nodeState.node.clone.appendChild(range.extractContents());
-						range.insertNode(nodeState.node.clone);
+						nodeState.node.element.appendChild(range.extractContents());
```
Golmote, line 110: @RunDevelopment Here, we could empty the element before appending the range, if we were to store it with its children.
```suggestion
						nodeState.node.element.innerHTML = '';
						nodeState.node.element.appendChild(range.extractContents());
```

plugins/keep-markup/prism-keep-markup.js
RunDevelopment: > We could store directly the element with its children, and only empty it just before we append the range.

Yup, that's what I meant to suggest. This has the nice property that we don't need to do any cloning at all.

> But we'd store unneeded information in data.

Not that it's really a problem. Assuming that we keep all the markup in a code block, we would need to store all (element) nodes anyway. The only unneeded information is the tree structure between those nodes + the text, and that's just not that much in terms of memory.

Golmote:
@RunDevelopment PR updated as per your comments.

RunDevelopment (approved):
Looks great! Thanks, @Golmote!

LeaVerou (approved):
Amazing that all it took was 3 loc! ❤️

LeaVerou:
Thank you so much @Golmote!! Can't wait to use this in combination with Mavo to make live demos with code that is editable through selects, ranges etc without having to employ crazy hacks (like doing the highlighting myself manually)!

Tuneabq (commented):
Yee haw

PR #3367, opened 2022-03-05
Core: Added support for language ids in `inside`

RunDevelopment:
This implements the addition to core described in #2948.

The basic idea is that the `inside` property of grammar tokens should allow language ids (strings) to make it easy to embed other languages and to create recursive languages. This is a relatively common problem and the solutions we have right now aren't pretty.

This is a breaking change because this PR changes the type definition of a grammar. Code relying on grammars having this specific shape will break.

---

I did not change any existing languages to use the new feature in this PR. We can still do this after merging this one.


JaKXz (commented):
Sorry I missed this! Looks great, but I would want to see a minimal example in a test ideally.

RunDevelopment:
> but I would want to see a minimal example in a test ideally.

Good point. I added a test case that checks the behavior of a minimal grammar.

JaKXz (approved):
🔥

JaKXz:
@RunDevelopment anything preventing this from being merged? I think it'd be a cool release for v1.29 :)

JaKXz:
@RunDevelopment ignore my previous comment - I misread `non-breaking change` instead of `breaking-change`.

Should we start a separate branch for v2.0 and start to get these merged? :)

RunDevelopment:
Implemented in v2.

PR #3377, opened 2022-03-09
properties: Use `key`, `value` for token names, `attr-...` as aliases

hoonweiting:
This PR replaces `attr-name` and `attr-value` with `key` and `value`, and uses `attr-...` as aliases instead.

Overall, this brings properties in line with similar languages (INI, editorconfig, systemd...).

### DIFF
--- a/components/prism-properties.js
+++ b/components/prism-properties.js
@@ -1,9 +1,13 @@
 Prism.languages.properties = {
 	'comment': /^[ \t]*[#!].*$/m,
-	'attr-value': {
+	'value': {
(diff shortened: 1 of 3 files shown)


RunDevelopment (approved):

RunDevelopment:
Thank you for the PR @hoonweiting!

I think that was the last config-type language with that problem, right?

hoonweiting:
I'm not really sure to be honest! I was just looking at the list of languages that haven't been completed, so I might have missed out on some config-type languages.

I'm looking at another language right now, but I'll go through the list again another day to check if there are more config-type languages! :)

RunDevelopment:
Thank you!

PR #3381, opened 2022-03-12
Ada: Change `attr-name` to `attribute`; Use `attr-name` as alias

hoonweiting:
From the small discussion [here](https://github.com/PrismJS/prism/issues/3170#issuecomment-1065859554)


RunDevelopment (approved):

RunDevelopment:
Thank you!

PR #3382, opened 2022-03-12
Erlang: Alias `quoted-atom` and `atom` to `symbol`

hoonweiting:
In [Erlang](https://www.erlang.org/doc/reference_manual/data_types.html#atom),

> An atom is a literal, a constant with name.

which matches the standard token `symbol`. This PR adds `symbol` as an alias to `quoted-atom` and `atom`.


RunDevelopment:
I don't think this is a good idea. While atoms are indeed close to symbols, they are also very common, like, almost everything is a symbol kinda common. If we added this alias, then all symbols would get a color which would leave no character un-highlighted. And if everything is highlighted, nothing is. [Example](https://prismjs.com/test.html#language=erlang&text=%25%20https%3A%2F%2Ferlangbyexample.org%2Fpattern-matching%0A%0Arun()%20-%3E%0A%20%20%20%20io%3Aformat(%22~p~n%22%2C%20%5Bpromote(%7Bceo%2C%20female%7D)%5D)%2C%0A%20%20%20%20io%3Aformat(%22~p~n%22%2C%20%5Bpromote(%7Bceo%2C%20male%7D)%5D)%2C%0A%20%20%20%20io%3Aformat(%22~p~n%22%2C%20%5Bpromote(%7Bceo%2C%20other%7D)%5D)%2C%0A%0A%20%20%20%20io%3Aformat(%22~p~n%22%2C%20%5Bdemote(%7Bceo%2C%20kid%7D)%5D)%2C%0A%0A%20%20%20%20Pid%20%3D%20spawn(%3FMODULE%2C%20postman%2C%20%5B%5D)%2C%0A%20%20%20%20Pid%20!%20%7Bsend%2C%20%7B%22Joe%22%2C%20%22Jane%22%2C%20%22cool%20gift%22%7D%7D%2C%0A%0A%20%20%20%20timer%3Asleep(100)%2C%0A%20%20%20%20Pid%20!%20stop%2C%0A%0A%20%20%20%20is_even(3)%2C%0A%20%20%20%20head(%5B101%2C%20201%2C%20301%5D).).

Also, aren't atoms closer to identifiers when compared to a C-like language?

hoonweiting:
@RunDevelopment That's a really good point, and I agree with where you're coming from!

I suppose then it depends on whether it is more important to guarantee the token types, or to maintain a reasonable amount of highlighting. I think the latter would be preferable for users, so I guess I would be in favour of closing this PR.

> Also, aren't atoms closer to identifiers when compared to a C-like language?

Without having studied computer science, I don't think I can answer this satisfactorily 😅 I'll give it a go anyway, and I'll be happy to be corrected.
I gave it a quick google, and here's the first sentence of the [C Identifiers article in Microsoft docs](https://docs.microsoft.com/en-us/cpp/c-language/c-identifiers?view=msvc-170):

> "Identifiers" or "symbols" are the names you supply for variables, types, functions, and labels in your program.

Here's the first two sentences from [Erlang's Types Specifications](https://www.erlang.org/doc/reference_manual/typespec.html#types-and-their-syntax):

> Types describe sets of Erlang terms. Types consist of, and are built from, a set of predefined types, for example, integer(), atom(), and pid().

From only these two sources, it appears to me that identifiers in C aren't the same as atoms in Erlang. I would think identifiers encompass atoms and other types. On the other hand, a symbol is a primitive data type found in some languages (which can be used as an identifier), so I would also think that identifiers can encompass symbols (this is getting confusing!).

However, I acknowledge that you mentioned C-like languages and not specifically C, so I'm not sure if I've made a very accurate assessment here.

RunDevelopment:
Oh, I completely overlooked that atoms are a type of data in Erlang. I should have looked at the title of the page you linked... My bad. You're right that atoms are like the symbols of many functional languages, and that they fit the definition of symbol we use for standard tokens.

However, I would still favor closing this PR, so we don't overload the highlighting. Sorry.

---

Also, you can still tick Erlang off the list in #3170. After all, using only non-standard tokens is perfectly compliant with our definition of standard token :)

hoonweiting:
Thank you for the discussion @RunDevelopment! :)

PR #3385, opened 2022-03-12
Autohotkey: Use standard tokens

RunDevelopment:
As suggested [here](https://github.com/PrismJS/prism/issues/3170#issuecomment-1065946061).


hoonweiting (commented):
These updated tokens are great! I'm just not very sure about `builtin-function`, since there is already a `function` for custom functions. Also, I think `builtin` is fine as it is, as we define `builtin` to be

> Functions/Methods/Classes/Types that are available out of the box.

What do you think?

RunDevelopment:
Good point.

hoonweiting (approved):
Looks great!

PR #3387, opened 2022-03-13
Erlang: Add `begin` keyword

hoonweiting:
Adding the `begin` keyword in [Erlang](https://www.erlang.org/doc/reference_manual/introduction.html#reserved-words).

`cond` and `let` are also reserved words in Erlang, but they're not currently being used. I'm not sure if they should be added, I think not, but I'm not sure!


RunDevelopment (approved):

RunDevelopment:
Thank you for the PR!

> `cond` and `let` are also reserved words in Erlang, but they're not currently being used.

Then we'll just wait. Languages typically have a few reserved words for future use. Frankly, it's up to us whether we want to highlight them. Since they are reserved, any usage of those unused reserved keywords is invalid syntax, and Prism is allowed to highlight invalid syntax as whatever we want.

That being said, I typically opt not to include revered-but-not-used keywords, simply because they (slightly) increase the file size of our languages.

PR #3389, opened 2022-03-13
VHDL: Add `private`, `view` keywords; Distinguish `attribute` from `keyword`

hoonweiting:
I'm not familiar with the language, but based on the file history, I presume that any changes to the language after 2015 are not reflected. There's been at least one new version since then ([VHDL-2019](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/VHDL2017), some [highlights](https://vhdlwhiz.com/vhdl-2019/) summarised on another site), and brings at least two new keywords: `private` and `view`. I'd confirm with the language reference manual but it costs like $500 and it's not quite worth the cost for what I'll use it for :P

I've also separated the attributes from the keywords, and used a generic pattern instead of a name list for the attributes (similar to Ada).


RunDevelopment:
The coverage test failing isn't your fault. 

That test uses a few different strategies to determine whether a pattern has been covered. The most basic strategy is to simply check that a regex produced some token. However, it uses a different strategy for keyword lists. It requires all keyword in the list to procure a token.

The test now fails because you simplified the `keyword` enough for the coverage test to recognize it as a keyword list, and now it wants you to test all of them.

I'm sorry about that, but could you please add the missing keywords to the test?

> but it costs like $500

lol. Yeah, that's not quite worth it.

> I've also separated the attributes from the keywords, and used a generic pattern instead of a name list for the attributes

Sounds good.

hoonweiting:
Ooh that's interesting! I'll update the tests shortly. My apologies for not running all the tests locally!

hoonweiting:
Ah hmm... I did not realise that `library` and `use` are also contained the in keyword list, I thought they were only considered `constant`s.

I'll update the keyword test, but I'm not sure whether to remove `constant`, or allow `library` and `use` to be assigned both `constant` and `keyword`, as they are both standard tokens. I'm leaning towards removing `constant`, but I wonder why those two are constants as well. I might have to look for whatever VHDL docs I can find about it.

hoonweiting:
@RunDevelopment I see! Thanks for searching that up :)

I ran the test on my laptop, it timed out the first time, but all of it passed after plugging my laptop in.

RunDevelopment (approved):

RunDevelopment:
Thank you!

> I ran the test on my laptop, it timed out the first time

Do you remember which one timed out?

hoonweiting:
Ah no, not really. I just tried to run it again after unplugging it, but it hasn't failed again yet.

I tried to google some of the words I remember seeing in the error, [this question on StackOverflow](https://stackoverflow.com/questions/53258510/why-am-i-keep-getting-error-timeout-of-2000ms-exceeded-from-mocha-on-async) seems to match it.

As for the file, I believe it was `coverage.js`, which makes sense because I only ran `npm run regex-coverage`.

RunDevelopment:
A timeout on coverage? Huh, that's strange. The test just runs all languages tests, so it doesn't do any heavy computations. Anyway, I'll keep that in mind for the future.

PR #3391, opened 2022-03-13
Excel: Use more fitting aliases for `function-name`, `range`, `cell`

hoonweiting:
`function-name`: Changed alias from `keyword` to `builtin`

`range` and `cell`: Changed alias from `property` to `selector`


RunDevelopment (approved):

RunDevelopment:
Thank you!

PR #3394, opened 2022-03-14
NEON: Change alias of `key` to `property`

hoonweiting:
NEON is a superset of JSON, so it makes sense for the alias of `key` to be `property`, like in JSON.


RunDevelopment (approved):

RunDevelopment:
Thank you @hoonweiting!

PR #3403, opened 2022-03-19
Added support for Arturo

drkameleon:
➕ Support for the [Arturo programming language](https://arturo-lang.io/) 

Screenshot 2022-03-19 at 12 31 06


components/prism-arturo.js
```diff
@@ -0,0 +1,50 @@
...
+        alias: 'keyword'
+    },
+
+    'sugar': {
+        pattern: /\->|=>|\||\:\:/,
+        alias: 'important'
+    },
+
+    'symbol': {
```
RunDevelopment, line 45: Prism has [standard tokens](https://prismjs.com/tokens.html) with specific semantic meanings. Standard tokens ensure that your language will look great with all of Prism's themes, so please use them.

Also, please try to avoid using two or more standard tokens at the same time. The `string` rule has a `function` alias. Please remove the `function` alias.

components/prism-arturo.js
```diff
@@ -0,0 +1,50 @@
...
+        pattern: /<\:|\-\:|ø|@|#|\+|\||\*|\$|\-|\%|\/|\.\.|\^|~|=|<|>|\\|\-\-\-/
```
RunDevelopment, line 46: ESLint gives us an error here because the `\-\-\-` alternative at the end is effective ignored because of the `\-` alternative. Try highlighting the text `---`. You'll see that it gets tokenized as 3 times `-`.

To fix this, just move the `---` alternative before the `-` alternative.

Also, please only escape punctuation if necessary. Unnecessary escaping just introduces noise that make regexes a little harder to read.
```suggestion
        pattern: /<:|-:|ø|@|#|\+|\||\*|\$|---|-|%|\/|\.\.|\^|~|=|<|>|\\/
```

RunDevelopment (changes requested):
Thank you for the PR @drkameleon!

I gave your PR a quick look and left you some comments. I'll re-review it again after the CI passes.

The CI currently fails because of the linter and coverage tests. 

Just run `npm run lint:fix` to fix all the indentation problems. Our ESLint config contains some very computationally intensive rules, so this might take a minute or two (yes, really). The error with the `symbol` regex is a bug, however. Please see my comment on this one.

To appease the coverage test, you just has to add a few more tests. To see what regexes still need tests, you can see the CI logs or run `npm run regex-coverage` locally. The coverage test requires that all keywords are tested, so please add all of them to the test files. You don't need a valid code snippet for each keyword. It's fine to just have them all as a list ([example](https://github.com/PrismJS/prism/blob/master/tests/languages/al/keyword_feature.test)).

examples/prism-arturo.html
```diff
@@ -0,0 +1,421 @@
+<h2>Full example</h2>
+<pre><code>; this is a comment
+    ; this is another comment
```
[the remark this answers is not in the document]
drkameleon: Nope! Good catch! :)

I guess I missed it while copy-pasting stuff around...

components/prism-arturo.js
drkameleon: I know, it looks messy. The result is like it is simply because I've used pretty-much the same syntax/patterns for different highlighters and I guess some of the syntax is a leftover of what other syntaxes expected.

Fixed :)

components/prism-arturo.js
drkameleon: I see. Admittedly it looks confusing, although the very reason for that is that I was - desperate - trying to enforce a specific look. But I do get your point.

I've changed a few things to *roughly* abide by the standard tokens you pointed me, though it's quite challenging to always find a 1-to-1 correspondence (Arturo is a REBOL/Red type of language, and unfortunately there is no such highlighter there already to get an implementation idea).

In any case, please let me know in case you think anything else is amiss. 😉

components/prism-arturo.js
```diff
@@ -0,0 +1,49 @@
...
+        alias: 'boolean'
+    },
+
+    'type': {
+        pattern: /:(?:[\w]+\b\??:?)/,
+        alias: 'boolean'
+    },
+
+    'keyword1': {
+        pattern: /\b(?:all|and|any|ascii|attr|attribute|attributeLabel|binary|block|char|contains|database|date|dictionary|empty|equal|even|every|exists|false|floating|function|greater|greaterOrEqual|if|in|inline|integer|is|key|label|leap|less|lessOrEqual|literal|logical|lower|nand|negative|nor|not|notEqual|null|numeric|odd|or|path|pathLabel|positive|prefix|prime|regex|same|set|some|sorted|standalone|string|subset|suffix|superset|symbol|symbolLiteral|true|try|type|unless|upper|when|whitespace|word|xnor|xor|zero)\?/,
```
RunDevelopment: Could we please have a dedicated `boolean` token for `false|maybe|true`?

components/prism-arturo.js
```diff
@@ -0,0 +1,49 @@
+Prism.languages.arturo = {
+    'comment': /;.*/,
+
+    'character': {
+        pattern: /`.`/,
```
RunDevelopment: Please make `comment` and `character` greedy as well. Right now, `';'` is highlighted as a comment.

RunDevelopment (changes requested):
I'm currently reading through the docs, but here are 2 things I found. Also, could we please have a `number` rule to highlight integers, floats, and versions?

components/prism-arturo.js
```diff
@@ -0,0 +1,49 @@
+Prism.languages.arturo = {
+	'comment': /;.*/,
+
+	'character': {
+		pattern: /`.`/,
+		alias: 'char'
+	},
+
+	'string': {
+		pattern: /"(?:[^"\\]|\\.)*"/,
```
RunDevelopment, line 10: This currently allows strings like
```
"foo
bar"
```
which seems to be incorrect.

Did you mean `/"(?:[^"\\\r\n]|\\.)*"/`?

RunDevelopment: On the topic of strings. How about supporting the other kinds of strings as well?

```js
'string': {
  pattern: /"..."|».*|\{:[\s\S]*?:\}|\{[\s\S]*?\}|^-{6}$[\s\S]*/m,
  greedy: true
},
```

RunDevelopment:
Could we also add support for regexes?

```js
'regex': { pattern: /\{\/.*?\/\}/, greedy: true }
```

drkameleon:
> I'm currently reading through the docs, but here are 2 things I found. Also, could we please have a `number` rule to highlight integers, floats, (and versions maybe?)?

Number rule added :wink

components/prism-arturo.js
drkameleon: You are right. I fixed it.

Regarding curly strings, etc I had thought of leaving it for some other day, but I guess your suggestion works beautifully. Thanks! 😄

drkameleon:
> Could we also add support for regexes?
> 
> ```js
> 'regex': { pattern: /\{\/.*?\/\}/, greedy: true }
> ```

Awesome! Done! 😉

components/prism-arturo.js
drkameleon: Done 😉

components/prism-arturo.js
drkameleon: Done 😉

drkameleon:
Since we're working on it, I have one question:

Arturo also support some special "switches" for multi-line curly-bracket strings, simply in order to enforce a specific inline syntax.

Let me give you an example:

```red
str: {!html
    <h1>This is some code that is to be highlighted as HTML
}
```

For the VSCode extension, I'm currently supporting `!html`, `!css`, `!js` and `!sql`, where we "include" one of the aforementioned styles in case the switch exists.

I thought that would be quite complicated, but I may be wrong.

Is it feasible with Prisma?

RunDevelopment:
> the very reason for that is that I was - desperate - trying to enforce a specific look.

Understandable. However, that's not the best way to do it. Language definitions are supposed to just tokenize code. Getting the look just right is the job of the theme you're using.

There are 2 ways to get fine-grained control, and both involve modifying the theme you're using.

1. Using language-specific CSS rules. 
   You can change the highlighting of token for a specific language by using a CSS override. E.g.
    ```css
    /* Strings get the color green. */
    .token.string { color: green; }

    /* Except in Arturo! There they are red. */
    .language-arturo .token.string { color: red; }
    ```
1. Using non-standard aliases.
    Since we control the language definition, we can also add specific aliases. Let's say you wanted to differentiate floats and integers from versions. You could do that like this:
    ```js
    'number': [
        {
            pattern: /\b\d+\.\d+\.\d+\b/,
            alias: 'version'
        },
        /\b\d+(?:\.\d+)?\b/
    ]
    ```
    This allows you to give versions special highlighting on your website, and all themes don't specifically support version will highlight them as numbers. It's essentially granular highlighting + fallback.

drkameleon:
I totally get your point. As a matter of fact, the way I've been doing it so far simply... doesn't look right.

As you have had a look at the official website, you must have already noticed that I *am* using Prism already. And obviously using all these weird tricks for token-color matching.

I guess once this PR is merged, I'll use this version for the highlighter (not my custom one; that is the point behind all this after all), doing it via CSS for the overrides, as you yourself suggest! :)

RunDevelopment:
> Arturo also supports some special "switches" for multi-line curly-bracket strings

Definitely possible. We'll have to deal with [optional dependencies](https://prismjs.com/extending.html#dependencies), but the rest is pretty straightforward.

```js
/**
 * @param {string} lang
 * @param {string} pattern
 */
function createLanguageString(lang, pattern) {
	return {
		pattern: RegExp(/\{!/.source + '(?:' + (pattern || lang)+ ')' + /$[\s\S]*\}/.source, 'm'),
		greedy: true,
		inside: {
			'embedded': {
				pattern: /(^\{!\w+\b)[\s\S]+(?=\}$)/,
				lookbehind: true,
				alias: 'language-' + lang,
				inside: Prism.languages[lang]
			},
			'string': /[\s\S]+/
		}
	}
}

// later...
'html-string': createLanguageString('html'),
'css-string': createLanguageString('css'),
// and so on...
'string': ...
```

And the optional dependencies in `components.json`:

```json
		"arturo": {
			"title": "Arturo",
			"alias": "art",
			"optional": ["markup", "css", ...]
			"owner": "drkameleon"
		},
```

components/prism-arturo.js
```diff
@@ -0,0 +1,72 @@
...
+		greedy: true
+	},
+
+	'regex': {
+		pattern: /\{\/.*?\/\}/,
+		greedy: true
+	},
+
+	'multistring': {
+		pattern: /"..."|».*|\{:[\s\S]*?:\}|\{[\s\S]*?\}|^-{6}$[\s\S]*/m,
```
RunDevelopment, line 24: Sorry, with `"..."` I meant the single-line strings. I only used three dots because I wasn't sure about the correct pattern at the time.
```suggestion
		pattern: /».*|\{:[\s\S]*?:\}|\{[\s\S]*?\}|^-{6}$[\s\S]*/m,
```

components/prism-arturo.js
```diff
@@ -0,0 +1,72 @@
...
+		pattern: /'(?:\w+\b\??:?)/,
+		alias: 'boolean'
+	},
+
+	'type': {
+		pattern: /:(?:\w+\b\??:?)/,
+		alias: 'boolean'
+	},
+
+	'keyword1': {
```
RunDevelopment, line 44: Right, about keyword 1 and 2. Could we find better names here? You mentioned in the docs that there are no keyword (in the sense of reserved identifiers), so what are those?

Keyword 2 seems to be built-in functions?

tests/languages/arturo/symbol_feature.test
```diff
@@ -0,0 +1,13 @@
+"hello" = " world"
```
RunDevelopment, line 1: Could we please also test all other symbols defined by the regex? This would have caught the `---` bug.

components/prism-arturo.js
drkameleon: Haha. I have to admit it totally confused me. And the absurd reason is: Arturo *did* use `...` as string-block delimiters at some point in the past and I was left like "ok, I forgot about this... I thought I had removed it!" lol

components/prism-arturo.js
drkameleon: basically, keyword1 are *predicate* built-in functions (all of them end with a `?` and return a logical value) and keyword2 are normal built-in functions

components/prism-arturo.js
```diff
@@ -0,0 +1,72 @@
...
+	'symbol': {
+		pattern: /<:|-:|ø|@|#|\+|\||\*|\$|---|-|%|\/|\.\.|\^|~|=|<|>|\\/
+	},
+
+	'boolean': {
+		pattern: /\b(?:false|maybe|true)\b/
+	},
+
+	'number': {
+		pattern: /\b\d+(?:\.\d+(?:\.\d+)?)?\b/,
```
RunDevelopment, line 68: Could we support pre-release versions as well? E.g. `1.2.3-rc1+build123`
```suggestion
		pattern: /\b\d+(?:\.\d+(?:\.\d+(?:-[\w+-]+)?)?)?\b/,
```

components/prism-arturo.js
RunDevelopment: lol That the first time I guess syntax correctly.

components/prism-arturo.js
RunDevelopment: Then could we call them `predicate` and `builtin-function`?

RunDevelopment:
Could we please also add support for punctuation? The theme you use doesn't highlight punctuation but many others do, so it would be nice to include it.

```js
'punctuation': /[()[\]:.#]/
```

I probably missed some in the regex.

---

Could we also add a pattern for colors? We have [a plugin that displays colors inline](https://prismjs.com/plugins/inline-color/), so it would be nice if it could support Arturo.

```js
'color': /#\w+/
```

(I'm not too sure about the specific syntax of colors, so I just went with "any word". I hope that's okay.)

components/prism-arturo.js
drkameleon: Done ;)

tests/languages/arturo/symbol_feature.test
drkameleon: Done ;) (hopefully...)

components/prism-arturo.js
drkameleon: Done ;)

components/prism-arturo.js
```diff
@@ -0,0 +1,99 @@
+/**
+ * @param {string} lang
+ * @param {string} pattern
+ */
+ function createLanguageString(lang, pattern) {
```
RunDevelopment, line 5: Right, I forgot to mention something. People typically import these language files directly, so this function is leaked into global scope. Please wrap the whole file into an IIFE. [Like this](https://github.com/PrismJS/prism/blob/d17e37629a9daa31d4a20ab4a282f54e42f8869b/components/prism-java.js#L1).

tests/languages/arturo/html-string_feature.test
```diff
@@ -0,0 +1,18 @@
+{!html
+    <h1>Header</h1>
+}
+
+----------------------------------------------------
+
+[
+    ["html-string", [
+            ["string", "{!html"],
+            ["embedded", "\r\n    <h1>Header</h1>\r\n"],
```
RunDevelopment, line 10: To properly test the HTML highlighting, do this:

1. Declare the optional dependencies in `components.json`. This is necessary for embedded languages to work in general.
2. Create a new folder `tests/languages/arturo!+markup` and move this file into it.

The embedded HTML will then be tokenized in the tests.

Please do this for the other languages as well.

drkameleon:
@RunDevelopment Apologies for the abrupt break but since it's weekend, I decided to take some time off.

I'll get back to working on the PR by tomorrow morning so that we can finally complete it; cleaning up the code/tests a bit and taking into account ofc all of your super-valuable suggestions.

Thanks a lot - I appreciate your assistance more than 100%! 😉

tests/languages/arturo/html-string_feature.test
drkameleon: All set :)

components/prism-arturo.js
drkameleon: Done 😉

drkameleon:
> Could we please also add support for punctuation? The theme you use doesn't highlight punctuation but many others do, so it would be nice to include it.
> 
> ```js
> 'punctuation': /[()[\]:.,#]/
> ```
> 
> I probably missed some in the regex.
> 
> Could we also add a pattern for colors? We have [a plugin that displays colors inline](https://prismjs.com/plugins/inline-color/), so it would be nice if it could support Arturo.
> 
> ```js
> 'color': /#\w+/
> ```
> 
> (I'm not too sure about the specific syntax of colors, so I just went with "any word". I hope that's okay.)

I did both 😉 

In the case of **punctuation**, I just removed `#` and `.`, since they are use for different things (the first one as a *symbol*, alias to the function `dictionary` and the second one is syntax for "attributes")

In the case of color, I think your regex will work fine. Actually the rules are quite complicated. What is recognized as a `:color` value:

- 3-digit RGB hex (e.g. `#F00`)
- 6-digit RGB hex (e.g. `#FF0000`)
- 8-digit RBBA hex (e.g. `#FF000088`)
- custom W3C-compliant color names (e.g. `#red` or `#darkOliveGreen`)

So, as a better approach without over-complicating things, I think your suggestion is fine. :)

tests/languages/arturo!+bash/sh-string_feature.test
```diff
@@ -0,0 +1,18 @@
+{!sh
+    ls -la test/*x
+}
+
+----------------------------------------------------
+
+[
+    ["sh-string", [
+            ["string", "{!sh"],
+            ["embedded", "\r\n    ls -la test/*x\r\n"],
```
RunDevelopment, line 10: Bash is supposed to highlight `ls`. This is because our Bash grammar doesn't have an `sh` alias (as I just found out), but it does have [`shell`](https://github.com/PrismJS/prism/blob/0d49553c6b8ce39ddb5c2d9ca6160ed2c263902c/components/prism-bash.js#L226), so please change the language definition to use `createLanguageString('shell')` instead.

RunDevelopment (changes requested):
Final round. I think that language definition is pretty much gtg right now.

Please fix the merge conflict. This will be a little tricky in `components.json` (because you have to do it manually), but after you resolved that file, you can just run `npm run build` and all other conflicts will be resolved automagically.

tests/languages/arturo!+bash/sh-string_feature.test
drkameleon: Done 😉

drkameleon:
> Final round. I think that language definition is pretty much gtg right now.
> 
> Please fix the merge conflict. This will be a little tricky in `components.json` (because you have to do it manually), but after you resolved that file, you can just run `npm run build` and all other conflicts will be resolved automagically.

I *think* I made it (I hope I didn't manage to break everything - Git is not my strongest point lol 🙄 )

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @drkameleon!

drkameleon:
> Thank you for contributing @drkameleon!

Thank *you* for the extremely helpful input and all the assistance!

You are a top coder - and I would be more than happy to see you contribute to Arturo too (in case you have time and wish, that is)

Have a great day! 😉

PR #3404, opened 2022-03-21
npm: Upgrade to lockfile version 2

RunDevelopment:
This changes our `package-lock.json` to you [the lockfile v2 format](https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile). This format is the default format for npm@7 and is supported by npm@6.

The reason for this PR is that whenever I change some of our deps, I have to manually install npm@6 as to not upgrade the lockfile. This is super inconvenient.


RunDevelopment:
I'm going to merge this now because I want to update and change a few dependencies. From a bit of googling, it seems that adopting lockfile v2 is fairly noncontroversial, so I'll assume that nobody will be against this.

PR #3407, opened 2022-03-21
Use terser

RunDevelopment:
This PR replaces UglifyJS 3 with Terser (5.x).

Since we want to transition to ES6 in v2, we'll need a minifier that supports ES6. We currently use [UglifyJS 3](https://github.com/mishoo/UglifyJS), but that is a pure ES5 minifier. Since transpiling our code into ES5 would add complexity and result in worse file sizes, I think the easiest solution is to just minify our ES6(+?) code.


RunDevelopment:
I had to disable unsafe transformations again because of a bug in the v8 version shipped in NodeJS 10.x. I even reported and helped fix this bug in terser a while ago, but it seems that there was a regression. It's a shame, but it's not worth the hassle now. Since this is a NodeJS 10.x specific bug, we might as well enable the transformation after dropping support for that version of Node.

JaKXz (approved):

Issue #2715, opened 2021-01-17
Rewrite prism in esm


hyrious:
**Motivation**
Many modules are supporting esm now, it is better for bundlers to do tree-shaking and do care node.js and browser environment with separated bundle.js.

**Description**
Bundlers like rollup or webpack want to see esm modules and they do tree-shaking for smaller size. Currently you're using cjs and supporting both node.js and browser by hand (instead of umd). This is bad bacause it exports some global variables like `_self` in browser and you won't be able to use it in the native esm way because `import`s order is not guaranteed. See [vite#1438](https://github.com/vitejs/vite/issues/1438) since vite depends on native esm in development mode. [This solution is also wrong.](https://github.com/PrismJS/prism/issues/1394#issuecomment-720857706)

As a result, your package.json should look like this:

```js
{
    // node's require and import, to override it, see "exports"
    "main": "dist/index.js",
    // https://nodejs.org/api/packages.html#packages_conditional_exports
    "exports": {
        "node": {
            "import": "./src/index.mjs",
            "require": "./dist/index.js"
        },
        "default": "./src/index.browser.mjs"
    },
    // front-end bundler read this (like vite or webpack)
    "module": "src/index.browser.mjs"
}
```

joshgoebel:
Is the `import` here just a small wrapper around the existing require CJS version then? Or ...?

hyrious:
> Is the `import` here just a small wrapper around the existing require CJS version then? Or ...?

Well, you don't have to write the wrapper by yourself. Rollup, esbuild and webpack both handle (most of the) cjs libraries (including [react](https://reactjs.org/)) well.

[...]

PR #3408, opened 2022-03-21
Drop support for NodeJS 10.x in development

RunDevelopment:
NodeJS 10.x lost support on [30 Apr 2021](https://endoflife.date/nodejs).

This PR removes NodeJS 10.x from your CI and changes the development requirements in the readme.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x, 16.x]
+        node-version: [12.x, 14.x, 16.x]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


mAAdhaTTah:
I feel like I said this when we dropped Node 8 support, but technically this is a breaking change.

JaKXz:
On a related /tangential/hopefully not too controversial note, can I bring up the possibility of moving to ESM for v2.0 as well? We may already have an issue for this but I'm not sure 😅

RunDevelopment:
> We may already have an issue for this

Yup, we do :) #2715. Moving to ESM is very much planned for v2.

JaKXz:
@RunDevelopment is this already in the v2 branch? If not you could change the base branch to make the change easy :)

RunDevelopment:
Yes, this is already implemented in v2. I'll close this now.

PR #3421, opened 2022-04-01
PHP: added 'never' return type + minor fix of named arguments

TomPavelec:


RunDevelopment:
Thanks for contributing @TomPavelec!

PR #3423, opened 2022-04-05
Copy-to-Clipboard: Add a text-override attribute

Scy147:
This PR adds a new data-attribute that allows overriding the text that is to be copied to the clipboard. This allows hiding credentials in the visible UI, but having them copied correctly.

An example of this could be a cURL command setting an API-Key as a Header, with this change the UI-visible Text could show `X-API-Key: XXXXXX` instead of the real APi-Key, but still copy the correct cURL command.

### DIFF
--- a/plugins/copy-to-clipboard/index.html
+++ b/plugins/copy-to-clipboard/index.html
@@ -34,6 +34,7 @@
 		<li><code class="token attr-name">data-prismjs-copy-error</code> — a message displayed after failing copying, prompting the user to press <code>Ctrl+C</code>;</li>
 		<li><code class="token attr-name">data-prismjs-copy-success</code> — a message displayed after a successful copying;</li>
 		<li><code class="token attr-name">data-prismjs-copy-timeout</code> — a timeout (in milliseconds) after copying. Once the timeout passed, the success or error message will revert back to the default message. The value should be a non-negative integer.</li>
+		<li><code class="token attr-name">data-prismjs-copy-text</code> — set a different string to be copied to the clipboard. This overrides the default behaviour of copying the shown text into the clipboard.</li>
 	</ul>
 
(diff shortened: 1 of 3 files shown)


RunDevelopment:
Thank you for the PR @Scy147.

However, I don't think that this is a useful feature. This is mostly because I don't understand your use case. Why would you want to allow users to copy a secrete key, but not see it? I mean, the security value of this zero, because users can see the contents of their clipboard by just pasting it into any text box. Did I misunderstand you

I also think that it would be very confusing for users if our "Copy" did not copy the text of the code block. Conceptually, clicking "Copy" and selecting all text + Ctrl+C should have the same result.

RunDevelopment:
> presenting a product to someone should not force you to expose secrets

But this PR enables exactly that. The secrete is exposed (via "Copy"), just not visually. That's why I don't understand the use case. 

Let's say I have a cURL call and an API key. If I *do not want* to expose the API key, then censoring it with e.g. XXXXX is fine. Copying the command should also copy the censored API key, because we do not want to leak it. If I *do want* to expose the API, then I can just include it in the cURL command as is.

Scy147:
> just not visually

That is exactly the use case i explained about showcasing a product. The actual user logged in has the right to view everything, but the viewer to which the product is being presented should not see it. Executing the command right then and there is not the goal.

RunDevelopment:
Okay, now I get it (I think). With "presenting" you meant "showing it to others during a presentation," PowerPoint and everything.

Assuming that that's the use case, then I will reject this PR. This use case is too specific to justify the added complexity.

---

Also, there is a simpler solution for this: Use [Keep Markup](https://prismjs.com/plugins/keep-markup/), add a `<span class="censor">` around the sensitive text, and add this CSS rule `.censor { filter: blur(0.2em); }`. The sensitive text will be blurred, so nobody can read it, but copying (with and without the Copy to clipboard plugin) will work as normal. The CSS rule can also be changed to support different censoring styles, aside from blurring.

Issue #3094, opened 2021-09-24
Language request: rescript


tkovs:
**Language**
ReScript is a robustly typed language that compiles to efficient and human-readable JavaScript.

**Additional resources**
https://rescript-lang.org/

vmarcosp:
#3435  closes this issue :)

RunDevelopment:
Thanks for the reminder!

PR #3435, opened 2022-04-15
Added support for ReScript

vmarcosp:
Added support for [ReScript](https://rescript-lang.org)

ReScript Prism Preview 1
ReScript Prism Preview 2
ReScript Prism Preview 3


components.json
```diff
@@ -1724,4 +1729,4 @@
 			"owner": "Golmote"
 		}
 	}
-}
```
RunDevelopment, line 1727: Please add a new line here.

components/prism-rescript.js
```diff
@@ -0,0 +1,44 @@
...
+Prism.languages.rescript = {
+	'comment': {
+		pattern: /\/\/(?:[^\r\n\\]|\\(?:\r\n?|\n|(?![\r\n])))*|\/\*[\s\S]*?(?:\*\/|$)/,
+		greedy: true
+	},
+	'class-name': className,
+	'function': {
+		pattern: /[a-zA-Z]\w*(?=\()|(\.)[a-z]\w*/,
+		lookbehind: true,
+	},
```
RunDevelopment, line 15: This regex doesn't have a lookbehind group.

```suggestion
	'function': /\b[a-zA-Z]\w*(?=\()|(\.)[a-z]\w*/,
```

components/prism-rescript.js
```diff
@@ -0,0 +1,44 @@
...
+					'punctuation': /\./g,
+					'operator': operator,
+					'number': number,
+					'keyword': keyword
+				}
+			},
+		}
+	},
+	'attr-value': /[A-Za-z]\w*(?==)/,
+	'constant': new RegExp('(?<=type\\s)[a-z]\\w*'),
```
RunDevelopment, line 35: Nice try, but you won't get around our ES5 requirement :)

While ES5 doesn't support lookbehind assertions, that's what Prism lookbehind groups are for:
```suggestion
	'constant': {
	pattern: /(\btype\s+)[a-z]\w*/,
	lookbehind: true
	},
```

components/prism-rescript.js
```diff
@@ -0,0 +1,44 @@
...
+	'tag': {
+		pattern: new RegExp('((?<=<)[a-z]\\w*)|((?:<\\/)[a-z]\\w*)'),
```
RunDevelopment, line 37: Same as `constant`.

components/prism-rescript.js
```diff
@@ -0,0 +1,44 @@
...
+		pattern: /\/\/(?:[^\r\n\\]|\\(?:\r\n?|\n|(?![\r\n])))*|\/\*[\s\S]*?(?:\*\/|$)/,
+		greedy: true
+	},
+	'class-name': className,
+	'function': {
+		pattern: /[a-zA-Z]\w*(?=\()|(\.)[a-z]\w*/,
+		lookbehind: true,
+	},
+	'number': /(?:\b0x(?:[\da-f]+(?:\.[\da-f]*)?|\.[\da-f]+)(?:p[+-]?\d+)?|(?:\b\d+(?:\.\d*)?|\B\.\d+)(?:e[+-]?\d+)?)[ful]{0,4}/i,
+	'string': {
```
RunDevelopment, line 17: Please move this token above `class-name`.

fdaciuk (approved):
🎉

components/prism-rescript.js
```diff
@@ -0,0 +1,44 @@
+var operator = /\.{3}|:[:=]|\|>|->|=(?:==?|>)?|<=?|>=?|[|^?'#!~`]|[+\-*\/]\.?|\b(?:asr|land|lor|lsl|lsr|lxor|mod)\b/;
```
RunDevelopment, line 1: This is a top-level variables that will leak into global scope. Please wrap the whole file into an IIFE. [Example](https://github.com/PrismJS/prism/blob/master/components/prism-java.js#L1).

RunDevelopment (changes requested):
Thank you for the PR @vmarcosp!

I gave the language a quick review and found a few things.

tests/languages/rescript/comment_feature.test
```diff
@@ -0,0 +1,22 @@
+// foobar
```
RunDevelopment, line 1: The comment pattern also includes stuff like

```
// foo\
bar
```

Is that correct? If so, please add a test case.

vmarcosp: I'm not sure, but probably doesn't include this kind of stuff.

components/prism-rescript.js
vmarcosp: Done

components/prism-rescript.js
vmarcosp: Done

components/prism-rescript.js
vmarcosp: Done

components/prism-rescript.js
vmarcosp: Done

vmarcosp:
@RunDevelopment some tests are broken, I'll be fix and ping you back soon.

fakenickels:
🔥🔥🔥

components/prism-rescript.js
```diff
@@ -0,0 +1,48 @@
...
+	var className = /(\b[A-Z]\w*)|(@[a-z.]*)|#[A-Za-z]\w*|#\d/;
+	var number = /(?:\b0x(?:[\da-f]+(?:\.[\da-f]*)?|\.[\da-f]+)(?:p[+-]?\d+)?|(?:\b\d+(?:\.\d*)?|\B\.\d+)(?:e[+-]?\d+)?)[ful]{0,4}/i;
+	var keyword = /\b(?:and|as|assert|begin|bool|class|constraint|do|done|downto|else|end|exception|external|float|for|fun|function|if|in|include|inherit|initializer|int|lazy|let|method|module|mutable|new|nonrec|object|of|open|or|private|rec|string|switch|then|to|try|type|val|virtual|when|while|with)\b/;
+
+	Prism.languages.rescript = {
+		'comment': {
+			pattern: /\/\/(?:[^\r\n\\]|\\(?:\r\n?|\n|(?![\r\n])))*|\/\*[\s\S]*?(?:\*\/|$)/,
+			greedy: true
+		},
+		'string': {
```
RunDevelopment, line 12: 2 things about strings:

1. Right now, `"${foo}"` is highlighted incorrectly. This can be fixed using an array:

```js
'string': [
  { pattern: /"..."/, greedy: true },
  { pattern: /`...`/, greedy: true, inside: ... }
]
```

2. The pattern for template strings doesn't actually handle interpolation. E.g. `` `foo${"`"}` `` is highlighted incorrectly.

To fix, I recommend just copying the [template string token from JS](https://github.com/PrismJS/prism/blob/master/components/prism-javascript.js#L123). Just replace the `rest: Prism.languages.javascript` with a `rest: null` and set the `rest` property after defining `Prism.languages.rescript`.

RunDevelopment:
> I'll be fix and ping you back soon

Take your time. I'm currently reading through the docs, so I'll probably add a few more comments.

tests/languages/rescript/comment_feature.test
RunDevelopment: The docs say that single line comments are the same as JS, so just `//.*` for single line comments should be fine. We don't need the whole backslash thing.

vmarcosp: You're right!  I'll fix

examples/prism-rescript.html
```diff
@@ -0,0 +1,57 @@
...
+type record = {
+  "field": string
+}
+
+let sum = (a, b) => a + b
+let sum2 = (a:int, b: int) => a + b
+</code></pre>
+
+<h2>Modules, JSX & Components</h2>
+<pre>
```
RunDevelopment, line 38: Test currently fail because of the new line here. There must not be any text between `pre` and `code`.

Make it `<pre><code>...</code></pre>` everywhere and the tests for example will pass.

vmarcosp: Oh, great!

components/prism-rescript.js
RunDevelopment: About [string interpolation](https://rescript-lang.org/docs/manual/latest/primitive-types#string-interpolation): How do j-strings (that's what I'll call `` j`$foo` `` strings) compare to regular backtick strings? Do j-strings just *add* `$name` interpolations or do the `$name` interpolations *replace* `${name}` interpolations?

RunDevelopment: My bad. This thread has already been resolved. Let's move the discussion about j-strings into its own thread.

components/prism-rescript.js
```diff
@@ -0,0 +1,57 @@
...
+			inside: {
+				'operator': /<|>|\//,
+			},
+		},
+		'keyword': keyword,
+		'operator': operator
+	};
+
+	Prism.languages.insertBefore('rescript', 'string', {
+		'template-string': {
```
RunDevelopment, line 32: About [string interpolation](https://rescript-lang.org/docs/manual/latest/primitive-types#string-interpolation): How do j-strings (that's what I'll call `` j`$foo` `` strings) compare to regular backtick strings? Do j-strings just *add* `$name` interpolations or do the `$name` interpolations *replace* `${name}` interpolations?

components/prism-rescript.js
```diff
@@ -0,0 +1,57 @@
...
+			pattern: /`(?:\\[\s\S]|\$\{(?:[^{}]|\{(?:[^{}]|\{[^}]*\})*\})+\}|(?!\$\{)[^\\`])*`/,
```
RunDevelopment, line 33: The docs says "no special character escaping." How would you do the JS string `` `a\`${b}` `` then?

Anyway, some parts of this regex handle backslash escapes. Do we need them?

components/prism-rescript.js
```diff
@@ -0,0 +1,57 @@
...
+	var operator = /\.{3}|:[:=]|\|>|->|=(?:==?|>)?|<=?|>=?|[|^?'#!~`]|[+\-*\/]\.?|\b(?:asr|land|lor|lsl|lsr|lxor|mod)\b/;
+	var className = /(\b[A-Z]\w*)|(@[a-z.]*)|#[A-Za-z]\w*|#\d/;
+	var keyword = /\b(?:and|as|assert|begin|bool|class|constraint|do|done|downto|else|end|exception|external|float|for|fun|function|if|in|include|inherit|initializer|int|lazy|let|method|module|mutable|new|nonrec|object|of|open|or|private|rec|string|switch|then|to|try|type|val|virtual|when|while|with)\b/;
+
+	Prism.languages.rescript = {
+		'comment': {
+			pattern: /\/\/(?:[^\r\n\\]|\\(?:\r\n?|\n|(?![\r\n])))*|\/\*[\s\S]*?(?:\*\/|$)/,
+			greedy: true
+		},
+		'string': /"(?:\\(?:\r\n|[\s\S])|[^\\\r\n"])*"/,
```
RunDevelopment, line 11: This still needs to be greedy. Otherwise, `"https://example.com"` will be highlighted incorrectly.

components/prism-rescript.js
vmarcosp: That's a good question. I'm not worried about these j-strings because everybody uses ``. These j`` strings are an old feature of the language that was "replaced" (but not deprecated yet) by the new template string (``).

RunDevelopment:
Could we also add support for ReScript's char syntax? It could be done like this:

```js
'comment': /.../,
'char': { pattern: /'.'/, greedy: true },
// ...
```

Also, chars seem to be heavily discouraged in the docs. Will this be removed eventually? The syntax does seem to somewhat overlap with generics.

components/prism-rescript.js
RunDevelopment: Alright, then let's just ignore them.

components/prism-rescript.js
vmarcosp: > How would you do the JS string `a\`${b}` then?

The same way, you can check [here](https://rescript-lang.org/try?code=DYUwLgBAhhC8ECIEChSTCAzpeADAOrgCQDeUAvhAJ4D2ArhAMZQB2EAJjRGABYCWmCLiA)

> Do we need them?

I guess so, since we can use the `\` inside of a ``

examples/prism-rescript.html
vmarcosp: Done

components/prism-rescript.js
```diff
@@ -0,0 +1,57 @@
+(function (Prism) {
+	var operator = /\.{3}|:[:=]|\|>|->|=(?:==?|>)?|<=?|>=?|[|^?'#!~`]|[+\-*\/]\.?|\b(?:asr|land|lor|lsl|lsr|lxor|mod)\b/;
+	var className = /(\b[A-Z]\w*)|(@[a-z.]*)|#[A-Za-z]\w*|#\d/;
+	var keyword = /\b(?:and|as|assert|begin|bool|class|constraint|do|done|downto|else|end|exception|external|float|for|fun|function|if|in|include|inherit|initializer|int|lazy|let|method|module|mutable|new|nonrec|object|of|open|or|private|rec|string|switch|then|to|try|type|val|virtual|when|while|with)\b/;
```
RunDevelopment, line 4: These variables are only referenced once each. Could we inline them?

vmarcosp: Sure!

components/prism-rescript.js
vmarcosp: Done

components/prism-rescript.js
RunDevelopment: > The same way

Nice, then we don't need to change any regexes here.

> [here](https://rescript-lang.org/try?code=DYUwLgBAhhC8ECIEChSTCAzpeADAOrgCQDeUAvhAJ4D2ArhAMZQB2EAJjRGABYCWmCLiA)

Ohh, that's useful.

vmarcosp:
> Could we also add support for ReScript's char syntax? It could be done like this:
> 
> ```js
> 'comment': /.../,
> 'char': { pattern: /'.'/, greedy: true },
> // ...
> ```
> 
> Also, chars seem to be heavily discouraged in the docs. Will this be removed eventually? The syntax does seem to somewhat overlap with generics.

Looks nice, I'll add.
Probably will be removed in the future, but since it's simple, I can add.

components/prism-rescript.js
```diff
@@ -0,0 +1,54 @@
+Prism.languages.rescript = {
+	'comment': {
+		pattern: /\/\/.*|\/\*[\s\S]*?(?:\*\/|$)/,
+		greedy: true
+	},
+	'string': {
+		pattern: /"(?:\\(?:\r\n|[\s\S])|[^\\\r\n"])*"/,
+		greedy: true
+	},
+	'class-name': /(\b[A-Z]\w*)|(@[a-z.]*)|#[A-Za-z]\w*|#\d/,
```
RunDevelopment, line 10: The capturing groups here are why the tests fail. Please only use capturing group in combination with backreferences or `lookbehind: true`.
```suggestion
	'class-name': /\b[A-Z]\w*|@[a-z.]*|#[A-Za-z]\w*|#\d/,
```

tests/languages/rescript/comment_feature.test
vmarcosp: Done

components/prism-rescript.js
vmarcosp: Done

tests/languages/rescript/class-name_feature.test
```diff
@@ -0,0 +1,35 @@
...
+	["class-name", "Button"],
+	["operator", "/"],
+	["operator", ">"],
+
+	["class-name", "@react.component"],
+
+	["keyword", "type"],
+	["constant", "poly"],
+	["operator", "="],
+	" [ ",
```
RunDevelopment, line 25: Could we please highlight those as punctuation? Just add a `'punctuation': /[(){}[\],;.]/` after `operaor`.

vmarcosp: Sure

RunDevelopment:
Another nice feature would be to highlight boolean literals. Please add `'boolean': /\b(?:false|true)\b/` after `number`.

RunDevelopment:
I have a question about regexes. Is `%re("/b/g")` special syntax, or is `%re` just a weird function name?

components/prism-rescript.js
```diff
@@ -0,0 +1,55 @@
...
+		pattern: /\/\/.*|\/\*[\s\S]*?(?:\*\/|$)/,
+		greedy: true
+	},
+	'char': { pattern: /'.'/, greedy: true },
+	'string': {
+		pattern: /"(?:\\(?:\r\n|[\s\S])|[^\\\r\n"])*"/,
+		greedy: true
+	},
+	'class-name': /\b[A-Z]\w*|@[a-z.]*|#[A-Za-z]\w*|#\d/,
+	'function': /[a-zA-Z]\w*(?=\()|(\.)[a-z]\w*/,
```
RunDevelopment, line 12: CI found another unused capturing group. It's probably meant to be a lookbehind, right?
```suggestion
	'function': { pattern: /[a-zA-Z]\w*(?=\()|(\.)[a-z]\w*/, lookbehind: true },
```

vmarcosp:
> I have a question about regexes. Is `%re("/b/g")` special syntax, or is `%re` just a weird function name?

`%re` Is a [ppx](https://ocamlverse.github.io/content/metaprogramming.html), but we can highlight this one as a function.

components/prism-rescript.js
vmarcosp: Right

vmarcosp: Done

tests/languages/rescript/class-name_feature.test
vmarcosp: Done

RunDevelopment:
Are all PPX of the form `%name`? If so, then we could highlight all of them as a function (or similar).

tests/languages/rescript/char_feature.test
```diff
@@ -0,0 +1,19 @@
...
+'\\'
+'\xff'
+'\o214'
+
+----------------------------------------------------
+
+[
+	["char", "'b'"],
+	["char", "'\\'"], ["operator", "'"],
+	["operator", "'"], "\\\\", ["operator", "'"],
```
RunDevelopment, line 12: Well, seems like my simple pattern was a little too simple. Could we replace the char regex with `'(?:[^\r\n\\]|\\(?:.|\w+))'`?

vmarcosp:
> Are all PPX of the form `%name`? If so, then we could highlight all of them as a function (or similar).

There are three forms of ppx: 
- `@` like @react.component - This one I'm highlighting as a `class-name` because it behaves like a decorator 
- `%%` like `%%raw` - This one we can highlight as a function because it behaves very similar to a function.
- `%` like `%re` - The same as `%%`

tests/languages/rescript/char_feature.test
vmarcosp: Yeah

tests/languages/rescript/operator_feature.test
```diff
@@ -0,0 +1,31 @@
+...
+:: :=
+= == ===
+< <= > >=
+| ^ ? '
```
RunDevelopment, line 5: Isn't `:` also an operator because of `cond ? a : b`?

RunDevelopment:
> There are three forms of ppx:
> 
> * `@` like @react.component - This one I'm highlighting as a `class-name` because it behaves like a decorator
> * `%%` like `%%raw` - This one we can highlight as a function because it behaves very similar to a function.
> * `%` like `%re` - The same as `%%`

Sounds good. Then let's highlight `%%name` and `%name` as functions.

tests/languages/rescript/operator_feature.test
vmarcosp: You're right, I'll add

RunDevelopment:
The coverage CI test is currently failing because:

1. The token `attr-value` is completely untested. Please add a test for this.
2. Not all keywords are tested. Please add a keyword test. [Example](https://github.com/PrismJS/prism/blob/master/tests/languages/cypher/keyword_feature.test).

components/prism-rescript.js
```diff
@@ -0,0 +1,60 @@
+Prism.languages.rescript = {
+	'comment': {
+		pattern: /\/\/.*|\/\*[\s\S]*?(?:\*\/|$)/,
+		greedy: true
+	},
+	'char': { pattern: /'(?:[^\r\n\\]|\\(?:.|\w+)')/, greedy: true },
```
RunDevelopment, line 6: Oh no, sorry. I mistyped.
```suggestion
	'char': { pattern: /'(?:[^\r\n\\]|\\(?:.|\w+))'/, greedy: true },
```

vmarcosp: I'll fix

vmarcosp:
> The coverage CI test is currently failing because:
> 
> 1. The token `attr-value` is completely untested. Please add a test for this.
> 2. Not all keywords are tested. Please add a keyword test. [Example](https://github.com/PrismJS/prism/blob/master/tests/languages/cypher/keyword_feature.test).

I'm running the tests locally just to make sure that everything is passing

RunDevelopment:
> I'm running the tests locally just to make sure that everything is passing

Some parts of the CI aren't part of `npm test`. Coverage (`npm run regex-coverage`) is one of them.

components/prism-rescript.js
vmarcosp: Done

components/prism-rescript.js
```diff
@@ -0,0 +1,60 @@
...
+	},
+	'tag': {
+		pattern: /(<)[a-z]\w*|(?:<\/)[a-z]\w*/,
+		lookbehind: true,
+		inside: {
+			'operator': /<|>|\//,
+		},
+	},
+	'keyword': /\b(?:and|as|assert|begin|bool|class|constraint|do|done|downto|else|end|exception|external|float|for|fun|function|if|in|include|inherit|initializer|int|lazy|let|method|module|mutable|new|nonrec|object|of|open|or|private|rec|string|switch|then|to|try|type|val|virtual|when|while|with)\b/,
+	'operator': /\.{3}|:[:=]|\|>|->|=(?:==?|>)?|<=?|>=?|[|^?'#!~`]|[+\-*\/]\.?|\b(?:asr|land|lor|lsl|lsr|lxor|mod)\b/,
```
RunDevelopment, line 31: `:` isn't tokenized as an operator.

```suggestion
	'operator': /\.{3}|:[:=]?|\|>|->|=(?:==?|>)?|<=?|>=?|[|^?'#!~`]|[+\-*\/]\.?|\b(?:asr|land|lor|lsl|lsr|lxor|mod)\b/,
```

tests/languages/rescript/keyword_feature.test
```diff
@@ -0,0 +1,105 @@
...
+	["keyword", "or"],
+	["keyword", "private"],
+	["keyword", "rec"],
+	["keyword", "string"],
+	["keyword", "switch"],
+	["keyword", "then"],
+	["keyword", "to"],
+	["keyword", "try"],
+	["keyword", "type"],
+	["constant", "val"],
```
RunDevelopment, line 96: Please add a comma (or some other punctuation) after `type`.

vmarcosp: Fixed

components/prism-rescript.js
vmarcosp: Fixed

RunDevelopment (approved):
Thank you, @vmarcosp! I think the language is ready to go. Please run `npm run build` to make the CI and then I'll merge this.

vmarcosp:
> Thank you, @vmarcosp! I think the language is ready to go. Please run `npm run build` to make the CI and then I'll merge this.

Thanks for your help 😄

vmarcosp:
@RunDevelopment is there an estimate for the next release that includes these changes?

RunDevelopment:
Probably this weekend.

vmarcosp:
Oh, great! Tks

vmarcosp:
I almost forgot, but since this pr is merged I think you can close this issue #3094

PR #3455, opened 2022-05-15
Added support for WGSL

Dr4gonthree:
This adds support for WGSL, the WebGPU Shading Language.

There is one Problem:
WGSL is still grayed out in the examples page, and i don't know why.


components/prism-wgsl.js
```diff
@@ -0,0 +1,94 @@
+Prism.languages.wgsl = {
+    'comment': {
+        pattern: /\/\/.*|\/\*[\s\S]*?(?:\*\/|$)/,
+        greedy: true,
+    },
+    'attributes': {
+        pattern: /(@)(?:(?:[_\p{XID_Start}][\p{XID_Continue}]+)|(?:[\p{XID_Start}]))/u,
```
RunDevelopment, line 7: Prism still supports browsers that don't support the u flag, so please don't use it.

While a lot more simplistic, if we only focus on ASCII text, the equivalent non-Unicode pattern would be:
```suggestion
        pattern: /(@)[_a-z]\w*/i,
```

components/prism-wgsl.js
```diff
@@ -0,0 +1,94 @@
...
+    ],
+    'decimal-float-literal-a': [
+        { pattern: /[0-9]*\.[0-9]+(?:[eE](?:\+|-)?[0-9]+)?[fh]?/, alias:'number' },
+        { pattern: /[0-9]+\.[0-9]*(?:[eE](?:\+|-)?[0-9]+)?[fh]?/, alias:'number' },
+        { pattern: /[0-9]+[eE](?:\+|-)?[0-9]+[fh]?/, alias:'number' },
+    ],
+    'variables': {
+        pattern: /\b(?:(?:[_\p{XID_Start}][\p{XID_Continue}]+)|(?:[\p{XID_Start}]))\b/u,
+        alias: 'variable',
+    },
```
RunDevelopment, line 67: This is now how [the `variable` token](https://prismjs.com/tokens.html#standard-tokens) is supposed to be used.

Please either remove the alias or the whole token. If you want to keep the token, please replace it with `/\b[_a-z]\w*\b/i`.

components/prism-wgsl.js
```diff
@@ -0,0 +1,94 @@
...
+        { pattern: /->/ },
+    ],
+    'punctuation': [
+        { pattern: /,/ },
+        { pattern: /[{}]/ },
+        { pattern: /[\(\)]/ },
+        { pattern: /;/ },
+        { pattern: /[\[\]]/ },
+        { pattern: /(?<!=)[<>]/ },
+    ],
```
RunDevelopment, line 93: We generally prefer to combine these things because the resulting regexes are a lot shorter and faster to read.

I also remove the lookbehind because it's unnecessary.

```suggestion
    'punctuation': /[(){}[\],;<>]/
```

components/prism-wgsl.js
```diff
@@ -0,0 +1,94 @@
...
+        inside: {},
+    },
+    'keyword': [
+        { pattern: /\b(?:bitcast|block|break|case|continue|continuing|default|discard|else|elseif|enable|fallthrough|for|function|if|loop|private|read|read_write|return|storage|switch|uniform|workgroup|write)\b/ },
+        { pattern: /\b(?:asm|const|do|enum|handle|mat|premerge|regardless|typedef|unless|using|vec|void|while)\b/ },
+        { pattern: /\b(?:let|var)\b/ },
+        { pattern: /\b(?:type)\b/ },
+        { pattern: /\b(?:enum)\b/ },
+        { pattern: /\bfn\b/ },
+    ],
```
RunDevelopment, line 24: Please combine all of those variants into a single pattern. The current categorization is somewhat arbitrary. Splitting this into multiple regexes also hinders our regex tooling.

components/prism-wgsl.js
```diff
@@ -0,0 +1,94 @@
...
+    'types': [
+        { pattern: /\b(?:bool|i32|u32|f32)\b/, alias: 'builtin' },
+        { pattern: /\b(?:i64|u64|f64)\b/, alias: 'builtin' },
+        { pattern: /\b(?:vec[2-4]|mat[2-4]x[2-4])\b/, alias: 'builtin' },
+        { pattern: /\b(?:atomic)\b/, alias: 'builtin' },
+        { pattern: /\b(?:array)\b/, alias: 'builtin' },
+        { pattern: /\b(?:override|ptr|sampler|sampler_comparison|staticAssert|struct)\b/, alias: 'builtin' },
+        { pattern: /\b(?:texture_1d|texture_2d|texture_2d_array|texture_3d|texture_cube|texture_cube_array|texture_multisampled_2d)\b/, alias: 'builtin' },
+        { pattern: /\b(?:texture_storage_1d|texture_storage_2d|texture_storage_2d_array|texture_storage_3d|texture_depth_2d)\b/, alias: 'builtin' },
+        { pattern: /\b(?:texture_depth_2d_array|texture_depth_cube|texture_depth_cube_array|texture_depth_multisampled_2d)\b/, alias: 'builtin' },
```
RunDevelopment, line 39: Same as with keywords for the builtin types.

components/prism-wgsl.js
```diff
@@ -0,0 +1,94 @@
...
+        { pattern: /\b(?:[A-Z][A-Za-z0-9]*)\b/, alias: 'class-name' },
+    ],
+    'built-in-values': [
+        { pattern: /\b(?:vertex_index|instance_index|position)\b/, alias: 'namespace' },
```
RunDevelopment, line 43: Please do not abuse the namespace token name and choose a different alias.

components/prism-wgsl.js
```diff
@@ -0,0 +1,94 @@
...
+        { pattern: /\b(?:front_facing|frag_depth|sample_index|sample_mask)\b/, alias: 'namespace' },
+        { pattern: /\b(?:local_invocation_id|local_invocation_index|global_invocation_id|workgroup_id|num_workgroups)\b/, alias: 'namespace' },
+    ],
+    'bool-literal': {
+        pattern: /\b(?:true|false)\b/,
+        alias: 'boolean',
+    },
+    'hex-int-literal': [
+        { pattern: /\b0[xX][0-9a-fA-F]+[iu]?(?!(?:\.|p|P))\b/, alias:'number' },
```
RunDevelopment, line 52:
```suggestion
        { pattern: /\b0[xX][0-9a-fA-F]+[iu]?\b(?![.pP])/, alias:'number' },
```

components/prism-wgsl.js
```diff
@@ -0,0 +1,94 @@
...
+    ],
+    'hex-float-literal': [
+        { pattern: /0[xX][0-9a-fA-F]*\.[0-9a-fA-F]+(?:[pP](?:\+|-)?[0-9]+[fh]?)?/, alias:'number' },
+        { pattern: /0[xX][0-9a-fA-F]+\.[0-9a-fA-F]*(?:[pP](?:\+|-)?[0-9]+[fh]?)?/, alias:'number' },
+        { pattern: /0[xX][0-9a-fA-F]+[pP](?:\+|-)?[0-9]+[fh]?/, alias:'number' },
```
RunDevelopment, line 57: Prism isn't a parser or compiler. We are a syntax highlighter. This means that we can do some things parsers can't do. E.g. ignore invalid syntax.

Prism is allowed to produce any highlighting for invalid syntax. This allows us to simplify many regexes.

This regex can also be simplified.

```suggestion
        { pattern: /0[xX][0-9a-fA-F]*(?:\.[0-9a-fA-F]*)?(?:[pP][+-]?\d+[fh]?)?/, alias:'number' },
```

components/prism-wgsl.js
```diff
@@ -0,0 +1,94 @@
...
+        alias: 'variable',
+    },
+    'decimal-float-literal-b': [
+        { pattern: /0[fh]/, alias:'number' },
+        { pattern: /[1-9][0-9]*[fh]/, alias:'number' },
+    ],
+    'int-literal': [
+        { pattern: /0[iu]?/, alias:'number' },
+        { pattern: /[1-9][0-9]*[iu]?/, alias:'number' },
+    ],
```
RunDevelopment, line 75: Same here and a few other tokens.

By ignore what highlighting will be produces for invalid syntax (e.g. `0000i`), we can simplify our regexes.

```suggestion
    'int-literal': { 
        pattern: /\b\d+[iu]?\b/,
        alias:'number'
    },
```

Dr4gonthree:
Thanks for all the feedback and suggestions @RunDevelopment.
I hope the new changes fix all the issues you pointed out.
All the ci jobs seem to have succeeded.

tests/languages/wgsl/attributes_feature.test
```diff
@@ -0,0 +1,75 @@
...
+	["attributes", "group"],
+	["punctuation", "("],
+	["punctuation", ")"],
+
+	"\r\n@",
+	["attributes", "id"],
+	["punctuation", "("],
+	["punctuation", ")"],
+
+	"\r\n@",
```
RunDevelopment, line 49: Could we highlight the `@` as well in some way? Maybe as part of the attribute or as punctuation.

components/prism-wgsl.js
```diff
@@ -0,0 +1,67 @@
...
+        { pattern: /(?:\^|~|\|(?!\|)|\|\||&&|<<|>>|!)(?!=)/ },
+        { pattern: /&(?![&=])/ },
+        { pattern: /(?:\+=|-=|\*=|\/=|%=|\^=|&=|\|=|<<=|>>=)/},
+        { pattern: /(?<![<>=!])=(?![=>])/ },
+        { pattern: /(?:==|!=|<=|(?<!=)>=)/ },
+        { pattern: /(?:(?:[+%]|(?:\*(?!\w)))(?!=))|(?:-(?!>))|(?:\/(?!\/))/ },
+        { pattern: /\.(?!\.)/ },
+        { pattern: /->/ },
+    ],
+    'punctuation': /[(){}[\],;<>]/,
```
RunDevelopment, line 66: Please also highlight `:` as punctuation.
image

components/prism-wgsl.js
```diff
@@ -0,0 +1,67 @@
...
+        pattern: /\b(fn\s+)[_a-zA-Z]\w*(?=[(<])/,
+        lookbehind: true,
+        alias: 'function',
+    },
+    'keyword': /\b(?:bitcast|break|case|const|continue|continuing|default|discard|else|enable|fallthrough|fn|for|function|if|let|loop|private|return|storage|switch|type|uniform|var|while|workgroup)\b/,
+    'function-calls': {
+        pattern: /\b[_a-z]\w*(?=\()/i,
+        alias: 'function',
+    },
+    'builtin': /\b(?:array|atomic|bool|i32|u32|f32|i64|u64|f64|vec[2-4]|mat[2-4]x[2-4]|override|ptr|sampler|sampler_comparison|staticAssert|struct|texture_1d|texture_2d|texture_2d_array|texture_3d|texture_cube|texture_cube_array|texture_multisampled_2d|texture_storage_1d|texture_storage_2d|texture_storage_2d_array|texture_storage_3d|texture_depth_2d|texture_depth_2d_array|texture_depth_cube|texture_depth_cube_array|texture_depth_multisampled_2d)\b/,
```
RunDevelopment, line 21: Please match bulti-ins before `function-calls`.
image

---

`struct` is a keyword, not a built-in type or function.

image

components/prism-wgsl.js
```diff
@@ -0,0 +1,67 @@
...
+        pattern: /\b(?:vertex_index|instance_index|position)\b/,
+        alias: 'attr-value',
+    },
+    'built-in-values-fragment': {
+        pattern: /\b(?:front_facing|frag_depth|sample_index|sample_mask)\b/,
+        alias: 'attr-value',
+    },
+    'built-in-values-compute': {
+        pattern: /\b(?:local_invocation_id|local_invocation_index|global_invocation_id|workgroup_id|num_workgroups)\b/,
+        alias: 'attr-value',
```
RunDevelopment, line 33: I'm getting false positives for these.

image
image

I think this could be fixed by saying that built-ins must not be followed by a `:` or preceded by a `.`.
```suggestion
    'built-in-values-vertex': {
        pattern: /((?:^|[^.\s])\s*)\b(?:vertex_index|instance_index|position)\b(?!\s*:)/,
        lookbehind: true,
        alias: 'attr-value',
    },
    'built-in-values-fragment': {
        pattern: /((?:^|[^.\s])\s*)\b(?:front_facing|frag_depth|sample_index|sample_mask)\b(?!\s*:)/,
        lookbehind: true,
        alias: 'attr-value',
    },
    'built-in-values-compute': {
        pattern: /((?:^|[^.\s])\s*)\b(?:local_invocation_id|local_invocation_index|global_invocation_id|workgroup_id|num_workgroups)\b(?!\s*:)/,
        lookbehind: true,
        alias: 'attr-value',
```

tests/languages/wgsl/operator_feature.test
```diff
@@ -0,0 +1,67 @@
...
+==
+!=
+<=
+>=
++
+%
+*
+-
+/
+.
```
RunDevelopment, line 30: `.` should be punctuation. That's how we do it in pretty much every other C-like language, so let's keep it consistent.

components/prism-wgsl.js
```diff
@@ -0,0 +1,67 @@
...
+    'int-literal': { 
+        pattern: /\b\d+[iu]?\b/,
+        alias:'number',
+    },
+    'operator': [
+        { pattern: /(?:\^|~|\|(?!\|)|\|\||&&|<<|>>|!)(?!=)/ },
+        { pattern: /&(?![&=])/ },
+        { pattern: /(?:\+=|-=|\*=|\/=|%=|\^=|&=|\|=|<<=|>>=)/},
+        { pattern: /(?<![<>=!])=(?![=>])/ },
+        { pattern: /(?:==|!=|<=|(?<!=)>=)/ },
```
RunDevelopment, line 61: We can't use lookbehind assertions.

```suggestion
        { pattern: /(^|[^<>=!])=(?![=>])/, lookbehind: true },
        { pattern: /(?:==|!=|<=|(^|[^=])>=)/, lookbehind: true },
```

RunDevelopment (changes requested):
I found some minor things.

RunDevelopment:
> All the ci jobs seem to have succeeded.

CI jobs can be abused to send spam, mine cryptocurrency, or other stuff. To prevent this, not all CI jobs trigger for first-time contributors automatically. A maintainer (e.g. me) has to manually approve that the CI jobs are allowed to run for each commit you make.

Unfortunately, this means that the CI may fail long after you make a commit. However, you can run tests (`npm run test`), lints (`npm run lint`), and coverage (`npm run regex-coverage`) locally. This is most of what the CI also does.

components/prism-wgsl.js
Dr4gonthree: Another options might be to only match them only inside the 'builtin' attribute, as they are only defined in that context. I think you can actually name a variable or a property 'position' but i havent tried it. Is this possible with inside?

RunDevelopment: Yes, that would be possible using `inside` for `attribute`.

Dr4gonthree: Ok i did it with `inside` now, but now in `builtin(notABuiltInValue)` the `notABuiltInValue` gets highlighted as attribute. Can i somehow achieve that everything inside the `inside` that doesn't get matched will not be highlighted as attribute. Should i use some kind of catch all pattern or is there another way to do this?

RunDevelopment: We could add a helper token for this.

```js
'builtin-attribute': {
  pattern: /(@)builtin\(.*?\)/,
  lookbehind: true,
 inside: {
    'attribute': /^builtin/, // TODO: alias
    'punctuation': /[(),]/,
    'built-in-values': /\b(?:...)\b/ // TODO: alias
  }
},    
'attributes': {
        pattern: /(@)(?:align|binding|const|group|id|interpolate|invariant|location|size|workgroup_size|vertex|fragment|compute)/i,
        lookbehind: true,
        alias: 'attr-name'
}
```

RunDevelopment:
Please run `npm run lint:fix` to fix the linting error.

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @Dr4gonthree!

PR #3467, opened 2022-06-09
Normalize Whitespace Plugin: add configuration via attributes

plasticrake:
Added the ability to set configuration using `data-` attributes. Previously you could only set them via page-wide defaults.

It appeared there was some unused code with `env.settings` left over from PR #851. But without that PR/plugin you cannot set anything at the block-level.

Original Plugin Author: @zeitgeist87


plugins/normalize-whitespace/prism-normalize-whitespace.js
```diff
@@ -161,6 +173,27 @@
...
+				if (pre.hasAttribute('data-' + key)) {
+					var value = pre.getAttribute('data-' + key);
+					switch (settingType) {
+						case 'boolean':
+							value = (value === 'true');
+							break;
+						case 'number':
+							value = (isNaN(+value) ? 0 : +value);
+							break;
+					}
```
RunDevelopment, line 191: This is too brittle. Why is the value initialized to the unparsed value? Why are invalid values just silently accepted and parsed using some kind of default value? Why is there no default case?

Anything that isn't a valid boolean should be ignored. Similarly, anything that isn't a valid number should be ignored (0 as a default doesn't even make sense for a lot of settings).

---

Also, users should be able to input `true` like this `<pre data-prop>`.

plugins/normalize-whitespace/index.html
```diff
@@ -103,6 +103,16 @@ <h1>Examples</h1>
 		};
 
 
+		var
```
RunDevelopment, line 106: Why did you add this here? This seems unrelated to your changes.

plasticrake: Are you asking about the `var` or the entire thing? this `var` is in initilzizing the `there_is_a_very_very_very_very_long_line_it_can_break_it_for_you` variable. The original without line breaks is in  `demo.html` line 25

The reason this block is here is it is demoing using `data-break-lines`

plugins/normalize-whitespace/prism-normalize-whitespace.js
plasticrake: I can add the `<pre data-prop>` use case. But I would do any kind of validation in the actual functions that are called. Otherwise you're validating the data-props but not when calling the functions vis JS. I was concerning this PR with adding the ability to pass props without using JS.

The reason for setting the number to 0 if invalid is all the functions that take a number in this plugin default to 0 if something is null/undefined. But I can change this to just ignore NaN (functionally it would be the same thing)

plasticrake: There is no default case as its impossible to not be a `boolean` or a `number`. This data is hard-coded on line 36.

plasticrake:
It is more strict parsing data-props. If a value is unexpected it is ignored. Boolean props set with `<pre data-prop>` are set to `true`.

```
const assert = require('assert').strict;

function t (value, settingType) {
  switch (settingType) {
    case 'boolean':
      value = (value === 'true' || value === '') || (value === 'false' ? false : undefined);
      break;
    case 'number':
      value = (value === '' ? undefined : (isNaN(+value) ? undefined : +value));
      break;
  }
  return value;
}

assert(t('', 'boolean') === true);
assert(t('true', 'boolean') === true);
assert(t('false', 'boolean') === false);
assert(t('invalid', 'boolean') === undefined);

assert(t('', 'number') === undefined);
assert(t('0', 'number') === 0);
assert(t('123', 'number') === 123);
assert(t('invalid', 'number') === undefined);
assert(t('123invalid456', 'number') === undefined);
```

plugins/normalize-whitespace/prism-normalize-whitespace.js
```diff
@@ -161,6 +173,29 @@
...
+			if (Object.hasOwnProperty.call(settingsConfig, key)) {
+				var settingType = settingsConfig[key];
+				if (pre.hasAttribute('data-' + key)) {
+					var value = pre.getAttribute('data-' + key);
+					switch (settingType) {
+						case 'boolean':
+							value = (value === 'true' || value === '') || (value === 'false' ? false : undefined);
+							break;
+						case 'number':
+							value = (value === '' ? undefined : (isNaN(+value) ? undefined : +value));
```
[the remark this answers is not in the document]
plasticrake: Fixed

plugins/normalize-whitespace/prism-normalize-whitespace.js
RunDevelopment: > The reason for setting the number to 0 if invalid is all the functions that take a number in this plugin default to 0 if something is null/undefined.

Their default values might change and should not affect argument parsing.

plugins/normalize-whitespace/prism-normalize-whitespace.js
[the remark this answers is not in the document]
plasticrake: You may want to turn on this rule so eslint catches it: https://eslint.org/docs/rules/no-implicit-coercion

plugins/normalize-whitespace/prism-normalize-whitespace.js
```diff
@@ -161,6 +173,29 @@
...
+						case 'boolean':
+							value = (value === 'true' || value === '') || (value === 'false' ? false : undefined);
+							break;
+						case 'number':
+							value = (value === '' || (isNaN(Number(value))) ? undefined : Number(value));
+							break;
+					}
+					if (value !== undefined) {
+						env.settings[key] = value;
+					}
```
RunDevelopment, line 194: Actually, how about this?
```suggestion
					try {
						var value = JSON.parse(pre.getAttribute('data-' + key) || 'true')
						if (typeof value === settingType) {
							env.settings[key] = value;
						}
					} catch (_error) {
						// ignore error
					}
```

plugins/normalize-whitespace/prism-normalize-whitespace.js
plasticrake: Yes I corrected this

plugins/normalize-whitespace/prism-normalize-whitespace.js
RunDevelopment: > if that matters to you

It doesn't. Highlighting and the normalize plugin itself will take orders of magnitude longer. There's no need to worry about this.

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @plasticrake!

PR #3473, opened 2022-06-14
hoon: process escaped strings correctly

matildepark:
As per urbit/urbit.org#770, this PR changes the string regex to handle escaped tapes (that is, "lists of strings") and strings (or "cords") correctly.


RunDevelopment (approved):

RunDevelopment:
Thank you @matildepark!

PR #3475, opened 2022-06-15
Ignore single-line highlight which is out of wrap section

lijyze:
When highlight a range, if its end line exceed the wrap section, prism highlight from start line of the rang to the end of wrap section. That's good.
When thing comes to a single line, prism highlight the last line of wrap section, that's weird.

So ,I make it a little different, if the single-line is out of wrap section, simply ignore this line.


RunDevelopment:
Thank you for the PR @lijyze!

Unfortunately, I do not understand the problem you fixed. Could you please explain it in more detail? Maybe even provide some screenshots?

lijyze:
@RunDevelopment Sorry for bad explanation.

Let's say, if my code looks like this, it has only 3 lines of code.
image

Now I want highlight some lines of it.
- If the end line of range exceed max line numbers of code, such as `data-line="2-10"`
    - Without `class="line-numbers"`: 
        image
    - And with `class="line-numbers"`: 
        image

I think this make sense, it means highlight from the start of the range to the end of the code.
But if the section is exceeded by a line rather than a range, things comes a little weird.
- Code section exceed by a line, such as `data-line="4"`
    - Without `class="line-numbers"`: 
        image
    - With `class="line-numbers"`:
        image

What have I done is ignore lines that exceed line numbers of code section when `class="line-number"` is present.

Thank you for reading this, I wish you a good day!

RunDevelopment:
Similarly, we could also ignore (= not display) any ranges or single-lines if their `start` is greater than the number of lines.

lijyze:
Yeah! Your solution is much more elegant.
Thanks you for giving pointer!!! I have done with code!

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @lijyze!

PR #3495, opened 2022-07-14
Add metadata property to Token

anastasiya1155:
Currently, in the "wrap" hook there is no way to know where the token is located. I would like to have this info to develop a plugin that can highlight ranges (not the whole line). This PR adds a new property to the Token class that can be used for storing any additional data about the token and it will be passed to the "wrap" hook. I believe this will allow greater flexibility when working with hooks. Any feedback is much appreciated!


RunDevelopment:
Hello @anastasiya1155!

While I agree that the wrap hook could provide more information to plugins, I think that token metadata is fundamentally not a good solution.

Token metadata has multiple problems:
1. Plugins have to explicitly support it. Metadata is lost if plugins that create/transform tokens don't diligently copy it over. This is a huge potential source of bugs as all code that handles tokens would have to account for metadata. So we would not only have to keep this in mind for future code, we would also have to look at/update all current code that handles tokens.
2. Metadata cannot be assumed. Since metadata is optional, you cannot rely on that information being present. So whatever plugin you develop, it is at the mercy of metadata not being lost/altered.
3. You cannot assume anything about metadata. Even if metadata is present, you cannot know what's in it. Even if your plugin adds metadata of a certain shape, other plugins might overwrite parts of your metadata. This creates yet another source of plugin incompatibility and source of bugs.
4. The created metadata might not even be meaningful. Prism recursively highlights text, and plugins are allowed to change/insert/remove text/tokens. Even if none of the above problems mattered, the metadata you get at the end might not even be correct. E.g. source position information is meaningless because Prism and plugins can freely change source code during highlighting (we only guarantee that after all these changes, we return the same source code at the very end).

For these reasons, I will reject the addition of token metadata.

---

> I would like to have this info to develop a plugin that can highlight ranges (not the whole line)

I don't know the specifics of what you're trying to do, but the wrap hook might not be the right one for the job. What exactly are you trying  to do?

Also, is what you want to do similar to what [Keep Markup](https://prismjs.com/plugins/keep-markup/) can do?

anastasiya1155:
> Hello @anastasiya1155!
> 
> While I agree that the wrap hook could provide more information to plugins, I think that token metadata is fundamentally not a good solution.
> 
> Token metadata has multiple problems:
> 
> 1. Plugins have to explicitly support it. Metadata is lost if plugins that create/transform tokens don't diligently copy it over. This is a huge potential source of bugs as all code that handles tokens would have to account for metadata. So we would not only have to keep this in mind for future code, we would also have to look at/update all current code that handles tokens.
> 2. Metadata cannot be assumed. Since metadata is optional, you cannot rely on that information being present. So whatever plugin you develop, it is at the mercy of metadata not being lost/altered.
> 3. You cannot assume anything about metadata. Even if metadata is present, you cannot know what's in it. Even if your plugin adds metadata of a certain shape, other plugins might overwrite parts of your metadata. This creates yet another source of plugin incompatibility and source of bugs.
> 4. The created metadata might not even be meaningful. Prism recursively highlights text, and plugins are allowed to change/insert/remove text/tokens. Even if none of the above problems mattered, the metadata you get at the end might not even be correct. E.g. source position information is meaningless because Prism and plugins can freely change source code during highlighting (we only guarantee that after all these changes, we return the same source code at the very end).
> 
> For these reasons, I will reject the addition of token metadata.
> 
> > I would like to have this info to develop a plugin that can highlight ranges (not the whole line)
> 
> I don't know the specifics of what you're trying to do, but the wrap hook might not be the right one for the job. What exactly are you trying to do?
> 
> Also, is what you want to do similar to what [Keep Markup](https://prismjs.com/plugins/keep-markup/) can do?

@RunDevelopment Thank you for pointing these out! All of your points are totally valid. The main problem I was trying to solve is the lack of context in the "wrap" hook, like getting info about the position of the token inside the text. I think you are right and I should find some other ways to solve the issue.

PR #3507, opened 2022-07-26
Java: Added support for constants

wangliang181230:
Java: Added support for constants.


RunDevelopment:
If users really wanted to add their own Java variables, they could just do `Prism.language.java['variable'] = /\b(?:foo|bar)\b/`. Why is your new `PrismConfig` necessary?

wangliang181230:
> If users really wanted to add their own Java variables, they could just do `Prism.language.java['variable'] = /\b(?:foo|bar)\b/`. Why is your new `PrismConfig` necessary?

Wait a minute, I also found this problem, so I'm developing a better solution.

wangliang181230:
> If users really wanted to add their own Java variables, they could just do `Prism.language.java['variable'] = /\b(?:foo|bar)\b/`. Why is your new `PrismConfig` necessary?



> If users really wanted to add their own Java variables, they could just do `Prism.language.java['variable'] = /\b(?:foo|bar)\b/`. Why is your new `PrismConfig` necessary?

done

components/prism-java.js
```diff
@@ -55,7 +55,9 @@
 		'operator': {
 			pattern: /(^|[^.])(?:<<=?|>>>?=?|->|--|\+\+|&&|\|\||::|[?:~]|[-+*/%&|^!=<>]=?)/m,
 			lookbehind: true
-		}
+		},
+		'constant': /\b[A-Z][A-Z_\d]{2,}\b/,
+		'variable': /\b[a-z]\w*\b/
```
RunDevelopment, line 60: I think the `variable` pattern has too many false positives to include. Class fields (instance and static) commonly use camelCase, so this regex will misidentify all of them as variables. This is just how Java's syntax is, so I don't think that we can fix this. Could you please remove the `variable` pattern?

components/prism-java.js
```diff
@@ -55,7 +55,9 @@
 		'operator': {
 			pattern: /(^|[^.])(?:<<=?|>>>?=?|->|--|\+\+|&&|\|\||::|[?:~]|[-+*/%&|^!=<>]=?)/m,
 			lookbehind: true
-		}
+		},
+		'constant': /\b[A-Z][A-Z_\d]{2,}\b/,
```
RunDevelopment, line 59: Why did you require constants to have at least 3 characters? There are a few English words with 2 characters (e.g. "no") that might be used as static final fields or enum variants.

RunDevelopment (changes requested):

components/prism-java.js
wangliang181230: I changed it to 2 characters

components/prism-java.js
wangliang181230: done

tests/languages/java/issue3345.test
```diff
@@ -10,7 +10,7 @@ public class Main {
     public static record A() {
     }
     // it cannot be used as class name in java syntax
-    public static class record {
+    public static class Record {
```
RunDevelopment, line 13: Could you please revert this change?

RunDevelopment:
One last nit. It's good to go otherwise.

tests/languages/java/issue3345.test
wangliang181230: done

wangliang181230:
@RunDevelopment

RunDevelopment:
> `BB`

Most of the class name detect works on naming conventions:
1. Class names start with an uppercase letter and contain at least one lowercase letter. E.g. `String`.
2. Class names are a single uppercase letter. E.g. `T` (for generics).

`BB` is a false negative that we cannot resolve without compiling the Java code. Just `BB` without context could be a static field (constant) or a type. There is no way to know based on the name alone.
Yes, `BB` is declared in `<A, BB>`, but we would need an AST to know that, and Prism doesn't build an AST. Prism only produces a token stream.

components/prism-java.js
```diff
@@ -55,7 +55,13 @@
 		'operator': {
 			pattern: /(^|[^.])(?:<<=?|>>>?=?|->|--|\+\+|&&|\|\||::|[?:~]|[-+*/%&|^!=<>]=?)/m,
 			lookbehind: true
-		}
+		},
+		'constant': /\b[A-Z][A-Z_\d]+\b/,
+		'variable': [
+			{
+				pattern: /\b(?:args|log(?:ger)?|value)\b/
```
RunDevelopment, line 62: Why did you add this again?

wangliang181230:
> > `BB`
> 
> Most of the class name detect works on naming conventions:
> 
>     1. Class names start with an uppercase letter and contain at least one lowercase letter. E.g. `String`.
> 
>     2. Class names are a single uppercase letter. E.g. `T` (for generics).
> 
> 
> `BB` is a false negative that we cannot resolve without compiling the Java code. Just `BB` without context could be a static field (constant) or a type. There is no way to know based on the name alone. Yes, `BB` is declared in `<A, BB>`, but we would need an AST to know that, and Prism doesn't build an AST. Prism only produces a token stream.

Need to add a pattern for the generic parameters?

components/prism-java.js
wangliang181230: I think `args`, `log`, `logger` and `value` are common, so this pattern is added.

RunDevelopment: I don't think this is sufficient justification to add arbitrary words. Please remove this.

components/prism-java.js
```diff
@@ -80,6 +86,7 @@
 		'generics': {
 			pattern: /<(?:[\w\s,.?]|&(?!&)|<(?:[\w\s,.?]|&(?!&)|<(?:[\w\s,.?]|&(?!&)|<(?:[\w\s,.?]|&(?!&))*>)*>)*>)*>/,
 			inside: {
+				'generic-parameter': /\b[A-Z]+\b/,
```
RunDevelopment, line 89: I don't think this is a good addition. While examples like `Foo<A, BB>` are not more correct, they are also inconsistent. `BB` is only highlighted as a generic parameter inside the `<>`, but not outside. E.g. `BB` will be inconsistent in this snippet:

```java
class Foo<BB> {
  foo: BB;
}
```

For syntax highlighting, consistency is very important. If `BB` is not highlighted in the same way, reader might believe that the two `BB`s might refer to different things.

components/prism-java.js
wangliang181230: done

components/prism-java.js
wangliang181230: outside the BB is not a generic-parameter.

图片

components/prism-java.js
```diff
@@ -120,4 +127,19 @@
 			}
 		}
 	});
+
+	// variable from window.PrismConfig
+	if (typeof window !== 'undefined') {
```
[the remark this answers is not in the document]
wangliang181230: removed

components/prism-java.js
wangliang181230: It is very difficult to distinguish between `generic-parameters` and `constants`

wangliang181230: At present, I have no solution, sorry.

RunDevelopment: It's not difficult, it's impossible without compilation. `BB` doesn't follow Java naming conventions anyway, so we can just ignore it.

Please remove this token.

wangliang181230: removed

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @wangliang181230!

PR #3509, opened 2022-07-27
Add Bash language alias "sh"

benhatsor:


RunDevelopment:
Hello @barhatsor!

I think this is a good addition, but it has to be implemented differently. Add the "sh" alias to [Bash's alias list here](https://github.com/PrismJS/prism/blob/05ee042a81bebd4ed47ce986f93644a37b4e7ca1/components.json#L206) (maybe even add an alias title) and run `npm ci && npm run build`. The build command will update Autoloader (and everything else that needs updating) automatically.

RunDevelopment (approved):

PR #3512, opened 2022-07-30
Improve BBj regexs

hyyan:
* [BBj: extend BBj keywords list](https://github.com/PrismJS/prism/commit/ca5e6a8191c966c5e7da8cba0652b30614ca8435)
* [BBj: fix strings regex does not match single quotes](https://github.com/PrismJS/prism/commit/f069d10e4faeb52254d0973d8a79643c686d6a4b)
* [BBj: add declare as keyword](https://github.com/PrismJS/prism/commit/358f56ffcc65b781c66108a4f357bd4edc7d319e)


components/prism-bbj.js
```diff
@@ -6,11 +6,11 @@
 			greedy: true
 		},
 		'string': {
-			pattern: /"(?:""|[!#$%&'()*,\/:;<=>?^\w +\-.])*"/,
+			pattern: /(['"])(?:(?!\1|\\).|\\.)*\1/,
```
RunDevelopment, line 9: In the tests, `"foo "" bar"` is highlighted as 2 strings. Is that intentional?

hyyan: @RunDevelopment That's intentional

RunDevelopment (approved):

Issue #3514, opened 2022-08-03
BQN Support


yewscion:
**Language**
[BQN](https://mlochbaum.github.io/BQN/) is a modern, powerful language created by @mlochbaum centered on Ken Iverson's array programming paradigm.

It supports:

- A simple, consistent, and stable array programming language
- A low-dependency C implementation using bytecode compilation: installation
- System functions for math, files, and I/O (including a C FFI)
- Documentation with examples, visuals, explanations, and rationale for features

**Additional resources**
There is a [userscript](https://gist.github.com/dzaima/35ca0ce12b5e215a62460f00e693984f) for highlighting BQN in a browser already written by @dzaima [info](https://aplwiki.com/wiki/Dzaima), who is also the maintainer of the officially-recommended implementation. 

While I'm sure the code in the userscript will need modifying for use in PrismJS, a large part of the work is already therefore done.

Other Links:

[BQN Homepage](https://mlochbaum.github.io/BQN/)
[Official Page Recommending the Above Userscript](https://mlochbaum.github.io/BQN/editors/index.html)

RunDevelopment:
Closed by #3515.

PR #3515, opened 2022-08-05
bqn: Add bqn language support.

yewscion:
Related to: #3514

This patch adds support for [the BQN language](https://mlochbaum.github.io/BQN/) to PrismJS, following the instructions in [the documentation](https://prismjs.com/extending.html). Let me know if anything needs fixed before merge.


components/prism-bqn.js
```diff
@@ -0,0 +1,61 @@
+Prism.languages.bqn = {
+	'shebang': {
+	pattern: /^#!\s*\/.*/,
+	alias: 'important'
+	},
+	'comment': /#.*$/m,
```
RunDevelopment, line 6: `$` (with the m flag present) is defined as `(?!.)`, so it will always trivially accept. It's essentially an unnecessary condition and can be removed.

```suggestion
	'comment': /#.*/,
```

RunDevelopment: Also, please make `comment` greedy.

components/prism-bqn.js
```diff
@@ -0,0 +1,61 @@
+Prism.languages.bqn = {
+	'shebang': {
```
RunDevelopment, line 2: Please make this token greedy.

components/prism-bqn.js
```diff
@@ -0,0 +1,61 @@
...
+	pattern: /^#!\s*\/.*/,
```
RunDevelopment, line 3: `\s` includes new lines, which is probably incorrect.
```suggestion
	pattern: /^#![ \t]*\/.*/,
```

RunDevelopment (changes requested):
Thank you for the PR @yewscion!

The code itself looks good, but please fix the formatting. We use ESLint to do that, so either install an ESLint plugin/extension for your editor or run `npm run lint:fix`.

RunDevelopment:
Coverage tests fail because the `punctuation` token is untested. Please add a test for it.

components/prism-bqn.js
```diff
@@ -0,0 +1,61 @@
...
+	'word-number': {
+		pattern: /[¯∞π]?(?:\d*\.?\b\d+(?:e[+¯]?\d+)?|¯|∞|π)(?:j¯?(?:(?:\d+(?:\.\d+)?|\.\d+)(?:e[+¯]?\d+)?|¯|∞|π))?/i,
+		alias: 'number'
+	},
+	'null-literal': {
+	pattern: /[@]/,
+	alias: 'string'
+	},
+	'primitive-functions': {
+	pattern: /[-+×÷⋆√⌊⌈|¬∧∨<>≠=≤≥≡≢⊣⊢⥊∾≍⋈↑↓↕«»⌽⍉/⍋⍒⊏⊑⊐⊒∊⍷⊔!]/u,
```
RunDevelopment, line 49: We still support browsers that don't support the u flag. Luckily, all these characters are one UTF16 char code, so we can safely remove the u flag without affecting the pattern.
```suggestion
	pattern: /[-+×÷⋆√⌊⌈|¬∧∨<>≠=≤≥≡≢⊣⊢⥊∾≍⋈↑↓↕«»⌽⍉/⍋⍒⊏⊑⊐⊒∊⍷⊔!]/,
```

yewscion: Unfortunately, it seems that there are surrogate pairs within the `special-name` tokens (`𝕨𝕩𝕗𝕘𝕤𝕣𝕎𝕏𝔽𝔾𝕊 _𝕣 _𝕣_`). If I should avoid the `/u` modifier, what is the way to deal with these?

Thank You for the prompt feedback!

RunDevelopment: Those have to be made into a simple groups, unfortunately. So `(?:𝕨|𝕩|𝕗|𝕘|𝕤|...)`.

yewscion: Ahh, Okay. Done and done.

components/prism-bqn.js
```diff
@@ -0,0 +1,65 @@
...
+		pattern: /^#![ \t]*\/.*/,
+		alias: 'important',
+		greedy: true
+	},
+	'comment': {
+		pattern: /#.*/,
+		greedy: true
+	},
+	'string-literal': {
+		pattern: /"(?:\\[\s\S]|\$(?!\()|`[^`]+`|[^"\\`$])*"/,
```
dzaima, line 12: BQNs string literals are way simpler - `/"(?:[^"]|"")*"/`

yewscion: Oh, thanks! The one I was using was from the APL highlighting, but I was having difficulty massaging it to recognized the double-double quote escape. This fixed it! Thanks again.

components/prism-bqn.js
```diff
@@ -0,0 +1,65 @@
...
+	},
+	'string-literal': {
+		pattern: /"(?:[^"]|"")*"/,
+		greedy: true,
+		alias: 'string'
+	},
+	'character-literal': {
+		pattern: /'.'/,
+		greedy: true,
+		alias: 'string'
```
RunDevelopment, line 19: `Char` is the [correct alias](https://prismjs.com/tokens.html) here.

```suggestion
		alias: 'char'
```

components/prism-bqn.js
```diff
@@ -0,0 +1,65 @@
...
+		pattern: /#.*/,
+		greedy: true
+	},
+	'string-literal': {
+		pattern: /"(?:[^"]|"")*"/,
+		greedy: true,
+		alias: 'string'
+	},
+	'character-literal': {
+		pattern: /'.'/,
```
RunDevelopment, line 17: This is slightly wrong because of non-Unicode regexes. The docs say: "[Character literals](https://mlochbaum.github.io/BQN/doc/token.html#characters-and-strings) start and end with `'`, and have exactly one character in between." 

So one [Unicode character](https://mlochbaum.github.io/BQN/doc/types.html#characters). To correctly match this, we have to handle surrogate pairs.

```suggestion
		pattern: /'(?:[\s\S]|[\uD800-\uDBFF][\uDC00-\uDFFF])'/,
```

components/prism-bqn.js
```diff
@@ -0,0 +1,65 @@
...
+		greedy: true,
+		alias: 'string'
+	},
+	'function': {
+		pattern: /•[\w¯.∞π]+[\w¯.∞π]*/i
+	},
```
RunDevelopment, line 23:
```suggestion
	'function': /•[\w¯.∞π]+[\w¯.∞π]*/i,
```

RunDevelopment (changes requested):

components/prism-bqn.js
RunDevelopment: Also, because of the `i` flag, this regex also matches "Π". Is that intentional?

components/prism-bqn.js
```diff
@@ -0,0 +1,65 @@
...
+	'special-name': {
+		pattern: /(?:𝕨|𝕩|𝕗|𝕘|𝕤|𝕣|𝕎|𝕏|𝔽|𝔾|𝕊|_𝕣_|_𝕣)/,
+		alias: 'keyword'
+	},
+	'dot-notation-on-name': {
+		pattern: /[A-Z_][\w¯∞π]*\./i,
+		alias: 'namespace'
+	},
+	'word-number-scientific': {
+		pattern: /\d+(?:\.\d+)?[e|E]¯?\d+/,
```
RunDevelopment, line 37: The `|` in `[e|E]` is probably wrong, isn't it?

components/prism-bqn.js
```diff
@@ -0,0 +1,65 @@
...
+		pattern: /[A-Z_][\w¯∞π]*/i,
+		alias: 'symbol'
+	},
+	'word-number': {
+		pattern: /[¯∞π]?(?:\d*\.?\b\d+(?:e[+¯]?\d+)?|¯|∞|π)(?:j¯?(?:(?:\d+(?:\.\d+)?|\.\d+)(?:e[+¯]?\d+)?|¯|∞|π))?/i,
+		alias: 'number'
+	},
+	'null-literal': {
+		pattern: /@/,
+		alias: 'string'
```
RunDevelopment, line 50: Isn't this also a character?
```suggestion
		alias: 'char'
```

components/prism-bqn.js
yewscion: Applied this change in most recent commit!

components/prism-bqn.js
yewscion: It was not, just pushed another commit to rectify this in both function names as well as words. 'Π' is never valid outside of a char/string literal.

components/prism-bqn.js
yewscion: Applied in recent patch.

components/prism-bqn.js
yewscion: No, both e and E work for scientific notation in BQN.

yewscion: Oh wait, that is a character set. Yes, that is wrong, I will fix it.

components/prism-bqn.js
yewscion: Technically, it is considered a token, I think: https://mlochbaum.github.io/BQN/spec/token.html 

However, it is not listed as a type: https://mlochbaum.github.io/BQN/spec/types.html

So, it should probably be a char as well.

components/prism-bqn.js
yewscion: Pushed Commit to that Effect.

components/prism-bqn.js
```diff
@@ -0,0 +1,63 @@
...
+	'dot-notation-on-brackets': {
+		pattern: /\{(?=.*\}\.)|\}\./,
+		alias: 'namespace'
+	},
+	'special-name': {
+		pattern: /(?:𝕨|𝕩|𝕗|𝕘|𝕤|𝕣|𝕎|𝕏|𝔽|𝔾|𝕊|_𝕣_|_𝕣)/,
+		alias: 'keyword'
+	},
+	'dot-notation-on-name': {
+		pattern: /[A-Z_][\w¯∞π]*\./i,
```
RunDevelopment, line 31: Is Π invalid here as well?

yewscion: Indeed, sorry. I should have noticed that; I keep forgetting it is a letter.

yewscion: Pushed in most recent commit.

RunDevelopment (approved):

RunDevelopment:
Thank you for contributing @yewscion!

yewscion:
Thank You @RunDevelopment! Sorry there were so many issues to fix; I will try to be more attentive in the future.

PR #3518, opened 2022-08-15
Line Highlight: Account for offset when clamping ranges

### LINKED ISSUE #3517: Some lines are not highlighted
**Information**
- Language: typescript, svelte
- Plugins: Line Highlight

**Description**
I'm using 10-line code slices and wanted to set an offset and highlight certain lines. It works fine until I need to highlight a line number greater than 10, then the line doesn't get highlighted.

For example,
`<pre data-line-offset={6} data-line={'9-12'} ...></pre>`
would highlight lines 9 and 10 but not 11 and 12

<details>
<summary>Line numbers that are greater than the snippet length are not highlighted.</summary>

</details>
Screen Shot 2022-08-13 at 8 15 23 PM

RunDevelopment:
This fixes #3517.

This was caused by #3475. I failed to take the offset into account when giving my suggestion.

Screenshot of the example in #3517 working.

image


JaKXz (commented):
I see that the line-highlight plugin does not have any tests. I think adding one would be good as part of this PR, if possible

RunDevelopment:
I added tests in #3524. This PR is now blocked by #3524.

RunDevelopment:
Alright, so https://github.com/PrismJS/prism/pull/3524 is not gonna happen until #3408 which require v2. So either we change our stance on #3408 or we merge this PR without tests and add them in later.

I don't want to have a bunch of long expected HTML in the test files, so I don't want to do these tests without snapshots.

Also, I'll likely have a bit of time on my hands next week, so I could start work with v2. So merging this PR now, making the last v1 release with weekend, and then starting with v2 would be my preferred timeline.

Thoughts? @JaKXz @mAAdhaTTah @Golmote @LeaVerou

mAAdhaTTah:
> we merge this PR without tests and add them in later.

That seems fine. Let's go w/ it.

JaKXz (approved):
Agreed with @mAAdhaTTah 👍🏽

PR #3524, opened 2022-08-18
Added snapshot plugin tests

RunDevelopment:
Since @JaKXz asked for Light Highlight tests in #3518, I added some. However, I thought that having a bunch of test HTML in source files wasn't ideal, so I added snapshot testing. Since this added a dependency and made quite a few changes, I didn't want this to be part of #3518, hence this PR.

Snapshots are pretty easy to use with `util.assert`. If the `expected` option is missing, it will use snapshots.

Right now, only the tests for Line Highlight and Highlight Keyword use snapshots. The other tests did not have long enough test HTML to warrant snapshots IMO.


RunDevelopment:
Sooo, it seems like this is blocked by #3408.

RunDevelopment:
Resolved by 5f5b6d21cac32a31d71288c006745a744f6c995c in v2.

Issue #2768, opened 2021-02-17
Lazily evaluate languages


RunDevelopment:
**Motivation**
Adding a new language to Prism takes time. While Prism grammars itself are cheap, our thousands of regexes are not, so loading a large number of grammars can take quite some time.

While Prism evaluates grammars, page loading and rendering are blocked. This means that having Prism choosing a Prism bundle with more languages hurts performance even if those languages end up unused.

**Description**
Add a way for Prism to lazily evaluate languages. (This should be doable with #2736.)

The performance gain is quite significant. Simply loading a Prism bundle (min) with all languages included takes about 200ms on my machine. Loading that same bundle with lazy evaluation only takes 10ms.

This can be a game-changer, especially on low-end and mobile devices.

<details>

Everything is eagerly evaluated:

![image](https://user-images.githubusercontent.com/20878432/108215154-4c28a380-7131-11eb-93bb-e010744ebc9f.png)

Only Markup, JS, CSS, and C-like are eagerly evaluated:
(Everything else is wrapped into a function.)

![image](https://user-images.githubusercontent.com/20878432/108214870-f81dbf00-7130-11eb-950b-67f5e0e39c1e.png)

---

Obviously, Prism Core is always eagerly evaluated.

</details>

mAAdhaTTah:
I think we'd need to dig into the perf impacts a bit. If we lazily evaluate a grammar then ends up being needed immediately anyway, are we getting any benefits from lazy eval? I suspicion is a loaded language is likely to be used on a given page so the benefits of lazy eval may be limited.

RunDevelopment:
> If we lazily evaluate a grammar that ends up being needed immediately anyway, are we getting any benefits from lazy eval?

Probably not. It's mainly useful when you have a large number of grammars and might not need them immediately.

The main use case for this is large bundles. With "large", I mean all languages. Those large bundles are easy to use because they have everything. Users don't have to worry about Autoloader or language dependencies. Everything just works. 
Bundle size doesn't matter too much with large bundles (because Prism also works when loaded async) but Prism blocking the page is a problem.

PR #3529, opened 2022-08-20
Roadmap

RunDevelopment:
I wrote a bit about our main goals for v2. I don't know if it's good enough for a roadmap though.

@LeaVerou @Golmote @mAAdhaTTah @JaKXz

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,7 +1,6 @@
 # Roadmap for v2
 
-Over the years, there have been many great suggestions that we were not able to implement due to them being a breaking change.
-Now, we are currently work on Prism v2.
+With Prism finally moving towards v2, we finally have an opportunity to revisit the many breaking changes we've received over the years. Below are some of the changes we'll be implementing, and how.
 
 ## ESM
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,6 +1,7 @@
 # Roadmap for v2
 
-With Prism finally moving towards v2, we finally have an opportunity to revisit the many breaking changes we've received over the years. Below are some of the changes we'll be implementing, and how.
+With Prism finally moving towards v2, we finally have an opportunity to revisit the many breaking changes we've received over the years.
+Below are some of the changes we'll be implementing, and how.
 
 ## ESM
 
@@ -25,7 +26,7 @@ This will make importing a component as simple as importing a single file.
 Since ESM isn't supported everywhere, we will offer 2 distributions: ESM and bundled.
 The bundle distribution of a component will be one monolithic file with all of its dependencies included.
 
-For more information see #2715, #2880, and #2736.
+For more information see [#2715], [#2880], and [#2736].
 
 ## Browser support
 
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -0,0 +1,43 @@
+# Roadmap for v2
+
+Over the years, there have been many great suggestions that we were not able to implement due to them being a breaking change.
+Now, we are currently work on Prism v2.
```
mAAdhaTTah, line 4:
```suggestion
With Prism finally moving towards v2, we finally have an opportunity to revisit the many breaking changes we've received over the years. Below are some of the changes we'll be implementing, and how.
```

ROADMAP.md
```diff
@@ -0,0 +1,43 @@
...
+
+## ESM
```
mAAdhaTTah, line 6:
```suggestion
## ESM for component dependencies
```

ROADMAP.md
```diff
@@ -0,0 +1,43 @@
...
+The bundle distribution of a component will be one monolithic file with all of its dependencies included.
+
+For more information see #2715, #2880, and #2736.
+
+## Public API
+
+We also plan a few other changes to the public API.
+These changes will generally make the API simpler and resolve accumulated tech debt.
+
+For more information see #3416, #3528, #2948, #3420, #2982, and #1844.
```
mAAdhaTTah, line 36: I think we should probably list these out here, or at least link them because [they aren't linked automatically](https://github.com/PrismJS/prism/blob/64404104d6d789758acb42b6c07bd73694cfbfcb/ROADMAP.md).

ROADMAP.md
```diff
@@ -0,0 +1,43 @@
...
+
+## IE11
```
mAAdhaTTah, line 38: We should probably expand this more broadly to be our browser support policy.

JaKXz (commented):
I think this makes sense! 

I would make it the [description of the milestone](https://github.com/PrismJS/prism/milestone/3) and leave this PR open so we can use it as a RFC :)

mAAdhaTTah (commented):
Couple other things we might want to include:

1. No longer committing build artifacts
2. Changes to the build process (if any)
3. ES6 syntax in general (Maybe this goes with our browser support policy?)

May we tweet this roadmap for feedback before we get too far along?

mAAdhaTTah:
Looking at the milestone, there are also things like #2768 that we should either include or close out.

ROADMAP.md
RunDevelopment: Not really true. Everything will use ESM.

ROADMAP.md
RunDevelopment: Well, then: What is our browser support policy?

RunDevelopment: I suppose it's: we require ES6. Everything that doesn't fully support ES6, isn't supported.

RunDevelopment:
> I would make it the [description of the milestone](https://github.com/PrismJS/prism/milestone/3) and leave this PR open so we can use it as a RFC :)

Prism v2 includes many different topics, so the comments could get messy very quickly. 

How about we make this an announcement in the discussion tab? We could share the roadmap and people can comment on it in a more structured way.

> Looking at the milestone, there are also things like #2768 that we should either include or close out.

Hmmm, I don't see the roadmap as be all and end all of what we'll do for v2. I view it as "these are the most important changes coming." Maybe we should add something like "and much more" and link to the milestone?

About #2768 specifically: we could technically implement the new language registry in a way that doesn't lazily evaluate all languages. I won't, but it's nice to explicitly have it there.

> May we tweet this roadmap for feedback before we get too far along?

My plan for this PR was to work together on the roadmap until we can publish it (somewhere, see above section). So I would hold that tweet until we do. I also plan to attach a link to the roadmap to the release notes of v1.last as well.

RunDevelopment:
I just made the announcement (#3531), so I'll close this PR now.

PR #3547, opened 2022-09-10
Type module

RunDevelopment:


RunDevelopment:
Okay, I can for the life of me not figure out how to make gulp work with type: module and esm. [Gulp's docs aren't very helpful either](https://gulpjs.com/docs/en/getting-started/javascript-and-gulpfiles/). I tried, babel, esm, and ts-node and nothing worked. 

The only way I found was to make the file end with `.cjs` and continue with CommonJS. Problem: This makes it impossible to require other ESM which we need to do. 

The thing is: The problem is 100% with gulp itself. You can run ESM scripts with node (and ts-node) just fine, no problem. I'm contemplating yeeting gulp and making our own simple task runner (grunt seems to have similar levels of ESM support as gulp; none).

I already wasted close to 2 days of this, so please help...

@LeaVerou @mAAdhaTTah @Golmote @JaKXz

LeaVerou:
IMO in v2 we should move away from gulp and to just npm scripts.

LeaVerou:
I’m probably missing something major here, but why do we need ad hoc scripts to generate the changelog? GitHub these days supports auto-generating a changelog (when you generate a release), which we could then paste into `CHANGELOG.md`. Admittedly, it will not be as nice, but it seems to require a substantial amount of effort to have our own custom one.

RunDevelopment:
> I’m probably missing something major here, but why do we need ad hoc scripts to generate the changelog?

GH generated changelogs are just the list of commits since the last release. Our changelogs are a bit nicer than that. Also, we already have it, so why not keep it?

Also, changelogs don't have a lot to do with this PR, so I don't really see how this is relevant...

RunDevelopment:
Closed in favor of https://github.com/PrismJS/prism/pull/3687.

PR #3626, opened 2023-01-08
Add Flat Assembler syntax highlighting

kspalaiologos:
Not as advanced as I would like just yet (no support for macros), but it's way more complete than using the NASM highlighter.
Before:
image
After:
image


doomed-neko:
no updates?

PR #3657, opened 2023-03-12
feat(treeview): Added support for ts extension

mkadirtan:


JaKXz:
> Should I also update treeview.min.css?

Yes please run the build command and include the changed files. Thanks!

PR #3662, opened 2023-03-21
Added `proto` alias for protobuf

dimitropoulos:
proto is the extension for protobuf files, and is used in [other](https://github.com/shikijs/shiki/blob/44b9b0c458744267d824a8283eac4d6393f65013/packages/shiki/src/languages.ts#L104) syntax highlighters as the language id.


dimitropoulos:
@RunDevelopment do have a spare moment to look at this quick PR?  Prism is a really popular choice and I'd love to make this easier for users.

PR #3617, opened 2022-12-27
Update prism-line-highlight.css

johnclyde:
color-adjust is deprecated

PR #3672, opened 2023-04-17
Patch for build libs warning (with PostCSS/autoprefixer) for line-highlight plugin's CSS color-adjust prop

oberocks:
For quite a while now, I've gotten vite/tailwind warnings during builds when using the ever awesome prism(js) as a dep.

Screen Shot 2023-04-17 at 11 55 57 AM

Today though, I found a random spare half hour to dig a bit on it...

spongebob-time-card-02

Ok so it turns out that it's possible to get rid of the warning and I found at least one other awesome dev that did a super [similar fix (3617)](https://github.com/PrismJS/prism/pull/3617).

I totally understand the hold for v2, and this isn't "security" per say... but then again I figured maybe [this pull release (3617)](https://github.com/PrismJS/prism/pull/3617) wasn't merged because the suggested change dropped the old `color-adjust: exact;` line. That might trigger a need to re-test/recompile/etc.

Now don't get me wrong, I totally did NOT check if a lack of `color-adjust: exact;` breaks old IE consumption of prismjs' line-highlight plugin's awesomeness. That's way over my budget for today's time investment. 

But I DO know that adding the newer `print-color-adjust: exact;` prop/value prior to the old line, does in fact fix the warning that vite/tailwind (and likely a few other build libs using autoprefixer and postcss) are throwing over the now depreciated `color-adjust` prop.

So hopefully this is an easy merge, even though it's not security related.

_(unless we consider "security" as a dev's sanity when non-coders can access our build logs and get panic attacks from thinking this warning will melt down reality... but I digress.)_

Please let me know if there's anything I can or need to do, as I understand how much work it is to make a major version update, albeit not for code as useful and awesome as this code is!

Cheers and thank you either way! 😄


rbirkgit:
Are there any plans to merge this? We are getting this warning with our Angular projects.

PR #3699, opened 2023-06-30
Add prepare script, making it easy to install Prism from outside of NPM.

trusktr:
The `prepare` script is the NPM-standard way in which outside-of-NPM packages can be build, for example with installed from a git location.

After merging this into the main branch (currently `master`), people will be able to do this:

```sh
npm install prismjs@github:PrismJS/prism
```

If they want to try it right now before v2 is merged (if this is merged into v2), they can do this:

```sh
npm install prismjs@github:PrismJS/prism#v2
```

Without this, such installations are not possible.

What NPM does during the `install` process is

- it clones the specified repo,
- checks out the specified git ref (f.e. the `v2` in `#v2`),
- installs dependencies including dev dependencies,
- runs the `prepare` script if any (the `prepare` script will *always* have dev dependencies available to it),
- and finally packages the result in the same way as `npm pack` such that build outputs are included in the final package that gets installed into the user's node_modules

This would be a way for people to easily test v2 right now and provide feedback.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
 	"scripts": {
 		"benchmark": "ts-node benchmark/benchmark.ts",
 		"build": "ts-node scripts/build.ts",
+		"prepare": "npm run build",
 		"lint": "eslint . --cache",
 		"lint:fix": "npm run lint -- --fix",
 		"lint:ci": "eslint . --max-warnings 0",


trusktr:
By the way, now that I've made this pull request, I can install and test this like so:

```js
npm install prismjs@github:trusktr/prism#patch-1
```

JaKXz (commented):
Thanks @trusktr -- were you able to confirm this works for you? I'm happy to ✅

trusktr:
Will take it for a spin soon (ironing out a few PRs in [Docsify](https://github.com/docsifyjs/docsify/) first, then we're due for a Prism update)

trusktr:
@JaKXz Here's an example that installs Prism from git, and then imports and initializers a `new Prism()` and calls `.highlightAll()`:

https://github.com/trusktr/prism-v2-test/tree/main

There's no syntax highlighting, and the docs aren't updated to describe how to use v2 yet, so I'm not yet sure how to get syntax coloring to be applied.

The build output also doesn't have type declaration outputs yet, so there wasn't yet intellisense or type checking in VS Code when I imported Prism. It guesses from the plain JS output:

Screenshot 2023-07-05 at 12 52 49 PM

Happy to help iron these details out.

I also think at this point in time we don't need to ship legacy CommonJS. To avoid maintenance, I'd say let other people build to CommonJS if they need to, f.e. they can use Rollup for that, and we can otherwise support the ES Module reality that all JS engines and build tools support nowadays.

trusktr:
This change is now included in https://github.com/PrismJS/prism/pull/3704

Issue #3702, opened 2023-07-05
[v2] type checking and intellisense support


trusktr:
Currently the v2 build provides no types for consumers. We should output declaration files that work well with vanilla ES Module usage (ensuring it works with vanilla ES Modules ensures the most compatibility: if it works with vanilla, it will work with everything else).

trusktr:
To make this have a high chance of working well in a way that is easy to understand, I recommend we change from the current chunked output, and simply output vanilla ES Modules that match the exact same structure as our source code.

Essentially, the `dist/` folder should simply be a mirror of our `src/` folder: a one-to-one mapping with merely types stripped, filenames converted from `.ts` to `.js`, plus type declaration files emitted for type checking and intellisense on the consumer end.

Anything else will be more complicated: more to think about, more likely someone looking at source gets confused when looking at dist/ output, etc.

trusktr:
Alright, here my initial vision for this simplification, including example app and Node ESM import test in the sample app:

https://github.com/PrismJS/prism/pull/3704

Note that with this setup, we get full intellisense and type checking.

<img width="790" alt="intellisense" src="https://github.com/PrismJS/prism/assets/297678/77ba79b5-5472-4918-ae6c-1578909e1427">

If we Go-To-Definition, we are taken to the TypeScript source (not just the declaration file):


https://github.com/PrismJS/prism/assets/297678/77cc0d1b-b638-49da-b7a0-fb3845a1bd77

Lastly, this exposes one problem with the current type definitions of languages. The type definitions for the language objects are not visible to VS Code for some reason, and we get no intellisense and `.grammar` methods are type `any`:


https://github.com/PrismJS/prism/assets/297678/63af5575-6a65-4933-852f-1d8ca0d65c16


I'm guessing this may have to do with the type casts (?), f.e. `as LanguageProto<"typescript">`.

Now that we have types flowing into consumer code in the simplest-possible ESM setup, we can start to work on the last things such as these missing types for end user code.

That's all I have time for today. I would love if you can take this for a spin and give some thoughts on it.

Last things needed on the PR's current path:

- ensure the `components.json` build is handled if needed
- fix last type definition issues

PR #3704, opened 2023-07-05
Simplified ESM build

trusktr:
Issue:

- https://github.com/PrismJS/prism/issues/3702 (read this first)

Continues from PR

- https://github.com/PrismJS/prism/pull/3699

To test this in an existing app, install Prism v2 like so:

```sh
npm install prismjs@github:trusktr/prism#simple-esm
```

To try an existing Prism v2 example along with a Node ESM test try this repo:

https://github.com/trusktr/prism-v2-test

```sh
npm install # this installs Prism from git, from this branch
npm test # this verifies that importing Prism in Node ESM works (we should perhaps move that test into this repo)
```

The gold standard today is that if a project is both as close to vanilla ESM as possible and importable in Node ESM, it is likely to work with today's build tools.

In particular, this change tries to stay as close to vanilla ESM as possible, by avoiding the use of package.json `exports` field (browsers do not read package.json, let along an exports field).

The example app above shows importing Prism without any configuration or build. In the worst case, if they wish to avoid putting `./node_modules` directly in their import statement, they can create a `<script type=importmap>` element to map `prismjs` to the `node_modules` location.

All Mocha unit tests pass except two of them. I also removed the components.json build, which I think needs to be restored for when new languages are added (? I'm not sure yet), but now we'll ensure that we restore what is needed on this simplified foundation.

---

The mantras here are:

- simple ES Module output that requires the least amount of tooling or configuration to import in vanilla ES Modules (namely, in a browser where this is most likely to be imported, where browser ESM is a subset of Node ESM).
- we start with most vanilla ESM setup before starting to add any build steps, so that we're staying on the most compatible ESM path.

This is as simple as it can possibly get for browser ES Modules (apart from us also including a sample `importmap` that they can paste into their HTML, which we can do.

This is still WIP, we need to look at the components.json build requirement.

---

- [ ] ensure the `components.json` build is handled if needed
- [ ] fix last type definition issues (described in https://github.com/PrismJS/prism/issues/3702#issuecomment-1622713226)

---

Comments in the diff have some more details.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -15,7 +15,7 @@ module.exports = {
 		'eqeqeq': ['error', 'always', { 'null': 'ignore' }],
 
 		// imports
-		'import/extensions': ['warn', 'never'],
+		'import/extensions': ['warn', 'always'], // "always" is most compatible with default vanilla ES Modules
 		'import/order': [
 			'warn',
 			{
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,0 +1,7 @@
+{
+	"Xrequire": "@babel/register",
+		"Xexperimental-specifier-resolution=node": "",
+	"node-option": [
+		"loader=ts-node/esm"
+	]
+}
--- a/package.json
+++ b/package.json
@@ -2,14 +2,15 @@
 	"name": "prismjs",
 	"version": "1.29.0",
 	"description": "Lightweight, robust, elegant syntax highlighting. A spin-off project from Dabblet.",
-	"main": "prism.js",
+  "type": "module",
+	"main": "dist/core/prism.js",
 	"style": "themes/prism.css",
 	"engines": {
 		"node": ">=14"
 	},
 	"scripts": {
 		"benchmark": "ts-node benchmark/benchmark.ts",
-		"build": "ts-node scripts/build.ts",
+		"build": "ts-node-esm scripts/build.ts && npm run tsc",
 		"prepare": "npm run build",
 		"lint": "eslint . --cache",
 		"lint:fix": "npm run lint -- --fix",
@@ -24,7 +25,8 @@
 		"test:plugins": "ts-mocha tests/plugins/**/*.ts",
 		"test:runner": "ts-mocha tests/testrunner-tests.ts",
 		"test": "npm-run-all test:*",
-		"tsc": "tsc && tsc -p tests/tsconfig.json"
+    "typecheck": "tsc --noEmit && tsc -p tests/tsconfig.json --noEmit",
+		"tsc": "tsc"
 	},
 	"repository": {
 		"type": "git",
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
 	"scripts": {
 		"benchmark": "ts-node benchmark/benchmark.ts",
 		"build": "ts-node scripts/build.ts",
+		"prepare": "npm run build",
 		"lint": "eslint . --cache",
 		"lint:fix": "npm run lint -- --fix",
 		"lint:ci": "eslint . --max-warnings 0",
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -1,34 +1,36 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
-import rollupTerser from '@rollup/plugin-terser';
-import rollupTypescript from '@rollup/plugin-typescript';
+// import rollupTerser from '@rollup/plugin-terser';
+// import rollupTypescript from '@rollup/plugin-typescript';
 import CleanCSS from 'clean-css';
 import fs from 'fs';
 import { mkdir, readFile, readdir, rm, writeFile, } from 'fs/promises';
-import MagicString from 'magic-string';
+// import MagicString from 'magic-string';
 import path from 'path';
-import { rollup } from 'rollup';
+// import { rollup } from 'rollup';
+import { fileURLToPath } from 'url';
 import { webfont } from 'webfont';
-import { toArray } from '../src/shared/util';
-import { components } from './components';
-import { parallel, runTask, series } from './tasks';
-import type { ComponentProto } from '../src/types';
-import type { Plugin, SourceMapInput } from 'rollup';
+// import { toArray } from '../src/shared/util.js';
+// import { components } from './components.js';
+import { parallel, runTask, series } from './tasks.js';
+// import type { ComponentProto } from '../src/types';
+// import type { Plugin, SourceMapInput } from 'rollup';
 
 
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const SRC_DIR = path.join(__dirname, '../src/');
-const languageIds = fs.readdirSync(path.join(SRC_DIR, 'languages')).map((f) => f.slice('prism-'.length).slice(0, -'.js'.length)).sort();
+// const languageIds = fs.readdirSync(path.join(SRC_DIR, 'languages')).map((f) => f.slice('prism-'.length).slice(0, -'.js'.length)).sort();
 const pluginIds = fs.readdirSync(path.join(SRC_DIR, 'plugins')).sort();
 
-async function loadComponent(id: string) {
-	let file;
-	if (pluginIds.includes(id)) {
-		file = path.join(SRC_DIR, `plugins/${id}/prism-${id}.ts`);
-	} else {
-		file = path.join(SRC_DIR, `languages/prism-${id}.ts`);
-	}
-	const exports = (await import(file)) as { default: ComponentProto };
-	return exports.default;
-}
+// async function loadComponent(id: string) {
+// 	let file;
+// 	if (pluginIds.includes(id)) {
+// 		file = path.join(SRC_DIR, `plugins/${id}/prism-${id}.ts`);
+// 	} else {
+// 		file = path.join(SRC_DIR, `languages/prism-${id}.ts`);
+// 	}
+// 	const exports = (await import(file)) as { default: ComponentProto };
+// 	return exports.default;
+// }
 
 async function minifyCSS() {
 	const input: Record<string, string> = {};
@@ -42,7 +44,7 @@ async function minifyCSS() {
 	for (const id of pluginIds) {
 		const file = path.join(SRC_DIR, `plugins/${id}/prism-${id}.css`);
 		if (fs.existsSync(file)) {
-			input[`plugins/prism-${id}.css`] = file;
+			input[`plugins/${id}/prism-${id}.css`] = file;
 		}
 	}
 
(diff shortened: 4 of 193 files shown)


package.json
```diff
@@ -2,14 +2,16 @@
 	"name": "prismjs",
 	"version": "1.29.0",
 	"description": "Lightweight, robust, elegant syntax highlighting. A spin-off project from Dabblet.",
-	"main": "prism.js",
+  "type": "module",
+	"main": "dist/core/prism.js",
```
trusktr, line 6: Opt into the modern Node ESM standard, and point to the main file to get Prism from (there was no more `./prism.js`).

package.json
```diff
@@ -2,14 +2,16 @@
...
 	"style": "themes/prism.css",
 	"engines": {
 		"node": ">=14"
 	},
 	"scripts": {
 		"benchmark": "ts-node benchmark/benchmark.ts",
-		"build": "ts-node scripts/build.ts",
+		"build": "ts-node-esm scripts/build.ts && npm run tsc",
```
trusktr, line 13: Running with `ts-node-esm` runs files as Node ESM, rather than CommonJS.

package.json
```diff
@@ -2,14 +2,16 @@
...
+		"prepare": "npm run build",
```
trusktr, line 14: This is from #3699

package.json
```diff
@@ -23,7 +25,8 @@
 		"test:plugins": "ts-mocha tests/plugins/**/*.ts",
 		"test:runner": "ts-mocha tests/testrunner-tests.ts",
 		"test": "npm-run-all test:*",
-		"tsc": "tsc && tsc -p tests/tsconfig.json"
+    "typecheck": "tsc --noEmit && tsc -p tests/tsconfig.json --noEmit",
+		"tsc": "tsc"
```
trusktr, line 29: We use a plain and simple `tsc` step to build dist/ output. The new `typecheck` script replaces the old `tsc` script for type checking without emitting output.

We should probably cover the other TS files as well, currently this covers only src/ and tests/.

.eslintrc.cjs
```diff
@@ -15,7 +15,7 @@ module.exports = {
 		'eqeqeq': ['error', 'always', { 'null': 'ignore' }],
 
 		// imports
-		'import/extensions': ['warn', 'never'],
+		'import/extensions': ['warn', 'always'], // "always" is most compatible with default vanilla ES Modules
```
trusktr, line 18: Having `.js` extensions, even if not to taste, is needed for the simplest ESM setup, and TypeScript officially recommends writing `.js` directly in TypeScript import statements for this purpose. So, we force it here so we don't mess up.

If we remove the `.js` extensions, the vanilla ESM example over at https://github.com/trusktr/prism-v2-test will break.

.eslintrc.cjs
```diff
@@ -15,7 +15,7 @@ module.exports = {
 		'eqeqeq': ['error', 'always', { 'null': 'ignore' }],
```
trusktr, line 15: This file is the only remnant of CommonJS in the repo (apart from the dependencies which I haven't updated yet).

In the upcoming ESLint v9, the new flat config format will also be in Node ESM format, and we'll have to move ESLint config to ESM at that point.

.mocharc.json
```diff
@@ -0,0 +1,7 @@
+{
+	"Xrequire": "@babel/register",
+		"Xexperimental-specifier-resolution=node": "",
```
trusktr, line 3: - [ ] cleanup

.mocharc.json
```diff
@@ -0,0 +1,7 @@
...
+	"node-option": [
+		"loader=ts-node/esm"
```
trusktr, line 5: This makes `ts-mocha` run our files as Node ESM files.

scripts/build.ts
```diff
@@ -1,34 +1,36 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
-import rollupTerser from '@rollup/plugin-terser';
-import rollupTypescript from '@rollup/plugin-typescript';
+// import rollupTerser from '@rollup/plugin-terser';
+// import rollupTypescript from '@rollup/plugin-typescript';
```
trusktr, line 3: - [ ] cleanup: delete unused code

scripts/build.ts
```diff
@@ -116,119 +118,119 @@ async function treeviewIconFont() {
...
+// 		const data = await Promise.all([...languageIds, ...pluginIds].map(async (id) => {
+// 			const proto = await loadComponent(id);
+// 			return { id, alias: toArray(proto.alias) };
+// 		}));
+// 		return Object.fromEntries(data.flatMap(({ id, alias }) => alias.map((a) => [a, id])));
+// 	},
+// 	all_languages_placeholder: () => Promise.resolve(languageIds),
+// 	title_placeholder: async () => {
+// 		const rawTitles = new Map<string, string>();
+// 		for (const [id, entry] of Object.entries(components.languages)) {
```
trusktr, line 132: - [ ] todo: figure if/how to handle components.json. For now I simply disabled `buildJS`.

scripts/build.ts
```diff
@@ -42,7 +44,7 @@ async function minifyCSS() {
 	for (const id of pluginIds) {
 		const file = path.join(SRC_DIR, `plugins/${id}/prism-${id}.css`);
 		if (fs.existsSync(file)) {
-			input[`plugins/prism-${id}.css`] = file;
+			input[`plugins/${id}/prism-${id}.css`] = file;
```
trusktr, line 47: Here we output the `.css` files so they are in the same location in `dist/` as they are in `src/`.

Less confusion: For anything in `src/**/*`, just replace `src/` with `dist/` to arrive at the output path. Easier mental model.

tests/components-test.ts
```diff
@@ -1,9 +1,11 @@
 import { assert } from 'chai';
 import { readFileSync } from 'fs';
 import path from 'path';
-import { forEach, noop, toArray } from '../src/shared/util';
-import { getComponent, getComponentIds, getLanguageIds } from './helper/prism-loader';
+import { fileURLToPath } from 'url';
+import { forEach, noop, toArray } from '../src/shared/util.js';
+import { getComponent, getComponentIds, getLanguageIds } from './helper/prism-loader.js';
 
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
trusktr, line 8: `__dirname` is not available in Node ESM files. Instead we have `import.meta.url` from which to base path handling.

tsconfig.json
```diff
@@ -25,35 +25,31 @@
     // "moduleDetection": "auto",                        /* Control what method is used to detect module-format JS files. */
 
     /* Modules */
-    "module": "commonjs",                                /* Specify what module code is generated. */
+    "module": "ESNext",                                  /* Specify what module code is generated. */
```
trusktr, line 28: Switch to ESM output format. All JS engines support top-level await (esnext).

tsconfig.json
```diff
@@ -25,35 +25,31 @@
...
     // "rootDir": "./",                                  /* Specify the root folder within your source files. */
-    // "moduleResolution": "node",                       /* Specify how TypeScript looks up a file from a given module specifier. */
+    "moduleResolution": "node",                          /* Specify how TypeScript looks up a file from a given module specifier. */
```
trusktr, line 30: This is not part of vanilla ESM, but it is very common to have this set to `node` even for vanilla web projects, using import maps as needed, because this is the default and simplest way in which type definitions flow into code in the ESM format (TS automatically grabs types from node_modules, which is the most common way people currently install libraries).

This covers non-relative bare module specifiers in our tests (f.e. `import ... from 'some-lib'`), while code inside of `src/` does not actually have any such library imports. If code in `src/` did have library imports, I would have needed to add an `importmap` to my example at https://github.com/trusktr/prism-v2-test.

Screenshot 2023-07-05 at 5 35 39 PM

tsconfig.json
```diff
@@ -25,35 +25,31 @@
...
     // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
     // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
     // "rootDirs": [],                                   /* Allow multiple folders to be treated as one when resolving modules. */
     // "typeRoots": [],                                  /* Specify multiple folders that act like './node_modules/@types'. */
     // "types": [],                                      /* Specify type package names to be included without being referenced in a source file. */
-    // "allowUmdGlobalAccess": true,                     /* Allow accessing UMD globals from modules. */
+    "allowUmdGlobalAccess": false,                        /* Allow accessing UMD globals from modules. */
```
trusktr, line 36: Globals are typically considered bad. Disable type defs for globals by default, so that they don't sneak into our scope unintentionally from node_modules. Add them explicitly when/if needed.

tsconfig.json
```diff
@@ -25,35 +25,31 @@
...
     // "moduleSuffixes": [],                             /* List of file name suffixes to search when resolving a module. */
     // "resolveJsonModule": true,                        /* Enable importing .json files. */
     // "noResolve": true,                                /* Disallow 'import's, 'require's or '<reference>'s from expanding the number of files TypeScript should add to a project. */
 
-    /* JavaScript Support */
-    "allowJs": true,                                  /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */
-    "checkJs": true,                                  /* Enable error reporting in type-checked JavaScript files. */
```
trusktr, line 43: There's no JS, not needed.

tsconfig.json
```diff
@@ -25,35 +25,31 @@
...
-    // "maxNodeModuleJsDepth": 1,                        /* Specify the maximum folder depth used for checking JavaScript files from 'node_modules'. Only applicable with 'allowJs'. */
-
     /* Emit */
-    // "declaration": true,                              /* Generate .d.ts files from TypeScript and JavaScript files in your project. */
-    // "declarationMap": true,                           /* Create sourcemaps for d.ts files. */
+    "declaration": true,                                 /* Generate .d.ts files from TypeScript and JavaScript files in your project. */
+    "declarationMap": true,                              /* Create sourcemaps for d.ts files. */
     // "emitDeclarationOnly": true,                      /* Only output d.ts files and not JavaScript files. */
-    // "sourceMap": true,                                /* Create source map files for emitted JavaScript files. */
+    "sourceMap": true,                                   /* Create source map files for emitted JavaScript files. */
```
trusktr, line 45: `declaration` plus `declarationMap` enables the feature where Go-To-Definition leads back to source code instead of declaration files. Really nice!

`sourceMap` enables source mapping for anyone who will further process the files. For example if they ever import `from 'prismjs'` and then compile their app with Webpack or Rollup, it is nice to be able to see errors/etc relative to the original source positions. If someone ever reports a bug, this will be handy because they can show the original location and source in the bug report, etc.

tsconfig.json
```diff
@@ -25,35 +25,31 @@
...
-    // "outDir": "./",                                   /* Specify an output folder for all emitted files. */
+    "outDir": "dist",                                    /* Specify an output folder for all emitted files. */
     // "removeComments": true,                           /* Disable emitting comments. */
-    "noEmit": true,                                   /* Disable emitting files from a compilation. */
+    // "noEmit": true,                                   /* Disable emitting files from a compilation. */
     // "importHelpers": true,                            /* Allow importing helper functions from tslib once per project, instead of including them per-file. */
-    // "importsNotUsedAsValues": "remove",               /* Specify emit/checking behavior for imports that are only used for types. */
+    // "importsNotUsedAsValues": "error",                /* Specify emit/checking behavior for imports that are only used for types. */
+	"verbatimModuleSyntax": true,
```
trusktr, line 52: `importsNotUsedAsValues` is deprecated. `verbatimModuleSyntax` is the new option. The rule is simple:

Any import with a `type` label gets removed. The rest is left behind. This is the least confusing compared to before: whatever imports you write are what you get (minus type imports). No room for error. It is now our responsibility to write or not write the imports we want, without TypeScript trying to guess.

Someone can use a tool like Webpack or Rollup to apply module tree shaking at their own risk of dealing with breakage, if they want to.

tsconfig.json
```diff
@@ -70,7 +66,7 @@
 
     /* Interop Constraints */
     // "isolatedModules": true,                          /* Ensure that each file can be safely transpiled without relying on other imports. */
-    // "allowSyntheticDefaultImports": true,             /* Allow 'import x from y' when a module doesn't have a default export. */
+    "allowSyntheticDefaultImports": true,                /* Allow 'import x from y' when a module doesn't have a default export. */
```
trusktr, line 69: Required for compatibility with Node ESM. In Node ESM, CommonJS `module` is always the `default` export, which is what this enforces.

trusktr:
Alrighty, I commented in all the relevant files to explain the changes. The rest of the vast majority of uncommented files are all just format updates (adding `.js` extensions, tweaking one or two imports for Node ESM compat, fixing `__dirname`).

PR #3814, opened 2024-07-21
Create generator-generic-ossf-slsa3-publish.yml

amin1419:


Querela:
This is just some unmodified template that basically contributes nothing.

PR #3821, opened 2024-08-19
Add inline comment syntax to liquid

vrugtehagel:
This PR adds support for `{% # inline comment %}` syntax in Liquid.


vrugtehagel:
Never mind, I should read READMEs 😛 this repository no longer accepts PR, at least not until v2 releases (if ever)

PR #3863, opened 2025-03-04
check that currentScript is set by a script tag

lkuechler:
Should resolve: https://osv.dev/vulnerability/GHSA-x7hr-w5r2-h6wg
Based on: https://gist.github.com/jackfromeast/aeb128e44f05f95828a1a824708df660


components/prism-core.js
```diff
@@ -222,7 +222,7 @@ var Prism = (function (_self) {
 				if (typeof document === 'undefined') {
 					return null;
 				}
-				if ('currentScript' in document && 1 < 2 /* hack to trip TS' flow analysis */) {
+				if ('currentScript' in document && 1 < 2 && document.currentScript && document.currentScript.tagName.toUpperCase() === 'SCRIPT' /* hack to trip TS' flow analysis */) {
```
LeaVerou, line 225: Seems like this would suffice?
```suggestion
				if (document.currentScript && 1 < 2 /* hack to trip TS' flow analysis */ && document.currentScript.tagName === 'SCRIPT') {
```

- Do we need both an `in` check and a `document.currentScript` check?
- In HTML `tagName` is always in uppercase per spec, see https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName 

or even just 

```suggestion
				if (document?.currentScript?.tagName === 'SCRIPT' && 1 < 2 /* hack to trip TS' flow analysis */ && ) {
```

Dhiraj-JC:
I appreciate the effort put into managing this open-source project. I'd love to get this PR merged and published. If any assistance is needed to move it forward, I'm happy to help.

Thanks! 😊

components/prism-core.js
lkuechler: Hi. Thanks for the review. The optional chaining was also something that I was considering but the project eslint already complained and I did not want to risk to limit the browser compatibility in any way compared to before. But you are right the check before was a bit unnecessary and should now be a bit more clean :)

LeaVerou: The first version of Prism was written in 2012. We can absolutely decide that certain browsers that were current then are out of scope in 2025. Folks can always use polyfills or pass it through babel.
I’m not saying we should go all cutting edge, but in this case, [any browser released in the last five years has supported optional chaining](https://caniuse.com/?search=optional-chaining).
And we need to depend on lookbehind anyway which already cuts down browser support to [this smaller set](https://caniuse.com/js-regexp-lookbehind)

lkuechler: All very good points but probably not something that should be changed when fixing a security issue :)
Is there still something open in the PR that is blocking it from being merged that I should change?

lkuechler: I added this PR that includes the same changes but with optional chaining and also the changes needed to make eslint not complain anymore: https://github.com/PrismJS/prism/pull/3869

I am okay with either being merged. If I can support in any other way to get this security issue solved I am willing to help out :)

LeaVerou (approved):
Indeed, I had missed that this was against v1 (but totally fine with fixing a security issue on v1). Yes, we do want to keep v1's browser support as-is, so I'd be inclined to merge this as-is, and repurpose https://github.com/PrismJS/prism/pull/3869 to be against v2. Would you be ok with doing that @lkuechler?

DmitrySharabin:
> Should this PR still target the master branch or should it target the v1 branch?

Please target the master `branch`. We removed the `v1` branch. Sorry for the confusion.

lkuechler:
@DmitrySharabin The PR is still focusing the master branch. Could you press the merge button? I do not have the permission to merge something :)

chadlwilson:
Thanks for the effort there, folks. Much appreciated.

Looking forward to a v1 release being cut!

carlosmintfan:
Looking forward for the fixed version to be published on npm!

LeaVerou:
@DmitrySharabin I just invited you as a maintainer to the prismjs npm package, once you accept can you release a new minor version?

DmitrySharabin:
> @DmitrySharabin I just invited you as a maintainer to the prismjs npm package, once you accept can you release a new minor version?

Done!

image

PR #3869, opened 2025-03-07
check that currentScript is set by a script tag with optional chaining

lkuechler:
Based on this PR but using optional chaining: https://github.com/PrismJS/prism/pull/3863


LeaVerou (changes requested):
Hi @lkuechler, would you be able to repurpose this to be against v2? See comment on #3863

lkuechler:
@LeaVerou I also applied the same changed to the v2 version now. There where a few more cases but I think I got all of them

PR #3876, opened 2025-03-11
Initial bit of prettier, with recommended .prettierrc

robert-j-webb:
Note: I only ran prettier on `src/core`. The reason for this is that this prettier config will mangle some code.

It is not the fault of a single prettier plugin, or the prettier version. Instead, it is when 3 plugins are used together that this issue happens.

It is kind of difficult to describe - you can see the error on the bash language definition. I just ran this:

`npx prettier src/languages/prism-bash.ts --write`

Additionally, I did not want to work super hard on forming eslint to the prettier config until we committed to this approach. I would love feedback.


src/core/hook-state.ts
```diff
@@ -6,11 +6,11 @@ export type StateKey<T> = (string | symbol) & { __keyType?: T };
 export class HookState {
 	private _data = new Map<string | symbol, {}>();
 
-	has(key: StateKey<{}>): boolean {
+	has (key: StateKey<{}>): boolean {
 		return this._data.has(key);
 	}
 
-	get<T extends {}>(key: StateKey<T>, defaultValue: T) {
+	get <T extends {}>(key: StateKey<T>, defaultValue: T) {
```
DmitrySharabin, line 13: This is a (known) bug in the prettier-plugin-space-before-function-paren plugin I still don't know how to fix. The correct formatting should be: `get<T extends {}> (key: StateKey<T>, defaultValue: T)`.

src/core/linked-list.ts
```diff
@@ -36,11 +36,14 @@ export class LinkedList<T> {
...
-	addAfter(node: LinkedListHeadNode<T> | LinkedListMiddleNode<T>, value: T): LinkedListMiddleNode<T> {
+	addAfter (
+		node: LinkedListHeadNode<T> | LinkedListMiddleNode<T>,
+		value: T,
+	): LinkedListMiddleNode<T> {
 		// assumes that node != list.tail && values.length >= 0
 		const next = node.next;
 
-		const newNode = { value, prev: node, next };
+		const newNode = { value , prev : node , next  };
```
DmitrySharabin, line 46: This also seems like a bug with prettier-plugin-space-before-function-paren. I'll take a look at that.

DmitrySharabin: Fixed in [`v0.0.6`](https://www.npmjs.com/package/prettier-plugin-space-before-function-paren) of the plugin.

DmitrySharabin:
> It is kind of difficult to describe - you can see the error on the bash language definition. I just ran this:
> 
> `npx prettier src/languages/prism-bash.ts --write`

It seems like this file is broken. For example, on line 166, there is not enough code. Here is what is supposed to be there:
https://github.com/PrismJS/prism/blob/76dde18a575831c91491895193f56081ac08b0c5/components/prism-bash.js#L154-L161

And everything starting from line 243 should be deleted: it's a fragment of the code above.

With these two changes, the file will be read and parsed correctly.

.prettierrc
```diff
@@ -0,0 +1,17 @@
...
+		"prettier-plugin-brace-style",
+		"prettier-plugin-space-before-function-paren",
+		"prettier-plugin-merge"
+	],
+	"braceStyle": "stroustrup",
+	"arrowParens": "avoid",
+	"bracketSpacing": true,
+	"endOfLine": "auto",
+	"semi": true,
+	"singleQuote": false,
```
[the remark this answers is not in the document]
robert-j-webb: Yes my nature is to prefer single quotes too, but I wasn't sue.

I copied the prettier config from here: https://github.com/color-js/color.js/blob/main/.prettierrc.json

src/core/linked-list.ts
robert-j-webb: Updated and it does fix, thanks

robert-j-webb:
> > It is kind of difficult to describe - you can see the error on the bash language definition. I just ran this:
> > `npx prettier src/languages/prism-bash.ts --write`
> 
> It seems like this file is broken. For example, on line 166, there is not enough code. Here is what is supposed to be there:
> 
> https://github.com/PrismJS/prism/blob/76dde18a575831c91491895193f56081ac08b0c5/components/prism-bash.js#L154-L161
> 
> And everything starting from line 243 should be deleted: it's a fragment of the code above.
> 
> With these two changes, the file will be read and parsed correctly.

Yes now it's working. I think the  v0.0.5 `prettier-plugin-space-before-function-paren` version I was using was causing the mangling.

.prettierrc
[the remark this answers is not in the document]
DmitrySharabin: Looking at the code base, I think sticking to single quotes makes perfect sense. In our other projects, though, we use double quotes. Anyway, it's just a matter of preference and agreement in the team. 🙂

DmitrySharabin:
> I think the v0.0.5 `prettier-plugin-space-before-function-paren` version I was using was causing the mangling.

I wouldn't have expected it to be so severe. 😱 My test didn't show that, but I can't deny I might have missed something.

package.json
```diff
@@ -48,7 +49,7 @@
 		"@types/jsdom": "^20.0.0",
 		"@types/mocha": "^9.1.1",
 		"@types/node": "^20.2.5",
-		"@types/prettier": "^2.7.0",
+		"@types/prettier": "^2.7.3",
```
DmitrySharabin, line 52: Shouldn't we use the same version as Prettier, at least v3.x.y? Or is it not necessary in this case?

robert-j-webb: Good catch - I don't think it's necessary anymore since I switched over to running prettier sync.  It looks like `@prettier/sync` ships with its own type defintion.

tests/helper/util.ts
```diff
@@ -101,7 +101,7 @@ export function getTrailingSpaces(string: string): string {
 }
 
 export function formatHtml(html: string): string {
-	return Prettier.format(html, {
+	return synchronizedPrettier.format(html, {
```
DmitrySharabin, line 104: Prettier doesn't do its job here anymore? Why do we need to switch to another package?

robert-j-webb: Specifically I couldn't install the plugins:

```
 npm install -D prettier@^2 prettier-plugin-brace-style prettier-plugin-space-before-function-paren prettier-plugin-merge

npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: prismjs@1.29.0
npm error Found: prettier@2.8.8
npm error node_modules/prettier
npm error   dev prettier@"^2" from the root project
npm error
npm error Could not resolve dependency:
npm error peer prettier@">=3.0.0" from prettier-plugin-space-before-function-paren@0.0.6
npm error node_modules/prettier-plugin-space-before-function-paren
npm error   dev prettier-plugin-space-before-function-paren@"*" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
npm error
npm error
npm error For a full report see:
npm error /Users/robertwebb/.npm/_logs/2025-03-11T17_00_55_750Z-eresolve-report.txt
npm error A complete log of this run can be found in: /Users/robertwebb/.npm/_logs/2025-03-11T17_00_55_750Z-debug-0.log
```

DmitrySharabin: Oh, it's expected. `prettier-plugin-space-before-function-paren` [works with Prettier v3+ only](https://github.com/LeaVerou/prettier-plugin-space-before-function-paren).

Anyhow, those changes LGTM. I figure we need some input from our more experienced colleagues and are good to go.

Let's ping @LeaVerou, for example.

robert-j-webb: It's used in `tests/helper/test-case.ts` which exports a test runner that formats HTML before it is printed.

It's also used in `test/helper/prism-dom-util.ts` to create test assertion helpers that allow comparing formatted HTML with HTML after it's been highlighted by prism. This is used a lot for plugin tests. You can see it's usage by searching for `util.assert`

I can't really tell how much effort it would be to make it work for async prettier, it just seems like a lot, especially compared to using `@prettier/sync`.

LeaVerou: How about we keep the sync Prettier localized to this particular thing, so that the rest of the project is not held back by it? I.e. use regular Prettier everywhere else, and treat sync Prettier as a dependency of this particular part of the testsuite.

If it's _only_ used for formatting HTML, it probably can get away without any plugins, so the issues around plugins are moot, and keeping it localized to this makes it easier to eventually get rid of it.
We may even move the testsuite to another repo down the line, in which case that would go with it.

robert-j-webb: > How about we keep the sync Prettier localized to this particular thing, so that the rest of the project is not held back by it? I.e. use regular Prettier everywhere else, and treat sync Prettier as a dependency of this particular part of the testsuite.

I agree completely. Prettier sync *should not* cause any issues but I wouldn't lean on it. 

> We may even move the testsuite to another repo down the line, in which case that would go with it.

I'd love to hear more about this idea, usually having tests being in the same commit as source changes is preferred since they tend to be tied to each other.

src/core/hook-state.ts
DmitrySharabin: It should be fixed in [`v0.0.7`](https://www.npmjs.com/package/prettier-plugin-space-before-function-paren). @robert-j-webb, could you check if it work as expected, please?

.prettierignore
```diff
@@ -0,0 +1 @@
+/website
```
LeaVerou, line 1: Isn't the website in a separate repo now anyway? And in general, it seems good practice to base `.prettierignore` on `.gitignore`. For now, we could do something like [this](https://stackoverflow.com/questions/65635648/how-can-i-base-a-prettierignore-file-on-gitignore) to start, and if we need to add additional files to `.prettierignore` we can write a build script that copies over `.gitignore` into it.

.prettierrc
```diff
@@ -0,0 +1,17 @@
...
+	"braceStyle": "stroustrup",
+	"arrowParens": "avoid",
+	"bracketSpacing": true,
+	"endOfLine": "auto",
+	"semi": true,
+	"singleQuote": true,
+	"tabWidth": 4,
+	"useTabs": true,
+	"trailingComma": "all",
+	"printWidth": 100
```
LeaVerou, line 16: I see from the PR that runs Prettier that it's adding quotes around properties whether they are needed or not. I would suggest using the `preserve` setting: https://prettier.io/docs/options#quote-props

src/core/hook-state.ts
LeaVerou: Can we make sure to pull the latest version of this before merging?

tests/helper/util.ts
```diff
@@ -101,7 +101,7 @@ export function getTrailingSpaces(string: string): string {
...
 		printWidth: 100,
 		tabWidth: 4,
 		useTabs: true,
```
LeaVerou, line 107: Non-blocking, but it would be nice if we could import our Prettier config here (and remove the plugins via JS since they don't work with sync prettier), instead of duplicating it.

LeaVerou (changes requested):
Finally found some time to review this, thanks everyone for your patience!

Some small tweaks and we're ready to go. I’ve left some comments on the diff.

Some additional comments:
- If we're using Prettier, we should remove any rules from ESLint that are covered by Prettier, or ditch it entirely. I'm fine with either of these options, but currently we have rules in `eslintrc` that are covered by Prettier (e.g. `curly`)

Not strictly related with this PR, just some things I noticed when going through the diff: 
-I think we should ditch babel & minification in v2. If folks want to transpile or minify they can do their own transformations (which for minification is AFAIK the currently recommended way, as minifying twice is a recipe for bugs)
- I wonder if we could switch the language definitions back to JS. Their use of types is minimal, and TS imposes barriers to contributions, and these are the most frequently contributed files.

.prettierignore
robert-j-webb: It actually already loads `.gitignore` and uses that by default. `Prettierignore` is added on top of that.

robert-j-webb: I'm leaving website out of formatting because it errors when prettier runs, and since we're going to be rewriting it anyway, it's not worth the struggle.

.prettierrc
robert-j-webb: Done

robert-j-webb:
Ok, most recent update - to make things less confusing, I removed any formatting of `src/core/*` - all of that will be in the follow up to make things more clear - if you want to see how formatting is changing files, check [out the other PR./(https://github.com/robert-j-webb/prism/pull/1/files) which I have just updated with the new config

Everything is now up to date and working! thanks so much for the feedback. 

> * If we're using Prettier, we should remove any rules from ESLint that are covered by Prettier, or ditch it entirely. I'm fine with either of these options, but currently we have rules in `eslintrc` that are covered by Prettier (e.g. `curly`)

Done!

tests/helper/util.ts
```diff
@@ -93,23 +101,19 @@ export function parseRegex(regex: RegExp): LiteralAST {
...
+export function getTrailingSpaces (string: string): string {
 	return /\s*$/.exec(string)?.[0] ?? '';
 }
 
-export function formatHtml(html: string): string {
-	return Prettier.format(html, {
-		printWidth: 100,
-		tabWidth: 4,
-		useTabs: true,
-		htmlWhitespaceSensitivity: 'ignore',
```
DmitrySharabin, line 108: I don’t see this rule in `.prettierrc`. I know it wasn’t there in the first place, but if we get rid of all this code, shouldn’t we add it there?

.prettierignore
DmitrySharabin: What do you think of removing this file after you formatted all the code if it only contains `website`?

tests/helper/util.ts
robert-j-webb: Yeah that's a good point, I'll add the option back

.prettierignore
robert-j-webb: > What do you think of removing this file after you formatted all the code if it only contains website?

Without this file, if you run `npx prettier . --write`, then prettier will have an error.  I wouldn't remove it until we removed the `/website` folder

DmitrySharabin: Working through the dependent PR, It dawned on me that we should also ignore `src/plugins`: we have already moved them to a separate repo. Sorry for not mentioning it before. 🙈

robert-j-webb: Yeah I can add ignore for plugins too. Plugins just worked so it never took any time to format them.

.eslintrc.js
```diff
@@ -2,17 +2,16 @@
...
 	parserOptions: {
 		ecmaVersion: 'latest',
-		sourceType: 'module'
+		sourceType: 'module',
 	},
 	rules: {
-		'no-use-before-define': ['error', { 'functions': false, 'classes': false }],
-		'eqeqeq': ['error', 'always', { 'null': 'ignore' }],
+		'no-use-before-define': ['error', { functions: false, classes: false }],
+		eqeqeq: ['error', 'always', { null: 'ignore' }],
```
DmitrySharabin, line 14: Do I get it right that quotes should be preserved here (and in some other places down there)? Or was this change made before we added the corresponding option to `.prettierrc`? Is there a chance that we should roll back all the changes and re-run prettier from scratch?

LeaVerou (approved):
LGTM! Thanks for your patience everyone. 

Non-blocking: We may want to add a `.vscode/settings.json` with:

```json
{
	"editor.formatOnSave": true,
	"editor.defaultFormatter": "esbenp.prettier-vscode",
	"prettier.enable": true,
}
```

That way, folks that don't have Prettier enabled by default, can still use it.

.eslintrc.js
DmitrySharabin: Other than that, LGTM!

robert-j-webb: Yes I will do that, great find

PR #3878, opened 2025-03-26
Ran prettier on all files

robert-j-webb:
Here's the original PR that is in my private fork: https://github.com/robert-j-webb/prism/pull/1


.github/workflows/test.yml
```diff
@@ -1,84 +1,79 @@
...
-        with:
-          node-version: 14.x
-      - run: npm ci
-      - run: npm run regex-coverage
+    tests:
+        runs-on: ubuntu-latest
+
+        strategy:
+            matrix:
+                node-version: [16.x, 18.x, 20.x]
```
robert-j-webb, line 18: On node version 14.x, I was seeing an error due to `prettier-syncrhonized`. I'm not sure why it wasn't caught in the previous PR.

The issue is, that it was using the `??=` operator [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment) which isn't supported in v14
 
I am not sure how many people are using node v14. The main result of this, is they wouldn't be able to run prettier tests. 


This PR actually changes the support matrix from 14,16,18 -> 16,18,20, but it's really just a suggestion. We have other options

tests/helper/util.ts
```diff
@@ -111,6 +111,7 @@ export function getTrailingSpaces (string: string): string {
 export function formatHtml (html: string): string {
 	return synchronizedPrettier.format(html, {
 		...synchronizedPrettier.resolveConfig('./prettierrc'),
+		parser: 'html',
```
robert-j-webb, line 114: I actually needed this line! tests were not passing without it.

.github/dependabot.yml
```diff
@@ -1,6 +1,6 @@
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
-    schedule:
-      interval: "weekly"
+    - package-ecosystem: 'github-actions'
```
LeaVerou, line 3: The indentation for YAML seems weird. Possible misconfiguration?

themes/prism.css
```diff
@@ -4,8 +4,8 @@
  * @author Lea Verou
  */
 
-code[class*="language-"],
-pre[class*="language-"] {
+code[class*='language-'],
```
LeaVerou, line 7: Let's add an override for CSS files? It seems double quotes are more common (which makes sense, as they mirror HTML, where double quotes are far more common).

LeaVerou (changes requested):
Took a quick look, left some comments! Very few things. 
@DmitrySharabin can you take a look as well? Since this is a pretty big PR, it's good to have another set of eyes.

.github/dependabot.yml
DmitrySharabin: I'd suggest overriding some rules for YAML files: it seems like two-space indentation is much more common. We won't get that many changes to those files if we do. I tried this and got no weird indentation:

```json
"overrides": [
	{
		"files": ["*.yml", "*.yaml"],
		"options": {
			"tabWidth": 2
		}
	}
]
```

DmitrySharabin:
Okay, I reviewed all the changes. Except for a couple of comments, LGTM.

I noticed that even though there seem to be a lot of changes in the diff, they are not so drastic in most cases. Especially in the language definitions, it's mostly moving code to a new line, trailing commas, spaces, and braces.

.github/dependabot.yml
robert-j-webb: Done!

themes/prism.css
robert-j-webb: Done!

src/shared/meta/alias-data.ts
```diff
@@ -3,8 +3,9 @@
  *
  * Note: This record is automatically generated by the build process.
  */
-const knownAliases: Readonly<Partial<Record<string, string>>> = /* aliases_placeholder[ */{}/* ] */;
+const knownAliases: Readonly<Partial<Record<string, string>>> =
+	/* aliases_placeholder[ */ {}; /* ] */
```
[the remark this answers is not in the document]
DmitrySharabin: Awesome! Thank you so much for having looked into it. Since the possible change is orthogonal to this PR, I wouldn't bother about it for now. If Prettier doesn't make the code broken, I'm happy with that. 🙂

DmitrySharabin (approved):
LGTM!

LeaVerou (approved):

PR #3884, opened 2025-04-03
[Action] Rebuild the website on push

DmitrySharabin:
Use Netlify build hook.

### DIFF
--- a/.github/workflows/website-rebuild.yml
+++ b/.github/workflows/website-rebuild.yml
@@ -1,0 +1,14 @@
+name: Rebuild Website
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


LeaVerou (commented):
I think Netlify build hooks are meant to be private, otherwise someone could write a script that triggers it every second for example.

DmitrySharabin:
> I think Netlify build hooks are meant to be private, otherwise someone could write a script that triggers it every second for example.

Your concern makes perfect sense. I removed the previous hook and made a new one a repo secret, so it's not exposed anymore. Could you please review the PR one more time?

PR #3885, opened 2025-04-03
Update documentation for auto-start and prism.ts

robert-j-webb:


src/core/prism.ts
```diff
@@ -162,7 +162,13 @@ export class Prism {
...
 	 * @returns The highlighted HTML.
 	 * @example
-	 * Prism.highlight('var foo = true;', 'javascript');
+	 * import { Prism } from './src/core/prism';
+	 * import javascript from './src/languages/prism-javascript';
+	 * const prism = new Prism();
+	 * prism.components.add(javascript);
+	 * prism.highlight('var foo = true;', 'javascript')
+	 * // Returns:
+	 * `<span class="token keyword">var</span> foo <span class="token operator">=</span> <span class="token boolean">true</span><span class="token punctuation">;</span>`
```
DmitrySharabin, line 171: I wonder if it makes sense also to use modern JS in examples, like here, to use `let` instead of `var`?

src/auto-start.ts
```diff
@@ -4,23 +4,27 @@ import autoloader from './plugins/autoloader/prism-autoloader';
...
 	 *
 	 * By setting this value to `true`, Prism will not automatically highlight all code elements on the page.
 	 *
-	 * You obviously have to change this value before the automatic highlighting started. To do this, you can add an
-	 * empty Prism object into the global scope before loading the Prism script like this:
+	 * You have to change this value before the automatic highlighting starts. To do this, you can add an
+	 * empty Prism object into the global scope before the Prism script loads like this:
 	 *
 	 * ```js
 	 * window.Prism = window.Prism || {};
```
DmitrySharabin, line 18: Do we still rely on `window`? Shouldn't we use `globalThis` here or something?

src/auto-start.ts
```diff
@@ -4,23 +4,27 @@ import autoloader from './plugins/autoloader/prism-autoloader';
...
 	 * Prism.manual = true;
 	 * // add a new <script> to load Prism's script
 	 * ```
 	 *
+	 * or you can set `data-manual` on the script tag used to load PrismJs:
+	 * ```html
+	 * <script src="prism.js" data-manual></script>
```
DmitrySharabin, line 25: It feels like examples of that kind depend on what Prism would look like in v2. I thought it should become ESM-first, no?

src/core/prism.ts
```diff
@@ -162,7 +162,13 @@ export class Prism {
 	 * Usually a language definition like `Prism.languages.markup`.
 	 * @returns The highlighted HTML.
 	 * @example
-	 * Prism.highlight('var foo = true;', 'javascript');
+	 * import { Prism } from './src/core/prism';
```
DmitrySharabin, line 165: I might miss some essential things, but those imports should be shorter if I could say so. Like, `import Prism, { Token } from "prismjs"`, `import javascript from "prismjs/languages"`. Is this something we should fix in `package.json`? Do I understand correctly that we use these examples to show how to use Prism as an NPM package?

src/core/prism.ts
```diff
@@ -199,14 +205,22 @@ export class Prism {
...
+	 * const prism = new Prism();
+	 * prism.components.add(javascript);
+	 *
+	 * const tokens = prism.tokenize(`var foo = 0;`, prism.components.getLanguage('javascript')!);
+	 * tokens.forEach((token: Token | string) => {
+	 *   if (token instanceof Token && token.type === 'number') {
+	 *     console.log(`Found numeric literal: ${token.content}`);
+	 *   }
 	 * });
+	 * // Logs: Found numeric literal: 0
```
DmitrySharabin, line 221: I'd mark it as a code block (```ts in this case). That way, TypeDoc will know this snippet should be highlighted appropriately.

src/core/prism.ts
```diff
@@ -199,14 +205,22 @@ export class Prism {
...
-	 *         console.log(`Found numeric literal: ${token.content}`);
-	 *     }
+	 * import { Token } from './src/core';
+	 * import { Prism } from './src/core/prism';
+	 * import javascript from './src/languages/prism-javascript';
+	 *
+	 * const prism = new Prism();
+	 * prism.components.add(javascript);
+	 *
+	 * const tokens = prism.tokenize(`var foo = 0;`, prism.components.getLanguage('javascript')!);
```
DmitrySharabin, line 215: Nit, why not single quotes here instead of \` and \`?

src/auto-start.ts
```diff
@@ -4,23 +4,27 @@ import autoloader from './plugins/autoloader/prism-autoloader';
...
+	 * ```
+	 *
 	 * @default false
 	 * @public
```
DmitrySharabin, line 29: I'd remove it. If I'm not mistaken, the property visibility is already described in the code (if not specified, it's public by default). TypeDoc also [mentions](https://typedoc.org/documents/Tags._public.html) that this tag is redundant (and not desirable). And the result TypeDoc produces is not that good:
image

src/core/prism.ts
DmitrySharabin, line 171: I think
```
* ```ts
* import { Prism } from './src/core/prism';
* import javascript from './src/languages/prism-javascript';
* const prism = new Prism();
* prism.components.add(javascript);
* prism.highlight('var foo = true;', 'javascript')
* ```
* Returns:
* ```html
* <span class="token keyword">var</span> foo <span class="token operator">=</span> <span class="token boolean">true</span><span class="token punctuation">;</span>
* ```
```
produces a slightly better result
image


What do you think?

src/auto-start.ts
LeaVerou: You can’t get a reference to the script element that triggered loading in ESM. It would need to be an attribute placed anywhere (eg data-prism-manual) or we’d key on the fil name or we’d also support non-ESM imports (by having an entry point that uses dynamic import to load everything else). Or all of the above 😀

src/core/prism.ts
LeaVerou: On a higher leve, these are meant to be small snippets, not full code examples. I think it’s ok if this doesn’t include the import code, which is different depending on tooling/env.

LeaVerou (changes requested):
Hi there, thanks for working on this! I left some comments, and I see Dmitry did too. The most important one I think is that code examples above functions don’t need to be complete with imports etc, especially since import code differs by environment (eg build tools or no build tools, JS vs TS, etc).

src/auto-start.ts
robert-j-webb: This is a really good point, I'll try redoing the examples as an actual 3rd party project instead of doing it locally

PR #3889, opened 2025-04-12
Add types

DmitrySharabin:
Based on #3887 by @robert-j-webb 

Main proposals:

- Use `__dirname` (I'm surprised it doesn't work on your side; I don't see any issues with that; in case it's relevant, I'm using Node v22.14.0)
- Generate all types during the build (see `buildTypes()` in `scripts/build.ts`. Don't use Rollup (let it build bundles); use the TypeScript compiler directly.
- Reference types in `projects.json` to improve DX for Prism users and developers.
 
image

Robert, I thought I could send it as a PR directly to your PR, but it didn't work. Anyway, it's easier to see all the changes I propose in one place than in comments in your PR (it's a lot of code 😅). What do you think?

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -397,5 +397,5 @@ async function buildJS() {
 }
 
 runTask(
-	series(clean, parallel(parallel(buildTypes, buildJS), series(treeviewIconFont, minifyCSS)))
+	series(clean, parallel(buildTypes, buildJS, series(treeviewIconFont, minifyCSS)))
 );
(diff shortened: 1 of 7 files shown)


scripts/build.ts
robert-j-webb, line 1: I think we still need the `commonjs` plugin here for the CJS bundle

DmitrySharabin: You are probably right. I don't have a strong opinion here. :)

DmitrySharabin:
Close in favor of https://github.com/PrismJS/prism/pull/3887.

PR #3891, opened 2025-04-16
Ditch Babel, fix tests, and benchmark

DmitrySharabin:
### Summary

- Ditch Babel
- Ditch `ts-mocha` in favor of the original Mocha. Update packages, types, and scripts.
- Update the `yargs` package and adopt the updated API.
- Use the `v2` branch for benchmark.
- Change the default branch from `master` to `main`.
- Ditch non-existent files.
- Use the website files directly (we moved the website to a separate repo and will eventually remove the `website` folder).


DmitrySharabin:
Though we still need to fix CI.

tests/helper/prism-loader.ts
```diff
@@ -108,18 +110,28 @@ export function createPrismDOM (): PrismDOM<{}> {
 		const g = global as unknown as Record<string, unknown>;
```
robert-j-webb, line 110: FYI we actually don't need the typing here, `const g = global` still works for everything

tests/helper/prism-loader.ts
```diff
@@ -108,18 +110,28 @@ export function createPrismDOM (): PrismDOM<{}> {
...
+				writable: true,
 			});
+
+			undo = overwriteProps(g, globals);
 			fn();
 		}
 		finally {
 			undo?.();
+			// Clean up navigator property
+			delete (g as any).navigator;
```
robert-j-webb, line 134: You can do a little better here like: 
```ts
			delete (g as Partial<typeof global>).navigator;
```

But really it's not necessary, casting to `any` is totally fine here.

DmitrySharabin: Clever!

robert-j-webb (approved):

LeaVerou (approved):

PR #3893, opened 2025-04-18
Fix the `type-check` action

DmitrySharabin:
We ignore the non-existent capturing group TS error (by expecting it) where it’s irrelevant and preserve the existing `eslint-disable` rules. There was no way to stack the existing ESLint rule with the new one (they should be applied to the same line), so I placed the ESLint rule on the same line it disables.

I also bumped up the TS version. Because why not? It's a good chance to make the code base up-to-date step-by-step. 🙃

### DIFF
--- a/src/languages/prism-cshtml.ts
+++ b/src/languages/prism-cshtml.ts
@@ -94,33 +94,33 @@ export default {
 			(/[^<]/.source +
 				'|' +
 				// all tags that are not the start tag
-				// eslint-disable-next-line regexp/strict
-				/<\/?(?!\1\b)/.source +
+				// @ts-expect-error TS(2532): Ignore the non-existent capturing group error.
+				/<\/?(?!\1\b)/.source + // eslint-disable-line regexp/strict
 				tagContent +
 				'|' +
 				// nested start tag
 				nested(
-					// eslint-disable-next-line regexp/strict
-					/<\1/.source +
+					// @ts-expect-error TS(2532): Ignore the non-existent capturing group error.
+					/<\1/.source + // eslint-disable-line regexp/strict
 						tagAttrs +
 						/\s*>/.source +
 						'(?:' +
 						(/[^<]/.source +
 							'|' +
 							// all tags that are not the start tag
-							// eslint-disable-next-line regexp/strict
-							/<\/?(?!\1\b)/.source +
+							// @ts-expect-error TS(2532): Ignore the non-existent capturing group error.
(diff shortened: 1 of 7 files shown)


LeaVerou (approved):

src/languages/prism-cshtml.ts
```diff
@@ -94,33 +94,33 @@ export default {
 			(/[^<]/.source +
 				'|' +
 				// all tags that are not the start tag
-				// eslint-disable-next-line regexp/strict
-				/<\/?(?!\1\b)/.source +
+				// @ts-expect-error TS(2532): Ignore the non-existent capturing group error.
+				/<\/?(?!\1\b)/.source + // eslint-disable-line regexp/strict
```
robert-j-webb, line 98: These comments no longer do anything, if I remove them I get this as the output from lint: 

```

/Users/robertwebb/prism/src/languages/prism-cshtml.ts
   35:33  error  Unnecessary escape character: \[                                             regexp/no-useless-escape
   57:11  error  Unnecessary escape character: \.                                             regexp/no-useless-escape
   57:14  error  Unnecessary escape character: \[                                             regexp/no-useless-escape
   82:21  error  Unnecessary escape character: \/                                             regexp/no-useless-escape
   86:34  error  Unnecessary escape character: \/                                             regexp/no-useless-escape
   98:13  error  Invalid legacy octal escape sequence '\1'. Use a hexadecimal escape instead  regexp/strict
  104:8   error  Invalid legacy octal escape sequence '\1'. Use a hexadecimal escape instead  regexp/strict
  112:16  error  Invalid legacy octal escape sequence '\1'. Use a hexadecimal escape instead  regexp/strict
  118:11  error  Invalid legacy octal escape sequence '\1'. Use a hexadecimal escape instead  regexp/strict
  123:8   error  Invalid legacy octal escape sequence '\1'. Use a hexadecimal escape instead  regexp/strict
```

You had the right idea to change from `eslint-disable-next-line` to `eslint-disable-line` but somehow the eslint rule isn't erroring anymore?

DmitrySharabin: image

I think it does. See line 98 in your output. This is exactly what I see if I remove the `eslint-disable` rule. Or am I missing something?

DmitrySharabin: TS and ESLint see this error differently. So that's why we need both these rules.

robert-j-webb: Oh you're totally right, I misread the line numbers

robert-j-webb (approved):

PR #3894, opened 2025-04-18
Fix paths in the Autoloader plugin and add a build for `prism.js` (that can be used as in Prism v1)

DmitrySharabin:
### Summary

- ~Prism plugins moved (starting from v2) to `plugins.prismjs.com`~
- ~In `src/prism-autostart.ts`, we import the Autoloader plugin~
- ~To make it work, we need to import it from a URL (`https://plugins.prismjs.com/autoloader/prism-autoloader.js`)~
- ~We have an ESLint rule that doesn't allow importing from `.js` files, so we need to disable the line where we import the plugin~
- ~TS doesn't know the type of the imported plugin, so we need to declare it (in `src/url-imports.d.ts`)~
- ~We have a new file with types that need to be copied to the `types` folder, and the plugin URL should be added as an external dependency to the Rollup config~
- ~Since we need to change the `buildJS()` function, I decided to follow Lea's advice and use a data structure for bundles we generate and extract the common (repeating) parts. I also added another bundle for the future global build (commented out for now)~
- ~This PR depends on https://github.com/PrismJS/plugins/pull/23/files, where we tweak the language definitions path that the Autoloader plugin uses~

~Let's discuss whether this approach will work and how to make it work if it won't. 🙃~

~**EDIT:** Well, I know that it already works (I checked it with two local servers: one served the plugin, the other served Prism), I'm just curious if there's something fundamental I'm missing that should be done differently.~

I updated this summary based on the discussion below.

- As Lea mentioned, the autoloader functionality is commonly needed, so we should probably make it a core functionality.
I suggest leaving it a (core) plugin and letting it live with the Prism core in one repo (with other core plugins, like, e.g., the [Custom Class](https://plugins.prismjs.com/custom-class/) plugin). For now, the autoloader is used in `src/auto-start.ts`.
- This PR introduces another build (named _global_) that generates the `prism.js` file from `auto-start.ts`. With that build, we allow users to work with Prism v2 like they do with Prism v1—simply including two files (`prism.js` and a theme). The only difference is that starting from v2, they'd have to add `type="module"` to the `<script>` tag including `prism.js`.
You can see it in action in [this pen](https://codepen.io/dmitrysharabin/pen/xbbdXZb?editors=1000).
- To make the new build work, I had to fix some paths in the Autoloader plugin source code: The languages are no longer in the `components` folder. By default, they are in the `languages` folder, which is in the same folder as `prism.js`.
- Since we now have three different bundles, I followed Lea's advice and used a data structure for them.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -406,13 +406,15 @@ async function buildJS() {
 				dir: './dist/cjs',
 			},
 		},
-		// global: {
-		// 	rollupOptions: defaultRollupOptions,
-		// 	outputOptions: {
-		// 		...defaultOutputOptions,
-		// 		dir: './dist/global',
-		// 	},
-		// }
+		global: {
+			rollupOptions: {
+				...defaultRollupOptions,
+				input: {
+					'prism': path.join(SRC_DIR, 'auto-start.ts'),
+				}
+			},
+			outputOptions: defaultOutputOptions,
+		}
 	};
 
 	try {
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -329,7 +329,7 @@ async function buildTypes() {
 	await mkdir('./types');
 
 	// Copy existing type definitions
-	const typeFiles = ['types.d.ts', 'known-plugins.d.ts', 'url-imports.d.ts'];
+	const typeFiles = ['types.d.ts', 'known-plugins.d.ts'];
 
 	await Promise.all(
 		typeFiles.map(file => copyFile(path.join(SRC_DIR, file), path.join('./types', file)))
@@ -373,8 +373,6 @@ async function buildJS() {
 
 	const defaultRollupOptions: RollupOptions = {
 		input,
-		// TODO: Remove “v2.” from the URL once Prism v2 is released
-		external: ['https://v2.plugins.prismjs.com/autoloader/prism-autoloader.js'],
 		plugins: [rollupTypescript({ module: 'esnext' })],
 	};
 
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -15,7 +15,7 @@ import { toArray } from '../src/shared/util';
 import { components } from './components';
 import { parallel, runTask, series } from './tasks';
 import type { ComponentProto } from '../src/types';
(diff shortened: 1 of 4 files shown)


LeaVerou:
Also, the autoloader seems to be functionality that is so commonly needed, I wonder if we should just make it part of the core.

DmitrySharabin:
> Also, the autoloader seems to be functionality that is so commonly needed, I wonder if we should just make it part of the core.

Exactly! I'd do that. Otherwise, it looks like we're working around our own issues.


> Struggling a bit to wrap my head around what problem is being solved.

Ideally, this PR should make us a step closer to the following: including the script below in a page should make Prism work with _any language_ out of the box (if we suggest that `src/auto-start.ts` will be compiled to something accessible as `prism.js`):

```html
<script src="https://dev.prismjs.com/prism.js" type="module"></script>
```

But if we do that as part of the core, we can close this (draft) PR. :)

DmitrySharabin:
> One thing I wondered is, what if we had a convenience method that allows plugin ids to be specified for official plugins and takes care of importing them? Would that help?

We should probably start designing changes to the API and stop making any other changes. As my experiments show, Prism v2 already works and is usable. For example, this code correctly highlights all JS code on the page:

```html
<script type="module">
    import Prism from 'https://dev.prismjs.com/index.js';
    import javascript from 'https://dev.prismjs.com/languages/prism-javascript.js';
    let prism = new Prism();
    prism.components.add(javascript);
    prism.highlightAll();
</script>
```
So, we are ready to move on. The design decisions we make will determine all other changes.

LeaVerou:
> ```
> <script type="module">
>     import Prism from 'https://dev.prismjs.com/index.js';
>     import javascript from 'https://dev.prismjs.com/languages/prism-javascript.js';
>     let prism = new Prism();
>     prism.components.add(javascript);
>     prism.highlightAll();
> </script>
> ```


Yeah, this is an extremely heavyweight method to be the default. It would be okay for cases where tree-shaking is desired, but for the common case, there should be something much simpler.

LeaVerou (approved):

PR #3899, opened 2025-04-26
Ditch the `prism-` prefix from language filenames

DmitrySharabin:
### Summary

- Rename all language files
- Fix paths: `languages/prism-ID.ts` → `languages/ID.ts`


LeaVerou (approved):

LeaVerou:
We can probably do the same for languages.

DmitrySharabin:
> We can probably do the same for plugins.

Seconded. I was also considering it as one of the following steps.

PR #3902, opened 2025-04-26
Delete `bower.json`

LeaVerou:
No-one uses Bower anymore, even its own webpage advises folks not to use it:

image


DmitrySharabin (approved):

eighty4:
Don't trust anything that suggests using yarn!

PR #3905, opened 2025-04-26
[WIP] Simplify core

LeaVerou:


src/core/hooks.ts
```diff
@@ -60,102 +56,4 @@ export class Hooks {
...
-
-export interface WrapEnv {
-	type: TokenName;
-	content: string;
-	tag: string;
-	classes: string[];
-	attributes: Record<string, string>;
-	language: string;
-}
+export type HookCallback<HookName extends string> = (env: Record<string, any>) => void;
```
DmitrySharabin, line 59: Do we need `HookName` here? It doesn't seem to be used anywhere? Does callback depend on the hook name it's associated with?

src/core/hooks.ts
```diff
@@ -1,7 +1,3 @@
-import type { Grammar, TokenName } from '../types';
-import type { HookState } from './hook-state';
-import type { TokenStream } from './token';
-
 export class Hooks {
 	// eslint-disable-next-line func-call-spacing
 	private _all = new Map<string, ((env: unknown) => void)[]>();
```
DmitrySharabin, line 3: What if we switch to native private properties and methods? I think TS will still correctly infer the property visibility. Or is there a reason to stick with the `_` convention?

DmitrySharabin:
I can't comment on the corresponding line since it's not in the PR, but I can't understand why we need a `list` variable if we already have `hooks`. Moreover, we do `const list = hooks`. What am I missing?

```ts
add<Name extends string> (name: Name, callback: HookCallback<Name>): () => void {
	let hooks = this._all.get(name);
	if (hooks === undefined) {
		hooks = [];
		this._all.set(name, hooks);
	}
	const list = hooks;

	list.push(callback as never);

	return () => {
		const index = list.indexOf(callback as never);
		if (index !== -1) {
			list.splice(index, 1);
		}
	};
}
```

src/core/registry.ts
```diff
@@ -1,7 +1,7 @@
 import { extend } from '../shared/language-util';
 import { forEach, kebabToCamelCase } from '../shared/util';
 import type { ComponentProto, Grammar } from '../types';
-import type { Prism } from './prism';
+import type { Prism } from './prism-class';
```
DmitrySharabin, line 5:
```suggestion
import type Prism from './prism-class';
```

Another option would be:

```ts
import type { Prism } from './prism';
```

src/core.ts
```diff
@@ -1,2 +1,2 @@
-export { Prism } from './core/prism';
+export { Prism } from './core/prism-class';
```
DmitrySharabin, line 1:
```suggestion
export { default as Prism } from './core/prism-class';
```

src/global.ts
```diff
@@ -1,10 +1,8 @@
-import { Prism } from './core/prism';
+import globalPrism, { Prism } from './core/prism-class';
```
DmitrySharabin, line 1:
```suggestion
import globalPrism, { type Prism } from './core/prism';
```

src/global.ts
```diff
@@ -1,10 +1,8 @@
...
 
-const globalSymbol = Symbol.for('Prism global');
-
-// eslint-disable-next-line no-undef
-const namespace = globalThis as Partial<Record<typeof globalSymbol, Prism>>;
-const globalPrism = (namespace[globalSymbol] ??= new Prism());
+declare global {
+	var Prism: Prism | undefined
+}
```
DmitrySharabin, line 5:
```suggestion
	const Prism: Prism | undefined;
```

Is there a reason for using `var` I might miss?

src/types.d.ts
```diff
@@ -1,4 +1,4 @@
-import type { Prism } from './core/prism';
+import type { Prism } from './core/prism-class';
```
DmitrySharabin, line 1:
```suggestion
import type Prism from './core/prism-class';
```

src/core/prism-class.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Prism: Lightweight, robust, elegant syntax highlighting
+ *
+ * @license MIT <https://opensource.org/licenses/MIT>
+ * @author Lea Verou <https://lea.verou.me> and contributors <https://github.com/PrismJS/prism/graphs/contributors>
+ */
+export default class Prism {
+	hooks = new Hooks();
+	components = new Registry(this);
```
DmitrySharabin, line 19: It looks like we have only languages there (plugins have their own property). Why not use languages instead of components here?

src/core/prism-class.ts
```diff
@@ -0,0 +1,49 @@
...
+	 */
+	highlightElement (element: Element, options: HighlightElementOptions = {}) {
+		return highlightElement.call(this, element, options);
+	}
+
+	/**
+	 * See {@link highlight}
+	 */
+	highlight (text: string, language: string, options?: HighlightOptions): string {
+		return highlight.call(this, text, language, options);
```
DmitrySharabin, line 40:
```suggestion
	highlight (text: string, language: string, options: HighlightOptions = {}): string {
```

LeaVerou: I think this suggestion is on the wrong line

src/core/registry.ts
LeaVerou: We want to import from the lowest level module we can

src/core/prism-class.ts
LeaVerou: Agreed. Can you do it?

src/core/prism-class.ts
DmitrySharabin: It's. Weird ... I'll fix it

src/core/prism-class.ts
DmitrySharabin: Sure, will do

src/core/prism-class.ts
```diff
@@ -0,0 +1,49 @@
...
+	 * See {@link highlightElement}
+	 */
+	highlightElement (element: Element, options: HighlightElementOptions = {}) {
+		return highlightElement.call(this, element, options);
+	}
+
+	/**
+	 * See {@link highlight}
+	 */
+	highlight (text: string, language: string, options?: HighlightOptions): string {
```
DmitrySharabin, line 39:
```suggestion
	highlight (text: string, language: string, options: HighlightOptions = {}): string {
```

src/global.ts
LeaVerou: Nope, that's just what the code example I found for extending `global` was using 😛 But this suggestion is also on the wrong line.

src/core/hooks.ts
LeaVerou: No, it's unused.

src/core/hooks.ts
LeaVerou: `_all` indicates that it's private, but is still readable by any code doing some weird low-level thing. That said, it doesn't seem to be read by any other part of the codebase, so we could, though it also doesn't provide much of a benefit so I don't think it's particularly important, so I'm inclined to leave it as-is.

src/global.ts
```diff
@@ -1,10 +1,8 @@
-import { Prism } from './core/prism';
+import globalPrism, { type Prism } from './core/prism';
 
-const globalSymbol = Symbol.for('Prism global');
-
-// eslint-disable-next-line no-undef
-const namespace = globalThis as Partial<Record<typeof globalSymbol, Prism>>;
-const globalPrism = (namespace[globalSymbol] ??= new Prism());
+declare global {
+	var Prism: Prism | undefined
```
DmitrySharabin, line 4:
```suggestion
	const Prism: Prism | undefined;
```

src/core/stringify.ts
```diff
@@ -0,0 +1,58 @@
...
+	if (Array.isArray(o)) {
+		let s = '';
+		o.forEach(e => {
+			s += stringify(e, language, hooks);
+		});
+		return s;
+	}
+
+	const env: Record<string, any> & { classes: string[]; attributes: Record<string, string> } = {
+		type: o.type,
```
DmitrySharabin, line 29: `o` is a string. Does it have a `type` property?

src/core/stringify.ts
```diff
@@ -0,0 +1,58 @@
...
+		content: stringify(o.content, language, hooks),
+		tag: 'span',
+		classes: ['token', o.type],
```
DmitrySharabin, line 32: And here

src/core/stringify.ts
```diff
@@ -0,0 +1,58 @@
...
+		attributes: {},
+		language,
+	};
+
+	const aliases = o.alias;
```
DmitrySharabin, line 37: And here

src/core/stringify.ts
LeaVerou: It's not just a string. If it's a string, we've already returned in line 18.

src/core/highlight.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Usually a language definition like `Prism.languages.markup`.
+ * @returns The highlighted HTML.
+ * @example
+ * Prism.highlight('var foo = true;', 'javascript');
+ */
+export function highlight (
+	this: Prism,
+	text: string,
+	language: string,
+	options?: HighlightOptions
```
DmitrySharabin, line 28:
```suggestion
	options: HighlightOptions = {}
```

src/core/highlight.ts
```diff
@@ -0,0 +1,55 @@
...
+): string {
+	const prism = this ?? singleton;
+
+	const languageId = prism.components.resolveAlias(language);
+	const grammar = options?.grammar ?? prism.components.getLanguage(languageId);
```
DmitrySharabin, line 33:
```suggestion
	const grammar = options.grammar ?? prism.components.getLanguage(languageId);
```

We don't need this if we accept the previous suggestion.

src/core/stringify.ts
DmitrySharabin: Oh, right. Here, it's a `TokenStream` or a `Token`. My bad.

src/core/stringify.ts
```diff
@@ -0,0 +1,58 @@
...
+	const attributes =
+		Object.entries(env.attributes)
+			.map(([name, value]) => ` ${name}=${(value ?? '').replace(/"/g, '&quot;')}"`)
+			.join('') || '';
+
+	return `<${env.tag} class="${env.classes.join(' ')}"${attributes}>${env.content}</${env.tag}>`;
+}
+
+export { stringify };
+export default stringify;
```
DmitrySharabin, line 58: The answer to my following question will probably become TIL. Why do we need to export the same thing in two different ways? To make different ways of importing possible?

src/plugins/keep-markup/prism-keep-markup.ts
```diff
@@ -77,11 +76,11 @@ export default {
 
 				if (data.length) {
 					// data is an array of all existing tags
-					env.state.set(markupData, data);
+					env.markupData = data;
 				}
 			},
 			'after-highlight': (env) => {
-				const data = env.state.get(markupData, []);
+				const data = env.markupdata ?? [];
```
DmitrySharabin, line 83:
```suggestion
				const data = env.markupData ?? [];
```

src/core/stringify.ts
LeaVerou: Yes

src/known-plugins.d.ts
DmitrySharabin, line 1: We also need to delete a mention of this file from `scripts/build.ts` (line 332)

LeaVerou: Can you do it?

DmitrySharabin: Done

eslint.config.mjs
```diff
@@ -171,6 +171,10 @@ const config = [
 
 			'@typescript-eslint/no-unsafe-call': 'off',
 			'@typescript-eslint/no-explicit-any': 'off',
+			'@typescript-eslint/no-unsafe-assignment': 'off',
+			'@typescript-eslint/no-unsafe-member-access': 'off',
+			'@typescript-eslint/no-unsafe-return': 'off',
+			'@typescript-eslint/no-unsafe-argument': 'off',
```
DmitrySharabin, line 177: Can't comment on line 182 (it's not in the PR), but I'd suggest adding `src/config.ts` to the list of (core) files to calm ESLint down about the undefined `document`.

src/config.ts
```diff
@@ -0,0 +1,65 @@
+const hasDOM = typeof document !== 'undefined' && typeof window !== 'undefined';
+const scriptElement: HTMLScriptElement = hasDOM && document.currentScript;
+// @ts-ignore
+const globalConfig: Record<String, string | string[] | null> =
+	globalThis.Prism?.constructor?.name === 'Object' ? globalThis.Prism : {};
```
DmitrySharabin, line 5:
```suggestion
const globalConfig: Record<string, string | string[] | null> =
```

Ignore my suggestion if using uppercase `String` was intentional.

src/config.ts
```diff
@@ -0,0 +1,65 @@
...
+	languagePath?: string;
+}
+
+export const globalDefaults: PrismConfig = {
+	manual: getGlobalBooleanSetting('manual', !hasDOM),
+	languages: getGlobalArraySetting('languages'),
+	plugins: getGlobalArraySetting('plugins'),
+	languagePath: getGlobalSetting('language-path') ?? './languages/',
+	pluginPath: getGlobalSetting('plugin-path') ?? './plugins/',
+};
```
DmitrySharabin, line 63:
```suggestion
	languagePath: getGlobalSetting('language-path') as string | undefined ?? './languages/',
	pluginPath: getGlobalSetting('plugin-path') as string | undefined ?? './plugins/',
```

src/config.ts
```diff
@@ -0,0 +1,65 @@
+const hasDOM = typeof document !== 'undefined' && typeof window !== 'undefined';
+const scriptElement: HTMLScriptElement = hasDOM && document.currentScript;
+// @ts-ignore
+const globalConfig: Record<String, string | string[] | null> =
```
DmitrySharabin, line 4:
```suggestion
const globalConfig: Record<string, string | string[] | null> =
```

Ignore my suggestion if using `String` was intentional. If so, I'd love to know why. 🙂

src/config.ts
```diff
@@ -0,0 +1,65 @@
+const hasDOM = typeof document !== 'undefined' && typeof window !== 'undefined';
+const scriptElement: HTMLScriptElement = hasDOM && document.currentScript;
+// @ts-ignore
```
DmitrySharabin, line 3:
```suggestion
const scriptElement: HTMLOrSVGScriptElement | null = document?.currentScript;
```

From my perspective we get the same result with simpler types, no?

DmitrySharabin, line 3:
```suggestion
const scriptElement: HTMLOrSVGScriptElement | null = document?.currentScript;
```

From my perspective, we get the same result with a bit simpler types, no?

src/config.ts
```diff
@@ -0,0 +1,65 @@
+const hasDOM = typeof document !== 'undefined' && typeof window !== 'undefined';
+const scriptElement: HTMLScriptElement = hasDOM && document.currentScript;
```
DmitrySharabin, line 2:
```suggestion
const scriptElement: HTMLOrSVGScriptElement | null = document?.currentScript;
```

From my perspective, we get the same result with a bit simpler types, right?

src/config.ts
```diff
@@ -0,0 +1,65 @@
...
+	pluginPath?: string;
+	languagePath?: string;
+}
+
+export const globalDefaults: PrismConfig = {
+	manual: getGlobalBooleanSetting('manual', !hasDOM),
+	languages: getGlobalArraySetting('languages'),
+	plugins: getGlobalArraySetting('plugins'),
+	languagePath: getGlobalSetting('language-path') ?? './languages/',
+	pluginPath: getGlobalSetting('plugin-path') ?? './plugins/',
```
DmitrySharabin, line 62:
```suggestion
	languagePath: getGlobalSetting('language-path') as string | null ?? './languages/',
	pluginPath: getGlobalSetting('plugin-path') as string | null ?? './plugins/',
```

src/config.ts
```diff
@@ -0,0 +1,65 @@
...
+
+	if (camelCaseName in globalConfig) {
+		return globalConfig[camelCaseName];
+	}
+	else if (name in globalConfig) {
+		return globalConfig[name];
+	}
+	else if (hasDOM) {
+		return (
+			scriptElement.dataset[camelCaseName] ??
```
DmitrySharabin, line 18:
```suggestion
			scriptElement!.dataset[camelCaseName] ??
```

If I'm not mistaken, `scriptElement` will be defined here, but TS can't figure it out from the code, so we need to tell it we're sure with what we do.

src/core/classes/prism.ts
```diff
@@ -0,0 +1,132 @@
+import globalDefaults, { type PrismConfig } from './config';
```
DmitrySharabin, line 1:
```suggestion
import globalDefaults, { type PrismConfig } from '../../config';
```

src/config.ts
DmitrySharabin, line 5:
```suggestion
	globalThis.Prism?.constructor?.name === 'Object' ? globalThis.Prism.config : {};
```

src/config.ts
```diff
@@ -0,0 +1,65 @@
+const hasDOM = typeof document !== 'undefined' && typeof window !== 'undefined';
+const scriptElement: HTMLScriptElement = hasDOM && document.currentScript;
+// @ts-ignore
+const globalConfig: Record<String, string | string[] | null> =
+	globalThis.Prism?.constructor?.name === 'Object' ? globalThis.Prism : {};
+
+function getGlobalSetting (name: string) {
+	let camelCaseName = name.replace(/-([a-z])/g, g => g[1].toUpperCase());
```
DmitrySharabin, line 8:
```suggestion
	let camelCaseName = name.replace(/-([a-z])/g, g => g[1].toUpperCase()); // eslint-disable-line regexp/no-unused-capturing-group
```

ESLint still can't analyze the code properly. 😅

src/config.ts
```diff
@@ -0,0 +1,65 @@
...
+export interface PrismConfig {
+	manual?: boolean;
+	plugins?: string[];
+	languages?: string[];
+	pluginPath?: string;
+	languagePath?: string;
+}
+
+export const globalDefaults: PrismConfig = {
+	manual: getGlobalBooleanSetting('manual', !hasDOM),
```
DmitrySharabin, line 58:
```suggestion
	manual: getGlobalBooleanSetting('manual', !hasDOM),
	silent: getGlobalBooleanSetting('silent', false),
```

And here. `errorHandler` should probably stay `undefined`. It will be set in the Prism class constructor.

src/core/classes/prism.ts
```diff
@@ -0,0 +1,132 @@
...
+	waitFor: Promise<unknown>[] = [nextTick()];
+	ready: Promise<unknown> = allSettled(this.waitFor);
+
+	constructor (config: PrismConfig = {}) {
+		this.config = Object.assign({}, globalDefaults, config);
+		this.config.errorHandler ??= this.config.silent ? () => undefined : console.error;
+
+		const reportError: (reason: any) => PromiseLike<never> = this.config.errorHandler;
+
+		this.languageRegistry = new LanguageRegistry(this.config.languagePath ?? './languages/');
```
DmitrySharabin, line 38:
```suggestion
		this.languageRegistry = new LanguageRegistry(this.config.languagePath;
```

We probably don't need it since it's already in `globalDefaults`.

src/core/classes/prism.ts
```diff
@@ -0,0 +1,132 @@
...
+		this.pluginRegistry = new PluginRegistry(this.config.pluginPath ?? './plugins/');
```
DmitrySharabin, line 39:
```suggestion
		this.pluginRegistry = new PluginRegistry(this.config.pluginPath);
```

The same is here.

src/core/classes/prism.ts
```diff
@@ -0,0 +1,132 @@
...
+
+		// Preload languages
+		if (this.config.languages?.length > 0) {
+			this.languageRegistry.loadAll(this.config.languages);
```
DmitrySharabin, line 43:
```suggestion
			this.languageRegistry.loadAll(this.config.languages!);
```

TS can't get that `languages` is not `undefined` here.

src/core/classes/prism.ts
```diff
@@ -0,0 +1,132 @@
...
+		}
+
+		this.languagesReady = this.languageRegistry.ready;
+		this.waitFor.push(this.languagesReady);
+
+		if (this.config.plugins?.length > 0) {
+			let pluginsReady = this.languagesReady
+				.then(() => {
+					return this.pluginRegistry.loadAll(this.config.plugins);
```
DmitrySharabin, line 52:
```suggestion
					return this.pluginRegistry.loadAll(this.config.plugins!);
```

src/core/classes/prism.ts
```diff
@@ -0,0 +1,132 @@
...
+
+	config: PrismConfig;
+
+	languagesReady: Promise<unknown>;
+	waitFor: Promise<unknown>[] = [nextTick()];
+	ready: Promise<unknown> = allSettled(this.waitFor);
+
+	constructor (config: PrismConfig = {}) {
+		this.config = Object.assign({}, globalDefaults, config);
+		this.config.errorHandler ??= this.config.silent ? () => undefined : console.error;
```
DmitrySharabin, line 34:
```suggestion
		this.config.errorHandler ??= (
			this.config.silent ? () => undefined : console.error
		) as PrismConfig['errorHandler'];
```

The code is already formatted with Prettier.

src/core/classes/prism.ts
```diff
@@ -0,0 +1,132 @@
...
+
+		const reportError: (reason: any) => PromiseLike<never> = this.config.errorHandler;
```
DmitrySharabin, line 36:
```suggestion
		const reportError: PrismConfig['errorHandler'] = this.config.errorHandler;
```

src/core/classes/prism.ts
```diff
@@ -0,0 +1,132 @@
...
+
+		this.languageRegistry = new LanguageRegistry(this.config.languagePath ?? './languages/');
+		this.pluginRegistry = new PluginRegistry(this.config.pluginPath ?? './plugins/');
+
+		// Preload languages
+		if (this.config.languages?.length > 0) {
```
DmitrySharabin, line 42:
```suggestion
		// @ts-expect-error: TypeScript does not narrow the type of the languages property when using optional chaining in the condition.
		if (this.config.languages?.length > 0) {
```

TS drives me crazy sometimes. 🤪 We can either go the proposed way and ignore this error, since our code clearly shows our intention. Or we can help TS to narrow down types like so:

```ts
const languages = this.config.languages;
if (languages && languages.length > 0) {
	this.languageRegistry.loadAll(languages);
}
```

In that case, we can ignore the following suggestion (about adding `!`).

src/config.ts
```diff
@@ -0,0 +1,65 @@
...
+	if (value === null) {
+		return defaultValue;
+	}
+
+	return !(value === false || value === 'false');
+}
+
+function getGlobalArraySetting (name: string): string[] {
+	const value = getGlobalSetting(name);
+	if (value === null || value === false || value === 'false') {
```
DmitrySharabin, line 36:
```suggestion
	if (value === undefined || value === false || value === 'false') {
```

src/shared/hooks-util.ts
```diff
@@ -1,4 +1,4 @@
-import type { HookCallback, HookEnvMap, Hooks } from '../core/hooks';
+import type { HookCallback, HookEnvMap, Hooks } from '../core/classes/hooks';
```
DmitrySharabin, line 1: Is `HooksEnvMap` something to be added later? I can't find this type anywhere.

I can't add a suggestion to line 17, so show it here for not to lose:

```diff
- export function addHooks (hooks: Hooks, hooksMap: { [K in keyof HookEnvMap]?: HookCallback<K> }) {
+ export function addHooks (hooks: Hooks, hooksMap: { [K in keyof HookEnvMap]?: HookCallback }) {
```

src/core/classes/hooks.ts
```diff
@@ -0,0 +1,59 @@
+export class Hooks {
+	// eslint-disable-next-line func-call-spacing
```
DmitrySharabin, line 2:
```suggestion
```

No such error in the following line.

src/types.d.ts
```diff
@@ -18,10 +17,9 @@ export interface ComponentProtoBase<Id extends string = string> {
 export interface LanguageProto<Id extends string = string> extends ComponentProtoBase<Id> {
 	grammar: Grammar | ((options: GrammarOptions) => Grammar);
```
DmitrySharabin, line 18:
```suggestion
	grammar: Grammar | ((options?: GrammarOptions) => Grammar);
```

Otherwise, we can't do `grammar.call(this)` without passing mandatory parameters.

src/core/classes/language.ts
```diff
@@ -0,0 +1,78 @@
...
+
+			if (typeof grammar === 'object') {
+				// the grammar is a simple object, so we don't need to evaluate it
+				this.evaluatedGrammar = Object.assign(baseGrammar, grammar);
+			}
+			else if (typeof grammar === 'function') {
+				this.evaluatedGrammar = grammar.call(this);
+			}
+
+			if (this.evaluatedGrammar.$insertBefore) {
```
DmitrySharabin, line 44:
```suggestion
			if (this.evaluatedGrammar?.$insertBefore) {
```

src/core/classes/language.ts
```diff
@@ -0,0 +1,78 @@
...
+				for (let key in this.evaluatedGrammar.$insertBefore) {
+					const tokens = this.evaluatedGrammar.$insertBefore[key];
+					if (tokens) {
+						insertBefore(this.evaluatedGrammar, key, tokens);
+					}
+				}
+				delete this.evaluatedGrammar.$insertBefore;
+			}
+
+			if (this.evaluatedGrammar.$delete) {
```
DmitrySharabin, line 54:
```suggestion
			if (this.evaluatedGrammar?.$delete) {
```

src/core/classes/language.ts
```diff
@@ -0,0 +1,78 @@
...
+			// TODO extends should also be able to
+			let baseGrammar = def.extends ? { ...def.extends } : {};
+
+			if (typeof grammar === 'object') {
+				// the grammar is a simple object, so we don't need to evaluate it
+				this.evaluatedGrammar = Object.assign(baseGrammar, grammar);
+			}
+			else if (typeof grammar === 'function') {
+				this.evaluatedGrammar = grammar.call(this);
+			}
```
DmitrySharabin, line 42: We use `else if` here, which means that `this.evaluatedGrammar` might be `undefined` after this block. It `undefined` is a possible value, we should fix the returned type to `Grammar | undefined`.

Do I get it right that if the `grammar` getter returns `undefined`, something is wrong?

src/types.d.ts
```diff
@@ -114,7 +112,11 @@ export interface GrammarSymbols {
 	[rest]?: Grammar | string | null;
 	[tokenize]?: (code: string, grammar: Grammar, Prism: Prism) => TokenStream;
 }
-export type Grammar = GrammarTokens & GrammarSymbols;
+export interface GrammarSpecial {
+	$insertBefore?: GrammarTokens;
+	$delete: string[];
```
DmitrySharabin, line 117:
```suggestion
	$delete?: string[];
```

Otherwise, we can't do `delete grammar.$delete`.

src/languages/c.ts
```diff
@@ -5,9 +5,13 @@ import type { GrammarToken, LanguageProto } from '../types';
 export default {
 	id: 'c',
 	require: clike,
+	extends: clike,
 	optional: 'opencl-extensions',
-	grammar ({ extend, getOptionalLanguage }) {
-		const c = extend('clike', {
+	grammar ({ getOptionalLanguage }) {
```
DmitrySharabin, line 10:
```suggestion
	grammar ({ getOptionalLanguage }: GrammarOptions) {
```

src/languages/c.ts
```diff
@@ -28,78 +32,66 @@ export default {
...
+					greedy: true,
+					alias: 'property',
+					inside: {
+						'string': [
+							{
+								// highlight the path of the include statement as a string
+								pattern: /^(#\s*include\s*)<[^>]+>/,
+								lookbehind: true,
+							},
+							c['string'] as GrammarToken,
```
DmitrySharabin, line 58: We should define `c` somehow. Previously, it was defined as `const c = extend(...)`.

src/languages/chaiscript.ts
```diff
@@ -7,6 +7,7 @@ import type { LanguageProto } from '../types';
 export default {
 	id: 'chaiscript',
 	require: [clike, cpp],
+	extends: clike,
 	grammar ({ extend, getLanguage }) {
```
DmitrySharabin, line 11:
```suggestion
	grammar ({ extend, getLanguage }: GrammarOptions): Grammar {
```

src/plugins/toolbar/prism-toolbar.ts
```diff
@@ -1,6 +1,6 @@
 import { getParentPre } from '../../shared/dom-util';
 import { noop } from '../../shared/util';
-import type { CompleteEnv, HookCallback } from '../../core/hooks';
+import type { CompleteEnv, HookCallback } from '../../core/classes/hooks';
```
DmitrySharabin, line 3: Some new type not yet defined? I mean `CompleteEnv`.

DmitrySharabin:
Code suggestion for `types.d.ts`, line 106:

```diff
- Record<TokenName, RegExpLike | GrammarToken | (RegExpLike | GrammarToken)[]>
+ Record<TokenName, RegExpLike | GrammarToken | GrammarTokens | (RegExpLike | GrammarToken | GrammarTokens)[]>
```

This allows us to cover cases like (in `languages/c.ts`, lines 35–42):

```ts
$insertBefore: {
				'string': {
					'char': {
						// https://en.cppreference.com/w/c/language/character_constant
						pattern: /'(?:\\(?:\r\n|[\s\S])|[^'\\\r\n]){0,32}'/,
						greedy: true,
					},
				},
...
```

src/global.ts
```diff
@@ -1,10 +1,8 @@
...
+import globalPrism, { type Prism } from './core/prism';
 
-const globalSymbol = Symbol.for('Prism global');
-
-// eslint-disable-next-line no-undef
-const namespace = globalThis as Partial<Record<typeof globalSymbol, Prism>>;
-const globalPrism = (namespace[globalSymbol] ??= new Prism());
+declare global {
+	const Prism: Prism | undefined;
+}
```
DmitrySharabin, line 5:
```suggestion
declare global {
	interface globalThis {
		Prism: Prism | undefined;
	}
}
```

This seems to be a better approach. However, I still don't understand why TS complains in `config.ts`. I tried adding the index signature to the `globalThis` interface, but it didn't help. 🤷‍♂️

```ts
declare global {
	interface globalThis {
		Prism: Prism | undefined;
		[key: string]: unknown;
	}
}
```

src/types.d.ts
```diff
@@ -18,10 +17,9 @@ export interface ComponentProtoBase<Id extends string = string> {
 export interface LanguageProto<Id extends string = string> extends ComponentProtoBase<Id> {
 	grammar: Grammar | ((options: GrammarOptions) => Grammar);
 	plugin?: undefined;
+	extends?: LanguageProto;
 }
-type PluginType<Name extends string> = Name extends keyof KnownPlugins
-	? KnownPlugins[Name]
-	: unknown;
+type PluginType<Name extends string> = unknown;
```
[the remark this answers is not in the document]
LeaVerou: Can you do it?

eslint.config.mjs
LeaVerou: I wonder if we need to tweak the config somehow, I would expect `requestAnimationFrame()` would be treated as a known global.

src/types.d.ts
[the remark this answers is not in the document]
DmitrySharabin: Sure. Just mark the changes you approve, and I’ll do that. 🙂

src/core/classes/prism.ts
LeaVerou: Let's do that (`languages && languages.length > 0`)

src/config.ts
```diff
@@ -0,0 +1,65 @@
...
+			scriptElement.dataset[camelCaseName] ??
+			document.querySelector(`[data-prism-${name}]`)?.getAttribute(`data-prism-${name}`)
+		);
+	}
+}
+
+function getGlobalBooleanSetting (name: string, defaultValue: boolean): boolean {
+	const value = getGlobalSetting(name);
+
+	if (value === null) {
```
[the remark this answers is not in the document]
LeaVerou: We should probably do both — a missing attribute will be `null`, not `undefined`.

src/shared/hooks-util.ts
LeaVerou: I think it's part of the stuff that got removed.

src/plugins/diff-highlight/prism-diff-highlight.ts
```diff
@@ -1,7 +1,7 @@
 import { Token, getTextContent } from '../../core/token';
 import diff, { PREFIXES } from '../../languages/diff';
 import { addHooks } from '../../shared/hooks-util';
-import type { BeforeSanityCheckEnv, BeforeTokenizeEnv } from '../../core/hooks';
+import type { BeforeSanityCheckEnv, BeforeTokenizeEnv } from '../../core/classes/hooks';
```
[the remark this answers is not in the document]
LeaVerou: Leftovers I believe. Can you fix?

src/plugins/toolbar/prism-toolbar.ts
LeaVerou: No idea.

src/languages/c.ts
LeaVerou: Or eliminate the need to refer to it. 🤔

src/languages/c.ts
LeaVerou: Can you do it?

src/plugins/diff-highlight/prism-diff-highlight.ts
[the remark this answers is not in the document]
DmitrySharabin: They are used to type the `env` parameter of a function. I think we need to define a type for `env` in one place and reference it in others, even if it is just `any`. Later, we can fix it once, and everything should just work. We have a lot of that stuff in plugins.

src/core/classes/language.ts
```diff
@@ -0,0 +1,96 @@
...
+
+		let { id, grammar } = def;
+		let base = this.base;
+
+		if (typeof grammar === 'function') {
+			grammar = grammar.call(this, {
+				base,
+				getLanguage (id: string) {
+					return this.registry.get(id);
+				},
```
DmitrySharabin, line 55: I wonder if this should be an error function? Will `registry` be accessible if not?

src/core/classes/language.ts
```diff
@@ -0,0 +1,96 @@
...
+	set grammar (grammar: Grammar) {
+		Object.defineProperty(this, 'grammar', { value: grammar, writable: true });
+	}
+
+	get effect () {
+		return this.def.effect;
+	}
+
+	get resolvedGrammar () {
+		return (this.resolvedGrammar = resolveGrammar(this.grammar));
```
DmitrySharabin, line 81: `resolveGrammar` mutates the passed grammar and doesn't return anything. Do we need to mutate `this.grammar`? If yes, then I can suggest we need something like:

```ts
		resolveGrammar(this.grammar);
		return (this.resolvedGrammar = this.grammar);
```

LeaVerou: They were meant to be separate. Resolve grammar applies the combined patches (from `$insertBefore`, `$delete`, `$merge`) against the grammar. I'll need to think more about this

src/core/classes/language.ts
LeaVerou: Oh, yes it does need to be an arrow function! Feel free to commit directly to this PR

DmitrySharabin:
I tried to transform the ASP.NET (C#) language, and [here](#3920) is what I came up with. I used `self` to reference the grammar being defined. I know it's not a part of the new API, but I don't know how to reference the grammar inside the grammar definition the other way. I might miss something.

This was the trickiest part for me (and I'm still not sure if I did it correctly)

```ts
const tag = aspnet['tag'] as GrammarToken & {
	inside: { 'attr-value': { inside: Grammar } };
};

// Regexp copied from markup, with a negative look-ahead added
tag.pattern =
	/<(?!%)\/?[^\s>\/]+(?:\s+[^\s>\/=]+(?:=(?:("|')(?:\\[\s\S]|(?!\1)[^\\])*\1|[^\s'">=]+))?)*\s*\/?>/;

// match directives of attribute value foo="<% Bar %>"
insertBefore(tag.inside['attr-value'].inside, 'punctuation', {
	'directive': aspnet['directive'],
});
```

I transformed it into this:

```ts
$merge: {
	'tag': {
		// Regexp copied from markup, with a negative look-ahead added
		pattern: /<(?!%)\/?[^\s>\/]+(?:\s+[^\s>\/=]+(?:=(?:("|')(?:\\[\s\S]|(?!\1)[^\\])*\1|[^\s'">=]+))?)*\s*\/?>/,
		inside: {
			'attr-value': {
				inside: {
					$insertBefore: {
						'punctuation': {
							// match directives of attribute value foo="<% Bar %>"
							'directive': self['directive'],
						},
					}
				}
			}
		}
	}
},
```

src/core/classes/language.ts
```diff
@@ -0,0 +1,96 @@
...
+	get grammar (): Grammar {
+		// Lazily evaluate grammar
+		const def = this.def;
+
+		let { id, grammar } = def;
+		let base = this.base;
+
+		if (typeof grammar === 'function') {
+			grammar = grammar.call(this, {
+				base,
```
DmitrySharabin, line 52: We should probably pass the required languages as well, so we can destructure them in the `grammar()` function and use the way we can use `base`. It can be helpful, for example, when defining the `JSDoc` language: there will be no need to get the needed languages using `getLanguage()`. See https://github.com/PrismJS/prism/blob/44429cac5aa9e2ae56d8adab20195edb1496fd30/src/languages/jsdoc.ts#L11-12.

src/shared/language-util.ts
```diff
@@ -104,100 +105,82 @@ export function insertBefore (grammar: Grammar, before: string, insert: GrammarT
  *     'color': /\b(?:red|green|blue)\b/
  * });
  */
-export function extend (grammar: Grammar, id: string, reDef: Grammar): Grammar {
-	const lang = cloneGrammar(grammar, id);
+export function extend (base: Grammar, id: string, grammar: Grammar): Grammar {
```
DmitrySharabin, line 108: We can probably ditch `id` here (it's not used anywhere in the function body). `deepClone()` doesn't need it as a parameter anymore.

src/core/classes/language.ts
```diff
@@ -0,0 +1,177 @@
...
+						this.languages[def.id] = language;
+					}
+					else {
+						this.registry.add(def as LanguageProto);
+						defineLazyProperty(this.languages, def.id, () => {
+							return this.registry.getLanguage(def.id)!;
+						});
+					}
+				},
+			});
```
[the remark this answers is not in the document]
LeaVerou: Oh that's clearly a mistake, yes. I think the outer defineProperty() can just be removed.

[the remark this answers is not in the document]
DmitrySharabin: Ok, I'll submit a PR so we are good with that change.

src/core/highlight.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @example
+ * Prism.highlight('var foo = true;', 'javascript');
+ */
+export function highlight (
+	this: Prism,
+	text: string,
+	language: string,
+	options: HighlightOptions = {}
+): string {
+	const prism = this ?? singleton;
```
DmitrySharabin, line 30: We introduce a cycle by importing `singleton` in all' highlight' methods. This is how it looks:

```
src/core/prism.ts
  └──> src/core/classes/prism.ts
         └──> src/core/highlight.ts
                └──> src/core/prism.ts
```

Moving it to the file with the Prism class won't help. We probably need to break it somehow. Possibly, this is not only the issue in the build step. It might cause some other issues with types? 🤔

src/core/classes/language-registry.ts
```diff
@@ -0,0 +1,114 @@
...
+		}
+		else {
+			// eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+			throw new Error('Invalid argument type: ' + ref);
+		}
+
+		id = this.aliases[id] ?? id;
+		def ??= this.cache[id];
+
+		return { id, def };
```
DmitrySharabin, line 63:
```suggestion
		let language = this.instances[id];

		return { id, def, language };
```

If there's already a resolved language corresponding to `id`, we should probably return it as well.

src/core/highlight.ts
```diff
@@ -0,0 +1,70 @@
...
+	this: Prism,
+	text: string,
+	languageRef: string | LanguageLike,
+	options: HighlightOptions = {}
+): string {
+	const prism = this ?? singleton;
+
+	const { id, def, language } = prism.languageRegistry.resolveRef(languageRef);
+
+	let env: HookEnv['after-tokenize'] = {
```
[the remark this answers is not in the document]
DmitrySharabin: I think it's a typo: we define `env` right before we run the `before-tokenize` hook.

src/core/highlight.ts
```diff
@@ -0,0 +1,70 @@
...
+		languageId: id,
+		languageDef: def,
+		language,
+	};
+
+	prism.hooks.run('before-tokenize', env);
+
+	if (!env.language) {
+		throw new Error(`No language definition found for ${env.languageId}.`);
+	}
```
DmitrySharabin, line 61: Does this mean that `language` can be added to `env` during the `before-tokenize` hook run? Otherwise, we probably would have performed this check before it.

src/core/highlight.ts
```diff
@@ -0,0 +1,70 @@
...
+
+	env = env as unknown as HookEnv['after-tokenize'];
+	env.tokens = tokenize.call(prism, env.code, env.language!.resolvedGrammar);
```
DmitrySharabin, line 64: It's only for TS, right? As an alternative, I'd suggest explicitly adding a `token` property to `env`:

DmitrySharabin, line 64: It's only for TS, right? As an alternative, I'd suggest explicitly adding a `token` property to `env`:

```suggestion
	env = { ...env, tokens: undefined } as HookEnv['after-tokenize'];
```

src/core/highlight.ts
```diff
@@ -24,32 +40,31 @@ import type { Grammar, GrammarToken, GrammarTokens, RegExpLike } from '../types'
...
 	prism.hooks.run('before-tokenize', env);
 
-	if (!env.grammar) {
-		throw new Error('The language "' + env.language + '" has no grammar.');
+	if (!env.language) {
+		throw new Error(`No language definition found for ${env.languageId}.`);
 	}
 
-	env.tokens = prism.tokenize(env.code, env.grammar.resolvedGrammar);
+	env = env as unknown as HookEnv['after-tokenize'];
```
DmitrySharabin, line 63: It's only for TS, right? As an alternative, I'd suggest explicitly adding a `token` property to `env`:

```suggestion
	env = { ...env, tokens: undefined } as HookEnv['after-tokenize'];
```

src/core/highlight.ts
```diff
@@ -24,32 +40,31 @@ import type { Grammar, GrammarToken, GrammarTokens, RegExpLike } from '../types'
...
+	env.tokens = tokenize.call(prism, env.code, env.language!.resolvedGrammar);
```
DmitrySharabin, line 64: Do we really need to use `call` here? `tokenize` already expects a Prism instance as the first argument. Or do you have some changes to its signature in mind?

DmitrySharabin: Or probably it's a TIL thing: if we call the first argument of a function `this`, we can use it inside the function body to assign some default context the function will work with? And to pass this context explicitly, we use `call`?

src/languages/c.ts
```diff
@@ -1,22 +1,31 @@
-import { insertBefore } from '../shared/language-util';
+import { rest } from '../shared/symbols';
 import clike from './clike';
-import type { GrammarToken, LanguageProto } from '../types';
+import type { Grammar, GrammarOptions, GrammarToken, LanguageProto } from '../types';
 
 export default {
 	id: 'c',
 	require: clike,
```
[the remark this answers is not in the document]
LeaVerou: Correct.

src/core/classes/hooks.ts
```diff
@@ -0,0 +1,104 @@
...
+			for (let n of name) {
+				this.add(n, callback);
+			}
+		}
+		else if (typeof name === 'object') {
+			// Multiple hooks
+			let hooks = name;
+
+			for (let name in hooks) {
+				this.add(name, hooks[name]);
```
DmitrySharabin, line 38:
```suggestion
				this.add(name, hooks[name as string]);
```

src/core/highlight.ts
LeaVerou: It's definitely confusing syntax though!

src/core/highlight.ts
LeaVerou: Yes

src/core/highlight.ts
LeaVerou: Cycles are not a problem in ESM, they happen all the time. Does Rollup still struggle with them? I seem to recall it had an issue in the past.

DmitrySharabin: It logs a warning, but builds the code fine.

LeaVerou: Cool, no issue then? :D

src/core/highlight.ts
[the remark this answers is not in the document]
DmitrySharabin: Oh, I see. So, it just should be a variable that covers all cases. Got it. My suggestion is irrelevant then.

src/languages/coffeescript.ts
```diff
@@ -38,67 +37,58 @@ export default {
...
-			},
+				'string': {
+					'inline-javascript': {
+						pattern: /`(?:\\[\s\S]|[^\\`])*`/,
+						inside: {
+							'delimiter': {
+								pattern: /^`|`$/,
+								alias: 'punctuation',
+							},
+							'script': 'javascript',
```
DmitrySharabin, line 65: Shouldn't that be something like

```js
'script': { $inside: 'javascript` }
```

LeaVerou: No, I think that happens automatically. Or at least it should (though I’m starting to think special casing languages as strings is not a good idea but let’s get the branch merged first before changing that)

src/core/tokenize/match.ts
```diff
@@ -46,11 +42,21 @@ export function _matchGrammar (
...
+			const insideGrammar = resolve.call(prism, inside);
+
+			let flagsToAdd = '';
 
 			if (greedy && !pattern.global) {
 				// Without the global flag, lastIndex won't work
-				patternObj.pattern = pattern = RegExp(pattern.source, pattern.flags + 'g');
+				flagsToAdd += 'g';
+			}
+			if (pattern.source.indexOf('(?<') && pattern.hasIndices === false) {
```
DmitrySharabin, line 53: Looking at the expression, it seems that `includes` is intended to be used instead of `indexOf`, no?

src/core/tokenize/match.ts
```diff
@@ -146,13 +153,37 @@ export function _matchGrammar (
...
+								localInsideGrammar = resolve.call(
+									prism,
+									localInsideGrammar(match.groups)
+								);
+							}
+
+							content = tokenize.call(prism, content, localInsideGrammar);
+						}
+
+						return typeof arg === 'object' && arg.type
```
DmitrySharabin, line 177: Why not `type` instead of `arg.type` here and in the following line?

src/languages/coffeescript.ts
DmitrySharabin: Okay. But what is the difference between `inside` and `$inside`? Is `inside` something that should be replaced with `$inside` eventually (like it happened with `rest` that became `$rest`)?

src/core/tokenize/match.ts
LeaVerou: Yes

src/core/tokenize/match.ts
LeaVerou: TS was complaining

src/languages/coffeescript.ts
LeaVerou: Yes, we should add the `language-xxx` class automatically.

src/core/classes/component-registry.ts
```diff
@@ -0,0 +1,171 @@
...
+	/** All components that are currently being loaded */
+	loading: Record<string, Promise<T>> = {};
+
+	/**
+	 * Same data as in loading, but as an array, used for aggregate promises.
+	 * IMPORTANT: Do NOT overwrite this array, only modify its contents.
+	 */
+	private loadingList: Promise<T>[] = [];
+
+	ready: Promise<T[]>;
```
DmitrySharabin, line 37: Shouldn't `ready` be a getter to become reactive? We change dynamically `loadingList` (`ready` depends on it). However, we set the value of `ready` once when creating an instance of a registry and never update it. As a result, we have `ready` that is always fulfilled, even though there might be promises that are not fulfilled in the `loadingList` array.
Or should we update `ready` every time we update `loadingList`? Or is there another option I'm not seeing?

DmitrySharabin:
Things to implement before merging:

- [ ] `$self`
- [ ] Correctly resolve `inside: 'language-id'`

Something else?

LeaVerou:
For `$self`, we need to figure out what that means when we have nested languages (e.g. highlighting JS in HTML). Probably the closest language.

`inside: 'language-id'` worked before, so that must be a regression.

DmitrySharabin:
> `inside: 'language-id'` worked before, so that must be a regression.

Yup, this is related to something we've already [started to discuss](https://github.com/PrismJS/prism/pull/3942#discussion_r2106217545).

LeaVerou:
> > `inside: 'language-id'` worked before, so that must be a regression.
> 
> Yup, this is related to something we've already [started to discuss](https://github.com/PrismJS/prism/pull/3942#discussion_r2106217545).

That does not address why it stopped working. This was even working in v1. And the code to make it work is still there I believe (in tokenize)

PR #3914, opened 2025-04-28
Ditch the `eslint-plugin-import` and use Prettier to order imports

DmitrySharabin:


.prettierrc
```diff
@@ -1,9 +1,20 @@
 {
 	"plugins": [
+		"@trivago/prettier-plugin-sort-imports",
 		"prettier-plugin-brace-style",
 		"prettier-plugin-space-before-function-paren",
 		"prettier-plugin-merge"
 	],
+	"importOrder": [
```
LeaVerou, line 8: How does this work? What order is it specifying?

LeaVerou (approved):

.prettierrc
DmitrySharabin: Added comments in this PR: https://github.com/PrismJS/prism/pull/3917

PR #3915, opened 2025-04-28
Some more suggestions in one place

DmitrySharabin:


src/config.ts
```diff
@@ -1,10 +1,11 @@
 const hasDOM = typeof document !== 'undefined' && typeof window !== 'undefined';
-const scriptElement: HTMLScriptElement = hasDOM && document.currentScript;
+const scriptElement: HTMLOrSVGScriptElement | null = document?.currentScript;
```
DmitrySharabin, line 2: From my perspective, we get the same result with a bit simpler types, right?

src/config.ts
```diff
@@ -1,10 +1,11 @@
...
 // @ts-ignore
-const globalConfig: Record<string, PrismConfig[keyof PrismConfig]> =
-	globalThis.Prism?.constructor?.name === 'Object' ? globalThis.Prism : {};
+const globalConfig: Record<string, PrismConfig[keyof PrismConfig] | null> =
+	globalThis.Prism?.constructor?.name === 'Object' ? globalThis.Prism.config : {};
 
 function getGlobalSetting (name: string) {
+	// eslint-disable-next-line regexp/no-unused-capturing-group
```
DmitrySharabin, line 8: ESLint still can't analyze the code properly. 😅

src/config.ts
```diff
@@ -15,7 +16,7 @@ function getGlobalSetting (name: string) {
 	}
 	else if (hasDOM) {
 		return (
-			scriptElement.dataset[camelCaseName] ??
+			scriptElement!.dataset[camelCaseName] ??
```
DmitrySharabin, line 19: If I'm not mistaken, `scriptElement` will be defined here, but TS can't figure it out from the code, so we need to tell it we're sure with what we do.

src/config.ts
LeaVerou: Why haven't we disabled this rule?

src/config.ts
LeaVerou: Maybe add a comment to that effect?

src/config.ts
```diff
@@ -1,10 +1,11 @@
 const hasDOM = typeof document !== 'undefined' && typeof window !== 'undefined';
-const scriptElement: HTMLScriptElement = hasDOM && document.currentScript;
+const scriptElement: HTMLOrSVGScriptElement | null = document?.currentScript;
 // @ts-ignore
-const globalConfig: Record<string, PrismConfig[keyof PrismConfig]> =
-	globalThis.Prism?.constructor?.name === 'Object' ? globalThis.Prism : {};
+const globalConfig: Record<string, PrismConfig[keyof PrismConfig] | null> =
+	globalThis.Prism?.constructor?.name === 'Object' ? globalThis.Prism.config : {};
```
LeaVerou, line 5: These are not equivalent. `Prism` here is not the global instance, this is before initialization, so it's just an object literal that folks can define (hence the constructor check)

src/config.ts
LeaVerou: Also, `scriptElement` won't necessarily be defined. `hasDOM` is a necessary, not a sufficient condition for it. It's also `null` when running code in ESM.

```suggestion
			scriptElement?.dataset[camelCaseName] ??
```

src/config.ts
LeaVerou:
```suggestion
	globalThis.Prism?.constructor?.name === 'Object' ? globalThis.Prism: {};
```

src/config.ts
LeaVerou: Not the same — if `document` is undefined referring to `document` will throw.

LeaVerou:
```suggestion
const scriptElement: HTMLOrSVGScriptElement | null = hasDOM && document.currentScript;
```

LeaVerou (approved):

PR #3919, opened 2025-04-29
Some suggestions

DmitrySharabin:


src/types.d.ts
```diff
@@ -1,11 +1,24 @@
-/* eslint-disable @typescript-eslint/no-redundant-type-constituents */
-import type Prism from './core/classes/prism';
-export type * from './classes/grammar';
-export type * from './classes/language';
-export type * from './classes/registry';
-export type * from './classes/plugin-registry';
-export type * from './classes/language-registry';
-export type * from './util/types';
```
DmitrySharabin, line 8: Unfortunately, this doesn't work with types. Even removing `type` doesn't solve the issue. We need to manually re-export the needed types.

src/core/classes/language-registry.ts
```diff
@@ -18,7 +18,7 @@ export default class LanguageRegistry extends ComponentRegistry<LanguageLike> {
 		if (!this.instances[canonicalId]) {
 			if (this.cache[canonicalId]) {
 				let def = this.cache[canonicalId];
-				let language = new Language(def, this);
+				let language = new Language(def as LanguageProto, this);
```
LeaVerou, line 21: Why is def not already LanguageProto?

LeaVerou (approved):

src/core/classes/language-registry.ts
DmitrySharabin: `def` has type of `LanguageLike`, which is a union of `Language` and `LanguageProto`, so we need to specify it here.

PR #3922, opened 2025-04-30
First pass on transforming `php` and `php-extras`

DmitrySharabin:


src/languages/php.ts
```diff
@@ -1,14 +1,13 @@
...
 import { embeddedIn } from '../shared/languages/templating';
 import { tokenize } from '../shared/symbols';
 import markup from './markup';
-import type { LanguageProto } from '../types';
+import type { Grammar, LanguageProto } from '../types';
 
 export default {
 	id: 'php',
-	require: markup,
+	base: markup,
```
LeaVerou, line 8: So far I've been using *both* `require` and `base`, we can later remove `require` but some languages do have requires beyond their base.

LeaVerou:
```suggestion
	require: markup,
	base: markup,
```

LeaVerou (approved):

PR #3925, opened 2025-05-01
Suggestions to the `simplify` branch

DmitrySharabin:


eslint.config.mjs
```diff
@@ -27,6 +27,10 @@ const config = [
 		languageOptions: {
 			ecmaVersion: 'latest',
 			sourceType: 'module',
+			globals: {
+				...globals.browser,
+				...globals.node,
+			},
```
DmitrySharabin, line 33: All files (not only from `core`) should not complain about `document`, `requestAnimationFrame()`, etc. I kept the other settings for now. We can revise them later.

src/core/classes/language-registry.ts
```diff
@@ -75,6 +79,7 @@ export default class LanguageRegistry extends ComponentRegistry<LanguageLike> {
 	 * Get resolved language or null if it doesn't exist
 	 * If definition is loaded but not yet resolved, it will be resolved and returned.
 	 */
+	// FIXME: This method overrides the get() method from the base class (ComponentRegistry), but it has a different signature. TS complains about it.
```
LeaVerou, line 82: Ooh I forgot there was a base `get()`, we should name it something else I guess. Maybe `resolve()`?

src/core/classes/language.ts
```diff
@@ -22,17 +22,17 @@ export default class Language extends EventTarget {
...
 		for (let def of this.require) {
-			let language = this.registry.peek(def);
+			let language = this.registry.peek(def as LanguageProto);
 			if (language) {
 				// Already resolved
 				this.languages[def.id] = language;
 			}
 			else {
-				this.registry.add(def);
+				this.registry.add(def as LanguageProto);
```
LeaVerou, line 35: But `def` is already defined as `LanguageProto`, why is this needed?

src/core/classes/prism.ts
```diff
@@ -134,3 +135,5 @@ export default class Prism {
 		return tokenize.call(this, text, grammar);
 	}
 }
+
+export type { Prism };
```
LeaVerou, line 139: Wouldn't this suffice?
```suggestion
export { Prism };
```

LeaVerou (approved):

src/core/classes/language.ts
DmitrySharabin: Here, `def` has the `LanguageLike` type, so we need to narrow it down.

PR #3926, opened 2025-05-02
Some more suggestions to the `simplify` branch

DmitrySharabin:


src/core/classes/prism.ts
```diff
@@ -67,6 +67,7 @@ export default class Prism {
 		if (!this.config.manual) {
 			this.waitFor.push(documentReady());
 
+			this.ready = allSettled(this.waitFor);
```
DmitrySharabin, line 70: For now, this seems to resolve on the next tick, and we don't consider all other pending promises.

src/util/objects.ts
```diff
@@ -99,7 +99,7 @@ export interface CloneOptions {
 }
 
 export function deepClone (obj: any, options: CloneOptions = {}) {
-	if (!obj) {
+	if (!obj || typeof obj !== 'object') {
```
DmitrySharabin, line 102: We should consider primitive values as well. If not, if we get, for example, `true` as a value of `obj`, we'll try to use it as a `_clones` key, and the code will throw.

src/core/tokenize.ts
```diff
@@ -240,12 +240,12 @@ function toGrammarToken (pattern: GrammarToken | RegExpLike): GrammarToken {
...
 
 function resolve (
-	components: Registry,
+	languageRegistry: LanguageRegistry,
 	reference: Grammar | string | null | undefined
 ): Grammar | undefined {
 	if (reference) {
 		if (typeof reference === 'string') {
-			return components.getLanguage(reference);
+			return languageRegistry.getLanguage(reference)?.grammar;
```
DmitrySharabin, line 248: I have doubts about this change. I saw you left a placeholder for a `resolve()` method in the `Language` class. Is it something that should be used here instead?

src/languages/markup.ts
```diff
@@ -103,7 +103,7 @@ export default {
 					'style': inlineEmbedded('style', 'css'),
 					'script': inlineEmbedded('script', 'javascript'),
 				},
-				'tag/attr-value': {
+				'tag/inside/attr-value': {
```
DmitrySharabin, line 106: It was `tag.inside` in the original grammar.

scripts/build.ts
```diff
@@ -141,7 +141,7 @@ const dataToInsert = {
 		const data = await Promise.all(
 			[...languageIds, ...pluginIds].map(async id => {
 				const proto = await loadComponent(id);
-				return { id, alias: toArray(proto.alias) };
+				return { id, alias: toArray(proto?.alias) };
```
DmitrySharabin, line 144: It's a temporary fix to make the `build` script work again.

src/core/classes/registry.ts
```diff
@@ -135,7 +135,7 @@ export default class Registry<T extends ComponentProto> extends EventTarget {
 			return this.loading[id];
 		}
 
-		let loadingComponent = import(this.path + id)
+		let loadingComponent = import(this.path + id + '.js')
```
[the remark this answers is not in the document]
LeaVerou: I think for now we can hardcode `.js`.

src/core/classes/prism.ts
LeaVerou: Is there a promise in the array when this runs? If it resolves before any other promises get added then it won't ever un-resolve.

src/core/classes/language.ts
```diff
@@ -44,24 +44,6 @@ export default class Language extends EventTarget {
...
-					}
-					else {
-						this.registry.add(def as LanguageProto);
-						defineLazyProperty(this.languages, def.id, () => {
-							return this.registry.getLanguage(def.id)!;
-						});
-					}
-				},
-			});
-
```
LeaVerou, line 64: Wait, why was this entire thing deleted? We still need this part:

```
defineLazyProperty(this.languages, def.id, () => {
	return this.registry.getLanguage(def.id)!;
});
```

src/languages/markup.ts
LeaVerou: The idea is that `inside` will be applied automatically (since there's no point in manipulating that object)

LeaVerou:
```suggestion
				'tag/attr-value': {
```

LeaVerou (approved):

src/languages/markup.ts
DmitrySharabin: > The idea is that `inside` will be applied automatically (since there's no point in manipulating that object)

@LeaVerou, is this something you are planning to implement? Because without the fix, we get this error when trying to highlight a markup code:

image

And none of the code blocks that follow the markup code are highlighted.

src/core/classes/prism.ts
DmitrySharabin: There is, and I think it's a dummy promise we need so `this.ready` has an initial value:

image

src/languages/markup.ts
LeaVerou: Yes

LeaVerou: Or you can :)

PR #3928, opened 2025-05-05
Suggestions (next iteration)

DmitrySharabin:


src/core/classes/language.ts
```diff
@@ -11,7 +11,7 @@ export default class Language extends EventTarget {
 	registry: LanguageRegistry;
 	resolvedGrammar?: Grammar;
 	require: List<LanguageLike> = new List();
-	optional: List<string> = new List();
+	optional: List<string | LanguageLike> = new List();
```
DmitrySharabin, line 14: We add languages from `extends` to this list, and `extends` has a broader type we should match.

src/core/classes/language.ts
```diff
@@ -48,22 +48,22 @@ export default class Language extends EventTarget {
 				let language = this.registry.peek(def as LanguageProto);
 				if (language) {
 					// Already resolved
-					return language.resolvedGrammar;
+					return language.resolvedGrammar as Grammar;
```
DmitrySharabin, line 51: In this and the following suggestions, we do that to exclude the case when `resolvedGrammar` is `null`.

src/core/classes/registry.ts
```diff
@@ -13,6 +13,12 @@ export interface ComponentProtoBase<Id extends string = string> {
 	optional?: string | readonly string[];
 }
 
+export interface AddEventPayload<T extends ComponentProto> {
+	id: string;
+	type?: string;
+	component: T;
+}
```
DmitrySharabin, line 20: Just to make working with custom events a bit more type-safe.

src/core/classes/registry.ts
```diff
@@ -146,8 +152,8 @@ export default class Registry<T extends ComponentProto> extends EventTarget {
 				return null;
 			});
 
-		this.loading[id] = loadingComponent;
-		this.loadingList.push(loadingComponent);
+		this.loading[id] = loadingComponent as Promise<T>;
```
DmitrySharabin, line 155: We need a type assertion here since we excluded `null` from `Promise<T>` (it was `Promise<T | null>` before), but in the code, we still expect that some promises might reject with `null`.

src/util/insert.ts
```diff
@@ -31,68 +31,14 @@ import type { Grammar, GrammarTokens } from '../types';
...
-			}
-		}
-
-		// Do not insert tokens which also occur in `insert`. See #1525
-		if (!Object.hasOwn(insert, key)) {
-			Object.defineProperty(grammar, key, descriptors[key]);
-		}
-	}
+export function insertBefore (grammar: Grammar, beforeKey: string, tokens: GrammarTokens) {
+	insert(grammar, beforeKey, tokens, 'before');
```
DmitrySharabin, line 37: If I'm not mistaken, we can use the (defined lower) `insert` utility function here.

src/core/classes/language.ts
LeaVerou: Why would that happen?

DmitrySharabin: Oh, sorry. Not `null`, `indefined`. Because of the way we defined this property with TS—it's optional:

image

src/util/grammar-patch.ts
```diff
@@ -17,8 +17,9 @@ export function grammarPatch (grammar: Grammar, patch: Grammar = grammar): Gramm
 
 			if (key?.includes('/')) {
 				// Deep key
-				const path = key.split('/');
+				let path = key.split('/');
 				const lastKey = path.pop();
+				path = path.flatMap(key => [key, 'inside']); // add `inside` after each key
```
DmitrySharabin, line 22: Is this something you had in mind to transform deep keys like `foo/bar/baz/yolo` into

```js
'foo': {
	inside: {
		'bar': {
			inside: {
				'baz': {
					inside: {
						'yolo': {...}
					}
				}
			}
		}
	}
}
```

src/util/insert.ts
LeaVerou: Haha that was the reason I wrote it and then I got distracted 🤣

src/core/classes/language.ts
LeaVerou: We should define it differently then — it is always present! In the interest of time I'm gonna merge this, but can you check that I fixed this and then revert the changes that worked around it? (I might get sidetracked)

LeaVerou (approved):

src/util/grammar-patch.ts
DmitrySharabin: > Possibly? Does it work?

It does.

> Also, can you check if we have any case where we may want to insert something at the same level as the object that has the `inside` key?

Will do

PR #3930, opened 2025-05-07
Place plugin docs along with their source code

DmitrySharabin:


src/plugins/normalize-whitespace/demo.html
```diff
@@ -0,0 +1,48 @@
+<body>
```
LeaVerou, line 1: Why HTML? We can use HTML elements in Markdown, wouldn't that work?

src/plugins/toolbar/README.md
```diff
@@ -0,0 +1,99 @@
...
+		alert(`This code snippet is written in ${env.language}.`);
+	},
+});
+```
+
+See how the above code registers the `Hello World!` button? You can use this in your plugins to register your own buttons with the toolbar.
+
+If you need more control, you can provide a function to `registerButton` that returns either a `span`, `a`, or `button` element.
+
+```js
```
LeaVerou, line 51: This applies to other `demo.js` files (but possibly not all): Instead of including a separate `demo.js` that basically duplicates what is here, can we have a class or something that we can add to code examples and have them get a "Run" button?

LeaVerou (approved):
See comments, but I'm fine to merge this and work on them later.

LeaVerou:
Also, we'll eventually do this for v1, right?

src/plugins/toolbar/README.md
DmitrySharabin: Yes, that makes sense. Though I think there is a `demo.js` file that should have another name: it's just a script that runs some functionality on the docs page, not related to demos (IIRC, it's related to the autoloader plugin). I should revert its name to `script.js`.

DmitrySharabin:
> See comments, but I'm fine to merge this and work on them later.

Yeah, I'd rather work on proposed improvements a bit later.

> Also, we'll eventually do this for v1 too, right?

Something like that, yes. Except we don't build the v1 website with 11ty; I figure that's fine, though, since eventually, it might become some sort of archive website.

src/plugins/normalize-whitespace/demo.html
DmitrySharabin: I just left it in its previous format. But yeah, switching to `demo.md` is no problem. Done!

LeaVerou:
> Except we don't build the v1 website with 11ty; I figure that's fine, though, since eventually, it might become some sort of archive website.

We should. We've already done most of the work! And we'd want to update the v1 website even before v2 launches, e.g. to add a banner to let people know that we're working on v2, or to improve its IA and the list of websites using Prism, etc.

src/plugins/normalize-whitespace/demo.html
LeaVerou: I meant as a part of its README, not a separate `demo.md`. Ideally none of them should have anything other than a `README.md`.

DmitrySharabin: Hm, I think there might be a reason for this, but I'm not sure. I'll figure it out.

DmitrySharabin:
> > Except we don't build the v1 website with 11ty; I figure that's fine, though, since eventually, it might become some sort of archive website.
> 
> We should. We've already done most of the work! And we'd want to update the v1 website even before v2 launches, e.g. to add a banner to let people know that we're working on v2, or to improve its IA and the list of websites using Prism, etc.

Let's do it, then! 🙂

PR #3932, opened 2025-05-09
Transform plugin docs to READMEs

DmitrySharabin:
This is the first step in moving the Prism v1 website to 11ty. The source `*.html` files are still in the repo. I'll remove them once the website is ready (I'm almost done, though there are still a few issues to fix).


plugins/normalize-whitespace/demo.md
```diff
@@ -0,0 +1,53 @@
+---
```
LeaVerou, line 1: As I said in the same v2 change, I think this should be part of the readme, possibly as a code snippet with a certain class that we pick up on the website and transform it to a live example.

LeaVerou (approved):
Same comments apply as the same change for v2, but otherwise LGTM!

plugins/normalize-whitespace/demo.md
DmitrySharabin: Yup, I remember that. Will do. We just need to start the process, so we have something to work with.

PR #3935, opened 2025-05-14
Update GDScript for Godot 4.x

Cykyrios:
I don't know if you're still accepting language PRs for v1, but I needed up-to-date GDSCript syntax highlighting, so I made the changes so it is compatible with Godot 4.x (this also removes some Godot 3.x-specific keywords).

Apart from the string and operators regex which I mostly reused as is, I overhauled the entire lexer taking inspiration from Godot's own pygments implementation for its official documentation. The following GDScript features are now available:
* Builtin types, builtin objects (all 1003 of them at the time of writing), builtin functions.
* Specific colors for control-flow keywords vs standard keywords, function definition vs function call, documentation vs standard comments.
* Annotations (`@export` and the like).
* NOTE, TODO, DANGER and other notes in comments and doc comments.
* nodepaths, string names, placeholders in strings (Godot itself doesn't color them, this is a bonus feature).
* `$` and `%` shorthands for nodepaths are highlighted as `variable` tokens.
* Binary and hexadecimal numbers are their own tokens, in case some themes want to color them differently.
* `INF`, `NAN`, `PI` and `TAU` are `builtin-number` tokens aliased as `keyword`, `null` is included in `builtin-pseudo` along with `true` and `false`, which are aliased as `boolean`.
* Signals are highlighted properly when declared.
* Properties/class members are also highlighted.

From my testing, the only thing that does not highlight the way Godot does is class names, where Prism is not able to know they are classes, the rest should work fine.

For reference, here are comparison screenshots with the code in the example html file, in both Godot (using passivestar's theme with the CommitMonoNerd font (bold)) and an appropriately-themed Docusaurus page.

image
image

### DIFF
--- a/components/prism-gdscript.js
+++ b/components/prism-gdscript.js
@@ -1,27 +1,258 @@
+const commentCritical = /(?:ALERT|ATTENTION|CAUTION|CRITICAL|DANGER|SECURITY)/;
+const commentWarning = /(?:BUG|DEPRECATED|FIXME|HACK|TASK|TBD|TODO|WARNING)/;
+const commentNotice = /(?:INFO|NOTE|NOTICE|TEST|TESTING)/;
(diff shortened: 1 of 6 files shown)


Cykyrios:
I updated with added string escapes and the corresponding tests.

PR #3943, opened 2025-05-26
Bug fixes (the `simplify` branch)

DmitrySharabin:
### Summary

- Fix the C, C#, F#, Markdown, PHP, and VB.NET language definitions
- Fix the issue with deeply nested properties in `extend()`


src/languages/c.ts
```diff
@@ -37,48 +37,48 @@ export default {
...
-						/(^[\t ]*)#\s*[a-z](?:[^\r\n\\/]|\/(?!\*)|\/\*(?:[^*]|\*(?!\/))*\*\/|\\(?:\r\n|[\s\S]))*/im,
-					lookbehind: true,
-					greedy: true,
-					alias: 'property',
-					inside: {
-						'string': [
-							{
-								// highlight the path of the include statement as a string
-								pattern: /^(#\s*include\s*)<[^>]+>/,
+					'macro': {
```
DmitrySharabin, line 40: `macro` should be inserted before `string`, not that something should be inserted before `macro`.

src/languages/aspnet.ts
```diff
@@ -1,55 +1,44 @@
 import csharp from './csharp';
 import markup from './markup';
-import type { Grammar, GrammarOptions, LanguageProto } from '../types';
+import type { Grammar, GrammarToken, LanguageProto } from '../types';
```
DmitrySharabin: By bad. I transformed this language incorrectly before.

src/languages/markdown.ts
```diff
@@ -95,7 +95,7 @@ export default {
 					// code block
 					// ```
 					pattern:
-						/^```(?<codeLanguage>[a-z-]+).+(?:\n|\r\n?)(?<codeBlock>[\s\S]*)?(?:\n|\r\n?)```$/i,
+						/^```(?<codeLanguage>[a-z-]+)(?:.+)?(?:\n|\r\n?)(?<codeBlock>[\s\S]*)?(?:\n|\r\n?)```$/i,
```
DmitrySharabin, line 98: Cover cases when there is no space (only a line break) after the language code, and its last symbol might not be treated as part of the language code.

src/languages/aspnet.ts
```diff
@@ -1,55 +1,44 @@
...
-			pattern: /<%.*%>/,
-			alias: 'tag',
-			inside: {
-				'directive': {
-					pattern: /<%\s*?[$=%#:]{0,2}|%>/,
-					alias: 'tag',
-				},
-				$rest: 'csharp',
+	grammar ({ base }): Grammar {
+		const pageDirectiveInside: Grammar = {
```
LeaVerou: Why did we rename this?

DmitrySharabin: Hm, something is wrong with the changes I propose. I took the language definition from the `v2` branch, not the `simplify` branch. Let me remove it from the PR for now. We'll return to it later.

LeaVerou (approved):
See question but otherwise LGTM

DmitrySharabin:
> See question but otherwise LGTM

If you are referring to the [ASP.NET language and why we renamed the variable](https://github.com/PrismJS/prism/pull/3943#discussion_r2111947471), I removed this language from this PR to update it correctly later.

PR #3950, opened 2025-05-29
Fix the Markdown language

DmitrySharabin:
- Add the missing inside token and fix the existing one
- Fix the regex that is too greedy now: if we have more than one code block next to each other, all of them starting from the second one will be interpreted as a code block of the first one
- Ignore spaces before the language code

However, to fully support it, we need to support functional tokens first. For now, we don't, and we work with it as it is a regex, or an object with the `pattern` property. In both cases, `matchPattern()` (in `tokenize/match.ts`) throws.

### DIFF
--- a/src/languages/markdown.ts
+++ b/src/languages/markdown.ts
@@ -95,9 +95,10 @@ export default {
 					// code block
 					// ```
 					pattern:
-						/^```(?<codeLanguage>[a-z-]+)(?:.+)?(?:\n|\r\n?)(?<codeBlock>[\s\S]*)?(?:\n|\r\n?)```$/i,
+						/^```(?<codeLanguage>[a-z-]+)(?:.+)?(?:\n|\r\n?)(?<codeBlock>[\s\S]*?)(?:\n|\r\n?)```$/im,
 					inside: {
-						'code-block': groups => groups.codeLanguage,
+						'code-language': groups => groups.codeLanguage,
+						'code-block': groups => groups.codeBlock,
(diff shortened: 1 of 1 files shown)


LeaVerou (approved):
I thought I added code in tokenize() to support functional tokens?

DmitrySharabin:
> I thought I added code in tokenize() to support functional tokens?

You did. It covers functional `inside` (without arguments) and cases where _every part_ of a string we tokenize belongs to a named group (inside the corresponding pattern).

If, OTOH, we have parts that are not in a group (like ”```“ in a code block in Markdown), we tokenize it by passing the same grammar (with functional tokens):

https://github.com/PrismJS/prism/blob/b86c4e0f46f1e01133ce2d356fd3b8286f2632d4/src/core/tokenize/match.ts#L174

While tokenizing, we iterate over the grammar’s tokens:

https://github.com/PrismJS/prism/blob/b86c4e0f46f1e01133ce2d356fd3b8286f2632d4/src/core/tokenize/match.ts#L27

These functional tokens are not regular expressions, and when we call `exec()` on them (inside `matchPattern()`), our code throws:

https://github.com/PrismJS/prism/blob/b86c4e0f46f1e01133ce2d356fd3b8286f2632d4/src/core/tokenize/match.ts#L131

https://github.com/PrismJS/prism/blob/b86c4e0f46f1e01133ce2d356fd3b8286f2632d4/src/core/tokenize/match.ts#L224

PR #3951, opened 2025-05-29
Before highlighting code, resolve the language if its definition is already loaded, and it is not yet resolved

DmitrySharabin:
For now, one can write a code that looks correct but will throw (necessary imports are omitted):

```js
prism.languageRegistry.add(javascript);
prism.highlight("let foo = 42;", javascript);
```

### DIFF
--- a/src/core/highlight.ts
+++ b/src/core/highlight.ts
@@ -54,6 +54,10 @@ export function highlight (
 		language,
 	};
 
+	if (env.languageDef && !env.language) {
+		env.language = prism.languageRegistry.getLanguage(env.languageDef);
+	}
+
 	prism.hooks.run('before-tokenize', env);
 
 	if (!env.language) {


src/core/highlight.ts
```diff
@@ -54,6 +54,10 @@ export function highlight (
 		language,
 	};
 
+	if (env.languageDef && !env.language) {
+		env.language = prism.languageRegistry.getLanguage(env.languageDef);
+	}
+
 	prism.hooks.run('before-tokenize', env);
```
DmitrySharabin, line 61: I thought the hook should have the full context during the run, so I decided to add this code before calling it. Am I right?

src/core/highlight.ts
```diff
@@ -54,6 +54,10 @@ export function highlight (
 		language,
 	};
 
+	if (env.languageDef && !env.language) {
+		env.language = prism.languageRegistry.getLanguage(env.languageDef);
```
LeaVerou, line 58:
```suggestion
		// NOTE is this the best place to manage this logic?
		env.language = prism.languageRegistry.getLanguage(env.languageDef);
```

LeaVerou (approved):
Not 100% sure this is the best place to manage this logic, but in the interest of progress, let's merge it.

DmitrySharabin:
Me neither. I just couldn't find a better place. 🤷‍♂️

PR #3952, opened 2025-05-29
Make `nextTick()` work in environments other than browsers

DmitrySharabin:
For now, when we call `nextTick()`, the returned promise is always rejected.


LeaVerou (approved):
Oooh nice! TIL about `setImmediate()`.

PR #3957, opened 2025-06-01
Another attempt to transform the ASP.NET language

DmitrySharabin:
Using `$insertBefore` inside `$merge` was a mistake I made. Let’s try to fix it.


LeaVerou:
Oh but maybe it should work, the intent seems clear

Issue #3960, opened 2025-06-12
Things blocking the `simplify` branch from being merged


DmitrySharabin:
1. Transform the `cshtml`, `jsx`, `tsx`, and `typescript` languages. For details, see https://github.com/PrismJS/prism/pull/3948#issuecomment-2916895852.

---

2. Correctly resolve grammars of required languages used as `inside` grammars.

Suppose we have the following grammar:

```js
{
  id: "foo",
  require: bar
  grammar: {
    token: {
      ...
      inside: "bar"
    }
  }
}
```

The `bar` language is _imported_ and can be accessed via the `languages` object associated with the `foo` language. However, `languages["bar"]` is evaluated lazily. Until it's accessed for the first time, the `bar` language won't be added to the `languageRegistry`, and `resolve(prism, "bar")` will **always** return `undefined`. (@LeaVerou had an [idea](https://github.com/PrismJS/prism/pull/3942#discussion_r2106219376) on how it can be fixed)

In [`resolve()`](https://github.com/PrismJS/prism/blob/37bdeeb77c2a43496ab6f56db95d9f2372b60771/src/core/tokenize/util.ts#L5), we don't have access to the language whose required dependencies are used as inside grammars (via `inside`). As a result, we don't have access to its `languages` property and can't get the resolved grammar. I attempted to address the issue in https://github.com/PrismJS/prism/pull/3956, although I'm unsure if it's the correct approach.

OTOH, we have a [`getLanguage()`](https://github.com/PrismJS/prism/blob/37bdeeb77c2a43496ab6f56db95d9f2372b60771/src/core/classes/language-registry.ts#L95) method to get languages from the language registry. However, with `getLanguage()`, we get _already resolved_ languages, but the required ones _might not be resolved_ until accessed at least once via the `languages` property.

---

3. [`ready`](https://github.com/PrismJS/prism/blob/37bdeeb77c2a43496ab6f56db95d9f2372b60771/src/core/classes/prism.ts#L35) and [`languagesReady`](https://github.com/PrismJS/prism/blob/37bdeeb77c2a43496ab6f56db95d9f2372b60771/src/core/classes/prism.ts#L33) in the `Prism` class and [`ready`](https://github.com/PrismJS/prism/blob/37bdeeb77c2a43496ab6f56db95d9f2372b60771/src/core/classes/component-registry.ts#L37) in the `ComponentRegistry` class should probably be reactive? 

[...]

PR #3961, opened 2025-06-16
In addition to waiting for all promises to settle, handle post-hoc additions/removals

DmitrySharabin:
Credit: https://github.com/mavoweb/mavo/blob/78efe2b9cadd09c1d131b8afd5fe2f38d5cfa8c7/src/mavo.js#L845

Partially addresses #3960

### DIFF
--- a/src/core/classes/prism.ts
+++ b/src/core/classes/prism.ts
@@ -73,7 +73,6 @@ export default class Prism {
 		if (!this.config.manual) {
 			this.waitFor.push(documentReady());
 
-			this.ready = allSettled(this.waitFor);
 			this.ready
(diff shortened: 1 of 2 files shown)


src/core/classes/prism.ts
```diff
@@ -73,7 +73,6 @@ export default class Prism {
 		if (!this.config.manual) {
 			this.waitFor.push(documentReady());
 
-			this.ready = allSettled(this.waitFor);
```
DmitrySharabin, line 76: We don't need this workaround anymore: we set the initial value of `ready` on line 35.

LeaVerou (approved):

PR #3966, opened 2025-06-20
Add utils for working with objects to a separate module

DmitrySharabin:
Adjust TS and Prettier settings accordingly.

For now, the `type-check` task fails. It will be fixed after merging #3965.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 /src/plugins
+tsconfig.json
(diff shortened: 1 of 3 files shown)


LeaVerou (changes requested):
Are these used anywhere? Each PR should leave the codebase in a better state than it was. It seems this one would leave it in a state of having utils that are unused, no?

DmitrySharabin:
Oh, you are absolutely right: those utils are not used anywhere now. I’ll get back to it a bit later, then.

DmitrySharabin:
Or better, I'm gonna close it for now.

PR #3973, opened 2025-06-24
Simplify the `Prism` class even more

DmitrySharabin:
- Move `tokenize()`, `_matchGrammar()` (and related code), and `resolve()` to separate modules.
- Ditch the `tokenize` symbol in favor of the `$tokenize` special property. Since symbols are excluded from the `for…in` loop, and `$tokenize` is not a symbol, to preserve the previous behavior, we need to handle it in `_matchGrammar` explicitly.
- Switch from `[tokenize]` to `$tokenize` in languages.


src/core/tokenize/match.ts
```diff
@@ -0,0 +1,206 @@
...
+	grammar: GrammarTokens,
+	startNode: LinkedListHeadNode<string | Token> | LinkedListMiddleNode<string | Token>,
+	startPos: number,
+	rematch?: RematchOptions
+): void {
+	const prism = this ?? singleton;
+
+	for (const token in grammar) {
+		const tokenValue = grammar[token];
+		if (!grammar.hasOwnProperty(token) || token === '$tokenize' || !tokenValue) {
```
DmitrySharabin: This is where we should handle special properties explicitly: when `tokenize` was a symbol, we didn't have to do that since it wasn't enumerable. Now, it shouldn't be matched as regular tokens. Probably because we don't do that in the `simplify` branch, we have tests that fail, and before, I couldn't understand why. Now I know. 🙂

Probably we should skip all special properties here (in the future), not only `$tokenize`? Shouldn't we replace `token === '$tokenize'` with something like `token.startsWith('$')`?

LeaVerou: Yes, I would use `startsWith('$')`. That was the whole reason to use that convention, that we can tell what is a special property even if we don't know what it does.

DmitrySharabin: Done! Thanks.

LeaVerou (approved):

PR #3976, opened 2025-06-26
`insert`, `insertAfter`, and new object utils

DmitrySharabin:
Adjust `tsconfig.json` to support modern syntax.


DmitrySharabin:
Close it for now. We can return to it later when we introduce other changes from the `simplify` branch.

PR #3979, opened 2025-06-27
Make the global bundle of Prism v2 work the same way as Prism v1 does

DmitrySharabin:
Now, the user can include Prism the way they used to (except v2 now uses the Autoloader plugin under the hood, so more languages can be highlighted right out of the box):

```html
<link rel="stylesheet" href="themes/prism.css" />
```

Besides, they have access to the Prism instance via `globalThis`.

---
- <kbd>&nbsp;2&nbsp;</kbd> #3979 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #3978

### DIFF
--- a/src/auto-start.ts
+++ b/src/auto-start.ts
@@ -4,49 +4,10 @@ import { documentReady } from './util/async';
 
 Prism.components.add(autoloader);
 
-export const PrismConfig = {
-	// TODO: Update docs
-	/**
-	 * By default, Prism will attempt to highlight all code elements (by calling {@link Prism#highlightAll}) on the
-	 * current page after the page finished loading. This might be a problem if e.g. you wanted to asynchronously load
-	 * additional languages or plugins yourself.
-	 *
-	 * By setting this value to `true`, Prism will not automatically highlight all code elements on the page.
-	 *
-	 * You obviously have to change this value before the automatic highlighting started. To do this, you can add an
-	 * empty Prism object into the global scope before loading the Prism script like this:
-	 *
-	 * ```js
-	 * window.Prism = window.Prism || {};
-	 * Prism.manual = true;
-	 * // add a new <script> to load Prism's script
-	 * ```
-	 *
-	 * @default false
-	 */
-	manual: false,
-};
-
-if (typeof document !== 'undefined' && typeof window !== 'undefined') {
-	// Get current script and highlight
-	const script = document.currentScript as HTMLScriptElement | null;
-	if (script && script.hasAttribute('data-manual')) {
-		PrismConfig.manual = true;
+void documentReady().then(() => {
+	if (!Prism.config.manual) {
+		Prism.highlightAll();
 	}
-
-	void documentReady().then(() => {
-		if (!PrismConfig.manual) {
-			Prism.highlightAll();
-		}
-	});
-
-	// Make Prism available globally
-	if (typeof globalThis !== 'undefined') {
-		(globalThis as any).Prism = Prism;
-	}
-}
-else {
-	PrismConfig.manual = true;
-}
+});
 
 export default Prism;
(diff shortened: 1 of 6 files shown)


src/auto-start.ts
```diff
@@ -39,6 +39,11 @@ if (typeof document !== 'undefined' && typeof window !== 'undefined') {
 			Prism.highlightAll();
 		}
 	});
+
+	// Make Prism available globally
+	if (typeof globalThis !== 'undefined') {
+		(globalThis as any).Prism = Prism;
+	}
```
DmitrySharabin: Even though I included this chunk in the PR, I wonder if we want to expose the Prism instance globally. 🤔

LeaVerou: We don’t, but the iife version should!

LeaVerou: You can check if `document.currentScript` is defined and expose it as a global if so. This means that the script is included via a non-module script and that it's in a browser, not Node.

DmitrySharabin: > We don’t, but the iife version should!

This is the issue with Rollup I still can't solve. It doesn't work as expected. I might miss something. I'll give it another look.

> You can check if `document.currentScript` is defined and expose it as a global if so. This means that the script is included via a non-module script and that it's in a browser, not Node.

Oh, will try

DmitrySharabin: > This is the issue with Rollup I still can't solve. It doesn't work as expected. I might miss something. I'll give it another look.

I managed to fix it: there should be another option enabled (`extend: true`), and we should use the default export, not the named one.

src/auto-start.ts
```diff
@@ -4,44 +4,10 @@ import { documentReady } from './util/async';
...
-	 * ```js
-	 * window.Prism = window.Prism || {};
-	 * Prism.manual = true;
-	 * // add a new <script> to load Prism's script
-	 * ```
-	 *
-	 * @default false
-	 */
-	manual: false,
-};
```
LeaVerou, line 28: Where did all this move to? I can't find it

DmitrySharabin: It's in `src/config.ts` (below). The code there allowed the simplifying of `auto-start.ts` and moving towards the changes introduced in the `simplify` branch.

LeaVerou: I'm asking about the long comment

src/config.ts
```diff
@@ -0,0 +1,43 @@
+const hasDOM = typeof document !== 'undefined' && typeof window !== 'undefined';
+const scriptElement: HTMLOrSVGScriptElement | null = hasDOM ? document.currentScript : null;
+const globalConfig: Record<string, PrismConfig[keyof PrismConfig] | null> =
+	// @ts-ignore
+	globalThis.Prism?.constructor?.name === 'Object' ? globalThis.Prism : {};
+
+function getGlobalSetting (name: string) {
```
LeaVerou, line 7: Are these functions new? If not, where were they before?

src/auto-start.ts
DmitrySharabin: I was using the `simplify` branch as the source of truth, and since you removed this comment there, I thought we might do the same here. If you feel it's still relevant, I can bring it back, though I'm not sure `v2` allows disabling manual highlighting that way. Only the global build supports that.

src/config.ts
DmitrySharabin: They are. They are from the `simplify` branch. I decided to introduce them here because they allow us to simplify `auto-start.ts` and prepare for adding support for other options.

LeaVerou: Can you link me to where they are in the `simplify` branch?

DmitrySharabin: Here you go:

https://github.com/PrismJS/prism/blob/7c7c598d69ea5c669deea900597647ac22f8dfea/src/config.ts#L7-L23

and

https://github.com/PrismJS/prism/blob/7c7c598d69ea5c669deea900597647ac22f8dfea/src/config.ts#L25-L33

LeaVerou (approved):

PR #3984, opened 2025-07-12
[Line Numbers] Automatically attach to the global Prism instance

DmitrySharabin:
---
- <kbd>&nbsp;2&nbsp;</kbd> #3984 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #3983

### DIFF
--- a/src/plugins/line-numbers/prism-line-numbers.ts
+++ b/src/plugins/line-numbers/prism-line-numbers.ts
@@ -1,4 +1,4 @@
-import Prism from '../../global';
+import prism from '../../global';
 import { getParentPre, isActive } from '../../shared/dom-util';
 import { isNonNull, noop } from '../../shared/util';
 import { combineCallbacks } from '../../util/combine-callbacks';
@@ -276,4 +276,4 @@ const Self = {
 
 export default Self;
 
-Prism.components.add(Self);
+prism.components.add(Self);
--- a/src/plugins/line-numbers/prism-line-numbers.ts
+++ b/src/plugins/line-numbers/prism-line-numbers.ts
@@ -1,6 +1,6 @@
+import Prism from '../../global';
 import { getParentPre, isActive } from '../../shared/dom-util';
 import { isNonNull, noop } from '../../shared/util';
-import { documentReady } from '../../util/async';
 import { combineCallbacks } from '../../util/combine-callbacks';
 import type { PluginProto } from '../../types';
 
@@ -276,11 +276,4 @@ const Self = {
 
 export default Self;
 
-if (typeof document !== 'undefined') {
-	void documentReady().then(() => {
-		const Prism = (globalThis as any).Prism;
-		if (Prism) {
-			Prism.components.add(Self);
-		}
-	});
-}
+Prism.components.add(Self);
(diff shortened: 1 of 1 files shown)


src/plugins/line-numbers/prism-line-numbers.ts
```diff
@@ -272,3 +273,14 @@ export default {
 		return combineCallbacks(removeListener, completeHook);
 	},
 } as PluginProto<'line-numbers'>;
+
+export default Self;
+
+if (typeof document !== 'undefined') {
```
DmitrySharabin: Without this check, Rollup chokes on bundling the code (when splitting it into chunks): in `documentReady`, it tries to access a property of an undefined `document`.

LeaVerou (approved):

LeaVerou:
Sorry, I meant to approve the other one!

src/plugins/line-numbers/prism-line-numbers.ts
```diff
@@ -272,3 +273,14 @@ export default {
...
+	void documentReady().then(() => {
+		const Prism = (globalThis as any).Prism;
+		if (Prism) {
+			Prism.components.add(Self);
```
LeaVerou: We should not be inflicting all this boilerplate on plugin authors!
As a design principle, syntax should not be more than the minimum required to declare intent. Boilerplate should be handled by suitable abstractions.

Besides the length and maintainability, whoever is including a plugin (either a Prism user or the plugin author adding the plugin to the global instance) should not need to care about when the document is ready. The Prism method one calls to add a plugin should automatically take care of that.

Also, ESM automatically takes care of running JS when the document is ready, so when using ESM one rarely needs to use a `documentReady()` event. Is this for the global Rollup build? If so, we should approach it differently (but still in an abstraction, not every single plugin). Additionally, not all plugins even need to wait for the document to be ready, since most just add code to hooks, and the hooks already run at the right time.

As a priority of constituencies, it's users > Prism users > Plugin users > Prism maintainers. 

All a plugin should need to do is something like:

```js
import prism from "/path/to/global/prism"

export default const Self = { /* plugin */ };

prism.components.add(Self);
```

or even

```js
import prism from "/path/to/global/prism"

export default prism.components.add({ /* plugin */ });
```

LeaVerou (changes requested):
See comment

src/plugins/line-numbers/prism-line-numbers.ts
DmitrySharabin: All excellent points. I tried to address your feedback and removed the boilerplate, but now this PR depends on #3985.

src/plugins/line-numbers/prism-line-numbers.ts
```diff
@@ -199,7 +200,7 @@ export class LineNumbers {
 	}
 }
 
-export default {
+const Self = {
```
DmitrySharabin, line 203: `export default const Self = { ... }` isn't allowed. 🤷‍♂️

src/plugins/line-numbers/prism-line-numbers.ts
```diff
@@ -1,3 +1,4 @@
+import Prism from '../../global';
```
LeaVerou: I would lowercase to not confuse this with the `Prism` class:

```suggestion
import prism from '../../global';
```

LeaVerou (approved):

PR #3985, opened 2025-07-14
ESM bundle build optimization

DmitrySharabin:
Rel to #3984.

- Ditch the `global` bundle (see below) in favor of a JS file
- Remove chunks (use native ESM features)
- Additionally, output `auto-load.ts` to `prism.js` (when we did that in a separate bundle, another instance of Prism was created automatically when attaching plugins to the global instance of Prism, which is wrong). As a result, `v2` can be used the same way (including automatically attaching plugins) as `v1` ~but should be included as a module~, i.e., `<script src="prism.js"></script>`

~I wonder if we still need to expose a Prism instance via `globalThis`? It seems like no `v2` functionality depends on it. By merging this PR, we won't do that (for now, since we have a separate `glo~al` bundle, we expose `Prism`).~

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -386,7 +386,6 @@ async function buildJS () {
 
 	const defaultOutputOptions: OutputOptions = {
 		dir: './dist',
-		chunkFileNames: '_chunks/[name]-[hash].js',
 		validate: true,
 		sourcemap: 'hidden',
 	};
@@ -400,7 +399,13 @@ async function buildJS () {
 		}
 	> = {
 		esm: {
-			rollupOptions: defaultRollupOptions,
+			rollupOptions: {
+				...defaultRollupOptions,
+				input: {
+					...input,
+					'prism': path.join(SRC_DIR, 'auto-start.ts'),
+				},
+			},
 			outputOptions: defaultOutputOptions,
 		},
 		cjs: {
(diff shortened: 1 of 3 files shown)


scripts/build.ts
```diff
@@ -411,21 +416,7 @@ async function buildJS () {
...
-					'prism': path.join(SRC_DIR, 'auto-start.ts'),
-				},
-			},
-			outputOptions: {
-				...defaultOutputOptions,
-				format: 'iife',
-				name: 'Prism',
-				exports: 'default',
-				extend: true,
+				chunkFileNames: '_chunks/[name]-[hash].js',
```
LeaVerou: Why are we using chunks, and why only in CJS?

DmitrySharabin: Honestly, IDK. This is the first project where I work with bundlers and chunks. Let me remove it even from the CJS bundle and see what we've got.

DmitrySharabin: I removed chunks from all the bundles

scripts/build.ts
```diff
@@ -413,28 +412,22 @@ async function buildJS () {
...
 	try {
 		for (const bundle of Object.values(bundles)) {
 			bundle.build = await rollup(bundle.rollupOptions);
 			await bundle.build.write(bundle.outputOptions);
 		}
+
+		// Global “bundle”
+		await writeFile(
+			path.join(DIST_DIR, 'prism.js'),
+			`if (globalThis.document?.currentScript) {
```
LeaVerou: Why are we writing it like this, instead of having an actual file?

src/prism.ts
```diff
@@ -0,0 +1,4 @@
+if (globalThis.document?.currentScript) {
```
LeaVerou: I would rename this to `prism.global.js`

LeaVerou: Also, I guess we don't really need the currentScript check. If someone doesn't want a global, don't import this file. 🤷🏼

DmitrySharabin: All makes sense. Fixed!

LeaVerou (approved):

PR #3986, opened 2025-07-15
Fix issue with relative URL in dynamic import

DmitrySharabin:
For now, when using the global bundle, we get the `Uncaught (in promise) TypeError: Failed to resolve module specifier './index.js'. The base URL is about:blank because import() is called from a CORS-cross-origin script.` error.

### DIFF
--- a/src/prism.global.ts
+++ b/src/prism.global.ts
@@ -1,2 +1,4 @@
 // In browser and imported via non-ESM
-import('./index.js').then(({ default: prism }) => ((globalThis as any).Prism = prism));
+const url = new URL('./index.js', (globalThis.document.currentScript as HTMLScriptElement).src);
+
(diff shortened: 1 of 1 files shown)


src/prism.global.ts
```diff
@@ -1,2 +1,4 @@
 // In browser and imported via non-ESM
-import('./index.js').then(({ default: prism }) => ((globalThis as any).Prism = prism));
+const url = new URL('./index.js', (globalThis.document.currentScript as HTMLScriptElement).src);
```
LeaVerou: We need to make sure this exists first.

DmitrySharabin: Fixed!

LeaVerou (approved):

PR #4000, opened 2025-08-10
Switch from TS to JS

DmitrySharabin:
### Summary

- All core files, language definitions, plugins, tests, scripts (build, changelog, etc.), benchmark are now JS
- Types (where possible; some TS types don't have analogues in JSDoc) are in JSDoc comments
- TypeScript is still in use inside the `build` script to build types (`.d.ts` files)
- Internally, instead of building `components.js` from `components.json`, [import attributes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with) are used to read the JSON file directly
- ESLint rules are adjusted: all TS dependencies and rules are removed
- All TS-related packages are uninstalled, `tsconfig.json` files are removed
- All NPM scripts in `package.json` adjusted accordingly (no more script for checking TS)
- Step for checking TS is removed from the GitHub workflow
- All tests, builds, regex coverage, etc., still pass. The only check that fails is ESLint

Prism `v2` can still be used in the same way as `v1`. To see it in action (with files from this PR), go to: https://codepen.io/dmitrysharabin/pen/QwjqywJ?editors=1000


scripts/build.js
```diff
@@ -327,6 +352,32 @@ async function copyComponentsJson () {
...
+		if (stats.isDirectory()) {
+			files.push(...(await findFiles(fullPath)));
+		}
+		else if (path.extname(item) === ext) {
+			files.push(fullPath);
+		}
+	}
+
+	return files;
+}
```
DmitrySharabin: I adjusted the process of building types, so it would be nice if you could review this part (1/3).

Full code is [here](https://github.com/PrismJS/prism/blob/5ccca819dad0da8cd58497b8b83db3a5966ecf46/scripts/build.js#L355-L413).

scripts/build.js
```diff
@@ -337,17 +388,14 @@ async function buildTypes () {
...
-	const compilerOptions: ts.CompilerOptions = {
-		...parsedConfig.options,
+	/** @type {import('typescript').CompilerOptions} */
+	const compilerOptions = {
+		target: ts.ScriptTarget.ES2022,
+		module: ts.ModuleKind.ES2022,
+		moduleResolution: ts.ModuleResolutionKind.Node,
+		esModuleInterop: true,
+		allowJs: true,
+		checkJs: false,
```
DmitrySharabin: This one (2/3).

scripts/build.js
```diff
@@ -356,57 +404,49 @@ async function buildTypes () {
 		noEmitOnError: false,
 	};
 
-	const program = ts.createProgram(parsedConfig.fileNames, compilerOptions);
+	// Get all .js files from the src directory
+	const files = await findFiles(SRC_DIR);
+
+	const program = ts.createProgram(files, compilerOptions);
```
DmitrySharabin: And the last one (3/3). 

It's straightforward, but you might find flaws we need to fix.

scripts/build.js
```diff
@@ -1,20 +1,23 @@
...
 import MagicString from 'magic-string';
 import { rollup } from 'rollup';
 import ts from 'typescript';
 import { webfont } from 'webfont';
-import { toArray } from '../src/util/iterables';
-import { components } from './components';
-import { parallel, runTask, series } from './tasks';
-import type { ComponentProto } from '../src/types';
-import type { OutputOptions, Plugin, RollupBuild, RollupOptions, SourceMapInput } from 'rollup';
+import components from '../src/components.json' with { type: 'json' };
```
DmitrySharabin, line 12: Here, I have switched to import attributes.

scripts/build.js
```diff
@@ -276,23 +289,35 @@ const inlineRegexSourcePlugin: Plugin = {
  * is a waste of CPU and memory, and it causes the JS thread to be block for roughly 200ms during page load.
  *
  * @see https://github.com/PrismJS/prism/issues/2768
+ *
+ * @type {Plugin}
  */
-const lazyGrammarPlugin: Plugin = {
+const lazyGrammarPlugin = {
```
DmitrySharabin, line 295: We still need the build step, not only because of types (we could do it in an NPM script), but because of this plugin (see [the comment above](https://github.com/PrismJS/prism/blob/5ccca819dad0da8cd58497b8b83db3a5966ecf46/scripts/build.js#L284-L291)).

package.json
```diff
@@ -30,21 +30,20 @@
 		"node": ">=18"
 	},
 	"scripts": {
-		"benchmark": "tsx benchmark/benchmark.ts",
-		"build": "tsx scripts/build.ts",
+		"benchmark": "node benchmark/benchmark.js",
+		"build": "node scripts/build.js",
```
DmitrySharabin, line 34: Here, we got rid of `tsx`.

package.json
```diff
@@ -30,21 +30,20 @@
...
-		"regex-coverage": "mocha tests/coverage.ts",
-		"test:components": "mocha tests/components-test.ts",
-		"test:core": "mocha tests/core/**/*.ts",
-		"test:identifiers": "mocha tests/identifier-test.ts",
-		"test:languages": "mocha tests/run.ts",
-		"test:patterns": "mocha tests/pattern-tests.ts",
-		"test:plugins": "mocha tests/plugins/**/*.ts",
-		"test:runner": "mocha tests/testrunner-tests.ts",
-		"test": "npm-run-all test:*",
-		"tsc": "tsc && tsc -p tests/tsconfig.json"
```
DmitrySharabin, line 47: And this script (`tsc`) is not needed anymore.

DmitrySharabin: All other scripts are not deleted, but the extension is changed from `.ts` to `.js`. I have no idea why they are shown as deleted in the diff.

LeaVerou:
I need to take a closer look, but some high level responses:

> ### Summary
> * All core files, language definitions, plugins, tests, scripts (build, changelog, etc.), benchmark are now JS

🎉 

> * Types (where possible; some TS types don't have analogues in JSDoc) are in JSDoc comments

For more complex types we actually _want_ to keep, we can have our own `.d.ts` files and _import_ them in JSDoc.

> * TypeScript is still in use inside the `build` script to build types (`.d.ts` files)
> * Internally, instead of building `components.js` from `components.json`, [import attributes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with) are used to read the JSON file directly

I'm not sure about this. Baseline 2025 might be a bit too recent for something used as widely as Prism and has no fallback. Does this file actually _need_ to be JSON? Can't we just make it JS?
But also, this seems entirely orthogonal to the TS -> JS conversion.

> * ESLint rules are adjusted: all TS dependencies and rules are removed
> * All TS-related packages are uninstalled, `tsconfig.json` files are removed

Don't we need them for typedoc?

> * All NPM scripts in `package.json` adjusted accordingly (no more script for checking TS)

I think it may be good to have something that catches inconsistencies in types, as long as it's opt-in and doesn't block CI.

> * Step for checking TS is removed from the GitHub workflow
> * All tests, builds, regex coverage, etc., still pass. The only check that fails is ESLint
> 
> Prism `v2` can still be used in the same way as `v1`. To see it in action (with files from this PR), go to: [codepen.io/dmitrysharabin/pen/QwjqywJ?editors=1000](https://codepen.io/dmitrysharabin/pen/QwjqywJ?editors=1000)

DmitrySharabin:
Thank you for the feedback. Let me think about it (and fix some obvious things). I’ll get back to you soon.

DmitrySharabin:
> > * Types (where possible; some TS types don't have analogues in JSDoc) are in JSDoc comments
> 
> For more complex types we actually _want_ to keep, we can have our own `.d.ts` files and _import_ them in JSDoc.

We already do that—most of the complex types are already in `types.d.ts`, and we import them in JS files. I'll revise the changes to see what other types make sense to move there for better compatibility.

> > * Internally, instead of building `components.js` from `components.json`, [import attributes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with) are used to read the JSON file directly
> 
> I'm not sure about this. Baseline 2025 might be a bit too recent for something used as widely as Prism and has no fallback.

Sorry for misleading you with the link. The import attributes are _not_ in the production code. They are used in the `build` and `create-changelog` scripts only. Both those scripts are supposed to be run in Node. Node supports import attributes starting from `v18.20` (see the History details [here](https://nodejs.org/api/esm.html#import-attributes)). Node 18+ is the minimal version we support, so we are safe here.

> Does this file actually _need_ to be JSON? Can't we just make it JS?

I have no strong opinion here. If you think it's better to switch to a JS file (I agree that it's more flexible than JSON), I'm up for it.

> But also, this seems entirely orthogonal to the TS -> JS conversion.

It might be. Should I restore it and use it instead of import attributes, then?

> > * All TS-related packages are uninstalled, `tsconfig.json` files are removed
> 
> Don't we need them for typedoc?

Not necessarily. Typedoc supports many options that can be passed to the TS compiler if the result it produces is not good enough.

> > * All NPM scripts in `package.json` adjusted accordingly (no more script for checking TS)
> 
> I think it may be good to have something that catches inconsistencies in types, as long as it's opt-in and doesn't block CI.

Agreed. I'll take a look at how we do that in our other projects.

LeaVerou:
Thanks! All sounds good.

> It might be. Should I restore it and use it instead of import attributes, then?

No, if they're only used in a build script it's fine

DmitrySharabin:
**Some updates**

- I ended up working with types, moving the complex ones to `types.d.ts` and importing them where needed
- All `.d.ts` files we generate automatically from JSDoc seem to be generated fine. However, there's an issue with the TypeScript compiler: if we have both `@private` and `@type` on class properties, it ignores either the property type (and doesn't emit it) or the property modifier. I would suggest using real JS private properties in such cases. For now, I think the correct types are more important. In most cases, we follow the underscore convention for naming such properties in the code base.

> > * All NPM scripts in `package.json` adjusted accordingly (no more script for checking TS)
> 
> I think it may be good to have something that catches inconsistencies in types, as long as it's opt-in and doesn't block CI.

This is the final step that should be taken. @LeaVerou, could you elaborate a bit? Do we want to check JS files using the TS compiler and show errors? Should it work the same way the `lint` script does (when we run ESLint in the terminal)?

scarf005:
may I ask why the Prism project is moving from TS to JS? It'd be great to have reasoning in the PR body since JSDoc is considered less ergonomic than Typescript.

DmitrySharabin:
> > > * All NPM scripts in `package.json` adjusted accordingly (no more script for checking TS)
> > 
> > 
> > I think it may be good to have something that catches inconsistencies in types, as long as it's opt-in and doesn't block CI.
> 
> This is the final step that should be taken. @LeaVerou, could you elaborate a bit? Do we want to check JS files using the TS compiler and show errors? Should it work the same way the `lint` script does (when we run ESLint in the terminal)?

Some more updates:

- Restore `tsconfig.json` but allow implicit `any`
- Add the `typecheck` NPM script (like `lint` but for types)

Now, all JS files (except `demo.js`, which comes with plugins) are type-checked, but it's not blocking the CI.

LeaVerou:
@scarf005 
> may I ask why the Prism project moving from TS to JS? It'd be great to have reasoning in the PR body since JSDoc is considered less ergonomic than Typescript.

**Please note that there has never been a released Prism version using TS, so the Prism project is not _moving_ to JS, it is simply _not moving_ to TS.** A past maintainer rewrote the codebase to use TS as it would help him move faster, and I agreed in the interest of making progress. Since he left and I took over again (with my apprentice @DmitrySharabin's help) we really gave TS a good shot since the v2 codebase was already using it and it was cheaper than switching at that point. 

However, the DX has been a constant uphill battle, and we found we were spending more time dealing with TS than actually making progress on v2. Besides the sheer time spent jumping through TS hoops to simply express how things worked, hoops that provided absolutely no benefit in terms of finding bugs more easily, there were also the pointless TS assertions for obvious (to a human) things that added noise all over the codebase. Go through the code this PR removes (e.g. in the grammar files) for many samples of this.

TS is fantastic for getting folks from classical strongly typed / statically typed languages into JS. But for JS-native folks, it means that many of the dynamic coding and metaprogramming paradigms that JS allows (and that Prism takes good advantage of) are very challenging to express in TS.

Additionally, Prism is the type of project where performance can be critical, so there is an advantage to avoiding such heavy abstractions unless absolutely necessary. Staying closer to the metal (at least as much as one can be with JS) allows to better optimize highlighting.

Do note that using JS doesn't mean the project is untyped, or that all types are now defined in JSDoc. JSDoc is used for simple types, whereas complex types are defined in `.d.ts` files that are imported in JSDoc. Type checking is there if we need it to, it's just not on the critical path for contributions or runtime. Many projects have been using this architecture, e.g. Svelte.

If you're a TS user, you could help by trying out this PR and reporting back on the experience!

DmitrySharabin:
All issues with types (which appeared after allowing TS checking of JS files) are now fixed. If we still had a CI task for checking types, we would pass it. That means the changes in this PR don't make the codebase worse than it was before.

I think the PR is ready for review.

scripts/build.js
LeaVerou: Lazy grammars are evaluated when needed, so not sure why we need a build step for them?

LeaVerou:
@DmitrySharabin I just went through the PR, though not very deeply. All LGTM, except the comment I left: why do we still need a build step given that lazy grammars are (supposed to be) evaluated when they're first used? That's the whole point of lazy grammars — otherwise they're not saving anything if we pre-build them.

scripts/build.js
DmitrySharabin: I might miss something, but as far as I understand it, not all grammars are lazily evaluated. We have grammars with bare grammar properties, not functions, and when the corresponding file is loaded, that might cause an issue when we have many such grammars loaded simultaneously. The plugin wraps all bare grammar objects into functions that return them. This is what I got from the comment above the plugin.

LeaVerou: Perhaps those grammars are too simple to cause slowdown, or just weren't converted to be lazy.
Having such a plugin is fine when there is *already* a build step, but that by itself doesn't suffice as the _only_ reason to have a build step.

DmitrySharabin: Oh, I might mislead you with the comment that we need the build step only because of that plugin. My bad. We have other plugins to apply, as well as build types, calculating file sizes (to be used on the website's Download page), building a treeview icon font, etc. So, wrapping bare grammar objects with functions is just one step of many other build steps. Probably, the build script should also be revised, but I'd leave it for another PR.

For context, this is why that plugin was introduced: https://github.com/PrismJS/prism/issues/2768

LeaVerou: There's a difference between a build step that runs when the codebase changes to build certain artifacts vs a build step that is needed to produce importable code. With TS, we needed a build step for importable JS to even be produced in the first place. By reverting back to JS, people should be able to import from `src` and it should just work, and I'm not sure if there's even a point of having a separate `/dist/` for ESM. All other things that you mentioned are things that produce _other_ artifacts based on the source files, they don't *transform* the code. So currently, the lazy grammars plugin sounds like it *is* the only plugin that actually transforms the code Prism consumers import.

But I agree with:

>  Probably, the build script should also be revised, but I'd leave it for another PR.

Yes.

LeaVerou (approved):
I haven't looked super deeply, but LGTM from a quick skim. We can fix any issues as they come up. Let's land this baby and kill TS with 🔥 ! Thanks for the hard work @DmitrySharabin!

(thinking @getify would approve of the spirit of this PR 😀)

DmitrySharabin:
> (thinking @getify would approve of the spirit of this PR 😀)

Seconded 😃

getify:
nicely done! 💯

PR #4001, opened 2025-08-11
[redirects] Simplify resource URLs

DmitrySharabin:
Instead of `https://dev.prismjs.com/dist/prism.js` and `https://dev.prismjs.com/dist/languages/c.js`, etc., one can write `https://dev.prismjs.com/prism.js` and `https://dev.prismjs.com/languages/c.js` respectively.

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -1,0 +1,5 @@
+# Don't redirect direct URLs
+/dist/* /dist/:splat 200
+
+# Otherwise, if one requests something from this server, that means they need something from the dist folder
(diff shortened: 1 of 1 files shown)


LeaVerou (commented):
I'm not sure. It's not a big issue to include `/dist/` and it makes it more robust — this could break the moment you add a file that actually resolves and happens to match a `/dist/` file name. 🤔

DmitrySharabin:
Okay, let’s not do it, then. 😊

PR #4020, opened 2025-10-11
Add support for Magik

### LINKED ISSUE #4008: Add support for Magik
**Language**

Magik is an [object-oriented programming](https://en.m.wikipedia.org/wiki/Object-oriented_programming) language that supports [multiple inheritance](https://en.m.wikipedia.org/wiki/Multiple_inheritance) and [polymorphism](https://en.m.wikipedia.org/wiki/Polymorphism_(computer_science)), and it is [dynamically typed](https://en.m.wikipedia.org/wiki/Dynamic_typing)

**Additional resources**

https://github.com/sebastiaanspeck/highlightjs-magik

https://github.com/krn-robin/tree-sitter-magik

sebastiaanspeck:
Closes #4008


sebastiaanspeck:
@LeaVerou Is there something we can do for getting this merged? Does a plan exist, when this might be done?

DmitrySharabin:
@sebastiaanspeck, I'm on it. I'll be back with the review results shortly.

DmitrySharabin (changes requested):
Foremost, thank you for putting effort into adding a new language to Prism!

After reviewing this definition against the [official Prism extending docs](https://prismjs.com/extending), I've spotted a few issues worth addressing.

The docs state:

> “All tokens before a greedy token should also be greedy” and “the greedy tokens should be the first tokens of a language.” (see https://prismjs.com/extending#greedy-matching)

Currently, non-greedy tokens (`keyword`, `builtin`, `boolean`) sit between greedy ones (`comment` and `char`), which violates this rule and can cause subtle tokenization bugs.

I'd suggest moving all greedy tokens to the top of the grammar, followed by all non-greedy tokens.

But before reordering the tokens, we need to remove `greedy: true` from some of them:

- No characters in a `number` literal start another greedy token, so there's no need in `greedy: true`.
- Moving `operator` before `keyword` and `property` before `operator` in the grammar eliminates the need for `greedy: true` in those tokens.
- Since Prism matches the text against tokens sequentially, we need to adjust the order of some operators within the `operator` array: move the compound assignment patterns (`_and<<`, etc.) *before* the plain `<<` and `\^<<` patterns. With that change in place, we don't need `greedy: true` there either.

In short, I'd recommend the following order of tokens in the grammar:

1. `comment` _(greedy)_
2. `char` _(greedy)_
3. `string` _(greedy)_
4. `regex` _(greedy)_
--- Remove all `greedy: true` from everywhere in the following tokens ---
5. `property`
6. `operator`
7. `keyword`
8. The rest: `builtin`, `boolean`, `variable`, `symbol`, `number`, `punctuation`

Please verify that with all the proposed changes in place, Prism highlights the language correctly.

sebastiaanspeck:
Thank you for the review. I will have a look if those changes still reflect the same thing as it does now.

sebastiaanspeck:
I am sorry, we had several fixes in v1, but didn't port them to v2. Did that now, but now it probably violates the docs again. Would you be so kind to help me review this PR to get it to a correct state?

DmitrySharabin:
Let's try to apply the suggestions from the previous review first. That might help us move further.

I'd incorporate the fixes from v1 into the PR _after_ applying the results of the first review. Then, we can bring the changes from v1 one by one and place them into the correct places right away. In case of doubts, we can discuss what the best place for them would be. What do you think?

sebastiaanspeck:
> Let's try to apply the suggestions from the previous review first. That might help us move further.
> 
> I'd incorporate the fixes from v1 into the PR _after_ applying the results of the first review. Then, we can bring the changes from v1 one by one and place them into the correct places right away. In case of doubts, we can discuss what the best place for them would be. What do you think?

Thank you for your corporation and plan. Will do!

sebastiaanspeck:
> Let's try to apply the suggestions from the previous review first. That might help us move further.
> 
> I'd incorporate the fixes from v1 into the PR _after_ applying the results of the first review. Then, we can bring the changes from v1 one by one and place them into the correct places right away. In case of doubts, we can discuss what the best place for them would be. What do you think?

Did that now. If the PR looks okay, I will add the fixes from v1.

DmitrySharabin:
LGTM! The checks seem to pass (ignore the lint check; I know that it fails). Let's move on to the next phase.

sebastiaanspeck:
I just pushed a commit which contains all fixes/enhancements from v1. Some tests are still failing, but I hope those will work out after we fixed all issues.

src/languages/magik.js
```diff
@@ -0,0 +1,137 @@
...
+export default {
+	id: 'magik',
+	grammar: {
+		'comment': [
+			{ pattern: /##.*/, greedy: true, alias: 'documentation' }, // documentation
+			{ pattern: /#(?!#).*/, greedy: true }, // comments
+		],
+
+		'char': {
+			pattern: /%(?:[a-z][a-z0-9_?!]*|.)/i,
```
DmitrySharabin: I'd suggest replacing `a-z0-9_` with [`\w`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_class_escape). Considering that you have the `i` flag, it means the same.

DmitrySharabin: Not only here, by the way. Everywhere in the grammar.

src/languages/magik.js
```diff
@@ -0,0 +1,137 @@
...
+				lookbehind: true
+			},
+			{
+				pattern: /\b_self\b/i,
+				greedy: true
+			}
+		],
+
+		'variable': [
+			{ pattern: /\|![a-z0-9_?!]+!\|/i }, // variable encased like |!var!|
```
DmitrySharabin: You don't have to use the object notation when the only property is `pattern`. You can use the regexes directly as values. And it's okay to have regexes and objects (when they have properties besides `pattern`) in one array.

src/languages/magik.js
```diff
@@ -0,0 +1,137 @@
...
+			inside: {
+				'modifier': /classify_level|topic|usage/,
+				'pragma-punctuation': { pattern: /[={},]/ }
+			}
+		},
+
+		'operator': [
+			/(?:\*\*\^?|\*\^?|\/\^?|-\^?|\+\^?)<</, // compound arithmetic assignment
+			/\^<</, /<</, // assignment operators
+			/>>/, /\b_return\b/, // return operators
```
DmitrySharabin: You have these two on L63. Is it intentional? Your tests expect it to be keywords, but since they are matched as `operator` (it comes before `keyword` in the grammar), your tests fail.

src/languages/magik.js
```diff
@@ -0,0 +1,137 @@
...
+		'boolean': /\b_(?:false|maybe|true)\b/i,
+
+		'symbol': {
+			pattern: /(^|[^a-z0-9_]):[a-z0-9_|]+(?:\([^)]*\)|\{[^}]*\}|\[[^\]]*\])?[a-z0-9_|]*/i,
+			lookbehind: true
+		},
+
+		'number': {
+			pattern: /\b\d+(?:\.\d+)?(?:[e&][+-]?\d+)?\b|\b(?:[2-9]|[12]\d|3[0-6])r[a-z0-9]+\b/i,
+			greedy: true
```
DmitrySharabin: Shouldn't be greedy.

src/languages/magik.js
```diff
@@ -0,0 +1,137 @@
...
+
+		'regex': {
+			pattern: /\/(?:(?!\/)(?:\\.|[^\\\/\r\n])*\/[qisdlmuCX]*|\/)/,
+			greedy: true
+		},
+
+		'declaration': [
+			{ pattern: /(\b_package\s+).*/i, greedy: true, lookbehind: true },
+			{ pattern: /(\b_global\s+)(?!_)\w+/i, greedy: true, lookbehind: true },
+			{ pattern: /(\b_constant\s+)[a-z_]+/i, greedy: true, lookbehind: true }
```
DmitrySharabin: Those shouldn't be greedy as well (as I see it).

src/languages/magik.js
```diff
@@ -0,0 +1,137 @@
...
+			alias: 'symbol'
+		},
+
+		'slot': {
+			pattern: /(^|[\s({])\.\s*[A-Z_]+/i,
+			lookbehind: true
+		},
+
+		'function': [
+			{ pattern: /\b_(?:abstract|endmethod|iter|method|private)\b/, greedy: true }, // method keywords
```
DmitrySharabin: And this one.

src/languages/magik.js
```diff
@@ -0,0 +1,137 @@
...
+			{
+				pattern: /(\b_method\b\s+)\S+(?=\.)/,
+				greedy: true,
+				lookbehind: true
+			},
+			{
+				pattern: /\b_self\b/i,
+				greedy: true
+			}
+		],
```
DmitrySharabin, line 116: Move before `function` and remove `greedy`.

tests/languages/magik/operator_feature.test
```diff
@@ -0,0 +1,77 @@
+^<<
+<<
+_and<<
```
DmitrySharabin, line 1: This and some others (`_or<<`, ` _mod<<`, `_div<<`, etc.) are not operators anymore: `_and` is now in `keyword`, and `<<` in `operator`, so they split. You need to adjust either the grammar or the tests. I don't know the language, so it's up to you.

DmitrySharabin:
Please, try to identify the issues with tests before we iterate over the PR one more time. You know the language and how it should be highlighted. If you're stuck, I'll help, but it's you who knows the stuff. 😉

sebastiaanspeck:
All tests are green now! Thanks for the quick review once more!

sebastiaanspeck:
> Please, try to identify the issues with tests before we iterate over the PR one more time. You know the language and how it should be highlighted. If you're stuck, I'll help, but it's you who knows the stuff. 😉

I am done now. You can review this PR again.

src/languages/magik.js
```diff
@@ -0,0 +1,129 @@
...
+		'number': /\b\d+(?:\.\d+)?(?:[e&][+-]?\d+)?\b|\b(?:[2-9]|[12]\d|3[0-6])r[a-z0-9]+\b/i,
+
+		'symbol': {
+			pattern: /(^|\W):(?:\|[^|]*\||\\.|[\w?!])+/,
+			lookbehind: true
+		},
+
+		'operator': [
+			/\^<</, /<</, { pattern: /_(?:and|andif|or|orif|xor)<</i, greedy: true }, { pattern: /(?:\*\*\^?|\*\^?|\/\^?|_mod\^?|_div\^?|-\^?|\+\^?)<</i, greedy: true }, // assignment operators
+			{ pattern: /(?:\*\*\^?|\*\^?|\/\^?|-\^?|\+\^?)<</, greedy: true }, // assignment operators
```
DmitrySharabin: Is there a reason why we have `greedy: true` here? Rearranging doesn't help?

DmitrySharabin (changes requested):
See the comment. And we seem to have a failing test. I'll approve the CI checks so you can see which one. I think it's `string_feature`.

sebastiaanspeck:
Thank you for your patience. Really appreciate it. It is the first time I write a new language for PrismJS

src/languages/magik.js
```diff
@@ -0,0 +1,129 @@
...
+
+		'number': /\b\d+(?:\.\d+)?(?:[e&][+-]?\d+)?\b|\b(?:[2-9]|[12]\d|3[0-6])r[a-z0-9]+\b/i,
+
+		'symbol': {
+			pattern: /(^|\W):(?:\|[^|]*\||\\.|[\w?!])+/,
+			lookbehind: true
+		},
+
+		'operator': [
+			/\^<</, /<</, { pattern: /_(?:and|andif|or|orif|xor)<</i, greedy: true }, { pattern: /(?:\*\*\^?|\*\^?|\/\^?|_mod\^?|_div\^?|-\^?|\+\^?)<</i, greedy: true }, // assignment operators
```
DmitrySharabin: Another two pairs of greedy patterns. Are those leftovers?

sebastiaanspeck: Yes, they are. Will remove them.

src/languages/magik.js
```diff
@@ -0,0 +1,129 @@
...
+		'number': /\b\d+(?:\.\d+)?(?:[e&][+-]?\d+)?\b|\b(?:[2-9]|[12]\d|3[0-6])r[a-z0-9]+\b/i,
+
+		'symbol': {
+			pattern: /(^|\W):(?:\|[^|]*\||\\.|[\w?!])+/,
+			lookbehind: true
+		},
+
+		'operator': [
+			/\^<</, /<</, /_(?:and|andif|or|orif|xor)<</i, /(?:\*\*\^?|\*\^?|\/\^?|_mod\^?|_div\^?|-\^?|\+\^?)<</i, // assignment operators
+			/(?:\*\*\^?|\*\^?|\/\^?|-\^?|\+\^?)<</, // assignment operators
```
DmitrySharabin: These two lines have two issues:

1. **Ordering**: \`\^<<\` and \`<<\` fire before the compound patterns. Prism processes patterns in array order, scanning all text for each one. So \`<<\` steals the \`<<\` from \`_and<<\` before the compound pattern matches. Compound patterns must come first.

2. **Duplicate**: Line 49 is a subset of line 48's second pattern (missing \`_mod\` and \`_div\`).

```suggestion
			/_(?:and|andif|or|orif|xor)<</i, /(?:\*\*\^?|\*\^?|\/\^?|_mod\^?|_div\^?|-\^?|\+\^?)<</i, /\^<</, /<</, // assignment operators
```

DmitrySharabin (commented):
## Operator compound assignment patterns: ordering fix

Per the [extending docs](https://prismjs.com/extending), Prism scans all remaining text for each pattern in sequence. Since `\^<<` and `<<` appear *before* the compound patterns (`_and<<`, `**<<`, etc.) in the array, `<<` consumes the `<<` portion from `_and<<` before the compound pattern gets a chance — splitting `_and<<` into `keyword(_and)` + `operator(<<)`.

The fix is to reorder compound patterns before `\^<<` and `<<`.

Also, line 49 is a duplicate of the pattern on line 48 but missing `_mod` and `_div`.

See inline suggestion below.

src/languages/magik.js
```diff
@@ -0,0 +1,128 @@
...
+
+		'number': /\b\d+(?:\.\d+)?(?:[e&][+-]?\d+)?\b|\b(?:[2-9]|[12]\d|3[0-6])r[a-z0-9]+\b/i,
+
+		'symbol': {
+			pattern: /(^|\W):(?:\|[^|]*\||\\.|[\w?!])+/,
+			lookbehind: true
+		},
+
+		'operator': [
+			/_(?:and|andif|or|orif|xor)<</i, /(?:\*\*\^?|\*\^?|\/\^?|\^?|_mod\^?|_div\^?|-\^?|\+\^?)<</i, /^<</, /<</, // assignment operators
```
DmitrySharabin: `\^?` matches an empty string. I think this is not what we want. And since we need `^` to be used literally, we should escape it in the regex.

```suggestion
			/_(?:and|andif|or|orif|xor)<</i, /(?:\*\*\^?|\*\^?|\/\^?|_mod\^?|_div\^?|-\^?|\+\^?)<</i, /\^<</, /<</, // assignment operators
```

DmitrySharabin: This also breaks the `regex-coverage` test. Run `npm run regex-coverage` to see it yourself.

DmitrySharabin:
> Thank you for your patience. Really appreciate it. It is the first time I write a new language for PrismJS

Don't mention it. We're almost there. Don't give up! 😀

src/languages/magik.js
```diff
@@ -0,0 +1,128 @@
...
+			/\b_primitive\b/i, // primitive
+			/\b_(?:endtry|try|when)\b/i, // try
+			/\b_(?:endprotect|locking|protect|protection)\b/i, // protect
+			/\b_(?:endlock|lock)\b/i, // lock
+			/\b_with\b/i, // standalone since _finally, _handling, _throw, _try, _leave and _continue all can have this
+			/\b_(?:allresults|gather|optional|scatter)\b/i // parameter options
+		],
+
+		'slot': {
+			pattern: /(^|[\s({])\.\s*[A-Z_]+/i,
```
DmitrySharabin: This is not an error, but just to be stylistically consistent, and since you are using the `i` flag, I'd suggest the following change:

```suggestion
			pattern: /(^|[\s({])\.\s*[a-z_]+/i,
```

src/languages/magik.js
```diff
@@ -0,0 +1,128 @@
...
+			{
+				pattern: /(\b_method\s+)\S+(?=\.)/,
+				lookbehind: true
+			},
+			/\b_self\b/i
+		],
+
+		'function': [
+			/\b_(?:abstract|endmethod|iter|method|private)\b/, // method keywords
+			/\b_(?:endproc|proc)\b/, // procedure
```
DmitrySharabin: I wonder if these should also have the `i` flag, like all other literal keywords do?

DmitrySharabin: I'd suggest checking if other patterns (not only in this token) miss one.

src/languages/magik.js
```diff
@@ -0,0 +1,128 @@
...
+			greedy: true
+		},
+
+		'string': {
+			pattern: /"(?:\\.|[^"\\\r\n])*"|'(?:\\.|[^'\\\r\n])*'/,
+			greedy: true
+		},
+
+		'regex': {
+			pattern: /\/(?:(?:\\.|[^\\/\r\n])+\/[qisdlmuCX]*|\/)/,
```
DmitrySharabin: I'm a bit confused that we have mixed case in here: `[qisdlmuCX]`. Since I don't know the language, this might be correct, but it's worth checking.

DmitrySharabin:
I also asked Claude if something is missing from the language. It pointed out the `_recursive` keyword. I don't know if it's correct. Just a heads-up.

src/languages/magik.js
sebastiaanspeck: Yes, this is correct. It is based on the flags for Java regex: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/regex/Pattern.html

image

sebastiaanspeck: just verified that `C` and `X` can't be set as would be the case with `i`. Those need to be set in a different way. Removed them from the grammar.

sebastiaanspeck:
> I also asked Claude if something is missing from the language. It pointed out the `_recursive` keyword. I don't know if it's correct. Just a heads-up.

Claude doesn't know anything about Magik 😉 `_recursive` isn't a keyword.

src/languages/magik.js
```diff
@@ -18,7 +18,7 @@ export default {
 		},
 
 		'regex': {
-			pattern: /\/(?:(?:\\.|[^\\/\r\n])+\/[qisdlmuCX]*|\/)/,
+			pattern: /\/(?:(?:\\.|[^\\/\r\n])+\/[qisdlmu]*|\/)/,
```
DmitrySharabin: Shouldn't they just be replaced with `cx` instead of removing them?

sebastiaanspeck:
```suggestion
			pattern: /\/(?:(?:\\.|[^\\/\r\n])+\/[qisdlmucx]*|\/)/,
```

DmitrySharabin:
The last step is to verify that the code is highlighted correctly in a live environment. I've built a tiny playground for the Magik language: https://codepen.io/dmitrysharabin/full/OPRwGKz. It always works with the most recent changes in this PR. Please check if the code in Magik is correctly highlighted with Prism. Give it a test drive so we make sure that the first iteration of the language is solid before we merge it.

sebastiaanspeck:
Looks good in the codepen :smile: Should we add the snippet thats there as a test case as well?

DmitrySharabin:
> Should we add the snippet thats there as a test case as well?

No need, I think. I just added it so that the playground works with at least some code when it loads. Tests should cover all that independently (on a token basis). Just check with your examples. And if everything's OK, and we have tests for all kinds of tokens, then we can merge it.

sebastiaanspeck:
Then I believe this is GTG 🚀

DmitrySharabin (approved):
LGTM! I hope to merge it today.

LeaVerou (approved):

DmitrySharabin:
@sebastiaanspeck, thank you for adding a new language to Prism, and congratulations! 🎉

PR #4029, opened 2025-11-08
Simplify core (the rest of the changes from the `simplify` branch)

DmitrySharabin:
Depends on #4018.

### Summary

Most of the code was authored by @LeaVerou.

- Create/update core classes + add corresponding utils + improve types
- [languages] Use the updated API
- [plugins] Fix auto-import + adopt the new API
- [tests] Adopt the new API + improve tests coverage in `bash` and `css-selector` (actually, we didn't have tests for it at all 🤷‍♂️)
- [build] Add the `global.js` entry point to work with bare Prism (without the `autoloader` plugin loaded automatically)
- [benchmark] Adopt the new API
- [GH actions] Bump up NodeJS versions
We use the `CustomEvent` class, which [became stable in `v20.13`](https://nodejs.org/api/events.html#class-customevent).

**Demo:** https://codepen.io/dmitrysharabin/pen/qEbeEXv

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #4032 
- <kbd>&nbsp;1&nbsp;</kbd> #4029 👈


DmitrySharabin:
Something weird is happening here: IDK why the tests fail. Locally, _all_ the tests pass. Hm, I need to investigate it.

**UPDATE:** Fixed! ~The `coverage` task fails expectedly—Prism found a few regexes not covered by the tests we have. I'm not sure if it's a regression or if we previously missed those cases. 🤔~

Anyway, it's a giant leap forward, since there are a few minor changes left in the `simplify` branch that we need to migrate, but they are tiny compared to that PR.

DmitrySharabin:
> Something weird is happening here: IDK why the tests fail. Locally, _all_ the tests pass. Hm, I need to investigate it.

Oh, I know why this happens. Locally, I have applied the changes from #4018. Therefore, the updated API leverages the new, more declarative language definitions. I'll update the PR's description to reflect that.

LeaVerou (approved):

PR #4045, opened 2025-11-28
Fix: TOML table to avoid splitting brackets from key

alasdairwilson:
I had discovered an issue where TOML table header parsing had undesired behaviour where table declarations like [table] and [[array]] were being split across multiple tokens, creating split spans when parsed. The bug was actually, for me, found in refractor but it uses prismjs language files upstream which cause the issue.

I would propose is to be consistent and copy the approach used in parsing .ini files, where the whole thing becomes a nested structure

using this:

```js
inside: {
	table-name': {
			pattern: RegExp('(^\\[\\[?\\s*)' + dottedKey),
			lookbehind: true,
			alias: 'class-name',
		},
		'punctuation': /\[|\]/,
	},
```

to produce this output:

```text
	["table", [
		["punctuation", "["],
		["table-name", "table"],
		["punctuation", "]"]
	]],
	["table", [
		["punctuation", "["],
		["punctuation", "["],
		["table-name", "array"],
		["punctuation", "]"],
		["punctuation", "]"]
	]],
```

This PR swaps to that approach and I personally think this is a superior approach.

### DIFF
--- a/src/languages/toml.js
+++ b/src/languages/toml.js
@@ -4,6 +4,7 @@ export default {
 	grammar () {
 		const key = /(?:[\w-]+|'[^'\n\r]*'|"(?:\\.|[^\\"\r\n])*")/.source;
 		const dottedKey = key + '(?:\\s*\\.\\s*' + key + ')*';
+		const trailingComment = '(?=\\s*(?:#.*)?$)';
 
 		return {
 			'comment': {
@@ -12,15 +13,18 @@ export default {
 			},
 			'table': {
 				// keep entire table header (including brackets) under one parent token
-				pattern: RegExp('(^[\\t ]*)\\[\\[?\\s*' + dottedKey + '\\s*\\]\\]?', 'm'),
+				pattern: RegExp(
+					'(^[\\t ]*)(?:\\[(?!\\[)\\s*' + dottedKey + '\\s*\\]' + trailingComment +
+						'|\\[\\[\\s*' + dottedKey + '\\s*\\]\\]' + trailingComment + ')',
+					'm'
+				),
 				lookbehind: true,
 				greedy: true,
-				alias: 'class-name',
 				inside: {
 					'table-name': {
 						pattern: RegExp('(^\\[\\[?\\s*)' + dottedKey),
 						lookbehind: true,
-						alias: 'class-name',
+						alias: 'variable',
 					},
 					'punctuation': /\[|\]/,
 				},
--- a/src/languages/toml.js
+++ b/src/languages/toml.js
@@ -3,15 +3,7 @@ export default {
 	id: 'toml',
 	grammar () {
 		const key = /(?:[\w-]+|'[^'\n\r]*'|"(?:\\.|[^\\"\r\n])*")/.source;
-
-		/**
-		 *
-		 * @param {string} pattern
-		 * @returns {string}
-		 */
-		function insertKey (pattern) {
-			return pattern.replace(/__/g, () => key);
-		}
+		const dottedKey = key + '(?:\\s*\\.\\s*' + key + ')*';
 
 		return {
 			'comment': {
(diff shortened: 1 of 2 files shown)


alasdairwilson:
Is there any way to get a review on this?

I have been reluctant to monkeypatch the language definition files in my affected application because I was hoping for an upstream fix. I think this would be a clear improvement in terms of how an unknown user would expect the parsing to behave as well as bring how toml is parsed more in line with other languages.

DmitrySharabin:
> Is there any way to get a review on this?

Hey, @alasdairwilson,

Thank you for your patience. I'll do my best to review it today. 🤞

src/languages/toml.js
```diff
@@ -3,35 +3,30 @@ export default {
...
 				pattern: /#.*/,
 				greedy: true,
 			},
 			'table': {
-				pattern: RegExp(
-					insertKey(/(^[\t ]*\[\s*(?:\[\s*)?)__(?:\s*\.\s*__)*(?=\s*\])/.source),
-					'm'
-				),
+				// keep entire table header (including brackets) under one parent token
+				pattern: RegExp('(^[\\t ]*)\\[\\[?\\s*' + dottedKey + '\\s*\\]\\]?', 'm'),
```
DmitrySharabin: Since we touch this regexp, why not fix the bug that seems to be standing: the not-balanced brackets? We can make Prism highlight only well-formed TOML table tokens, so `[name]]` and `[[name]` are not highlighted. I'm not sure whether we can reach that goal without explicitly stating two alternatives, though. What do you think?

src/languages/toml.js
```diff
@@ -3,35 +3,30 @@ export default {
...
 				lookbehind: true,
 				greedy: true,
 				alias: 'class-name',
+				inside: {
+					'table-name': {
+						pattern: RegExp('(^\\[\\[?\\s*)' + dottedKey),
+						lookbehind: true,
+						alias: 'class-name',
```
DmitrySharabin: Is it intentional that we duplicate the outer alias? If we would rather `table-name` not be highlighted differently, this line is redundant. But since it's a separate token, we probably need to give it a default highlighting that all themes support. `variable`? `constant`?

See https://prismjs.com/tokens as a reference.

tests/languages/toml/table_feature.test
DmitrySharabin, line 10: I wonder if it's worth adding a test for a dotted table name and malformed tables. I don't have a strong opinion here; it's up to you.

DmitrySharabin (changes requested):
Thank you for working on this! Almost there. See comments.

src/languages/toml.js
alasdairwilson: Ah good idea, I think I managed to make such a change. It meant having to specifically capture trailing comments though as initially `[name]]` would match `[name]` and `]`

src/languages/toml.js
alasdairwilson: Yeah I don't really know what I was doing here - with the duplicattion, i dont think it was a deliberate choice.

I am honestly torn as to whether it should be inherit or be separately defined, since table-name is now a separate token i've gone with an explicit alias and, to me, variable is a bit more generic than constant which is perhaps too specific.

I am totally happy to follow your lead on this though, I am just guessing.

tests/languages/toml/table_feature.test
alasdairwilson: I have never seen a project with easier to write tests, so I added some more to cover some additional cases. MAybe it will catch some regression some day.

src/languages/toml.js
DmitrySharabin: I think `variable` is the right call here (at least, I'd use it 😀). The good thing is that now the user will see the token highlighted. For those who need to style it differently, the `.token.table-name { ... }` CSS rule is for the rescue.

alasdairwilson:
Thanks very much for your review, think that is everything addressed.

src/languages/toml.js
```diff
@@ -3,35 +3,34 @@ export default {
...
 			'comment': {
 				pattern: /#.*/,
 				greedy: true,
 			},
 			'table': {
+				// keep entire table header (including brackets) under one parent token
 				pattern: RegExp(
-					insertKey(/(^[\t ]*\[\s*(?:\[\s*)?)__(?:\s*\.\s*__)*(?=\s*\])/.source),
+					'(^[\\t ]*)(?:\\[(?!\\[)\\s*' + dottedKey + '\\s*\\]' + trailingComment +
+						'|\\[\\[\\s*' + dottedKey + '\\s*\\]\\]' + trailingComment + ')',
```
DmitrySharabin: How about making the change minimal and cover (mostly) all the same cases and make the highlighting TOML-spec complient (`(?!\])` handles the case that the spec conciders as malformed)? This allows us not to introduce the concept of `trailingComment` at all (the `comment` token will do its job).

The only case that is not covered is `[name] foo = bar`: `[name]` will be highlighted even though the line is invalid if I understand the case correctly. But I think it's OK since Prism is a highlighter and not parser or whatever; it's up to the user to write correct TOML.

What do you think?

```suggestion
					'(^[\\t ]*)(?:\\[\\[\\s*' +
						dottedKey +
						'\\s*\\]\\]|\\[\\s*' +
						dottedKey +
						'\\s*\\])(?!\\])',
```

alasdairwilson: Yeah, your solution is simpler. I am inclined as well to agree that catching that syntax error is out of scope for prism.

src/languages/toml.js
```diff
@@ -3,35 +3,37 @@ export default {
...
-		/**
-		 *
-		 * @param {string} pattern
-		 * @returns {string}
-		 */
-		function insertKey (pattern) {
-			return pattern.replace(/__/g, () => key);
-		}
+		const dottedKey = key + '(?:\\s*\\.\\s*' + key + ')*';
+		const trailingComment = '(?=\\s*(?:#.*)?$)';
```
DmitrySharabin: Now, when we are agreed on the regex simplification, this is not needed anymore. 🙂

DmitrySharabin (approved):
Thank you for your contribution! LGTM!

PR #4073, opened 2026-04-09
magik: make function regex simpler and add extra test case

sebastiaanspeck:


sebastiaanspeck:
@DmitrySharabin after being very helpful in #4020 and adding support for Magik, we found two issues during daily use, can you have a look at this PR?

PR #4075, opened 2026-04-12
magik: define symbol before number to match symbols with numbers

sebastiaanspeck:


sebastiaanspeck:
@DmitrySharabin after being very helpful in #4020 and adding support for Magik, we found two issues during daily use, can you have a look at this PR?

DmitrySharabin:
Hey @sebastiaanspeck,

Is there a chance you could join your two PRs into one? Since both of them are improvements (fixes), it makes sense for them to come together.

sebastiaanspeck:
@DmitrySharabin done in #4076

PR #4076, opened 2026-04-13
Apply fixes for Magik language

sebastiaanspeck:


sebastiaanspeck:
#4075 and #4073 are included in this PR (and extended with more fixes)

sebastiaanspeck:
@DmitrySharabin as requested, a new PR which contains all found fixes.

src/languages/magik.js
```diff
@@ -22,28 +22,48 @@ export default {
 			greedy: true
 		},
 
-		'declaration': [
-			{ pattern: /(\b_package\s+).*/i, lookbehind: true },
-			{ pattern: /(\b_global\s+)(?!_)\w+/i, lookbehind: true },
-			{ pattern: /(\b_constant\s+)[a-z_]+/i, lookbehind: true }
-		],
+		'slot': {
+			pattern: /(^|[\s({])\.\s*[a-z][\w?!]+/i,
```
DmitrySharabin: With the new regex, `slot` requires at least 2 characters after `.` (see `[a-z][\w?!]+`). Is it intentional?

src/languages/magik.js
```diff
@@ -22,28 +22,48 @@ export default {
...
+		],
+
+		'dynamic-variable': [
+			{ pattern: /\|![\w?!]+!\|/, alias: 'variable' }, // variable encased like |!var!|
+			{ pattern: /\|![\w?!]+\|!/, alias: 'variable' }, // variable encased like |!var|!
+			{ pattern: /!\|[\w?!]+\|!/, alias: 'variable' }, // variable encased like !|var!|
+			{ pattern: /!\|\|!/, alias: 'variable' }, // empty variable !||!
+			{ pattern: /[a-z_]+:![a-z][\w?!]*!/, alias: 'variable' }, // variable with a prefix like sw:!var!
+			{ pattern: /![a-z][\w?!]*!/, alias: 'variable' }, // variable encased like !var!
+		],
```
DmitrySharabin, line 57: Are `dynamic-variable`s case-sensitive? The patterns don't have the `i` flag, but, e.g., `global-variable`s have.

In the first three patterns, a variable can start with a digit (`[\w?!]+` allows that), but not in the last two patterns. Is that correct?

src/languages/magik.js
```diff
@@ -22,28 +22,48 @@ export default {
...
+
+		'global-variable': [
+			{ pattern: /\b[a-z_]+:\w+\b/i, alias: 'variable' }, // variable with a prefix like sw:gis_program_manager
+			{ pattern: /[a-z_]+:\|\w+\|/i, alias: 'variable' }, // variable with a prefix like sw:|gis_program_manager|
```
DmitrySharabin: Just to make sure: we can't use `?` and `!` in the global vars like we do in, e.g., `symbol` or `dynamic-variable`?

src/languages/magik.js
```diff
@@ -22,28 +22,48 @@ export default {
...
+		},
 
 		'pragma': {
 			pattern: /_pragma.*/,
-			alias: 'prolog',
-			inside: {
-				'modifier': /classify_level|topic|usage/,
-				'pragma-punctuation': { pattern: /[={},]/ }
-			}
+			alias: 'prolog'
```
DmitrySharabin: You don't plan to tokenize the inner of `pragma`? If the pragma is covered by the other tokens of the language, then it makes sense to use `inside: 'magik'` to let Prism tokenize it correctly.

src/languages/magik.js
```diff
@@ -22,28 +22,48 @@ export default {
...
+			{ pattern: /![a-z][\w?!]*!/, alias: 'variable' }, // variable encased like !var!
+		],
+
+		'global-variable': [
+			{ pattern: /\b[a-z_]+:\w+\b/i, alias: 'variable' }, // variable with a prefix like sw:gis_program_manager
+			{ pattern: /[a-z_]+:\|\w+\|/i, alias: 'variable' }, // variable with a prefix like sw:|gis_program_manager|
+		],
+
+		'declaration': [
+			{ pattern: /(\b_package\s+).*/i, lookbehind: true },
```
DmitrySharabin: `.*` swallows the rest of the line. Consider making it more concrete if it makes sense (it's up to you) with `\w+` or `[\w.]+`, like you do in the following patterns.

src/languages/magik.js
DmitrySharabin, line 113: It's not in the PR. I just want to draw your attention to it: you aliased `_unset` and `_constant` to `symbol` even though other tokens started from `_` (e.g., `keyword-variable`s) are aliased to `keyword`. That means that they will be highlighted differently. If this is what you need (and it's semantically correct), then ignore this comment. The alternative solution is to add a CSS rule to highlight `_unset` differently: `token.unset.keyword { color: ...; }`. But it's up to you.

DmitrySharabin (changes requested):
I left some comments/questions. Please, take a look. I'd also suggest adding a few tests to catch the mentioned cases if they make sense.

src/languages/magik.js
sebastiaanspeck: Well, the first character needs to be a lowercase letter, it is okay if it is only a one-letter slot, but will change the regex to support both (one-letter and 2+ letters).

DmitrySharabin: In that case, you just need to replace `+` with `*`, and it should just work: `[a-z][\w?!]*`.

src/languages/magik.js
sebastiaanspeck: A pragma is some kind of documentation, put above a method definition;

```magik
_pragma(classify_level=basic, topic={prism, magik}, usage={subclassable})
```

But I chose to revert the change and indeed tokenize the inner as well.

src/languages/magik.js
sebastiaanspeck: It was mainly done for CSS to at least alias it to a known "keyword".

In a [theme](https://github.com/sebastiaanspeck/prism/blob/75cc0db411a4abc4d6a499fdc85a6e694266c653/themes/prism-deeper-blue.css#L120) we made, we use the same colour for symbol and unset/constant:

```css
.token.symbol,
.token.unset,
.token.constant,
.token.char {
    color: #a2cd5a;
}
```

src/languages/magik.js
sebastiaanspeck: Well-spotted! It is indeed possible.

src/languages/magik.js
```diff
@@ -37,13 +37,37 @@ export default {
...
+			{ pattern: /\|![\w?!]+!\|/, alias: 'variable' }, // variable encased like |!var!|
+			{ pattern: /\|![\w?!]+\|!/, alias: 'variable' }, // variable encased like |!var|!
+			{ pattern: /!\|[\w?!]+\|!/, alias: 'variable' }, // variable encased like !|var!|
+			{ pattern: /!\|\|!/, alias: 'variable' }, // empty variable !||!
+			{ pattern: /[a-z_]+:![a-z][\w?!]*!/i, alias: 'variable' }, // variable with a prefix like sw:!var!
+			{ pattern: /![a-z][\w?!]*!/i, alias: 'variable' }, // variable encased like !var!
+		],
+
+		'global-variable': [
+			{ pattern: /\b[a-z_]+:[\w?!]+\b/i, alias: 'variable' }, // variable with a prefix like sw:gis_program_manager
```
sebastiaanspeck:
```suggestion
			{ pattern: /[a-z_]+:[\w?!]+/i, alias: 'variable' }, // variable with a prefix like sw:gis_program_manager
```

sebastiaanspeck:
@DmitrySharabin I resolved your comments. Can you have a look if it looks good to you?

src/languages/magik.js
DmitrySharabin: And `_unset` and `_constant` are different kinds of keywords, and you would rather not use `keyword` as an alias for them, right? I just want to understand why you chose another alias if, in your theme, you style them similarly. 🤔

sebastiaanspeck: To have them styled in a default theme, with default keywords. AFAIK `unset` and/or `symbol` aren't default keywords.

DmitrySharabin: [They are not](https://prismjs.com/tokens). But to have the default highlighting, you can assign the `keyword` alias to all the tokens that start with `_` (to be consistent across the language definition).

sebastiaanspeck: How would that look like in code?

DmitrySharabin: You already have `_class`, `_dynamic`, etc., all aliased to `keyword`.
```js
'keyword-variable': {
	pattern: /\b_(?:class|dynamic|global|import|local)\b/i,
	alias: 'keyword' // ← here
},
```

If `_unset` and `_constant` should be highlighted the same way (even in the default theme), I'd suggest:

```diff
'unset': {
	pattern: /\b_unset\b/i,
-	alias: 'symbol'
+ 	alias: 'keyword'
},

'constant': {
	pattern: /\b_constant\b/i,
-	alias: 'symbol'
+ 	alias: 'keyword'
},
```

In that case, all `_foo` will have the same alias `keyword`. And it is consistent across the language definition.

src/languages/magik.js
```diff
@@ -37,13 +37,37 @@ export default {
...
+			{ pattern: /[a-z_]+:[\w?!]+/i, alias: 'variable' }, // variable with a prefix like sw:gis_program_manager
+			{ pattern: /[a-z_]+:\|[\w?!]+\|/i, alias: 'variable' }, // variable with a prefix like sw:|gis_program_manager|
+		],
+
+		'declaration': [
+			{ pattern: /(\b_package\s+)\w+\b/i, lookbehind: true },
+			{ pattern: /(\b_(?:constant|global|import|local)\s+)(?!_)\w+\b/i, lookbehind: true },
+		],
+
+		'number': /(?<!\|)\b\d+(?:\.\d+)?(?:[e&][+-]?\d+)?\b|\b(?:[2-9]|[12]\d|3[0-6])r[a-z0-9]+\b/i,
```
DmitrySharabin: I looked up, and if I'm not mistaken, `|16r0000FFFF|` is a valid variable name. For example, in
```magik
_global |16r0000FFFF| << colour.new_with_properties(:blue, 255)
```

If so, for now, it will be interpreted as a number (`16r0000FFFF`) between `|` and `|`. To avoid it, we need to fix the second alternative. Or better, move `(?<!\|)` outside both alternatives. One lookbehind guards both.

```suggestion
		'number': /(?<!\|)(?:\b\d+(?:\.\d+)?(?:[e&][+-]?\d+)?\b|\b(?:[2-9]|[12]\d|3[0-6])r[a-z0-9]+\b)/i,
```

I'd suggest adding a regression test to `variable_feature.test`.

DmitrySharabin (approved):
LGTM!

Nit: Could you run Prettier, please, on the updated files? It seems like it should fix some formatting issues.

sebastiaanspeck:
> Nit: Could you run Prettier, please, on the updated files? It seems like it should fix some formatting issues.

You mean on the test files?

```
➜  prism git:(fix/magik) npx prettier tests/languages/magik/*
tests/languages/magik/boolean_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/boolean_feature.test".
tests/languages/magik/builtin_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/builtin_feature.test".
tests/languages/magik/char_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/char_feature.test".
tests/languages/magik/comment_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/comment_feature.test".
tests/languages/magik/constant_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/constant_feature.test".
tests/languages/magik/declaration_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/declaration_feature.test".
tests/languages/magik/dynamic-variable_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/dynamic-variable_feature.test".
tests/languages/magik/function_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/function_feature.test".
tests/languages/magik/global-reference_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/global-reference_feature.test".
tests/languages/magik/global-variable_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/global-variable_feature.test".
tests/languages/magik/keyword-operator_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/keyword-operator_feature.test".
tests/languages/magik/keyword-variable_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/keyword-variable_feature.test".
tests/languages/magik/keyword_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/keyword_feature.test".
tests/languages/magik/number_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/number_feature.test".
tests/languages/magik/operator_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/operator_feature.test".
tests/languages/magik/pragma_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/pragma_feature.test".
tests/languages/magik/punctuation_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/punctuation_feature.test".
tests/languages/magik/regex_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/regex_feature.test".
tests/languages/magik/self_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/self_feature.test".
tests/languages/magik/slot_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/slot_feature.test".
tests/languages/magik/string_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/string_feature.test".
tests/languages/magik/symbol_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/symbol_feature.test".
tests/languages/magik/unset_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/unset_feature.test".
tests/languages/magik/variable_feature.test
[error] No parser could be inferred for file "/Users/sebastiaan.speck/src/prism/tests/languages/magik/variable_feature.test".
```

DmitrySharabin:
Yeah, I think it won't work on the test files. Let's just run Prettier on `magik.js`. It should be enough.

sebastiaanspeck:
> Yeah, I think it won't work on the test files. Let's just run Prettier on `magik.js`. It should be enough.

Which issues does it solve then? I ran it, but IMO the output was less readable than what we have now.

DmitrySharabin:
> > Yeah, I think it won't work on the test files. Let's just run Prettier on `magik.js`. It should be enough.
> 
> Which issues does it solve then? I ran it, but IMO the output was less readable than what we have now.

One is that if anyone ever wants to change the file, save the changes, and Prettier kicks in automatically (as it should according to the recommended IDE setup), they should undo the changes made by Prettier or commit/submit unrelated changes.
For example, if we introduce a new feature in Prism and we need the languages to adopt it, we tweak the language definitions. And before we added Prettier, we had many unrelated changes in the corresponding commits (that was a pain; trust me, once, I had to update all the languages Prism supports 😅). We try to avoid it. And yes, Prettier rules are opinionated, but we are fine with them in this project.

I should've asked you to do that in the first PR, but I somehow missed it.

sebastiaanspeck:
> > > Yeah, I think it won't work on the test files. Let's just run Prettier on `magik.js`. It should be enough.
> 
> > 
> 
> > Which issues does it solve then? I ran it, but IMO the output was less readable than what we have now.
> 
> 
> 
> One is that if anyone ever wants to change the file, save the changes, and Prettier kicks in automatically (as it should according to the recommended IDE setup), they should undo the changes made by Prettier or commit/submit unrelated changes.
> 
> For example, if we introduce a new feature in Prism and we need the languages to adopt it, we tweak the language definitions. And before we added Prettier, we had many unrelated changes in the corresponding commits (that was a pain; trust me, once, I had to update all the languages Prism supports 😅). We try to avoid it. And yes, Prettier rules are opinionated, but we are fine with them in this project.
> 
> 
> 
> I should've asked you to do that in the first PR, but I somehow missed it.

I'll run it then and put aside my own opinion.

DmitrySharabin:
Merged! Thank you.

sebastiaanspeck:
@DmitrySharabin I backported the changes to a branch for v1 so we can use Magik already in several plugins, but with the current grammar, the lint breaks, while in v2 it doesn't.

https://github.com/sebastiaanspeck/prism/actions/runs/24758384791

DmitrySharabin:
> @DmitrySharabin I backported the changes to a branch for v1 so we can use Magik already in several plugins, but with the current grammar, the lint breaks, while in v2 it doesn't.
> 
> https://github.com/sebastiaanspeck/prism/actions/runs/24758384791

The root cause is the lookbehind assertion (`(?<!\|)`) in the `number` token. This is an ES2018 feature. In Prism v1, we use ESLint v7. To make it support the lookbehind assertions in regexes, you can add 
```json
"parserOptions": {
  "ecmaVersion": 2018
}
```
to `.eslintrc.js` (I suppose, before the `rules` property should work).

PR #4078, opened 2026-05-02
Fix Magik to allow `?` and `!` in variable names

sebastiaanspeck:


sebastiaanspeck:
@DmitrySharabin can you please have a look? `npm run test:languages` does work, but `test:identifiers` fails somehow.

src/languages/magik.js
```diff
@@ -148,7 +148,7 @@ export default {
 
 		'variable': [
 			/\|[\w?!]+\|/, // variable encased like |var|, |0|, |123|
-			{ pattern: /(^|[^.])\b[a-z]\w*\b/i, lookbehind: true },
+			{ pattern: /(^|[^.])[a-z][\w?!]*/i, lookbehind: true },
```
DmitrySharabin: Indeed, we don't need `\b` on the right side of the regexp anymore (for Magik, `[\w?!]` technically means the same), but we still need to make sure that we match Magic variables on the word boundary, not mid-level (by removing `\b` we broke it; the `identifiers` tests catch exactly such cases). This change fixes it.

```suggestion
			{ pattern: /(^|[^.])\b[a-z][\w?!]*/i, lookbehind: true },
```

DmitrySharabin (commented):
Hey @sebastiaanspeck,

Thanks for being patient. I left a comment on why you face the issue and how to fix it. I hope it helps!

src/languages/magik.js
sebastiaanspeck: Thanks! I wasn't that sharp!

DmitrySharabin (approved):
LGTM!

DmitrySharabin:
Thank you! Merged.

PR #4079, opened 2026-05-12
Highlight `using` / `await using` declarations in JavaScript & TypeScript

Skn0tt:
Adds support for the ES2026 [Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management) `using` and `await using` declarations.

```js
using file = openFile();
await using conn = getConn();
for (using x of arr) {}
```

Previously, `using` was treated as a plain identifier in all positions; these snippets now highlight `using` as a keyword.

### DIFF
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -31,6 +31,11 @@ export default {
 			'keyword': [
 				{
 					pattern:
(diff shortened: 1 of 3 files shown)


DmitrySharabin (approved):

DmitrySharabin:
@Skn0tt, thank you for working on it! Merged.

PR #4080, opened 2026-05-19
Add support for Python's new `lazy import` syntax

ZeroIntensity:
Starting in Python 3.15, `lazy import xyz` and `lazy from xyz import abc` are valid under [PEP 810](https://peps.python.org/pep-0810/).


ZeroIntensity:
Friendly ping @DmitrySharabin

src/languages/python.js
```diff
@@ -67,4 +67,4 @@ export default {
 		'operator': /[-+%=]=?|!=|:=|\*\*?=?|\/\/?=?|<[<=>]?|>[=>]?|[&|^~]/,
 		'punctuation': /[{}[\];(),.:]/,
 	},
-};
+};
```
DmitrySharabin: Please remove such changes from the PR. It would be nice to keep the diff as minimal and not have unrelated changes if possible.

tests/languages/python/keyword_feature.test
```diff
@@ -9,7 +9,7 @@ pass print raise return
 try while with yield
 nonlocal
 and not or
-match case _:
+match case _: lazy
```
DmitrySharabin, line 12: Is there a reason to add the keyword to this line and not a separate one? Is it related to the rest of it (like `and`, `not`, and `or` are related and that's why they are on the same line)?

tests/languages/python/keyword_feature.test
```diff
@@ -25,9 +25,9 @@ match case _:
...
 	["keyword", "nonlocal"],
 	["keyword", "and"], ["keyword", "not"], ["keyword", "or"],
-	["keyword", "match"], ["keyword", "case"], ["keyword", "_"], ["punctuation", ":"]
+	["keyword", "match"], ["keyword", "case"], ["keyword", "_"], ["punctuation", ":"], ["keyword", "lazy"]
 ]
 
 ----------------------------------------------------
 
-Checks for all keywords.
+Checks for all keywords.
```
DmitrySharabin: See my first comment.

DmitrySharabin (changes requested):
Thank you for keeping the language definition up-to-date!

A few minor comments to address, and we are good to go.

tests/languages/python/keyword_feature.test
ZeroIntensity: Yeah, these are all the [soft keywords](https://docs.python.org/3.16/reference/lexical_analysis.html#soft-keywords). These are still valid identifiers, whereas things like `nonlocal` are not.

src/languages/python.js
ZeroIntensity: Sorry, I think my editor did this automatically. Is there a commit hook or something the repo uses for formatting that I can apply?

DmitrySharabin: Nothing that I'm aware of. Sorry. Probably a separate commit will do? We just need the opposite of what you already have. IIRC, editors should have an option to save without formatting. In VS Code, I do <kbd>Cmd + K</kbd>, <kbd>S</kbd>.

But, yeah, we need to do something to fix all such issues automatically. I'll add it to my list.

ZeroIntensity: Should be fixed now.

DmitrySharabin (approved):
LGTM! Thank you.

src/languages/python.js
DmitrySharabin: Thank you. Merged!

PR #4087, opened 2026-06-15
Build dist/prism.js as IIFE for synchronous globalThis.Prism

DmitrySharabin:
## Summary

- `prism.global.js` used `import()` to dynamically load `index.js`, making `globalThis.Prism` available only **asynchronously** — breaking backward compatibility with v1 where `Prism` was available immediately after the `<script>` tag
- Replace it with an IIFE build target that bundles `auto-start.js` (core + autoloader) into a single self-contained file (~20KB)
- Rollup's `name: 'Prism'` assigns the default export to `var Prism`, making it synchronously available as a global
- Languages are still loaded dynamically by the autoloader via `import()`
- `global.js` now bridges the IIFE and ESM singletons: when `globalThis.Prism` is already a Prism instance (set by the IIFE), ESM plugins reuse it instead of creating a second one

Live demo: https://codepen.io/dmitrysharabin/pen/xbbdXZb

## Test plan

- [x] Load `dist/prism.js` via `<script>` and verify `globalThis.Prism` is defined synchronously in the next `<script>` block
- [x] Verify autoloaded language highlighting still works (e.g. `<code class="language-javascript">`)
- [x] Verify ESM (`import Prism from 'prismjs'`) and CJS (`require('prismjs')`) still work
- [x] Verify pre-load config (`window.Prism = { manual: true }`) is still read correctly
- [x] Verify ESM plugins (`line-numbers`, `inline-color`, `show-invisibles`) loaded via `<script type="module">` work alongside the IIFE build
- [x] Verify Node.js (no IIFE) — `globalThis.Prism` stays `undefined`, ESM/CJS return a fresh instance

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -334,7 +334,8 @@ function toRenderedChunk (s) {
 	};
 }
 
-const terserPlugin = rollupTerser({
+/** @type {import('@rollup/plugin-terser').Options} */
+const terserOptions = {
 	ecma: 2020,
 	module: true,
 	compress: {
@@ -348,7 +349,8 @@ const terserPlugin = rollupTerser({
 		comments: false,
 	},
 	keep_classnames: true,
-});
+};
+const terserPlugin = rollupTerser(terserOptions);
 
 async function clean () {
 	const outputDir = path.join(__dirname, '../dist');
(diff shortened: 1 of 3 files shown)


LeaVerou (approved):
LGTM, only concern is that this will cause trouble if v1 Prism and v2 Prism co-exist in the same page. But if you do that, you'll probably have more issues anyway.

DmitrySharabin:
> But if you do that, you'll probably have more issues anyway.

Exactly!

PR #3078, opened 2021-09-14
Fix ReDoS

ready-research:
Fix ReDoS

PR #4103, opened 2026-09-02
Markup: match comments containing a nested "<!--"

spokodev:
Prism's `comment` rule for markup uses `/<!--(?:(?!<!--)[\s\S])*?-->/`, which refuses to match once it encounters a second `<!--` inside the comment body. Per the HTML tokenizer (WHATWG §13.2.5 comment states), a comment ends at the first `-->`; an embedded `<!--` has no special meaning and is just content.

So `<!-- foo <!-- bar -->` — commented-out markup that itself contains a comment marker — fails to tokenize as a comment and falls through to the `tag` rule, which mis-highlights the whole thing as a bogus HTML tag. Dropping the `(?!<!--)` guard matches the comment as a single token, in line with the spec. Added a test case.

### DIFF
--- a/src/languages/xml.js
+++ b/src/languages/xml.js
@@ -15,7 +15,7 @@ export default {
 
 		return {
 			'comment': {
-				pattern: /<!--(?:(?!<!--)[\s\S])*?-->/,
+				pattern: /<!--[\s\S]*?-->/,
 				greedy: true,
 			},
 			'prolog': {
--- a/tests/languages/markup/nested_marker_comment_feature.test
+++ b/tests/languages/markup/nested_marker_comment_feature.test
@@ -1,0 +1,13 @@
+<!-- foo <!-- bar -->
+
+----------------------------------------------------
+
+[
+	["comment", "<!-- foo <!-- bar -->"]
+]
+
+----------------------------------------------------
+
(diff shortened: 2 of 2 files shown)


jonpyt:
Ignoring comments containing `<!--` is done for performance reasons. The pattern `/<!--[\s\S]*?-->/` is vulnerable to ReDoS. See #3078 and GHSA-727m-83cp-w8cj.

spokodev:
You are right, and I should have checked before opening this. The lookahead is doing real work — it bounds how far each match attempt can scan, so removing it makes the pattern quadratic.

Measured on Node 24, input `"<!--".repeat(n)` (many starts, no `-->`):

| n | `<!--(?:(?!<!--)[\s\S])*?-->` | `<!--[\s\S]*?-->` |
|---:|---:|---:|
| 4000 | 0.0 ms | 10.9 ms |
| 8000 | 0.0 ms | 44.4 ms |
| 16000 | 0.1 ms | 177.8 ms |
| 32000 | 0.1 ms | 721.1 ms |

Four times the work per doubling, so the current guard is the thing keeping this linear. The divergence from the HTML tokenizer on a nested `<!--` is a deliberate trade-off, not an oversight.

Closing.

PR #4105, opened 2026-09-14
Implement `outer:inner` language ids for meta-languages

### LINKED ISSUE #4104: Implement `lang1:lang2` syntax for meta-languages
One of the most straightforward improvements from #3927 is the `outer:inner` syntax for meta-languages that can produce other languages.

This is distinct from embedded languages where the embedding is determined by the actual syntax.

Prime use cases for this are:
- Templating languages like `jinga2`, `njk`, `liquid` etc. These would default to `markup` as the output language, but an explicit e.g. `njk:css` would override.
- Meta-languages like `diff`. These would likely default to no embedded language, but e.g. `diff:js` would override. 
- Note that this needs to be recursive, e.g. `diff:njk:css` should work.

This will not only simplify a lot of existing code and make these languages more declarative, it will also simplify autoloading (e.g. currently `diff-css` is syntactically ambiguous: is it a diff of css, or a new `diff-css` language? Hence the awkwardness of custom code etc.

LeaVerou:
Languages can now declare that they embed or produce another language via a
new `inner` key on the language definition, and be used with compound ids such
as `diff:css`, `django:css` or, recursively, `diff:django:css`.

- `inner: markup` declares the default inner language (templating languages),
  `inner: null` declares a meta-language without a default (`diff`). `:none`
  opts out of a default. Aliases are resolved per part and ids are normalized
  (`django:markup` → `django`, `diff:none` → `diff`).
- New declarative grammar key `$inner`: everything that is not a token of the
  grammar is highlighted as the given language. It replaces the
  `$tokenize: embeddedIn(...)` custom tokenizers of all templating languages;
  a literal grammar of a meta-language gets it at the top level automatically.
  `$placeholder: false` takes tokens out without leaving a placeholder, for
  tokens that don't stand for anything (diff prefixes).
- `templating()` re-inserts tokens by position rather than by searching for
  placeholder text.
- `diff:xxx` replaces the `diff-xxx` pseudo-ids of the diff-highlight plugin,
  which is removed: `diff` highlights the code inside its blocks itself, and
  themes already style the `deleted`/`inserted` tokens.
- The autoloader, `loadLanguages()`, `LanguageRegistry.load()`, `has()` and
  the `language-xxx` class parsing understand compound ids; markdown fences
  accept them; show-language renders them as `Diff (JavaScript)`.

Closes #4104

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01RbhmLXUybTbxCBSwGa2YZr


src/shared/meta/title-data.js
```diff
@@ -20,9 +20,16 @@ function guessTitle (name) {
...
  *
+ * Compound ids (`outer:inner`) produce titles like `Diff (JavaScript)`.
+ *
  * @param {string} name
  * @returns {string}
  */
 export function getTitle (name) {
+	const index = name.indexOf(':');
+	if (index !== -1) {
+		return `${getTitle(name.slice(0, index))} (${getTitle(name.slice(index + 1))})`;
```
DmitrySharabin, line 31: We should probably cover the `outer:none` case explicitly. Otherwise, we'll get, e.g., `Diff (None)` as a title. Or is it okay? I'd expect to see simply `Diff` in that case.

DmitrySharabin:
**Removing `diff-highlight` drops its line backgrounds with no replacement.**

Themes style only the foreground of `.token.deleted` / `.token.inserted`. In a `diff:js` block the wrapper contains inner-language tokens that set their own colors:

```html
<span class="token deleted-sign deleted">
  <span class="token prefix deleted">-</span>
  <span class="token keyword">let</span> a<span class="token operator">=</span><span class="token number">1</span>
</span>
```

`let`, `=` and `1` keep their JavaScript colors, so the deleted color reaches only the spaces between them. That is what the plugin's opt-in `class="diff-highlight"` existed for.

Consequence: in `diff:<lang>` blocks, changed lines are distinguishable only by the one-character prefix. Default rendering is unaffected — this hits only users who opted in.

If it's restored as presentation-only, the CSS needs no changes: `diff.js:38` gives `deleted-sign` and `deleted-arrow` the `deleted` alias, so the existing `.token.deleted:not(.prefix)` still matches the wrappers and excludes the prefixes. `color: inherit` has to stay, or the inner syntax colors are lost. A CSS-only plugin isn't possible today — `scripts/build.js:150` and `:414` both require `plugins/<id>/<id>.js`.

src/shared/meta/title-data.js
DmitrySharabin: Proposed a fix in #4107.

LeaVerou:
I'm aware, but I think we need to find a better solution for that than a companion plugin. We should however file a bug and tag it with a v2 release milestone. But I think it's ok for the alpha. Worst case as a workaround we could add something to the default theme.

DmitrySharabin (approved):
LGTM! The API is awesome!

